PR #8, opened 2022-10-20
Show progress by default when run in a notebook

freider:
Looking into better ways of checking this... Might also want to use it to mute some warnings that are common to get in notebooks like overwriting stub functions

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -26,6 +26,7 @@ from .rate_limit import RateLimit
 from .schedule import Schedule
 from .secret import _Secret
 from .shared_volume import _SharedVolume
+from ._ipython import is_notebook
 
 _default_image = _Image.debian_slim()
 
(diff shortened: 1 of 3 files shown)


freider:
I'll try to add some notebook test. Perhaps a client test that runs through cells in a notebook and asserts that the output cells look like expected?

erikbern:
could also just put something in `_output.py` like this:

```python
def isatty(stdout):
    if stdout.isatty():
        return True
    try:
        import ipykernel.iostream

        return isinstance(stdout, ipykernel.iostream.OutStream)
    except ImportError:
        return False
```

modal/stub.py
```diff
@@ -462,11 +463,12 @@ def _add_function(self, function: _Function) -> _FunctionHandle:
         if function.tag in self._blueprint:
             old_function = self._blueprint[function.tag]
             if isinstance(old_function, _Function):
-                logger.warning(
-                    f"Warning: Tag {function.tag} collision!"
-                    f" Overriding existing function [{old_function._info.module_name}].{old_function._info.function_name}"
-                    f" with new function [{function._info.module_name}].{function._info.function_name}"
-                )
+                if not is_notebook():
```
erikbern, line 466: oh because you want to redefine functions?

erikbern: wonder if it makes sense to do something like @aksh-at mentioned where we have a helper function `stub.notebook()` or whatever which creates a background task that always syncs the stub with the running app

freider: Yeah, but those are two slightly unrelated problems. Redefining functions, even without having them repopulate a running app, is pretty common in notebooks as you iterate on the function definitions (and don't want to restart the interpreter/kernel every time you save the cell), so I think it makes sense to ignore these deprecations in notebook contexts.

Adding something like `stub.notebook()` (probably part of a notebook magic?) would be helpful too, in case the app has already been constructed and is live when the redefinition happens.

PR #22, opened 2022-10-21
Add 'modal --version' capability

thundergolfer:
```
(venv) ubuntu@ip-172-31-70-51:~/modal-client$ modal --version
modal client version: 0.38.101
```

I reached for this today. Already thought `--version` was far more command than `modal version`, and https://clig.dev/ is evidence enough to just choose that.

### DIFF
--- a/modal/cli/entry_point.py
+++ b/modal/cli/entry_point.py
@@ -8,6 +8,15 @@ from .secret import secret_cli
 from .token import token_cli
 from .volume import volume_cli
 
+
+def version_callback(value: bool):
+    if value:
+        from modal_version import __version__
+
+        typer.echo(f"modal client version: {__version__}")
+        raise typer.Exit()
+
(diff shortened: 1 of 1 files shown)


modal/cli/entry_point.py
```diff
@@ -20,6 +29,15 @@
     """,
 )
 
+
+@entrypoint_cli.callback()
+def root(
+    ctx: typer.Context,
+    version: bool = typer.Option(None, "--version", callback=version_callback),
```
freider, line 36: wow, hacky!

freider:
Nice!

modal/cli/entry_point.py
erikbern: haha yeah... would be nice if there's some slightly less hacky way of doing it. but not a big deal.

PR #27, opened 2022-10-24
Adds simple notebook test

freider:
Injects info about test server into a parameter-annotated notebook executed by ~~papermill~~ nbclient


client_test/notebook_test.py
```diff
@@ -0,0 +1,53 @@
...
+        return tagged_cells
+
+    return runner
+
+
+def test_notebook_outputs_status(notebook_runner, test_dir):
+    input_notebook_path = test_dir / "supports" / "notebooks" / "simple.ipynb"
+    tagged_cells = notebook_runner(input_notebook_path)
+    combined_output = "\n".join(c["data"]["text/plain"] for c in tagged_cells["main"]["outputs"])
+    assert "Initialized" in combined_output
```
freider, line 51: Actually not sure how this passes now, since this is before I merged my fix for auto-detecting output :thinking: Maybe nbclient doesn't manipulate sys.stdout like a vanilla notebook does...

client_test/notebook_test.py
```diff
@@ -0,0 +1,53 @@
...
+from nbclient.exceptions import CellExecutionError
+
+
+@pytest.fixture
+def notebook_runner(servicer):
+    import nbformat
+    from nbclient import NotebookClient
+
+    def runner(notebook_path: Path):
+        output_notebook_path = notebook_path.with_suffix(".output.ipynb")
```
erikbern, line 14: minor, but it would be nice to use jupytext or something similar instead of storing an .ipynb in the repo – makes it a bit easier to debug these things when things fail since you don't have to install and set up notebooks

erikbern (approved):

PR #32, opened 2022-10-25
Interactive functions

aksh-at:
Adding `interactive=True` runs it in a PTY. This means that `breakpoint()` works out of the box when called from inside that function:

https://user-images.githubusercontent.com/5786378/197664805-c51d67bc-3d02-4264-945d-6f2a1fd34f73.mov

Also installing ipython inside the image (but not adding to `setup.cfg`, so it's not a requirement to install Modal). `modal.ipython()` is a simple wrapper around `IPython.embed()` that you can drop in to open an IPython shell from an interactive function.

The biggest caveat is that interactive functions don't return their values anymore, because of the way the function runs in a child process. It's possible to do this better, but it's a bit gnarly so will let this rest for a bit. For now, adding a warning that lets the user know that the return value won't be sent back (if their function returns a value).

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2022
 from modal_version import __version__
 
+from ._pty import ipython
 from .app import App, container_app, is_local
 from .dict import Dict
 from .exception import Error
@@ -41,6 +42,7 @@ __all__ = [
     "container_app",
     "create_package_mount",
     "create_package_mounts",
+    "ipython",
     "is_local",
     "lookup",
     "ref",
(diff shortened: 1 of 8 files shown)


erikbern:
This looks super cool but I guess I'm a bit confused what's the user interaction mode? Like, is the point to make it easier to debug existing Modal code?

aksh-at:
Yeah the point is to debug or interact with my code in Modal. The thing I wanted to use it for was to open an IPython shell inside a function so I can figure out how to use this library that I don't want to install locally. (And it takes too long to run the whole function again every time I tweak a line). 

The other use case is just adding a `breakpoint()` (to drop into pdb), which quite a few people have asked for.

PR #36, opened 2022-10-25
Add 3.11

erikbern:
Got it working thanks to @aksh-at's trick adding `;python_version>='3.11'`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,8 @@ install_requires =
     aiostream
     asgiref
     certifi
-    cloudpickle>=2.0.0,<2.1.0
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -22,7 +22,8 @@ install_requires =
     aiostream
     asgiref
     certifi
-    cloudpickle>=2.0.0,<2.1.0
+    cloudpickle>=2.0.0;python_version<'3.11'
```
erikbern, line 25: not sure why we had it pinned to 2.0.*, but i hope it's fine to unpin it

erikbern:
Apparently we need to make sure not to run different minor versions of cloudpickle in the container vs in the client. So that's why we pin the minor version. Not super clean.

PR #45, opened 2022-10-27
link to ephemeral apps page on ephemeral app creation

thundergolfer:
This would be a small quality of life improvement (at least for me 🙂, but probably for others). 

> View app page at ...

image

**linking to:**

image

(Would be more interesting with a more interesting app...)


erikbern:
nice!

freider (approved):

PR #56, opened 2022-11-09
Add support and tests for Python 3.11

ekzhang:
Python 3.11 is between 10-60% faster than Python 3.10.

https://docs.python.org/3/whatsnew/3.11.html


ekzhang:
Does `grpcio` not support Python 3.11?

PR #60, opened 2022-11-10
[MOD-661] prevent coro task collections growing indefinitely

thundergolfer:
**Details** in [Linear MOD-661](https://linear.app/modal-labs/issue/MOD-661/modal-server-backgroundpy-memory-leak). 

Started looking at this because a long-running background k8s container kept growing its memory until it hit OOM and got restarted.

**TLDR** is that this change significantly reduces the memory usage reported by the [scalene](https://github.com/plasma-umass/scalene) profiler's memory profiling.

Getting review because this module is kinda complicated and I may have missed something when making this change.


erikbern (approved):
We have a bunch of tests for the `TaskContext` so I don't see any risks merging it!

PR #72, opened 2022-11-17
Extra tests of container requirements.txt

erikbern:
Spent a bunch of time trying to parse pip dependencies and whatnot but realized the most direct way to test that `modal/requirements.txt` works is to just run the container tests with it.

Could potentially also run all tests and use an environment variable to disable any tests that's "local only" (e.g. any test of stub.serve)


erikbern:
this is failing pretty hard because some dumb stuff. will revisit this later.

erikbern:
seems pretty hard to get this working in a clean way – closing it now

PR #74, opened 2022-11-18
Enable CORS by default on simple webhooks

ekzhang:
This change allows webhooks to be called from browsers, using an implementation of [cross-origin resource sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) standard headers.

CORS is primarily helpful to ensure that users of a website don't have their cookies or auth stolen by another website, but that's only relevant if you have a full web app and using `@stub.asgi` anyway — I don't see anyone, e.g., setting cookies in `@stub.webhook` that could be stolen or vulnerable to CSRF in a CORS context.

I also wrote some documentation for both webhook constructor functions and a test.

Resolves MOD-534

### DIFF
--- a/client_test/webhook_test.py
+++ b/client_test/webhook_test.py
@@ -15,3 +15,26 @@ async def f(x):
 async def test_webhook(servicer, aio_client):
     async with stub.run(client=aio_client):
         assert f.web_url
+
+
+def test_webhook_cors():
+    from fastapi.testclient import TestClient
+
+    from modal._asgi import webhook_asgi_app
+
+    def handler():
+        return {"message": "Hello, World!"}
+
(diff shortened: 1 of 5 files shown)


luiscape:
I think this is reasonable. The `@stub.webhook` is meant to give users a quick way for exposing functions, so enabling this by default seems the right right thing to do.

luiscape (approved):

ekzhang:
Thank you for the review!

erikbern:
A bit nervous about this long-term, at least if we enable custom domains. Maybe we can catch any security problems though? Like if the browser sends a cookie, we could reject the request unless CORS is explicitly enabled.

PR #75, opened 2022-11-18
[MOD-673] Provide instructions to a usr if a client auth fails with missing token

thundergolfer:
* Handled client-side to tie the local lack of creds to the server-side `AuthError`
* Hardcoding domain is a minor annoyance.


erikbern:
Nice!

PR #78, opened 2022-11-18
Use __future__.annotations to avoid needing typing.List, typing.Dict annotations

thundergolfer:
Our minimum supported Python version is **3.7**, so we can use `from __future__ import annotations` safely: https://docs.python.org/3/library/__future__.html#id2.

This changeset just does one module. Can do the rest in a follow-up.


erikbern:
oh cool

PR #80, opened 2022-11-21
Add a unit test for container_entrypoint webhooks

erikbern:
Was going to fix the lifecycle methods for webhooks, but realized there's zero unit tests for webhooks in general in the client. Added a basic one.

This is purely test code, no functional changes.


aksh-at (approved):
This is great, thanks!

PR #84, opened 2022-11-22
Fix lifecycle methods for webhooks

erikbern:
This ended up being more complex than I anticipated but net is actually only +3 lines – the rest of the additions is a test for it.

PR #88, opened 2022-11-22
Fix WSGI apps

erikbern:
I think this should resolve the issue mentioned here: https://github.com/modal-labs/modal-client/pull/84#issuecomment-1324018072


thundergolfer (approved):

erikbern:
Let's add tests specifically for this at some point – we do have integration tests for it though

PR #92, opened 2022-11-28
Poetry: fail on missing lockfile

erikbern:
Minor change prompted by discussion here: https://modal-com.slack.com/archives/C02ALG62GE7/p1669654602066509

Added a test as well

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -274,6 +274,8 @@ class _Image(Provider[_ImageHandle]):
     def poetry_install_from_file(
         self,
         poetry_pyproject_toml: str,
+        poetry_lockfile: Optional[str] = None,
+        ignore_lockfile = False,
     ):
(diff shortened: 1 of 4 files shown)


modal/image.py
```diff
@@ -274,6 +274,8 @@ def pip_install_from_requirements(
     def poetry_install_from_file(
         self,
         poetry_pyproject_toml: str,
+        poetry_lockfile: Optional[str] = None,
+        ignore_lockfile=False,
```
freider, line 278: Nit: Perhaps add some comments to the args, like the lockfile being automatically inferred and required by default?

freider (approved):

modal/image.py
erikbern: Yeah, good point, I'll add that!

PR #94, opened 2022-11-29
Fixes two issues with serialized functions

freider:
* There was a bug with get_function_mounts erroring for serialized functions (should return no code mount)
* Closure values were determined at postponed _load time instead of function registration time, so all functions decorated in a loop using the loop variable would get the value of the last loop var.
* Adds a name override option to the function decorator which sets the modal name/tag of the function.

TODO:
* There are obviously some missing tests for serialized functions since the first bug wasn't caught, so will add some now
* Lifecycle classes for function local serialized functions aren't working at the moment since it still relies on loading the class from global scope. Postponing a fix for that, as it might be simplified if we change lifecycle syntax.

### DIFF
--- a/modal/_function_utils.py
+++ b/modal/_function_utils.py
@@ -8,6 +8,8 @@ import typing
 from pathlib import Path
 from typing import Dict, Union, Any, Optional, Type, Callable
 
+import cloudpickle
+
 from modal_proto import api_pb2
 from .config import config, logger
 from .exception import InvalidError
@@ -54,12 +56,13 @@ def filter_safe_mounts(mounts: typing.Dict[str, _Mount]):
 
 
 class FunctionInfo:
-    """Class the helps us extracting a bunch of information about a function."""
+    """Class the helps us extract a bunch of information about a function."""
 
     # TODO: we should have a bunch of unit tests for this
     # TODO: if the function is declared in a local scope, this function still "works": we should throw an exception
-    def __init__(self, f, serialized=False):
-        self.function_name = f.__qualname__
+    def __init__(self, f, serialized=False, name_override: Optional[str] = None):
+        self.raw_f = f
+        self.function_name = name_override if name_override is not None else f.__qualname__
         self.signature = inspect.signature(f)
         module = inspect.getmodule(f)
 
@@ -85,6 +88,7 @@ class FunctionInfo:
             self.definition_type = api_pb2.Function.DEFINITION_TYPE_FILE
             self.is_package = True
             self.is_file = False
+            self.serialized_function = None
         elif hasattr(module, "__file__") and not serialized:
             # This generally covers the case where it's invoked with
             # python foo/bar/baz.py
@@ -94,12 +98,15 @@ class FunctionInfo:
             self.definition_type = api_pb2.Function.DEFINITION_TYPE_FILE
             self.is_package = False
             self.is_file = True
+            self.serialized_function = None
         else:
             self.module_name = None
             self.base_dir = os.path.abspath("")  # get current dir
             self.definition_type = api_pb2.Function.DEFINITION_TYPE_SERIALIZED
             self.is_package = False
             self.is_file = False
+            self.serialized_function = cloudpickle.dumps(self.raw_f)
+            logger.debug(f"Serializing {self.raw_f.__qualname__}, size is {len(self.serialized_function)}")
 
         if self.definition_type == api_pb2.Function.DEFINITION_TYPE_FILE:
             # Sanity check that this function is defined in global scope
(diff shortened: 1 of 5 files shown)


modal/_function_utils.py
```diff
@@ -186,10 +193,18 @@ def is_nullary(self):
         return all(param.default is not param.empty for param in self.signature.parameters.values())
 
 
+class LocalFunctionError(Exception):
```
erikbern, line 196: maybe inherit from `InvalidError`

modal/_function_utils.py
```diff
@@ -186,10 +193,18 @@ def is_nullary(self):
...
+    pass
+
+
 def load_function_from_module(module, qual_name):
     # The function might be defined inside a class scope (e.g mymodule.MyClass.f)
     objs: list[Any] = [module]
     for path in qual_name.split("."):
+        if path == "<locals>":
```
erikbern, line 204: there is already some code that checks this (and raises `InvalidError` in the `FunctionInfo` constructor – maybe worth making it a utility function

erikbern: Actually, I'm confused how this exception could happen? The `load_function_from_module` is only called when `definition_type == api_pb2.Function.DEFINITION_TYPE_FILE` right? and if that's the case, the function shouldn't have been possible to create in the first place (because of the check I just mentioned)

erikbern:
> There are obviously some missing tests for serialized functions since the first bug wasn't caught, so will add some now

+1 for more tests. Now that we don't have a monorepo, we need to add more unit tests for these things. The upside is that the client tests run super fast so it's much nicer to work with IMO

freider:
The fix is in moving `cloudpickle.dumps()` from `_load()` into `FunctionInfo` construction. Good point that this will indeed trigger cloudpickle serialization in containers as well (if those modules are loaded, which they won't if all of the methods in them are serialized). Not sure if that's a show stopper for this?

The following works after this fix, and is quite useful when writing libraries dynamically generating modal functions from some kind of external definition:

```python
def make_functions():
    for s in ["foo", "bar"]:
        @stub.function(name=f"add_{s}", serialized=True)
        def adder(x):
            return x + s

make_functions()
```

modal/_function_utils.py
```diff
@@ -189,7 +200,12 @@ def is_nullary(self):
 def load_function_from_module(module, qual_name):
     # The function might be defined inside a class scope (e.g mymodule.MyClass.f)
     objs: list[Any] = [module]
+    if not is_global_function(qual_name):
+        raise LocalFunctionError()
```
erikbern, line 204: probably want to add a message?

modal/_function_utils.py
```diff
@@ -94,17 +102,20 @@ def __init__(self, f, serialized=False):
...
             self.is_file = False
+            self.serialized_function = cloudpickle.dumps(self.raw_f)
+            logger.debug(f"Serializing {self.raw_f.__qualname__}, size is {len(self.serialized_function)}")
 
         if self.definition_type == api_pb2.Function.DEFINITION_TYPE_FILE:
             # Sanity check that this function is defined in global scope
             # Unfortunately, there's no "clean" way to do this in Python
-            if "<locals>" in self.function_name.split("."):
+            if not is_global_function(f.__qualname__):
                 raise InvalidError("Modal can only import functions defined in global scope")
```
erikbern, line 119: shouldn't this be `LocalFunctionError` too?

erikbern: separately, we _could_ let people know of `serialized=True` here? maybe? i don't know

modal/exception.py
```diff
@@ -42,6 +42,12 @@ class InvalidError(Error):
     """Raised when user does something invalid."""
 
 
+class LocalFunctionError(InvalidError):
```
erikbern, line 45: super minor preference for defining this in `_function_utils.py` since it's never used elsewhere, but important

modal/_function_utils.py
freider: I thought so too at first, but it's also called for *only* extracting the lifecycle class of a serialized function, and then only working if the serialized function is also global... definitely feels like a code wart though, I agree.

I have a recollection of working on an alternative way of getting lifecycle classes for serialized functions at some point, but it will be a lot easier once we have class decorators or similar.

modal/functions.py
```diff
@@ -772,11 +773,18 @@ async def _load(self, client, stub, app_id, loader, message_callback, existing_f
...
 
-            cls, _ = load_function_from_module(mod, self._raw_f.__qualname__)
+            try:
+                cls, _ = load_function_from_module(mod, self._raw_f.__qualname__)
+            except LocalFunctionError:
+                # if a serialized function is defined within a function scope
+                # we can't load it from the module and detect its parent class
+                # TODO: fix this somehow... maybe put the decorator on the
+                #       class instead for entrypoint classes
+                cls = None
```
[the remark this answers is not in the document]
freider: To some degree yeah, it's more of a graceful fallback for an error we can't currently handle for serialized functions. Would you prefer if I send in a flag whether to raise or not in case the class can't be loaded? Or some other structure suggestions?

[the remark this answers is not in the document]
erikbern: Don't think it's super important, but code like this looks like we're "suppressing" an error which I think can make debugging harder in some cases... that's why I don't love using exceptions for control flow. But it's a very minor thing.

erikbern:
Sorry for lots of minor nits here. Feel free to ignore a lot of it if you think it doesn't make sense!

PR #97, opened 2022-11-30
f.call(...) instead of f(...)

erikbern:
This is a very major change to the Modal SDK, but very small code delta.

It deprecates the old `f(...)` syntax in favor of `f.call(...)`

The ultimate goal is for `f(...)` to call a function _in the same process_. This means decorating a function in Modal conceptually just "adds" Modal functionality to it, rather than changing it.

Some reasons I think this is the way forward:
* I think it helps shape the mental model of what functions run in what containers. I've talked to several users who are confused about what needs to be decorated. If you have a Modal function `f` and it needs to call a utility function `g` in the same process, there's confusion whether `g` needs to be decorated (it doesn't). 
* It makes it easier to read code and see calls that cross container boundaries
* It opens the door for auto-decoration of functions, in particular it will be possible for a class decorator to auto-decorate every method on a class.
* Unit testing Modal functions will be simpler – you can just call `f(...)` rather than the awkward `f.get_raw_f()(...)` today. (although to be fair, we _could_ add a `f.local(...)` method)
* Decorating `f(...)` today will sometimes change the function signature in certain cases – e.g. if you call a sync function from an async context or vice versa. This is a pretty minor thing but seems ugly to me.

I'm not particularly attached to the name "call" – open to whatever (eg remote, invoke, ...).

I expect the deprecation path to be very long for something like this. Suggested schedule:

1. This PR: Add deprecation warnings to `f(...)` and nudge people to switch to `f.call(...)`
3. At some point in Jan 2023: Turn deprecation warnings into exceptions
4. At some point in Feb 2023: Add back the `__call__` method but this time it calls the underlying function in the same process

### DIFF
--- a/client_test/function_test.py
+++ b/client_test/function_test.py
@@ -25,7 +25,7 @@ def dummy():
 
 def test_run_function(client):
     with stub.run(client=client):
-        assert foo(2, 4) == 20
+        assert foo.call(2, 4) == 20
 
 
 def test_map(client):
@@ -194,7 +194,7 @@ def test_function_exception(client, servicer):
 
     with stub.run(client=client):
         with pytest.raises(CustomException) as excinfo:
-            failure_modal()
+            failure_modal.call()
         assert "foo!" in str(excinfo.value)
 
 
@@ -208,7 +208,7 @@ def test_function_relative_import_hint(client, servicer):
     import_failure_modal = stub.function(servicer.function_body(import_failure))
     with stub.run(client=client):
         with pytest.raises(ImportError) as excinfo:
-            import_failure_modal()
+            import_failure_modal.call()
         assert "HINT" in str(excinfo.value)
 
 
--- a/client_test/function_test.py
+++ b/client_test/function_test.py
@@ -96,7 +96,7 @@ def test_function_future(client, servicer):
     later_modal = stub.function(servicer.function_body(later))
 
     with stub.run(client=client):
-        future = later_modal.submit()
+        future = later_modal.spawn()
         assert isinstance(future, FunctionCall)
 
         servicer.function_is_running = True
@@ -108,6 +108,13 @@ def test_function_future(client, servicer):
         servicer.function_is_running = False
         assert future.get(0.01) == "hello"
 
+        with pytest.warns(DeprecationError):
+            future = later_modal.submit()
+            assert isinstance(future, FunctionCall)
+
+        servicer.function_is_running = True
+        assert future.object_id == "fc-2"
+
 
 @pytest.mark.asyncio
 async def test_function_future_async(client, servicer):
@@ -116,7 +123,7 @@ async def test_function_future_async(client, servicer):
     later_modal = stub.function(servicer.function_body(later))
 
     async with stub.run(client=client):
-        future = await later_modal.submit()
+        future = await later_modal.spawn()
         servicer.function_is_running = True
 
         with pytest.raises(TimeoutError):
@@ -148,7 +155,7 @@ async def test_generator_future(client, servicer):
 
     later_gen_modal = stub.function(later_gen)
     with stub.run(client=client):
-        assert later_gen_modal.submit() is None  # until we have a nice interface for polling generator futures
+        assert later_gen_modal.spawn() is None  # until we have a nice interface for polling generator futures
 
 
 async def slo1(sleep_seconds):
@@ -165,7 +172,7 @@ def test_sync_parallelism(client, servicer):
     with stub.run(client=client):
         t0 = time.time()
         # NOTE tests breaks in macOS CI if the smaller time is smaller than ~300ms
-        res = gather(slo1_modal.submit(0.31), slo1_modal.submit(0.3))
+        res = gather(slo1_modal.spawn(0.31), slo1_modal.spawn(0.3))
         t1 = time.time()
         assert res == [0.31, 0.3]  # results should be ordered as inputs, not by completion time
         assert t1 - t0 < 0.6  # less than the combined runtime, make sure they run in parallel
--- a/client_test/rate_limit_test.py
+++ b/client_test/rate_limit_test.py
@@ -26,5 +26,5 @@ def test_rate_limit(client):
         stub.function(dummy, rate_limit=modal.RateLimit(per_minute=15, per_second=5))
 
     with stub.run(client=client):
-        per_second_5_modal()
-        per_minute_15_modal()
+        per_second_5_modal.call()
+        per_minute_15_modal.call()
--- a/client_test/retries_test.py
+++ b/client_test/retries_test.py
@@ -52,7 +52,7 @@ def test_retries(client):
         stub.function(dummy, retries=modal.Retries(max_retries=2, backoff_coefficient=0.0))
 
     with stub.run(client=client):
-        default_retries_from_int_modal()
-        fixed_delay_retries_modal()
-        exponential_backoff_modal()
-        exponential_with_max_delay_modal()
+        default_retries_from_int_modal.call()
+        fixed_delay_retries_modal.call()
+        exponential_backoff_modal.call()
+        exponential_with_max_delay_modal.call()
--- a/client_test/shared_volume_test.py
+++ b/client_test/shared_volume_test.py
@@ -18,7 +18,7 @@ def test_shared_volume_files(client, test_dir, servicer):
     )(dummy)
 
     with stub.run(client=client):
-        dummy_modal()
+        dummy_modal.call()
 
 
 @pytest.mark.skipif(platform.system() == "Windows", reason="TODO: implement client-side path check on Windows.")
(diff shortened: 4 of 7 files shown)


erikbern:
Btw, regarding names, Ray uses `f.remote(...)` – however, this returns a future that you have to call `.get()` on, so it's a bit different.

erikbern:
Naming ideas:
* `.call`
* `.invoke`
* `.dispatch`
* `.remote`
* `.rpc`

tbenthompson:
Love this! Slight preference towards names that make the remote execution clear like `.remote` and `.rpc` just as a way of more clearly flagging to new users that aren't familiar with Modal that we're jumping across into the cloud.

anishpdalal:
I think this is definitely helps make the execution model more clear just by reading the code so I'm a fan of the change! I also prefer something like `remote ` since it makes it clearer where the code is running.

Given this name change, I'm also wondering if it might make sense to change the `.submit()` method to something like `.call_async()` ? From a code reading perspective `.call` and `.submit` are similar names but invoke functions very differently.

erikbern:
> Given this name change, I'm also wondering if it might make sense to change the `.submit()` method to something like `.call_async()` ? From a code reading perspective `.call` and `.submit` are similar names but invoke functions very differently.

Yeah good point! It would be incredibly confusing to have both `invoke` and `submit` doing slightly different things.

I think the main options are something like:

| calling        | mapping     | async call        |                                                    |
| ----------- | ----------- | -------------- | ------------------------------ |
| `f(...)`        | `f.map(...)` | `f.submit(...)` | the current situation                  |
| `f.call(...)` | `f.map(...)` | `f.submit(...)` | what this PR initially proposed  |
| `f.call(...)` | `f.map(...)` | `f.call_async...)` | what I think @anishpdalal suggested  |
| `f.rpc(...)` | `f.rpc_map(...)` | `f.rpc_async...)` | could maybe also work? logical but a bit ugly imo  |

erikbern:
Btw, Lambda uses the term `invoke` and has a separate argument `invocation_type` that's either `"RequestResponse"` or `"Event"`.

erikbern:
Planning to merge this tomorrow unless we have any last minute concerns!

koenvo:
In celery there is something like “eager” mode. When this is enabled all calls to functions are run eager in the current process instead of send to the broken. 

I was playing with unit tests and came to the conclusion I don’t want to have different ways of calling functions (especially when they call other remote functions again) but just want everything to run locally. 

The thing I tried was creating my own decorator which delegates a call to the modal decorated function or to the raw one depending on some settings. Only thing I couldn’t work around was deployment: it looks like a local run but functions need the modal decorator at that time.

Maybe my comment should be placed somewhere in a discussion thread and I didn’t want to spam this PR. If above doesn’t make sense please ignore :-)

erikbern:
> I was playing with unit tests and came to the conclusion I don’t want to have different ways of calling functions (especially when they call other remote functions again) but just want everything to run locally.

This PR will make it slightly easier to use Modal locally – you should be able to write unit tests for decorated functions for instance.

I think there's separately something we should probably build that changes the behavior of `f.call(...)` in runtime so that it just calls `f(...)` – for some cases where you need to test chains of calls locally. I think that changes is fairly unrelated to this PR though!

dongreenberg:
A few thoughts. There are a few modes to consider that are woven together:
1. Async vs. Sync - in PyTorch RPC we used "rpc_sync" and "rpc_async" to denote whether to block on the call or return a future.
2. Whether the returned object is local or remote - In PT RPC we had a separate call "remote" which returned a reference to a remote object which still lived on the worker that created it, so if it were passed into another worker it can stream over directly, rather than bounce off the user's laptop.
See [here](https://youtu.be/bWqpf8ISk_c?t=237).

Ray is opinionated that all calls that go through Ray should be async and remote, which is why they just have .remote. Ray also uses "submit" for running shell commands, but .submit (only on ActorPools) as an RPC-stream like call to enqueue a task that can be retrieved via "get_next" for results (and "map" for ActorPools in the same way Modal uses it). 

In Runhouse we have a bunch of variants of calling an RPCed function - __call__, map, starmap, enqueue (run with no resource contention in the VM so you don't overload or thrash GPU memory), async, and remote (right now this means if the result is a resource with filesystem assets, we convert those to SFTP references via fsspec). We also have "run" and "run_python" for running a shell or python command in the VM. But we don't decorate the function to make it into an RPC so there's no concept of "running it locally" (the raw fn is a totally different object).  So in the interest of wanting to be able to interop nicely with Modal, I would vote for "call" (denotes same behavior as __call__), "map," "call_async" or "call_nonblocking" (clearest IMO), and save "remote" for returning references if you do that later. That would make you consistent with Ray too, just because you don't overlap with their submit or remote, and you both would call the fn locally if just called with __call__. Spawn is a little confusing to me, because "call" also spawns a new process (and with a new interpreter, if you want to talk about relationship to multiprocessing's spawn), as would remote.

dongreenberg:
Also I definitely agree that a global config or env variable to switch between fn() and fn.call() to call to Modal would be lovely. 

I guess I'll also throw out the option to do what we do, which is no decorator, but I think that may be a pretty big change for Modal. To get an rpc you do `my_rpc_fn = rh.send(my_fn, hardware...)` so there's no ambiguity about what __call__does on the rpc, and you can rpc any function, even one that isn't yours to decorate (obviously you can just use decorators as functions to do this too, but you get what I mean). It also means that you can send your function to a few different kinds of hardware to get rpcs for each, whereas with a decorator you need to have one configuration for that function. *But*, you don't get the same elegance of being able to flip a switch to make everything run locally (not as obviously at least) for CI or debugging, and it's not as obvious how you would decorate a library of functions for someone else to use with their own Modal account. Those two use cases make a lot more sense for Modal than Runhouse. Also for the record I really like how you've designed your decorators.

erikbern:
> To get an rpc you do `my_rpc_fn = rh.send(my_fn, hardware...)`

This approach works well if you can cloudpickle `my_fn` and you're fine with the overhead of doing that for every function call, I think? Modal creates all functions attached to the stub in the cloud prior to calling them, so it wouldn't work quite as well.

erikbern:
the current `f.submit` actually does return a reference to an object that lives in the cloud (not just a local future), so maybe it's more like the Ray remote concept in that sense. But like you said, there's no blocking function call in Ray, so they only have one type of calls – we have two, and it would be confusing to call them "call" and "remote", I think.

We _could_ get rid of the `.call` and _only_ have `.remote`, I guess, but I wonder if that's easier. Then every blocking call would look like `f.remote(42).get()` instead.

dongreenberg:
> This approach works well if you can cloudpickle my_fn and you're fine with the overhead of doing that for every function call
Sorry, I wasn't super clear. The full flow is: 
`my_rpc_fn = rh.send(my_fn, hardware...)
res = my_rpc_fn(my_inputs)  # you can call this over and over`
We don't cloudpickle the function, just sync the code over to the cluster when the rpc is created, and then when it's called we send a gRPC request to the cluster which imports and calls the function. I think it's the same as you guys do (at least since last time we spoke about this). Either way none of this really matters because I was just throwing this out there to be intellectually honest 🙃, but it's not the right option for Modal. 

> the current f.submit actually does return a reference to an object that lives in the cloud (not just a local future)

tbh, I would actually then support leaving `call` but then just replacing `submit` with `remote.` I don't recall where but recently in the Ray slack some people were asking for a wrapper that did ray.get(fn.remote(my_ags)) because they were doing it all over the place (and I've personally done this a lot), and this is just `call`. IMO if you make `remote` the core usage path and sync calls second-class citizens, you're turning Modal in a non-eager framework by default. I'm obviously biased but I think the eagerness is a major ergonomic benefit today. It'll also be a lot weirder to run your code locally if you have fn.remote.get() all over the place but a flag is secretly making all of that happen locally (and doing the gets through a Ray-like KV store?).

If you had `call`, `map`, and `remote`, I think that would be super clear (`call` means give me results like this is a normal function call, and `remote` means call my function and leave the results remote until I ask for them) and consistent with Ray and PyTorch RPC.

dongreenberg:
And obviously do what you please, I don't mean to thrash this PR if you're already pretty settled.

aksh-at:
> remote means call my function and leave the results remote until I ask for them

Interesting, that's not what I'd guess `.remote()` to mean if I saw it out of context. Before looking at Ray docs, I thought it was just short for "remote call". I'm a bit wary of `call()` and `remote()` for this reason—the interpretations can be ambiguous. 

Originally, I was leaning towards `call` and `call_*` where * is some suffix that indicates that the call is not blocking. The benefit of this is that it highlights the similarity between the two methods (arguments are the same, but return types are different). The problem with this (that @ekzhang pointed out) is that `call_nonblocking` is a bit weird when used with async-flavored Modal, because you'd have to `await fn.call_nonblocking()` to get your future back.

dongreenberg:
Okay, I get that. I guess then the only thing which isn't too close to just being a synonym for "remote function call" but also doesn't make explicit reference to async or non-blocking would be `submit`.  I should have known that I'd eventually arrive back at whatever you gang have already figured out 😄

erikbern:
`remote` is a great name if it's the _only_ way to call a function, but it's a bit confusing if we have both `remote` and other names – it sort of implies the other calls are not remote

erikbern:
> `my_rpc_fn = rh.send(my_fn, hardware...)`
> `res = my_rpc_fn(my_inputs)  # you can call this over and over`

Yeah that makes sense. the verb "send" is a bit strange to me maybe? But other than that, it roughly looks similar to Modal today.

dongreenberg:
Ya it's a pretty different concept, where we're sending functions to existing / persistent compute, a little more like PyTorch's `.to`. At one point it was also supposed to be a contraction for "serverless endpoint" but that doesn't really make much sense considering how serverful it is. Anyway I'll stop spamming the whole Modal team's github notifications and tell you more about it later.

PR #98, opened 2022-11-30
Remove Image(version=) kwarg support

ekzhang:
As discussed on company Slack this argument isn't used for anything and is obsolete.

No examples use it and it isn't documented anywhere, so I think the risk of breakage is low (therefore no deprecation period). I'm still bumping the client version number though, since this is technically a breaking change (public API update).


erikbern:
Seems fine. I think it's always been there but never used. Don't think it's important to bump the client version but I also don't see any harm in doing it!

erikbern (approved):

PR #99, opened 2022-12-01
Remove closed channels from channel pool

freider:
Fixes the `AttributeError: 'Connection' object has no attribute '_transport'` error.

The connection pool will still raise an error if a connection is closed mid-request (StreamTerminatedError) or if it can't create a new connection, but that shouldn't happen unless something is wrong with the server/load balancer logic (I think)


erikbern (approved):
Nice! I wonder if we should try to get the ChannelPool into grpclib at some point

PR #100, opened 2022-12-01
Deprecate @stub.generator

erikbern:
It's rarely used and we can almost always infer it from the signature of the function. In some rare cases we can't, but we can catch the issue easily and tell the user to add `is_generator=True`.


erikbern:
This _might_ break our internal integration tests, which I'll fix separately

erikbern:
Merging this – should be relatively harmless

PR #102, opened 2022-12-01
Remove old deprecated code, add a date to deprecations going forward

erikbern:
We had a lot of deprecated code that was more than 3 months old – removing all of it.

Added a date argument to `deprecation_warning` to track _when_ we deprecate code


erikbern:
Let me actually turn these into exceptions first

PR #105, opened 2022-12-01
New app state DEPLOYING to support webhook UX

thundergolfer:
A deployment proceeds through these steps:

1. App creation (or reuse if previous deployment)
2. Object creation, including Function objects that are webhooks
3. App deployment

Currently webhook label assignment happens in step 2, but there's no data available to differentiate an ephemeral app from a new deployment. `app.name` allow differentiation, but it is set in step 3, once a deployment has succeeded (the name vs.description situation is already confusing, too).

This PR introduces a new app state that will be set in step 1, so that Function creation code has data to differentiate ephemeral vs. new deployment.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -20,6 +20,9 @@ enum AppState {
 
   // Stopped
   APP_STATE_STOPPED = 5;
+
+  // App is created and in process of deployment.
+  APP_STATE_DEPLOYING = 6;
 }
 
 enum ClientType {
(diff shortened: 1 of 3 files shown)


erikbern (approved):
I think this makes sense!

modal_proto/api.proto
```diff
@@ -20,6 +20,9 @@ enum AppState {
 
   // Stopped
   APP_STATE_STOPPED = 5;
+
+  // App is created and in process of deployment.
+  APP_STATE_DEPLOYING = 6;
```
erikbern, line 25: just so I understand it, this is only the first time you deploy an _new_ app right? when you're deploying to an existing app, it won't go to this state?

thundergolfer: Yeh that's the intention. Just need it to differentiate a new deploying app. Deployed apps are existing, and already in `DEPLOYED` state.

PR #108, opened 2022-12-03
fix: stub.serve() shutdown

thundergolfer:
`stub.serve()` is currently unable to `await app.disconnect()` and stop an app, because `async_utils.on_shutdown(AioClient.stop_env_client())` runs before the disconnect call and disconnects the `client`. (it's a race that the `disconnect()` never wins)

I tried a couple dozen things, but can't get around the fact that once <kbd>Ctrl</kbd> + <kbd>C</kbd> has initiated the async loop's shutdown, you can't control the ordering of the app disconnection and the client stop.

The proper fix is probably to rewrite all `client` usage to be done by context manager, but that's much more involved than this stop-gap I've done here. 

----

**Ref:** Shutdown hook is from https://github.com/modal-labs/modal/pull/2957.

### DIFF
--- a/client_test/stub_test.py
+++ b/client_test/stub_test.py
@@ -168,6 +168,7 @@ def test_serve(client):
     stub.serve(client=client, timeout=1)
 
 
+@skip_in_github
 def test_serve_teardown(client, servicer):
     stub = Stub()
     with modal.client.Client(servicer.remote_addr, api_pb2.CLIENT_TYPE_CLIENT, ("foo-id", "foo-secret")) as client:
--- a/client_test/stub_test.py
+++ b/client_test/stub_test.py
@@ -2,12 +2,15 @@
 import logging
 import os
 import pytest
+import sys
 
 from grpclib import GRPCError, Status
 
+import modal.app
 from modal import Stub
 from modal.aio import AioApp, AioQueue, AioStub, aio_lookup
 from modal.exception import InvalidError, NotFoundError
+from modal_proto import api_pb2
 from modal_test_support import module_1, module_2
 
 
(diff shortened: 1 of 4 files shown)


erikbern:
This doesn't seem 100% obvious to me – what if someone wants to call `stub.serve` from within their code? They might not expect the client to get torn down when stub.serve finishes. Maybe it's a weird case.

thundergolfer:
I think that's a weird case. The `.serve()` blocks and doesn't return, so it only really finishes when a termination is sent, such as SIGTERM or SIGINT or SIGKILL.

erikbern:
Would prefer to solve it in a nicer way – fine to merge this but let's at least add some comments explaining what's the purpose of this code. Would also prefer to add a test.

erikbern:
If it's hard to add the test, it's not the end of the world to skip it. Would be great to add some comments though!

client_test/stub_test.py
```diff
@@ -166,6 +169,22 @@ def test_serve(client):
...
+    sys.version.startswith("3.7"),
+    reason="AsyncMock only introduced in Python 3.8",
+)
+def test_serve_teardown(client, servicer):
+    from unittest.mock import AsyncMock, patch
+
+    stub = Stub()
+    fake_disconnect = AsyncMock()
+    with modal.client.Client(servicer.remote_addr, api_pb2.CLIENT_TYPE_CLIENT, ("foo-id", "foo-secret")) as client:
+        with patch.object(modal.app._App, "disconnect", fake_disconnect):
```
erikbern, line 182: Awesome with test!

Instead of mocking the app, you could probably just see if the `AppDisconnect` call was called [on the servicer fixture](https://github.com/modal-labs/modal-client/blob/main/client_test/conftest.py#L110)

thundergolfer: Cheers, I'll try that 👍

thundergolfer: A mock-y test that runs in CI is probably better than a non-mock test that doesn't run in CI, so I'll go back to that.

thundergolfer: Damn, the mock is flaky too.

freider:
Attempt at removing this here: https://github.com/modal-labs/modal-client/pull/108

PR #111, opened 2022-12-06
Update client proto to prepare for image build context mount addition

freider:


aksh-at (approved):
Cool! As you suggested on Slack, we might want to mount a subset of the build context (in case of `COPY`s) and not the entire directory. But that still works with this interface.

PR #125, opened 2022-12-13
client should be forward-compatible with new ProgressType

thundergolfer:
Just realized that old-clients aren't forward compatible with new progress types, they'll raise an exception. Fix this for clients  going forward. Will announce the function queuing progress logs in beta channel before releasing the feature server-side, as A100 users will need to upgrade their client.


erikbern:
Nice! Forward compatibility is tricky

PR #129, opened 2022-12-15
Adds basic copy support to Image

freider:
Lets users copy data from a Mount into an image during image build.
Either using the convenience `copy` method on Image, or using the more low level `context_mount=..., dockerfile_commands=["COPY ..."]` options for `.extend()`

Next step is to improve the semantics of creating Mounts to make mount composition easier than it is today.

This is what the new functionality looks like:
```python
mount = Mount(remote_dir="/", local_dir=Path(__file__).parent / "static")
Image.debian_slim().copy(mount, remote_path="/static")
```

### DIFF
--- a/client_test/image_test.py
+++ b/client_test/image_test.py
@@ -189,3 +189,24 @@ def test_poetry(client, servicer):
         layers = get_image_layers(running_app["image"].object_id, servicer)
         context_files = {f.filename for layer in layers for f in layer.context_files}
         assert context_files == {"/.poetry.lock", "/.pyproject.toml", "/modal_requirements.txt"}
+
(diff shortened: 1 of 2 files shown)


erikbern:
it's a bit confusing that there's both `remote_dir` and `remote_path` – I guess we can remove one of them at some point?

erikbern (approved):

freider:
> it's a bit confusing that there's both `remote_dir` and `remote_path` – I guess we can remove one of them at some point?

Yeah the remote_dir of mount is superfluous and doesn't do much. It will make more sense after refactoring the mount interface

PR #133, opened 2022-12-17
Graceful handling for KeyboardInterrupts with async functions

aksh-at:
We weren't canceling the task inside `asyncio.run` on receiving `KeyboardInterrupt`s. Created a bunch of log vomit for async functions (including all webhooks). 

Separately also handles a `CancelledError` that's raised in `_asgi` wrapper from the task being cancelled.

### Before 

(async function, not a webhook)

image

### After

image


erikbern:
oh interesting – and nice fix!

PR #137, opened 2022-12-19
Accept string values for `gpu` param

aksh-at:
Small QoL thing for GPU specification. We're keeping `GPUConfig` objects around, to allow GPU configuration beyond type (e.g. count or memory). But for most users typing out this config is onerous and hard to remember, so `gpu` now also accepts strings.

Adds a deprecation warning for old `gpu=True`, which should now be replaced in favor of `gpu='any'`.

### Usage

```python
@stub.function(gpu="any")
def f1():
    ...

@stub.function(gpu="A100")
def f2():
    ...

```


modal/gpu.py
```diff
@@ -30,3 +34,34 @@ def __init__(self, *, count: int = 1):
...
+        return value._to_proto()
+    elif isinstance(value, str):
+        if value.lower() not in STRING_TO_GPU_CONFIG:
+            raise InvalidError(
+                f"Invalid GPU type: {value}. Value must be one of {list(STRING_TO_GPU_CONFIG.keys())} (case-insensitive)."
+            )
+        return STRING_TO_GPU_CONFIG[value.lower()]._to_proto()
+    elif value is True:
+        deprecation_warning(
+            date(2022, 12, 19), 'Setting gpu=True is deprecated. Use `gpu="any"` or `gpu=modal.gpu.Any()` instead.'
```
erikbern, line 61: nice

PR #139, opened 2022-12-19
Fix ASGI lifecycles (WIP)

erikbern:
WIP – so far this just adds a failing test for lifecycles with ASGI apps.

The reason this fails is that the asgi-decorated function runs before the `__enter__` method is called, which I think is the opposite of what people would expect.

I'll think about how to fix it. It feels like the interface is already a bit awkward since asgi apps are already objects that have their own lifecycle. So maybe we should come up with something more elegant that uses that.

### DIFF
--- a/client_test/container_test.py
+++ b/client_test/container_test.py
@@ -330,6 +330,10 @@ def test_webhook_lifecycle(unix_servicer, event_loop):
     second_message = deserialize(items[1].result.data, client)
     assert json.loads(second_message["body"]) == {"hello": "space"}
 
+    cls = sys.modules["modal_test_support.functions"].WebhookLifecycleClass
+
+    assert cls._events == ["init", "enter", "call", "exit"]
+
 
 @skip_windows
 def test_serialized_function(unix_servicer, event_loop):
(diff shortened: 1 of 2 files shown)


aksh-at:
> It feels like the interface is already a bit awkward since asgi apps are already objects that have their own lifecycle. 

Yeah I'm not sure an enter hook is actually needed anymore. We can use the function that's decorated with @stub.asgi or @stub.wsgi (and actually returns the ASGI/WSGI app) to perform initialization before returning the app. 

FastAPI has [startup/shutdown events](https://fastapi.tiangolo.com/advanced/events/) too. The shutdown event should in theory be triggered correctly when the Modal container receives SIGINT, but should confirm this.

erikbern:
That makes sense. I also wonder if we should allow for calling stub.asgi with an asgi object rather than a closure returning an asgi handler. That way you can just do stub.asgi(web_app) in global scope.

whoiskatrin:
@erikbern @aksh-at, what about dependency injection (fastAPI)? What if we decouple lifecycle mgmt from the ASGI-decorated function and the '__enter__' completely?

erikbern:
> @erikbern @aksh-at, what about dependency injection (fastAPI)? What if we decouple lifecycle mgmt from the ASGI-decorated function and the '**enter**' completely?

What do you mean? A lot of use cases like model inference with costly initialization requires that the user can hook into the container lifecycle

whoiskatrin:
> > @erikbern @aksh-at, what about dependency injection (fastAPI)? What if we decouple lifecycle mgmt from the ASGI-decorated function and the '**enter**' completely?
> 
> 
> 
> What do you mean? A lot of use cases like model inference with costly initialization requires that the user can hook into the container lifecycle

My bad, "startup" and "shutdown" event handler seems to be the only way then

erikbern:
> My bad, "startup" and "shutdown" event handler seems to be the only way then

Are you talking about something like this? https://fastapi.tiangolo.com/advanced/events/

I think that would be nice to support! Not sure if there's something equivalent in all asgi/wsgi frameworks though

whoiskatrin:
> https://fastapi.tiangolo.com/advanced/events/

Yes, this is precisely it. Worst comes to worst - custom ones, but I think Fast, Starlette, and Quart has a similar approach, and I think Flask has decorators that can do similar things.

erikbern:
Cool will take a look at it. It would be great if this is a part of the asgi/wsgi standard and we can trigger using the unified interface

erikbern:
Will take another stab at this at some point!

PR #140, opened 2022-12-20
Local entrypoints + reworked CLI

freider:
Notably:
* Changes `modal app run` to `modal run`
* Adds `stub.local_entrypoint` that can be called via `modal run` so people don't have to do `if __name__ == __main__` everywhere. Conveniently works with async functions as well
* Change `modal run` to take the function name as an argument rather than an option (less typing)
* Add (somewhat experimental) support for parsing CLI arguments and passing them to functions using typer. Function arguments without type annotations are parsed as strings.  

--- 

It was kind of tricky to get typer to dynamically insert function arguments of something determined by previous arguments in the cli invocation.

In the end I opted for a pretty simple, albeit kind of hacky, solution:
Allow any arguments to the top level `modal run` and make a call to a new typer instance representing the selected function. Not optimal and we might need to rework this.
Another disadvantage with using typer for the arg parsing is that all of the arguments become *arguments* rather than *options*, which feels like a bit too many arguments to the command IMO
 

TODO:
- [x] Tests
- [x] Move to `modal run`
- [x] Support case of not specifying any function name (if there is a single function, or single local entrypoint)
- [x] Build up the click command for Modal functions ourselves instead of relying on Typer, since typer makes everything into positional arguments and has no built-in support for many of the types we might want to support (like dates)


modal/stub.py
```diff
@@ -508,6 +510,35 @@ def _add_function(self, function: _Function) -> _FunctionHandle:
...
+        def main():
+            some_modal_function()
+
+        To run the function, either call it directly from code:
+        ```python
+        main()  # needs no stub.run() context
+        ```
+        Or from the CLI:
+        ```
+        modal app run stub_module.py::main
```
erikbern, line 535: you don't need `:main` if there's just one entrypoint right?

erikbern:
random thought but i wonder if we should use `modal run` and `modal deploy` instead of `modal app run` and `modal app deploy`? marginally shorter

freider:
> 

Yeah I think that makes sense. We can keep the `modal app` command for interacting with running apps, e.g. modal app list

modal/stub.py
freider: Good point, I need to fix that!

freider: And the syntax here needs to be updated, since it's `modal run file_or_module[::stub_name] [function_name]` where the `[]` are optional

PR #142, opened 2022-12-21
remove deprecated fields

erikbern:
Super minor housekeeping

```
% git blame modal_proto/api.proto | grep deprecated
ac0eed18 modal_proto/api.proto       (Erik Bernhardsson 2022-08-27 22:47:33 +0200 140)   map<string, string> object_ids = 1 [deprecated=true];  // remove once support for 0.0.22 is dropped
0f7e4fe9 modal_proto/api.proto       (Erik Bernhardsson 2022-10-12 18:52:31 -0400 221)   int32 num_connections = 2 [deprecated=true];
05aac794 modal_proto/proto/api.proto (Akshat Bubna      2022-03-14 13:56:16 -0400 309)   string secret_id = 5 [deprecated=true];
df0bedf0 modal_proto/api.proto       (Akshat Bubna      2022-10-21 14:23:37 -0400 373)   float timeout = 4 [deprecated=true];
ccfc58e3 modal_proto/api.proto       (Elias Freider     2022-10-13 13:51:26 +0200 393)   bool return_empty_on_timeout = 5;  // deprecated - remove in followup PR
6b2daad0 modal_proto/api.proto       (Richard Gong      2022-10-10 14:50:36 -0400 455)   string task_id = 5 [deprecated=true];
5a70dd42 modal_proto/api.proto       (Akshat Bubna      2022-10-17 00:13:33 -0400 473)   string root_function_call_id = 1 [deprecated=true];
fb41811e modal_proto/api.proto       (Akshat Bubna      2022-10-18 14:27:44 -0400 621)   bool block = 2 [deprecated=true];
fb41811e modal_proto/api.proto       (Akshat Bubna      2022-10-18 14:27:44 -0400 624)   string idempotency_key = 5 [deprecated=true];
fb41811e modal_proto/api.proto       (Akshat Bubna      2022-10-18 14:27:44 -0400 633)   string idempotency_key = 3 [deprecated=true];
```

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -137,7 +137,7 @@ message AppGetObjectsItem {
 }
 
 message AppGetObjectsResponse {
-  map<string, string> object_ids = 1 [deprecated=true];  // remove once support for 0.0.22 is dropped
+  map<string, string> object_ids = 1 [deprecated=true]; // remove once support for 0.43.0 is dropped
   repeated AppGetObjectsItem items = 2;
 }
 
@@ -218,7 +218,6 @@ message ClientCreateResponse {
 
 message ClientHeartbeatRequest {
   string client_id = 1;
-  int32 num_connections = 2 [deprecated=true];
   string current_input_id = 3;
   double current_input_started_at = 4;
 }
@@ -306,7 +305,6 @@ message Function {
   string function_name = 2;
   repeated string mount_ids = 3;
   string image_id = 4;
-  string secret_id = 5 [deprecated=true];
   bytes function_serialized = 6;
 
   enum DefinitionType {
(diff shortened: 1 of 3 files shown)


erikbern:
update: realized we actually send both `object_ids` and `items` from the server for AppGetObjects and that clients still consume the former despite it being deprecated it. I updated it so we consume the latter but we'll have to keep sending the old one for another few months

modal_proto/api.proto
```diff
@@ -618,10 +612,8 @@ message QueueCreateResponse {
 
 message QueueGetRequest {
   string queue_id = 1;
-  bool block = 2 [deprecated=true];
   float timeout = 3;
   int32 n_values = 4;
-  string idempotency_key = 5 [deprecated=true];
```
freider, line 624: We should either keep it or add some kind of reservation dummy for the field id since it's the last one on the request and might get reused otherwise

freider (commented):
Minor thing, otherwise lgtm

modal_proto/api.proto
erikbern: yeah i thought about that, but this field hasn't been used for an extremely long time so I think it's fine?

PR #143, opened 2022-12-21
Fix client requirements for Python 3.11

aksh-at:
Realized aiohttp 3.8.3 actually does support Python 3.11. `ddtrace` (used for distributed tracing) isn't there yet, but it's an optional dependency and Modal can function without it.


erikbern:
dope

PR #149, opened 2022-12-23
Take args list for image build functions

ekzhang:
### Motivation

When building large images from Python I often have to write things like (this one taken from youtube face detection example):

```python
image = (
    modal.Image.debian_slim()
    .run_commands(
        [
            "apt-get install -y libgl1-mesa-glx libglib2.0-0 wget",
            f"wget https://raw.githubusercontent.com/opencv/opencv/master/data/haarcascades/{FACE_CASCADE_FN} -P /root",
        ]
    )
    .pip_install(["pytube", "opencv-python", "moviepy"])
)
```

These brackets aren't really necessary; they don't seem to help make the API clearer.

```python
image = (
    modal.Image.debian_slim()
    .run_commands(
        "apt-get install -y libgl1-mesa-glx libglib2.0-0 wget",
        f"wget https://raw.githubusercontent.com/opencv/opencv/master/data/haarcascades/{FACE_CASCADE_FN} -P /root",
    )
    .pip_install("pytube", "opencv-python", "moviepy")
)
```

If we remove them, we also get to eliminate another level of indentation from Black-formatted code.

### Description

Changes the following functions that build images:

- `apt_install`
- `pip_install`
- `run_commands`
- `conda_install`

The specific change is to the function signature, which is modified from

```python
def pip_install(
    self,
    packages: list[str],  # A list of Python packages, eg. ["numpy", "matplotlib>=3.5.0"]
    find_links: Optional[str] = None,
) -> "_Image":
```

to a new signature

```python
def pip_install(
    self,
    *packages: Union[str, list[str]],  # A list of Python packages, eg. ["numpy", "matplotlib>=3.5.0"]
    find_links: Optional[str] = None,
) -> "_Image":
```

We could gradually deprecate `list[str]` being accepted far in the future (e.g., by showing warnings), but I don't think it's important immediately. Just changing all the docs should be good enough.

### Compatibility

We still support code like `image.pip_install(["pip"])` due to the `Union[]` type and some flattening at runtime.

**However, this PR introduces a limited breaking change because it turns a positional argument into a keyword argument.**

This breaking change only affects `pip_install()` and `run_commands()`. Other functions already don't take extra positional arguments. In practice I think `pip_install()` will be the only function that causes real-world breakage because users are probably passing `secrets=[...]` to `run_commands`, not using the argument as positional.

I think this is minor enough that it's fine to just bump the release number though. The error happens immediately, and the fix is really simple.

```python
# broken
image.pip_install(["torch==1.12.1+cu116"], "https://download.pytorch.org/whl/torch_stable.html")

# fix
image.pip_install(["torch==1.12.1+cu116"], find_links="https://download.pytorch.org/whl/torch_stable.html")

# or, even better
image.pip_install("torch==1.12.1+cu116", find_links="https://download.pytorch.org/whl/torch_stable.html")
```

We could send a message in the beta testers Slack about this to make it clear.

(internal note: I need to add `find_links=` to two places in our monorepo's tests before we merge this)

### DIFF
--- a/client_test/image_test.py
+++ b/client_test/image_test.py
@@ -79,7 +79,14 @@ def test_image_requirements_txt(servicer, client):
 
 def test_empty_install(servicer, client):
     # Install functions with no packages should be ignored.
-    stub = Stub(image=Image.debian_slim().pip_install().pip_install([], []).apt_install([]))
+    stub = Stub(
+        image=Image.debian_slim()
+        .pip_install()
+        .pip_install([], [], [], [])
+        .apt_install([])
+        .run_commands()
+        .conda_install()
+    )
 
     with stub.run(client=client) as running_app:
         layers = get_image_layers(running_app["image"].object_id, servicer)
--- a/client_test/image_test.py
+++ b/client_test/image_test.py
@@ -44,7 +44,7 @@ def get_image_layers(image_id: str, servicer) -> List[api_pb2.Image]:
 
 def test_image_python_packages(client, servicer):
     stub = Stub()
-    stub["image"] = Image.debian_slim().pip_install(["numpy"])
+    stub["image"] = Image.debian_slim().pip_install("numpy")
     with stub.run(client=client) as running_app:
         layers = get_image_layers(running_app["image"].object_id, servicer)
         assert any("pip install numpy" in cmd for cmd in layers[0].dockerfile_commands)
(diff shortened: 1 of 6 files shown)


aksh-at (approved):
Awesome, this is a definite improvement. Unfortunate about the positional arguments, but they probably should have been keyword-only in hindsight anyway.

PR #151, opened 2022-12-23
Don't use synchronicity as a wrapper for cli commands that run user code

freider:
Uses _translate_in and _translate_out hacks instead to reliably run both async and sync versions of imported code. There might be a better way?

### DIFF
--- a/client_test/cli_test.py
+++ b/client_test/cli_test.py
@@ -120,3 +120,4 @@ def test_app_run_custom_stub(servicer, server_url_env, test_dir):
 def test_app_run_aiostub(servicer, server_url_env, test_dir):
     modal_file = test_dir / "supports" / "app_run_tests" / "async_stub.py"
     _run(["app", "run", modal_file.as_posix()])
+    assert len(servicer.client_calls) == 1
--- a/modal/cli/app.py
+++ b/modal/cli/app.py
@@ -13,6 +13,7 @@ from rich.panel import Panel
 from rich.prompt import Prompt
 from rich.table import Table
 from rich.tree import Tree
+from synchronicity import Interface
 
 from modal._output import OutputManager, step_progress
 from modal.cli.utils import timestamp_to_local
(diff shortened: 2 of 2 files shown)


client_test/cli_test.py
```diff
@@ -120,3 +120,4 @@ def test_app_run_custom_stub(servicer, server_url_env, test_dir):
 def test_app_run_aiostub(servicer, server_url_env, test_dir):
     modal_file = test_dir / "supports" / "app_run_tests" / "async_stub.py"
     _run(["app", "run", modal_file.as_posix()])
+    assert len(servicer.client_calls) == 1
```
freider, line 123: This checks that the .call() actually triggers something and doesn't cause an unawaited coroutine or something

modal/cli/app.py
```diff
@@ -63,10 +63,11 @@ async def run(
             f"No function `{function_name}` could be found in the specified stub. Registered functions are: {registered_functions_str}"
         )
         exit(1)
-
-    async with _stub.run(detach=detach) as app:
+    blocking_stub = synchronizer._translate_out(_stub, Interface.BLOCKING)
+    with blocking_stub.run(detach=detach) as app:
         func_handle = getattr(app, function_name)
-        await func_handle.call()
+        blocking_function = synchronizer._translate_out(synchronizer._translate_in(func_handle), Interface.BLOCKING)
```
freider, line 69: Not sure if we have to do the translation stuff for the function handle as well or if the blocking stub is smart enough to always return blocking version of FunctionHandle when doing getattr? @erikbern

erikbern: i think this looks right!

modal/cli/app.py
```diff
@@ -63,10 +63,11 @@ async def run(
...
+        blocking_function.call()
```
erikbern, line 70: does this work for a local function too?

erikbern (approved):
nice! i'm realizing we should make this stuff a bit nicer in synchronicity since all the translate stuff looks pretty ugly

we should probably also add a unit test for this (basically assert that the function isn't running inside the synchronizer event loop) but it's not urgent – can do that later

PR #153, opened 2022-12-27
Test that run and deploy commands don't run user code in synchronicity loop

freider:
…

### DIFF
--- a/client_test/cli_test.py
+++ b/client_test/cli_test.py
@@ -1,5 +1,7 @@
 # Copyright Modal Labs 2022
 import os
+import pytest
+import sys
 import traceback
 
 import pytest_asyncio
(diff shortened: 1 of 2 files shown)


client_test/cli_test.py
```diff
@@ -121,3 +123,26 @@ def test_app_run_aiostub(servicer, server_url_env, test_dir):
...
+def fresh_main_thread_assertion_module(test_dir):
+    modules_to_unload = [n for n in sys.modules.keys() if "main_thread_assertion" in n]
+    assert len(modules_to_unload) <= 1
+    for mod in modules_to_unload:
+        sys.modules.pop(mod)
+    yield test_dir / "supports" / "app_run_tests" / "main_thread_assertion.py"
+
+
+def test_no_user_code_in_synchronicity_run(servicer, server_url_env, test_dir, fresh_main_thread_assertion_module):
+    pytest._did_load_main_thread_assertion = False
```
freider, line 138: kind of hacky - used the pytest module as a namespace for a module load side effect to make sure that the module is loaded on each test and not cached, so we know the tests actually ran...

erikbern: I think you can also do something like this

https://github.com/modal-labs/modal-client/blob/main/modal_utils/package_utils.py#L86

But just mentioning it as an FYI – this works too.

PR #154, opened 2022-12-28
doc updates + fixes, mainly following up recent client API changes

thundergolfer:
Follows up:

* https://github.com/modal-labs/modal-client/pull/97
* https://github.com/modal-labs/modal-client/pull/104

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -62,6 +62,7 @@ from .shared_volume import _SharedVolume
 
 
 def exc_with_hints(exc: BaseException):
+    """mdmd:hidden"""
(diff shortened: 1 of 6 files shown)


modal/functions.py
```diff
@@ -644,6 +646,7 @@ class CloudProvider(Enum):
 
 
 def parse_cloud_provider(value: str, gpu_config: api_pb2.GPUConfig) -> "api_pb2.CloudProvider.V":
+    """mdmd:hidden"""
```
thundergolfer, line 649: Perhaps can be `_parse_cloud_provider` private? cc @aksh-at

aksh-at: ah yes good catch, it should be hidden

PR #155, opened 2022-12-28
Send client version on every request, move auth_metadata into modal

erikbern:
Minor prep work for changing the client connection logic. I'm going to get rid of the `Client` class eventually. This means we have to send the client version with every request. I'm making that change now already, so that we can rely on the version field in the server sooner. Relying on this to be set requires deprecating every client up until that version, so we want to start sending the version asap.

Also ended up moving `auth_metadata` out of `modal_utils` since it has modal-specific logic.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -1,4 +1,6 @@
 # Copyright Modal Labs 2022
+from __future__ import annotations
+
 import asyncio
 import platform
 import time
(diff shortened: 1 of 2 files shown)


erikbern:
merging this since it's fairly harmless (it's really just a one liner other than the refactor)

PR #158, opened 2022-12-29
Adds proto fields for specifying new async modes for webhooks

freider:


modal_proto/api.proto
```diff
@@ -72,6 +72,14 @@ enum WebhookType {
   WEBHOOK_TYPE_WSGI_APP = 3;
 }
 
+enum WebhookAsyncMode {
+  WEBHOOK_ASYNC_MODE_UNSPECIFIED = 0;
+  WEBHOOK_ASYNC_MODE_REDIRECT = 1;  // redirect immediately to result polling endpoint
+  WEBHOOK_ASYNC_MODE_DISABLED = 2; // no redirect, fail after timeout
+  WEBHOOK_ASYNC_MODE_TRIGGER = 3; // return immediately, same as old wait_for_response=False
```
freider, line 79: open to other name suggestions for this one... not that it matters too much

aksh-at (approved):

modal_proto/api.proto
```diff
@@ -72,6 +72,14 @@ enum WebhookType {
...
+  WEBHOOK_ASYNC_MODE_AUTO = 4; // redirect to polling endpoint if execution time nears the timeout
```
ekzhang, line 80: Could you give a code example of how to call a webhook in this async mode?

modal_proto/api.proto
```diff
@@ -72,6 +72,14 @@ enum WebhookType {
   WEBHOOK_TYPE_WSGI_APP = 3;
 }
 
+enum WebhookAsyncMode {
+  WEBHOOK_ASYNC_MODE_UNSPECIFIED = 0;
+  WEBHOOK_ASYNC_MODE_REDIRECT = 1;  // redirect immediately to result polling endpoint
+  WEBHOOK_ASYNC_MODE_DISABLED = 2; // no redirect, fail after timeout
```
ekzhang, line 78: If `DISABLED` is the default maybe it could be merged with `UNSPECIFIED = 0`, but I'm not sure about protobuf enum convention.

modal_proto/api.proto
freider: Calling the webhook is the same regardless of modes (just an http call to the webhook url, which doesn't change based on this mode), it's just the logic for when/how to return results that's different.

When AUTO is used, it will return the result the same way as today if the input finishes within the webhook timeout (150s) - otherwise it will return a redirect to the polling endpoint (the same url with added `__modal_function_call_id=fc-asdfasdf` query parameter) at that time.
Backend PR will have more info :)

modal_proto/api.proto
freider: It could, but I think our convention is to not use the unspecified/0-value for anything "meaningful". Good if we want to distinguish functions deployed before this change took place from those deployed after.

modal_proto/api.proto
ekzhang: I see, so the pseudocode would look something like:

```js
const resp = await fetch(webhookUrl);
if (resp.redirected) {
  const { redirectUrl } = await resp.json();
  while (true) {
    const resp = await fetch(redirectUrl);
    if (resp.status === 200) {
      return await resp.json();
    }
  }
} else {
  return await resp.json();
}
```

It seems like a rather complicated API to present to users. I wonder if they could get the same outcome by just redirecting immediately?

Does the polling URL block for a certain number of ms? That would produce the same outcome, I think.

I also am not sure if you need to _redirect_ necessarily so much as return the URL in the response body. Not all language HTTP clients make it easy to see what URL you were redirected to from a response, they just handle the 30x responses transparently.

freider: `ASYNC_MODE_ENABLED` would redirect immediately and is quite useful in many long-polling cases, but depending on what clients people use they might not want that. The redirect would be a 302 so most clients (like web browsers) would handle the redirect by following it automatically.

The idea is that the AUTO mode would just block most quick running requests to an asgi app until they finish (no redirect), but if there are specific paths in the app that take a longer time to finish those would redirect to the "polling" endpoint and hopefully finish within a few redirect passes.

The polling endpoint itself will not respond immediately but also wait 150s and do another redirect pass if it still hasn't finished (perhaps a maximum of ~10 times to prevent it from running into the browser max redirect limit?).

I can definitely see a point in having a way for clients to request the polling url directly instead of relying on 302 redirects in case you want to implement the polling "in code" instead - perhaps we could add a query parameter to trigger that behavior regardless of webhook mode? (unless it's async_disabled).

ekzhang: > The polling endpoint itself will not respond immediately but also wait 150s and do another redirect pass if it still hasn't finished (perhaps a maximum of ~10 times to prevent it from running into the browser max redirect limit?).

Thanks for the detailed explanation, that was really helpful.

I don't think it's ideal behavior for long-running responses to be handled by repeatedly sending 302 responses to the same URL. It will run into the redirect limit, as you mention, but also, any network issue in the middle of the webhook request would cause it to result in unrecoverable failure.

For a polling-mode job, I mean something like the following, supposing this takes 15 seconds:

```
> POST https://foo.modal.run/
{
  "question": "What is the height of the Eiffel Tower?"
}

Response: 202 Accepted
{
  "message": "Webhook invocation created",
  "url": "https://foo.modal.run/_call/fc-1eb2066dbdbe46"
}

> GET https://foo.modal.run/_call/fc-1eb2066dbdbe46

Response: 202 Accepted (not finished yet, same response)
{
  "message": "Webhook invocation created",
  "url": "https://foo.modal.run/_call/fc-1eb2066dbdbe46"
}

> GET https://foo.modal.run/_call/fc-1eb2066dbdbe46?wait=4000

(after 4 seconds)
Response: 204 No Content

> GET https://foo.modal.run/_call/fc-1eb2066dbdbe46?wait=10000000

Response: 400 Bad Request
(some error about wait time being higher than 150s)

> GET https://foo.modal.run/_call/fc-1eb2066dbdbe46?wait=40000

(after 11 more seconds)
Response: 200 OK
{
  "data": "The Eiffel Tower is 324 meters tall."
}
```

For a flow like this, when you manually open the webhook URL in their browser, you get a response URL that's tied to the current response. And you can refresh when it's finished, regardless of how long the job takes or browser redirect limits.

modal_proto/api.proto
ekzhang: Thanks, makes sense!

modal_proto/api.proto
erikbern: I think we should probably support both modes and it should be up to the caller to specify which one they want. I like that this PR enables long webhooks without requiring anything client side. Seems worth having a more "explicit" API at some point too.

ekzhang: It's a hack to repeatedly redirect the user but probably works well for _most cases_ given that the redirect limit is 20x. That sounds compelling to me.

I would still be confused that this workaround for network is called async_mode though. It doesn't change how the function is invoked and its result is awaited, which is why we're doing it. But when I read async_mode (before clarifications in this thread) it sounded like it would change how you call the function.

freider: I'm open to other name suggestions? At this point it's still not exposed in the client SDK, only the proto spec, so we can still rename it without affecting users. Merging now to unblock the server work

ekzhang: I might just be crazy 🤷‍♂️ if no one else finds it confusing or weird then it's fine

PR #159, opened 2022-12-29
Improve error messages for image syntax issues

ekzhang:
The original program that produced a confusing `AssertionError` looked like this:

```python
stub = modal.Stub(
    "example-import-torch",
    image=modal.Image.debian_slim().run_commands(["torch~=1.12.1"], secrets="https://download.pytorch.org/whl/cu116"),
)
```

I added some more type checks to the client code.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -107,11 +107,11 @@ class _Image(Provider[_ImageHandle]):
         base_images={},
         context_files={},
         dockerfile_commands: Union[list[str], Callable[[], list[str]]] = [],
-        secrets=[],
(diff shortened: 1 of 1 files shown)


erikbern:
Nice! Wonder if we should just use something like this at some point:

* https://github.com/FelixTheC/strongtyping
* https://typeguard.readthedocs.io/en/latest/userguide.html

ekzhang:
That sounds good! I can make a note to try it out later.

(This PR is linked to a specific ticket, by the way.)

PR #161, opened 2023-01-02
Let the container control its own heartbeats

erikbern:
This removes the use of `ClientHeartbeat` for containers, which now have their own heartbeat loop in `_container_entrypoint.py`

It will make it possible to clean up some of the global state where we set the current input id etc – I'll do it in a subsequent PR just to keep it simple and focused.

The other use of client heartbeats is for apps, so that will be its own PR as well.

### DIFF
--- a/client_test/client_test.py
+++ b/client_test/client_test.py
@@ -32,11 +32,9 @@ async def test_client(servicer, client):
 @pytest.mark.asyncio
 @skip_windows
 async def test_container_client(unix_servicer, aio_container_client):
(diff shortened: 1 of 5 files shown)


erikbern:
I have been testing this against prod using `MODAL_SYNC_ENTRYPOINT=1` and it's working with the examples. I can see `ContainerHeartbeat` requests on the server side (actually – just found a bug while doing this. Fixing it!)

freider (approved):
makes sense!

PR #165, opened 2023-01-02
[mod-747] webhook reload updates Modal Fn definitions on change

thundergolfer:
'interpreter surgery' based fix for `stub.serve` reloading, which currently doesn't reload providers on change. 

* don't have enough confidence that this approach is sound, or even practical, so holding off merge.
* but this is working fine when I test out basic webhooks, adding and editing function providers.
* the implementation has a couple of obvious optimizations that aren't yet done
* opportunity to catch `SyntaxError` and abort reload of broken code [mod-727]


erikbern:
This is maybe stupid but couldn't we just recreate the app every time the source code changes?

thundergolfer:
Closing in favor of #197

PR #169, opened 2023-01-03
Fix Secret type assertion

aksh-at:
Fixing assertion added in https://github.com/modal-labs/modal-client/pull/159 (breaks when secret passed was a Ref created through `modal.Secret.from_name()`).

Will add test in a follow-up.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -17,7 +17,7 @@ from modal_utils.grpc_utils import retry_transient_errors
 from .config import config, logger
 from .exception import InvalidError, NotFoundError, RemoteError
 from .mount import _Mount
-from .object import Handle, Provider
+from .object import Handle, Provider, Ref
 from .secret import _Secret
 
 
@@ -127,7 +127,7 @@ class _Image(Provider[_ImageHandle]):
             raise InvalidError("Cannot run a build function with multiple base images!")
 
         for secret in secrets:
-            if not isinstance(secret, _Secret):
+            if not isinstance(secret, _Secret) and not isinstance(secret, Ref):
                 raise InvalidError(f"Secret {secret!r} must be a modal.Secret object")
 
         if context_mount is not None and not isinstance(context_mount, _Mount):


modal/image.py
```diff
@@ -127,7 +127,7 @@ def __init__(
             raise InvalidError("Cannot run a build function with multiple base images!")
 
         for secret in secrets:
-            if not isinstance(secret, _Secret):
+            if not isinstance(secret, _Secret) and not isinstance(secret, Ref):
                 raise InvalidError(f"Secret {secret!r} must be a modal.Secret object")
 
         if context_mount is not None and not isinstance(context_mount, _Mount):
```
luiscape, line 133: Do we need to apply the same `Ref` check here in mounts?

aksh-at: yes you're right. Let me add that in the follow-up PR with tests.

aksh-at: (although I haven't seen anyone use a `Mount` as a ref, it's technically possible)

ekzhang: I think we need to fix the type annotations on this library as well, where are `Ref`s allowed? How do we write code if no one knows what type things are

aksh-at: Yeah we might want to define another type that's the `Union` of each object type and `Ref`? Not the cleanest solution. I think @erikbern had some ideas for having `Ref`s have better runtime types too.

Looks like we have this [cast](https://github.com/modal-labs/modal-client/blob/b2574e1ec16e554d823a6af8a982182cc659724c/modal/object.py#L180-L181) that casts `RemoteRef`s to the correct object type, but only for the type checker. Runtime types are unaffected (as we saw here).

erikbern: No, let's solve this at the foundational level – the correct solution is obviously that `Secret.from_name` returns a `Secret`

I don't know why it doesn't do that? Look at the signature in https://github.com/modal-labs/modal-client/blob/main/modal/object.py#L163

aksh-at: It's because of the thing I mentioned above:

> Looks like we have this [cast](https://github.com/modal-labs/modal-client/blob/b2574e1ec16e554d823a6af8a982182cc659724c/modal/object.py#L180-L181) that casts RemoteRefs to the correct object type, but only for the type checker. Runtime types are unaffected (as we saw here).

It makes the type checker happy, but the runtime type is still `RemoteRef`.

PR #171, opened 2023-01-03
Make Client (almost) stateless

erikbern:
This PR gets rid of the client_id and `ClientCreate`. It turns the `Client` into basically just a wrapper around a GRPC connection pool, with convenience functions for accessing the singleton and for verifying the connection.

Big reduction in total complexity imo. The only "state" left to manage is that we want to close the connection pool or else grpclib prints a warning.

Have been running a bunch of examples against prod with this client and it seems to work fine.

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -290,7 +290,8 @@ class _Stub:
                 if mode == StubRunMode.SERVE:
                     # Cancel logs loop since we're going to start another one.
                     logs_loop.cancel()
-                await app.disconnect()
+                else:
+                    await app.disconnect()
 
         if mode == StubRunMode.DEPLOY:
             output_mgr.print_if_visible(step_completed("App deployed! 🎉"))
@@ -385,6 +386,7 @@ class _Stub:
                     event = await event_agen.__anext__()
         finally:
             await event_agen.aclose()
+            await app.disconnect()
 
     async def deploy(
         self,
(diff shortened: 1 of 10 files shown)


modal_proto/api.proto
```diff
@@ -113,7 +113,6 @@ message AppDeployResponse {
 message AppGetByDeploymentNameRequest {
   DeploymentNamespace namespace = 1;
   string name = 2;
-  string client_id = 3;  // todo, not strictly needed
```
aksh-at, line 116: Probably should mark this deprecated so we don't accidentally use index 3 in a future change

modal/stub.py
```diff
@@ -290,8 +290,7 @@ async def _run(
                 if mode == StubRunMode.SERVE:
                     # Cancel logs loop since we're going to start another one.
                     logs_loop.cancel()
-                else:
-                    await app.disconnect()
+                await app.disconnect()
```
aksh-at, line 293: What's the reason for this? I think `stub.serve` intentionally doesn't disconnect the app because the app will be updated in a loop

aksh-at (approved):
Much cleaner! Might also want to run integration tests in the main repo before merge.

freider (approved):

modal/stub.py
erikbern: Hm ok maybe I misunderstood the code. I thought the existing code was there because of a race with the singleton closing (see the deleted pre_stop method). Will take a second look at it.

modal_proto/api.proto
erikbern: yeah i can do that

erikbern:
> Much cleaner! Might also want to run integration tests in the main repo before merge.

sure, i can try!

PR #172, opened 2023-01-04
Add `cloud_provider` to `SharedVolumeMount` message

aksh-at:


modal_proto/api.proto
```diff
@@ -788,6 +788,7 @@ message SharedVolumeListFilesResponse {
 message SharedVolumeMount {
   string mount_path = 1;
   string shared_volume_id = 2;
+  CloudProvider cloud_provider = 3;
```
erikbern, line 791: shouldn't we add region to this too everywhere? might as well add that now

aksh-at: yeah we'd need to do that. Probably worth waiting until we know how the regions will be setup and stuff though.

PR #174, opened 2023-01-04
Map return exceptions

tmwilliamlin168:


modal/functions.py
```diff
@@ -471,7 +478,7 @@ def web_url(self) -> str:
     def is_generator(self) -> bool:
         return self._is_generator
 
-    async def _map(self, input_stream: AsyncIterable, order_outputs: bool, kwargs={}):
+    async def _map(self, input_stream: AsyncIterable, order_outputs: bool, return_exceptions: bool, kwargs={}):
```
aksh-at, line 481: Let's also add this for `starmap` (should be at parity with the `map` interface)

aksh-at (approved):
Awesome, thanks! Just a couple nitty comments on the tests + realized we should also expose this for `starmap`.

PR #178, opened 2023-01-05
Repurpose the wait_for_response flag to use async modes under the hood

freider:
* wait_for_response=True - use new async mode 'auto' which uses 302 redirects to prolong request length
* wait_for_response=False - use new trigger mode instead

The new redirect mode should allow for webhooks of up to 50 min when served to a browser (150s * 20 max redirects)

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -757,7 +757,7 @@ class _Stub:
         [guide on web endpoints](https://modal.com/docs/guide/webhooks).
 
         The two `wait_for_response` modes for webhooks are as follows:
-        * wait_for_response=True - tries to fulfill the request on the original URL, but returns a 302 rediect after ~150s to a result url (original url with an added __modal_function_id=... query parameter)
+        * wait_for_response=True - tries to fulfill the request on the original URL, but returns a 302 redirect after ~150s to a result URL (original URL with an added `__modal_function_id=fc-1234abcd` query parameter)
         * wait_for_response=False - immediately returns a 202 ACCEPTED response with a json payload: {"result_url": "..."} containing the result "redirect" url from above (which in turn redirects to itself every 150s)
         """
         if image is None:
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -74,7 +74,6 @@ enum WebhookType {
 
 enum WebhookAsyncMode {
   WEBHOOK_ASYNC_MODE_UNSPECIFIED = 0;
-  WEBHOOK_ASYNC_MODE_REDIRECT = 1;  // redirect immediately to result polling endpoint
```
thundergolfer, line 77: This was never used I guess, so no need for `reserved 1;`?

modal/stub.py
```diff
@@ -745,12 +755,22 @@ def asgi(
 
         To learn how to use Modal with popular web frameworks, see the
         [guide on web endpoints](https://modal.com/docs/guide/webhooks).
+
+        The two `wait_for_response` modes for webhooks are as follows:
+        * wait_for_response=True - tries to fulfill the request on the original URL, but returns a 302 rediect after ~150s to a result url (original url with an added __modal_function_id=... query parameter)
```
thundergolfer, line 760:
```suggestion
        * wait_for_response=True - tries to fulfill the request on the original URL, but returns a 302 redirect after ~150s to a result URL (original URL with an added `__modal_function_id=fc-1234abcd` query parameter)
```

modal/stub.py
```diff
@@ -677,12 +677,25 @@ def webhook(
 
         To learn how to use Modal with popular web frameworks, see the
         [guide on web endpoints](https://modal.com/docs/guide/webhooks).
+
+        All webhook requests have a 150s maximum request time for the HTTP request itself. However, the underlying functions can
+        run for longer.
```
thundergolfer, line 682:
```suggestion
        run for longer and return results to the caller on completion.
```

thundergolfer: The suggestion is in the direction of reassuring the reader that the function calls aren't 'orphaned' from the HTTP client in any way.

PR #180, opened 2023-01-05
env method for building images

tmwilliamlin168:


modal/image.py
```diff
@@ -650,6 +650,27 @@ def my_build_function():
         """
         return self.extend(build_function=(raw_function, kwargs))
 
+    def env(self, **vars) -> "_Image":
```
luiscape, line 653: This seems a bit of an unusual interface to me. Wouldn't it be better to have this be a `dict[str, str]`? For example:

```python
modal.Image.conda().env({"CONDA_OVERRIDE_CUDA": "11.2"})
```

A dict is a more familiar interface to me because that's what [subprocess.run](https://docs.python.org/3/library/subprocess.html) uses.

luiscape (commented):
left question re interface

modal/image.py
```diff
@@ -650,6 +650,27 @@ def my_build_function():
         """
         return self.extend(build_function=(raw_function, kwargs))
 
+    def env(self, vars: dict[str, str]) -> "_Image":
+        """Sets the environmental variables of the image.
+
+        **Example**
+
+        ```python notest
```
ekzhang, line 658: I think this could be tested, since it's useful to check if it's working or not

modal/image.py
```diff
@@ -650,6 +650,27 @@ def my_build_function():
...
+        image = (
+            modal.Image.conda()
+                .env(CONDA_OVERRIDE_CUDA="11.2")
+                .conda_install(["jax", "cuda-nvcc"], channels=["conda-forge", "nvidia"])
+                .pip_install(["dm-haiku", "optax"])
```
ekzhang, line 663: These three lines need to be updated to use the right syntax.

`.env({"CONDA_OVERRIDE_CUDA": "11.2"})`

and no brackets around the list of packages for `conda_install` and `pip_install`

modal/image.py
```diff
@@ -650,6 +650,27 @@ def my_build_function():
...
+        )
+        ```
+        """
+        # TODO environment replacement, '$' is disallowed for now
+        if any("$" in val for (key, val) in vars.items()):
+            raise ValueError("'$' not allowed in values")
```
ekzhang, line 669: This would be better detected on the worker-side rather than the client-side, since there are other lower-level image builder functions like `from_dockerfile()` that check for it. If you raise an error in the builder, it should show up in the client.

Also this error message might be improved: rather than just saying `'$' not allowed in values`, you could give more context, like `environment variable substitution is not yet implemented in definition of "{VAR_NAME_HERE}"`

ekzhang (changes requested):

modal/image.py
tmwilliamlin168: The client currently puts the values in shlex.quote() before sending it, is this an issue?

.env({"foo": "$bar"}) becomes "ENV foo='$bar'"

ekzhang: Good point! I just tried it in Docker. It seems like the quotes themselves aren't a problem, but `shlex.quote()` won't do what you want with newlines.

```dockerfile
FROM alpine
ENV FOO="foo\n bar"
RUN echo $FOO
```

Then if I get a shell into the image and run it:

```
/ # echo $FOO
foo\n bar
```

modal/image.py
```diff
@@ -650,6 +650,24 @@ def my_build_function():
...
+        image = (
+            modal.Image.conda()
+                .env({"CONDA_OVERRIDE_CUDA": "11.2"})
+                .conda_install("jax", "cuda-nvcc", channels=["conda-forge", "nvidia"])
+                .pip_install("dm-haiku", "optax")
+        )
+        ```
+        """
+        return self.extend(
+            dockerfile_commands=["FROM base"] + [f"ENV {key}={shlex.quote(val)}" for (key, val) in vars.items()]
```
ekzhang, line 668: I think if `val` has newlines, this will literally put the string `\n` (two characters: backslash, n) inside the environment variable. But it also seems uncommon, so it's probably okay. We can always fix it later since this is just in the client library.

ekzhang (approved):
Great!

PR #185, opened 2023-01-07
image: pyproject optional dependency group installation

anthonycorletti:
small change to pull in optional dependency changes. testing out a ci flow with pytest.


modal/image.py
```diff
@@ -333,6 +339,13 @@ def pip_install_from_pyproject(
         pyproject_toml = os.path.expanduser(pyproject_toml)
 
         config = toml.load(pyproject_toml)
+        if optional_dependencies:
+            optionals = config["project"]["optional-dependencies"]
+            for dep in optional_dependencies:
+                if dep in optionals:
+                    config["project"]["dependencies"].extend(
```
aksh-at, line 346: nit: although we're not using the config anywhere else, still slightly prefer making a separate list instead of mutating config

client_test/image_test.py
```diff
@@ -142,6 +142,24 @@ def test_image_pip_install_pyproject(servicer, client):
         assert any("pip install 'banana >=1.2.0' 'potato >=0.1.0'" in cmd for cmd in layers[0].dockerfile_commands)
 
 
+def test_image_pip_install_pyproject_with_optionals(servicer, client):
```
aksh-at, line 145: nice test!

.python-version
```diff
@@ -0,0 +1 @@
+3.10.8
```
aksh-at, line 1: probably don't want to commit this file to the repo

aksh-at (approved):
Awesome, thank you for contributing this! Just a couple minor comments.

modal/image.py
```diff
@@ -322,8 +322,14 @@ def pip_install_from_requirements(
     def pip_install_from_pyproject(
         self,
         pyproject_toml: str,
+        optional_dependencies: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
anthonycorletti: 👍

modal/image.py
anthonycorletti: 👍

.python-version
anthonycorletti: whoops! should have removed that one!

anthonycorletti:
thanks @aksh-at ! added those changes.

excited about this because I've been experimenting with a way to run some of my ci (just linting and pytest really) on modal and its working great so far! would be really cool if the client had a kind of test mode where the user can hook in their test framework via a stub like interface. I'm doing a bit of extra typing here to specify things but I think this is a great step in the right direction for how an app and its tests can run the same way on modal. lmk what you think!

```python
import os

import modal
import pytest

from app._env import _Env

stub = modal.Stub("ci?")


stub["env"] = modal.Secret(_Env(_env_file=".env.local").dict())
image = modal.Image.debian_slim().pip_install(["pytest"])


@stub.function(
    image=image,
    mounts=[
        modal.Mount(
            local_dir=f"{os.environ['HOME']}/modal-async-pubsub-example/app",
            remote_dir="/root/modal-async-pubsub-example/app",
        ),
        modal.Mount(
            local_dir=f"{os.environ['HOME']}/modal-async-pubsub-example/tests",
            remote_dir="/root/modal-async-pubsub-example/tests",
        ),
    ],
    secret=stub["env"],
)
def _test() -> int:
    import os

    print(os.listdir("/root"))
    return pytest.main(["/root/modal-async-pubsub-example/tests"])


if __name__ == "__main__":
    with stub.run():
        rcode = _test()
        if rcode != 0:
            raise RuntimeError(f"pytest returned {rcode}")

```

aksh-at:
That's really neat! Yeah we could have some kind of out-of-the-box pytest integration to make this easier. Would also be cool if you can run the tests in parallel easily.

PR #190, opened 2023-01-10
Write to stdout in smaller chunks

aksh-at:
Fixes interactive shell freezing when outputting data that's too large. Was happening because `.write()` on a blocking underlying buffer blocks forever, and we were trying to write a payload that was too large for some buffer somewhere in the middle. Surprisingly, larger chunk sizes are too large and run into the issue. This is despite Python claiming the buffer has size `8192`. 

Experimented with making stdout non-blocking, but it doesn't seem to work. Might come back to it.

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -355,8 +355,12 @@ class OutputManager:
                             else:
                                 # If we're not showing progress, there's no need to buffer lines,
                                 # because the progress spinner can't interfere with output.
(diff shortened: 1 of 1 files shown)


aksh-at:
Thanks @ekzhang! FYI this PR was superseded by https://github.com/modal-labs/modal-client/pull/191

PR #191, opened 2023-01-12
Handle non-blocking write failure correctly for stdout

aksh-at:
More correct solution for interactive shell write issue (supersedes https://github.com/modal-labs/modal-client/pull/190).

 It turns out that stdout was already non-blocking (because we set stdin to non-blocking using fcntl). However, writing to stdout using the `BufferedIO` wrappers doesn't correctly raise `BlockingIOError` (it does when writing to `stdout.buffer` directly, but then in that case it never reports partial writes). Using `os.write` on the fd directly fixes it.

Also printing exceptions caught in the logs loop instead of failing silently. Would have helped fix this issue faster.

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -3,6 +3,7 @@ import asyncio
 import contextlib
 import functools
 import io
+import os
 import platform
 import re
 import sys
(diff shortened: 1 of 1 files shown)


aksh-at:
Just realized this has a very stupid bug. Fixing.

modal/_output.py
```diff
@@ -355,12 +356,15 @@ async def _get_logs():
...
                                 # because the progress spinner can't interfere with output.
-                                #
-                                # Flush data in chunks because write is blocking.
-                                # TODO: make non-blocking write work.
-                                for i in range(0, len(log.data), 64):
-                                    self.stdout.write(log.data[i : i + 64])
-                                    self.stdout.flush()
+
+                                written = 0
+                                while written < len(log.data):
```
freider, line 361: It probably doesn't matter in this case since it either writes all or errors in each loop iteration, but this should probably compare to the length of the utf8-encoded log.data to be correct.

aksh-at: Good point! Actually the third case is also possible: `os.write` does sometimes write part of the data you asked it to and return how many bytes it wrote. It's going to be quite tricky to trigger that intentionally, but it's a bug nonetheless. Fixing here: https://github.com/modal-labs/modal-client/pull/195

PR #195, opened 2023-01-12
Use encoded length for `stdout.write`

aksh-at:
Should add a test case that tries to print hundreds of emojis or something.


freider (commented):
lgtm!

PR #199, opened 2023-01-13
fix: handle non-app-IDs in 'app logs' command

thundergolfer:
Following up https://modalbetatesters.slack.com/archives/C031Z7H15DG/p1673415690367319.

This doesn't add support for getting logs by app name or description, but it does do better error handling when a user gives an app name as argument.

### DIFF
--- a/modal/cli/volume.py
+++ b/modal/cli/volume.py
@@ -10,6 +10,7 @@ from tempfile import NamedTemporaryFile
 from typing import Optional, Tuple
 
 import typer
+from click import UsageError
 from google.protobuf import empty_pb2
 from grpclib import GRPCError, Status
 from rich.console import Console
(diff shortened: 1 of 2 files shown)


modal/cli/volume.py
```diff
@@ -89,8 +90,7 @@ async def ls(volume_name: str, path: str = typer.Argument(default="/")):
         entries = await volume.listdir(path)
     except GRPCError as exc:
         if exc.status in (Status.INVALID_ARGUMENT, Status.NOT_FOUND):
-            print(exc.message, file=sys.stderr)
-            return
+            raise UsageError(exc.message)
```
thundergolfer, line 93: @freider seems we could use `UsageError` in a few places where we `print` and then `exit`. Is there any reason we shouldn't?

thundergolfer:
@freider I also looked at the `tree.add(tree)` bit of this code, which looks odd. I tried refactoring it to display the tree's message but I couldn't get something working, so might be worth you looking at it.

freider:
> @freider I also looked at the `tree.add(tree)` bit of this code, which looks odd. I tried refactoring it to display the tree's message but I couldn't get something working, so might be worth you looking at it.

Yeah I have no idea why it's doing tree.add(tree), sounds unintended 😬

PR #200, opened 2023-01-13
Add and improve various error messages for `modal run`

freider:


erikbern:
Do we support direct invocation of functions without the stub? Since they are tied to a stub, I think it should be fine right? I mean something like `modal run script.py::my_function`

freider:
> Do we support direct invocation of functions without the stub? Since they are tied to a stub, I think there's no ambiguity? I mean something like `modal run script.py::my_function`

Not in this version (except using the slightly hacky `modal run script.py::.my_function` (with a period before the function name)), but with my in-progress refactor that will work!

PR #202, opened 2023-01-13
refactor: use click.UsageError instead of print+exit stdlib fns

thundergolfer:
* Exits `2` now not `1`. The former is kind of standard for 'usage error'
* Errors now look like this:

```
$ MODAL_ENV=local modal volume get xyz /idontexist
Usage: modal volume get [OPTIONS] VOLUME_NAME REMOTE_PATH [LOCAL_DESTINATION]
Try 'modal volume get --help' for help.
╭─ Error ───────────────────────────────────────────────────────────────────────────────────────────────╮
│ No such file: /idontexist.                                                                            │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

### DIFF
--- a/modal/cli/volume.py
+++ b/modal/cli/volume.py
@@ -133,11 +133,9 @@ async def put(
         remote_path = remote_path + os.path.basename(local_path)
 
     if Path(local_path).is_dir():
-        print("Directory uploads are currently not supported", file=sys.stderr)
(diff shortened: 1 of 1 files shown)


luiscape (approved):
lgtm -- someone else that is more familiar may want to take a look

PR #203, opened 2023-01-14
Always use absolute paths for `commonpath`

aksh-at:
Surprised that `__file__` and `__path__` for the same module aren't always both absolute or relative.


erikbern:
nice!

PR #204, opened 2023-01-14
Fix `FunctionHandle` init inside the container

aksh-at:
We need to rethink how Object Handles are initialized inside containers, but this fixes the proximate issue (can't call `.starmap()` etc on `container_app.my_function`)

Integration test in https://github.com/modal-labs/modal/pull/4614

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -472,6 +472,11 @@ class _FunctionHandle(Handle, type_prefix="fu"):
     @property
     def web_url(self) -> str:
         """URL of a Function running as a web endpoint."""
+        from modal import is_local
+
+        if not is_local():
+            raise InvalidError("Function.web_url is not accessible from inside a container")
+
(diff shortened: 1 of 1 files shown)


modal/functions.py
```diff
@@ -471,6 +472,11 @@ def _get_context(self) -> Tuple[_Client, str]:
     @property
     def web_url(self) -> str:
         """URL of a Function running as a web endpoint."""
+        from modal import is_local
+
+        if not is_local():
```
erikbern, line 477: we should fix this though (although it's not trivial)

aksh-at: Yeah totally. Just a quick thing to at least warn users this is not possible for now.

erikbern: Yep sounds right. I think roughly what has to happen is `AppGetObjects` needs to return protobufs containing additional stuff.

PR #205, opened 2023-01-15
Retry transient errors in more places

aksh-at:
User reported running into a `StreamTerminatedError` on one of the `modal.Dict` operations. We should probably move this logic to be more of a middleware at this point.


erikbern:
yeah agree that it would be nice as middleware but not sure how hard it is to build that – any idea?

PR #211, opened 2023-01-17
Fixes poetry lock file existence check in container global scope

freider:
Also fixes minor bug in traceback parsing


aksh-at (approved):
Awesome, thanks!

PR #212, opened 2023-01-17
Simplify the `_load` arguments by using a helper class

erikbern:
There was 6 arguments to `_load` so I decided to create a helper class for the purpose instead.

This is a pure refactoring with zero intended side effects


erikbern:
There's an annoying recursive dependency right now between some objects, which we should try to fix later. It makes type annotations quite hard, among other things.

erikbern:
Merging this to unblock me from continuing. Happy to address any comments async though!

PR #214, opened 2023-01-17
Remove all Ref classes

erikbern:
This removes the special `Ref`, `RemoteRef`, and `PersistedRef`. Going forward, `Secret.from_name("my-secret")` will return an object of type `Secret`, not `RemoteRef` as previously. This both simplifies the code a bit, and removes some runtime type inconsistencies (we had several places with checks similar to `isinstance(obj, Secret) or isinstance(obj, Ref)`.

The trick is for the `Provider` classes to store a constructor function on the instance, rather than having a per-class overloaded abstract method. This means you can create any subclass of Provider with any construction logic.

This change makes the `Provider` class pretty "dumb", and less useful on its own – any subclass of `Provider[H]` is basically just a bunch of class methods that returns `Handle` creation closures. So we can potentially remove the Provider classes later.

### DIFF
--- a/client_test/object_test.py
+++ b/client_test/object_test.py
@@ -16,6 +16,8 @@ async def test_async_factory(servicer, client):
 @pytest.mark.asyncio
 async def test_use_object(servicer, client):
     stub = AioStub()
-    stub["my_q"] = AioQueue.from_name("foo-queue")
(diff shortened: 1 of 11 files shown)


freider (approved):
good stuff, nice to get rid of the ref classes!

PR #217, opened 2023-01-18
Adds clear_on_success flag for FunctionGetOutputs

freider:
This allows the backend to set a short ttl for clearing outputs soon after they have been consumed.

For now we use the flag to clear except when we poll a future

This is a mix of what you suggested @erikbern with the "ack" solution that I suggested. For single input function calls no ack is needed, but for maps and generators the same call is used to "ack" after consuming all outputs

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -42,10 +42,10 @@ from ._blob_utils import (
 )
 from ._call_graph import InputInfo, reconstruct_call_graph
 from ._function_utils import FunctionInfo, LocalFunctionError, load_function_from_module
-from ._resolver import Resolver
 from ._location import CloudProvider, parse_cloud_provider
 from ._output import OutputManager
 from ._pty import get_pty_info
+from ._resolver import Resolver
 from ._serialization import deserialize, serialize
 from ._traceback import append_modal_tb
 from .client import _Client
@@ -197,7 +197,7 @@ class _Invocation:
             raise Exception("Could not create function call - the input queue seems to be full")
         return _Invocation(client.stub, function_call_id, client)
 
-    async def pop_function_call_outputs(self, timeout: Optional[float] = None):
+    async def pop_function_call_outputs(self, timeout: Optional[float], clear_on_success: bool):
         t0 = time.time()
         if timeout is None:
             backend_timeout = 55.0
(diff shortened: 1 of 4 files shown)


modal/functions.py
```diff
@@ -355,6 +375,20 @@ async def get_all_outputs():
                     del pending_outputs[item.input_id]
                     yield item
 
+    async def get_all_outputs_and_clean_up():
+        try:
+            async for item in get_all_outputs():
+                yield item
+        finally:
+            # "ack" that we have all outputs we are interested in and let backend clear results
+            request = api_pb2.FunctionGetOutputsRequest(
```
erikbern, line 384: This feels a bit like "boolean flag" type code smell – where the function does two very different things based on whether the flag is present or not. I think it would be cleaner if it is a separate method?

erikbern:
I think one way we could avoid the extra call would be if the client tells the server how many outputs are "left"? If it asks for the last 8 outputs, and the server returns 8 outputs, then that should be enough for the server to set an expiration on the function call, I think?

modal/functions.py
freider: It might look like it's a flag to make the call into an "ack", but this *extra* acking call is only performed for maps and generators. The flag is an "actual flag" for the single-input use case, where there is no ack-call and everything is done in the same call. Does that make sense or do you think it's still a boolean flag type code smell?
Perhaps the flag is badly named? Maybe it should be named "final_request" or something, but it could be misleading as it only has effect if an actual output is returned (if not, further requests will be allowed)

freider:
> Nice – this looks conceptually good, although (a) I have a feeling we can avoid the extra RPC call (b) left a comment that I think the current call looks a bit like boolean flag smell.
> 
> I think one way we could avoid the extra call would be if the client tells the server how many outputs are "left"? If it asks for the last 8 outputs, and the server returns 8 outputs, then that should be enough for the server to set an expiration on the function call, I think?

The extra call is only for maps and generators, and for generators we don't know how many outputs there will be. Not sure if we always do for maps either, since we could still be injecting new inputs when the first request for output happens, so there is at least a theoretical risk that we receive outputs for all inputs that have been sent, without having actually queued all inputs?

erikbern (approved):
Thanks for additional feedback! LGTM

PR #220, opened 2023-01-18
Handle some weird-ass case with inspect.signature

erikbern:
Was debugging some wacky stuff and I think this is a problem deep in Python? For some reason `__annotations__` is a dict where values are types sometimes, and strings sometimes.

I injected a `print` right after this line: https://github.com/python/cpython/blob/main/Lib/inspect.py#L241 and it seems like it's completely inconsistent.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -24,7 +24,20 @@ from modal_utils.package_utils import NoSuchStub, import_stub, parse_stub_ref
 run_cli = typer.Typer(name="run")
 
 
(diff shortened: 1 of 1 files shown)


freider:
Lgtm. Afaik the different behavior can occur depending on if `from __future__ import annotations` is used or not (see PEP-563 https://peps.python.org/pep-0563/)

freider:
I think it's possible to pass `eval_str=True` (along with optional globals and locals...) to inspect.signature, which would make it always return the referenced object instead. It's not the cleanest part of Python :)

PR #222, opened 2023-01-19
`modal serve` command line

erikbern:
It supports a `--timeout` flag which is really only useful for tests

### DIFF
--- a/client_test/cli_test.py
+++ b/client_test/cli_test.py
@@ -224,3 +224,8 @@ def test_no_user_code_in_synchronicity_deploy(servicer, server_url_env, test_dir
     _run(["deploy", "--name", "foo", fresh_main_thread_assertion_module.as_posix()])
     assert pytest._did_load_main_thread_assertion
     print()
+
+
+def test_serve(servicer, server_url_env, test_dir):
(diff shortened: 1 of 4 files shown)


client_test/cli_test.py
```diff
@@ -224,3 +224,8 @@ def test_no_user_code_in_synchronicity_deploy(servicer, server_url_env, test_dir
     _run(["deploy", "--name", "foo", fresh_main_thread_assertion_module.as_posix()])
     assert pytest._did_load_main_thread_assertion
     print()
+
+
+def test_serve(servicer, server_url_env, test_dir):
+    stub_file = test_dir / "supports" / "app_run_tests" / "webhook.py"
+    _run(["serve", stub_file.as_posix(), "--timeout", "3"])
```
erikbern, line 231: Maybe there's some more things we should check? Like hot-reloading? But I think we already have tests for that elsewhere

erikbern:
happy to address comments async

client_test/cli_test.py
thundergolfer: Yeh there's already a live reloading integration test, and some other reloading tests in this codebase (`live_reload_test.py`, `watcher_test.py`)

PR #223, opened 2023-01-19
Guard against module having __package__ but no __path__

aksh-at:
Didn't know this was possible, but happens in the case of `duckdb`. Couldn't repro super easily to add to `mounted_files_test.py`.


erikbern:
Wonder if there's some 3rd party library we can use for this.

PR #224, opened 2023-01-19
Don't have a bunch of instance attributes on providers

erikbern:
This is a step towards getting rid of `Provider` and merging it into `Handle`. These classes are already fairly dumb, and this PR makes several subclasses even simpler by removing all internal state.

I think this will be helpful as a way to remove a source of bugs we've had – a lot of function attributes that are set in global scope when functions are defined, but aren't set from `lookup` or inside the container, leading to weird failures (like `.web_url` missing).

It will also remove lots of unnecessary code that writes a bunch of intermediate stuff to `self` then reads it back exactly once during app construction (I think we can delete 100-200 lines of code in the end).

I only did this for `Secret`, `Dict`, `Queue`, and `SharedVolume` for now. The two big ones are `Function` and `Image` but I don't want to touch them until someone else agrees this approach makes sense.

### DIFF
--- a/client_test/notebook_test.py
+++ b/client_test/notebook_test.py
@@ -3,7 +3,14 @@ from pathlib import Path
 
 import jupytext
 import pytest
-from nbclient.exceptions import CellExecutionError
+import warnings
+
+try:
+    from nbclient.exceptions import CellExecutionError
+except ModuleNotFoundError:
+    # TODO(erikbern): sometimes my local jupyter packages end up in a bad state,
+    # but we don't want that to cause pytest to fail on startup.
+    warnings.warn("failed importing nbclient")
 
 
 @pytest.fixture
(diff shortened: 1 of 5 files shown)


aksh-at (commented):
Nice, would be nice to remove all those extra lines in `functions.py`. 

> I think this will be helpful as a way to remove a source of bugs we've had – a lot of function attributes that are set in global scope when functions are defined, but aren't set from lookup or inside the container, leading to weird failures (like .web_url missing).

I don't quite follow how this helps with that in particular. In that case, the problem is that we initialize `Handle` subclasses without ever calling the constructor right?

erikbern:
> I don't quite follow how this helps with that in particular. In that case, the problem is that we initialize `Handle` subclasses without ever calling the constructor right?

Yeah you're right. We had more of those problems before we split up the classes, and we'll increase the risk of those problems again once we merge them. But you're right that it's not necessarily a problem right now.

PR #226, opened 2023-01-20
Export conda environment variables

freider:
These are read by some packages in order to find
the location of compiled libraries etc.

Normally these variables are exported by .bashrc
but since we don't run that in our container entrypoint the variables are not defined during container runtime without this.

This fixes `import bitsandbytes` on gpu instances with cudatoolkit

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -582,6 +582,15 @@ class _Image(Provider[_ImageHandle]):
         return _Image(
             dockerfile_commands=dockerfile_commands,
             context_files={"/modal_requirements.txt": requirements_path},
+        ).dockerfile_commands(
+            [
+                "ENV CONDA_EXE=/usr/local/bin/conda",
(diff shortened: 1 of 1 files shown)


modal/image.py
```diff
@@ -582,6 +582,15 @@ def conda(python_version: str = "3.9") -> "_Image":
         return _Image(
             dockerfile_commands=dockerfile_commands,
             context_files={"/modal_requirements.txt": requirements_path},
+        ).dockerfile_commands(
```
freider, line 585: Not sure if this makes sense, but I appended this as an sub-image rather than include it in the base to make rebuilds quicker. Let me know if you think I should rather move it to the base image instead

erikbern (approved):
lgtm but i think a bigger problem is that conda builds keep failing – we should look into that

PR #229, opened 2023-01-23
Move .lookup to sit on the class instead

erikbern:
This deprecates the top-level global `modal.lookup` in favor of a classmethod on each `Provider` class, e.g. `modal.Function.lookup(...)` etc.

Benefits:
* We can send the type to the server, which will let us separate the "namespaces"
* A bit more type safety in Python

This will cause warnings for everyone who currently uses `lookup` / `aio_lookup` and we'll have to support the "old" behavior for at least 1-2 months.

There's a bit of type janking right now caused by the fact that providers and handles are different classes. We should be able to remove most/all of it once we merge those two.

### DIFF
--- a/client_test/lookup_test.py
+++ b/client_test/lookup_test.py
@@ -1,22 +1,22 @@
 # Copyright Modal Labs 2023
 import pytest
 
-from modal.aio import AioQueue, AioStub, aio_lookup
-from modal.exception import NotFoundError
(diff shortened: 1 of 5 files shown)


freider (approved):
Very nice!

erikbern:
Tried to revisit this one but it's causing a bunch of conflicts now. Will try to rewrite it as a new PR instead.

PR #233, opened 2023-01-25
Clean up file watching mount logic

erikbern:
This is something I ran into while implementing #225

In general, we have a few places that touch "internal" attributes on Objects, which makes it harder to refactor things. Cleaning it up will make it easier to rewrite how functions are created, which will make it possible down the road to merge providers and handles.

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -78,8 +78,14 @@ class _Mount(Provider[_MountHandle]):
         self._remote_dir = remote_dir
         self._condition = condition
         self._recursive = recursive
+        self._is_local = True
         rep = f"Mount({self._local_file or self._local_dir})"
         super().__init__(self._load, rep)
+
+    def is_local(self):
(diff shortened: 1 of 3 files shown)


modal/mount.py
```diff
@@ -78,9 +78,15 @@ def __init__(
         self._remote_dir = remote_dir
         self._condition = condition
         self._recursive = recursive
+        self._is_local = True
         rep = f"Mount({self._local_file or self._local_dir})"
         super().__init__(self._load, rep)
 
+    def is_local(self):
+        # TODO(erikbern): since any remote ref bypasses the constructor,
+        # we can rely on it to be set. Let's clean this up later.
```
thundergolfer, line 87:
```suggestion
        # we can't rely on it to be set. Let's clean this up later.
```

I think?

thundergolfer (approved):

erikbern:
Thanks @thundergolfer !

PR #235, opened 2023-01-25
Rewrite make_function_panel to not touch function internals

erikbern:
Minor fix to avoid reading function internals – should make it a bit easier to merge #225


erikbern:
tested this locally with a complex app and the output looks identical :)

PR #236, opened 2023-01-25
Move _load to a closure

erikbern:
Replaces #225

Will do it more incrementally this time


erikbern:
Ugh – ran into something with `interactive_shell` this time. I'll pause this for a bit. Might revisit later.

PR #238, opened 2023-01-25
remove direct `Handle` constructor invocation

erikbern:
This removes all custom `__init__` methods on `Handle` objects. The reason is we need to support initializing `Handle` objects in several different ways, and having overloaded construction logic makes it a bit trickier.

There's a bunch of hacky logic with functions which unfortunate gets even hackier in this PR, but in the long term this hackiness is exactly what I'll be able to remove quite soon.


erikbern:
merging, should be a noop

PR #239, opened 2023-01-26
Sort of merge FunctionHandle into Function

erikbern:
My goal here is to merge back Handles and Providers.

This PRs does the first step for functions by tying the objects together – we now pre-create a `FunctionHandle` for every `Function`. It also cleans up the custom constructors for function objects so that everything goes through `_initialize_from_proto`, which means there's exactly one way to initialize a function handle. This is good because it removes a source of errors that we've had before.

The other object types are much easier to do the same thing with.


aksh-at (approved):
Nice, already much cleaner!

PR #240, opened 2023-01-26
Refactor `_Image.run_function`

erikbern:
This changes `_Image.run_function` so that it doesn't use the user-facing `stub.function` SDK, but instead it uses the internal lower level interface and creates a `_Function` directly.

Reason for doing this is
1. The resulting code fits more nicely the pattern I'm trying to get to with providers/handles (and more tactically it blocks #239 for related reasons) – it removes a bit of a hack (imo) where `_Image._load` would create other objects "dynamically"
2. I think we should try to avoid `kwargs` logic on the user-facing interface (even though unwrapping it does lead to a bit more redundancy). Makes autogenerated docs nicer.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -730,6 +730,7 @@ class _Function(Provider[_FunctionHandle]):
         self,
         function_info: FunctionInfo,
         image=None,
+        secret: Optional[_Secret] = None,
         secrets: Collection[_Secret] = (),
         schedule: Optional[Schedule] = None,
         is_generator=False,
@@ -766,7 +767,10 @@ class _Function(Provider[_FunctionHandle]):
 
         self._raw_f = raw_f
         self._image = image
-        self._secrets = secrets
+        if secret:
+            self._secrets = [secret, *secrets]
+        else:
+            self._secrets = secrets
 
         if retries:
             if isinstance(retries, int):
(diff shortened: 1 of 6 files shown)


modal/functions.py
```diff
@@ -996,6 +1000,18 @@ def tag(self):
         """mdmd:hidden"""
         return self._tag
 
+    def get_build_def(self):
```
aksh-at, line 1003:
```suggestion
    def get_build_def(self):
        """mdmd:hidden"""
```

modal/image.py
```diff
@@ -778,7 +781,28 @@ def my_build_function():
...
+            info,
+            image=self,
+            secret=secret,
+            secrets=secrets,
+            gpu=gpu,
+            base_mounts=base_mounts,
+            mounts=mounts,
+            shared_volumes=shared_volumes,
+            memory=memory,
+            timeout=86400,
```
aksh-at, line 801: minor diff from before, but the max timeout could be overridden if desired, but now it no longer can. Probably worth allowing for?

aksh-at (approved):
Nice, LGTM besides nits

modal/image.py
erikbern: makes sense!

modal/functions.py
erikbern: thanks!

erikbern:
Passes integration tests btw!

erikbern:
Thanks @aksh-at !

PR #253, opened 2023-01-30
Refactor CLI import refs - allow for `modal run my_file.py::my_func`

freider:
Tries to also increase code reuse across the different "run" style commands: run/shell/deploy/serve

There appears to be some weird outstanding bug with module imports breaking when all tests are run at the same time, but individual tests work (so there is something stateful about imports, breaking relative imports for external packages)

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2022
 import asyncio
 import inspect
+import json
 import os
 import platform
 import time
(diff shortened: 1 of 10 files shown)


modal/functions.py
```diff
@@ -474,11 +473,11 @@ def _set_output_mgr(self, output_mgr: OutputManager):
...
 
-    def _set_raw_f(self, raw_f):
-        self._raw_f = raw_f
+    def _set_info(self, function_info: FunctionInfo):
+        self._info = function_info
 
-    def _set_function(self, function: "_Function"):
-        self._function = function
+    def _set_stub(self, stub):
+        self._stub = stub
```
erikbern, line 480: would prefer to avoid this – it creates a circular reference between stubs and functions and those things tend to make things harder to refactor

if you have the function's name, you can look it up directly from the stub – that should be enough right?

freider: It would be nice to not have it, I agree, but the point of this refactor is to let users specify *only* the function path, and not have to go through the "stub path" in order to access it, e.g. `modal run foofile.py::myfunc`, and then we need a way to look up the stub given only the FunctionHandler 🤔

erikbern: ok, seems acceptable! I can't think of any other simpler ways to accomplish the same thing.

freider:
Time to merge? :) I think code can probably be cleaned up more, but this one seems to constantly get into merge conflicts due to the large surface area of the the import handling in the cli

modal/functions.py
```diff
@@ -316,18 +317,27 @@ async def drain_input_generator():
...
+            sent_inputs += len(items)
+            t0 = time.monotonic()
             resp = await retry_transient_errors(
                 client.stub.FunctionPutInputs,
                 request,
                 max_retries=None,
             )
+            lat = time.monotonic() - t0
+            lats.append(lat)
+            print("Sent", sent_inputs, lat)
```
gongy, line 335: I'm getting this when running anything - intentional?

freider: Nope, unintentional and unrelated debug message

PR #254, opened 2023-01-30
Set `stub.app` to the container app inside containers

erikbern:
This also uses the stub app to initialize function handles in the container.

I'm a bit annoyed this PR is +43 lines net, since it's supposed to be a reduction in complexity. I suspect we'll be able to clean up a lot more stuff once we merge `_Function` and `_FunctionHandle`

This will break some rare cases. If the stub is defined in the global scope, but not used for a function `g`, then `g` won't be able to call any other functions defined on the stub (in global scope). For instance this case below:
```python
stub = Stub()

@stub.function
def f():
    print("hello")

def g():
    f.call()  # this will fail

if __name__ == "__main__":
    g_modal = stub.function(g)
    with stub.run():
        g_modal()
```

This feels acceptable? The reason it breaks is that `g` isn't associated with the stub, so the container will fail to initialize the stub properly.

### DIFF
--- a/client_test/container_app_test.py
+++ b/client_test/container_app_test.py
@@ -20,13 +20,13 @@ def my_f_2(x):
 @pytest.mark.asyncio
 async def test_container_function_initialization(unix_servicer, aio_container_client):
     stub = AioStub()
-    # my_f_1_container = stub.function(my_f_1)
+    my_f_1_container = stub.function(my_f_1)
 
     unix_servicer.app_objects["ap-123"] = {
         "my_f_1": "fu-123",
         "my_f_2": "fu-456",
     }
-    await AioApp.init_container(aio_container_client, "ap-123")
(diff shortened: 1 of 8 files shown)


erikbern:
Just ran integration tests and realized that this breaks `stub.is_inside` in the global scope because of how the import order changes (this PR creates the container app _after_ the import, but `is_inside` runs _during_ import).

Will think a bit about how to fix this. I'll also see if we can add a client-side test for it.

erikbern:
rebased on #256 and it's failing now (as expected)

I think the approach here is a bit flawed because of what I ran into. Will think a bit more about it. Will also think about `is_inside` since i think it's been a problematic thing for a long time.

erikbern:
Closing this for now – I'm planning to resurrect this in a slightly different form

PR #258, opened 2023-01-31
simplification of is_inside

erikbern:
Simplifies by setting the default image to the stub if there is none. This removes a bunch of special workarounds.


erikbern:
This is succeeding client tests but failing integration tests, so it implies the client tests are not yet enough to cover all cases. Will try to repro the failure!

erikbern:
Going to close this – I realized there's a better way to implement this

PR #263, opened 2023-01-31
Get container app from stub

erikbern:
When a stub is created, try to get the container app. This is then used for (a) `is_inside` (b) initializing all functions.

The stub grabs the container app in a very greedy way. It's potentially problematic if there's multiple stubs defined and some of them do not belong to the running app. We can fix that later by checking the stub's name, making sure there's not multiple stubs, etc.

We don't (yet) expose the `stub._app` property but will do that later.

This gets rid of a bunch of hacky code in several places, although there are some things that are slightly more verbose afterwards. But there's +14 lines of test so this would be net -17 lines otherwise.

### DIFF
--- a/client_test/container_app_test.py
+++ b/client_test/container_app_test.py
@@ -102,6 +102,9 @@ async def test_is_inside_default_image(servicer, unix_servicer, aio_client, aio_
         default_image_id = default_image_handle.object_id
         app_id = app.app_id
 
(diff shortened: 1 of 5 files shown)


erikbern:
This PR supersedes both #254 and #258 – basically contains the same functionality and is substantially simpler.

erikbern:
This passes all integration tests as well

PR #264, opened 2023-01-31
new weburlinfo state — label_stolen

thundergolfer:
Used when we take a webhook label from a running _ephemeral_ app to give it to a newer ephemeral app.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -8,16 +8,7 @@ import warnings
 from dataclasses import dataclass
 from datetime import date, timedelta
 from pathlib import Path
-from typing import (
-    Any,
-    AsyncIterable,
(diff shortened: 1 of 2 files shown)


modal/functions.py
```diff
@@ -8,16 +8,7 @@
...
-    Any,
-    AsyncIterable,
-    Callable,
-    Collection,
-    Dict,
-    List,
-    Optional,
-    Union,
-)
+from typing import Any, AsyncIterable, Callable, Collection, Dict, List, Optional, Union
```
thundergolfer, line 11: This line is 88 chars now, and our project configured 120. I dunno why it was line wrapped before.

PR #267, opened 2023-02-01
Adds context_mount argument for dockerfile_commands and from_dockerfile

freider:


client_test/image_test.py
```diff
@@ -302,16 +303,23 @@ def test_image_build_with_context_mount(client, servicer, tmp_path):
...
-        assert layers[0].context_mount_id == "mo-123"
-        files = {f.rel_filename: f.content for f in data_mount._get_files()}
-        assert files["data.txt"] == b"hello"
-        assert files[os.path.join("data", "sub")] == b"world"
-        assert len(files) == 2
+        for (image_name, expected_layer) in [("copy", 0), ("dockerfile_commands", 1), ("from_dockerfile", 0)]:
+            layers = get_image_layers(running_app[image_name].object_id, servicer)
+            assert layers[expected_layer].context_mount_id == "mo-123", f"error in {image_name}"
+            assert "COPY . /dummy" in layers[expected_layer].dockerfile_commands
+            files = {f.rel_filename: f.content for f in data_mount._get_files()}
```
thundergolfer, line 319: This loop looks to be re-testing the same `data_mount` object which is not related to the 3 image objects. From poking into the code, I think it's `Image.context_files` that has the data to assert on?

modal/image.py
```diff
@@ -677,10 +679,10 @@ def from_dockerhub(tag: str, setup_commands: list[str] = [], **kwargs) -> "_Imag
         )
 
     @staticmethod
-    def from_dockerfile(path: Union[str, Path]) -> "_Image":
+    def from_dockerfile(path: Union[str, Path], context_mount: Optional[_Mount] = None) -> "_Image":
```
thundergolfer, line 682: Worth commenting the `context_mount` param imo. It combines Docker build context and Modal mount concepts, both of which are non-trivial concepts.

client_test/image_test.py
freider: Oh yeah the data_mount part shouldn't be part of the loop (since the same is used for all three images). good catch.
Image.context_files is something else though, that's the legacy payload for attaching context files (now mostly used for the modal requirements.txt)

modal/image.py
freider: For sure, adding comment

PR #273, opened 2023-02-07
noop mount refactor

erikbern:
Zero behavioral impact, just add types and break up the function.

The goal of this was is to rewrite the constructor but I'll do it in a subsequent PR

### DIFF
--- a/client_test/watcher_test.py
+++ b/client_test/watcher_test.py
@@ -7,15 +7,14 @@ from pathlib import Path
 from watchfiles import Change
 
 from modal._watcher import _watch_args_from_mounts
-from modal.mount import _Mount
 
 
 @pytest.mark.asyncio
 async def test__watch_args_from_mounts(monkeypatch, test_dir):
     paths, watch_filter = _watch_args_from_mounts(
-        mounts=[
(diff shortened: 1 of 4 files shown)


freider:
Oh, I didn't know you were working on mounts. Just realized I had this old draft PR unmerged, which does a larger refactor where the Mount constructor takes no arguments and instead uses a builder pattern similar to our Image class:
https://github.com/modal-labs/modal-client/pull/136

I fixed it up now so it should build, and I can rebase it on this change once you merge (the changes here aren't very related to my branch, but there is likely to be some conflicts so I don't want to merge and break this PR). What did you have in mind for the upcoming changed constructor?

erikbern:
@freider there's a bunch of different ways to construct handles and providers, and it might get more confusing once we merge them. I'm trying to avoid using constructors as much as possible, since static factory methods are a bit more flexible. also trying to remove a lot of object properties, since objects may be constructed in different ways (by importing from modal dynamically etc) and those properties may not always be defined. so I'm trying to get rid of having a `_load` method and instead defining that as a closure on construction.

it's not an absolute thing though. but in that spirit, this PR mostly just moves a lot of code out of the constructor into the global scope and will help me remove a few object properties

PR #274, opened 2023-02-07
Check for set equality

erikbern:
These tests have been failing for me locally for a long time and the structure of the tests have made it hard to understand why.

Checking for set equality is nicer because pytest will output the delta:

```
         # Assert we include everything from `pkg_a` and `pkg_b` but not `pkg_c`:
>       assert files == {"a.py", "c.py", "e.py", "script.py", "__init__.py", "f.py", "h.py"}
E       AssertionError: assert {'__init__.py...'g/h.py', ...} == {'__init__.py..., 'h.py', ...}
E         Extra items in the left set:
E         'g/h.py'
E         Extra items in the right set:
E         'h.py'
E         Full diff:
E         - {'a.py', 'h.py', '__init__.py', 'c.py', 'e.py', 'script.py', 'f.py'}
E         ?         --------                       --------
E         + {'a.py', '__init__.py', 'c.py', 'script.py', 'f.py', 'g/h.py', 'e.py'}
E         ?                                                    ++++++++++++++++++
```

### DIFF
--- a/client_test/mounted_files_test.py
+++ b/client_test/mounted_files_test.py
@@ -20,12 +20,17 @@ def venv_path(tmp_path):
 script_path = "pkg_a/script.py"
 
 
-def test_mounted_files_script(test_dir):
+@pytest.fixture
+def supports_dir(test_dir):
+    return test_dir / Path("supports")
+
+
+def test_mounted_files_script(supports_dir):
     p = subprocess.run(
         [sys.executable, str(script_path)],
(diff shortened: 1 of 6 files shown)


erikbern:
will fix this tomorrow but @freider lmk if you think the approach makes sense?

freider (commented):
Makes sense!

freider (approved):

PR #278, opened 2023-02-08
Mount: static factory methods

erikbern:
This replaces `Mount(local_dir="foo/bar")` with `Mount.from_local_dir("foo/var")` which is consistent with the new factory method `add_local_dir` that @freider just added in #136

For instance you can chain it using

```python
mount = Mount.from_local_dir("foo/bar").add_local_dir("foo/baz")
```

Minor improvement in terms of consistency, and getting rid of constructors also helps me doing some more refactoring later (merging providers and handles). This is also a net reduction in LOC eventually – we will be able to delete about 40 lines once we deprecate the old constructor fully.

A minor thing I realized is that `remote_path` in `add_local_file` is not the same as `remote_dir` in the `Mount` constructor – the former adds the basename of the file itself, the latter excludes it. Had to rewrite a few places, when changing code from `Mount(local_file=x, remote_dir=y)` to `Mount.from_local_file(x, remote_path=y)`

### DIFF
--- a/modal/_function_utils.py
+++ b/modal/_function_utils.py
@@ -144,18 +144,18 @@ class FunctionInfo:
     def get_mounts(self) -> Dict[str, _Mount]:
         if self.type == FunctionInfoType.PACKAGE:
             mounts = {
-                self.base_dir: _Mount(
-                    local_dir=self.base_dir,
-                    remote_dir=self.remote_dir,
+                self.base_dir: _Mount.local_dir(
+                    self.base_dir,
+                    remote_path=self.remote_dir,
                     recursive=True,
                     condition=package_mount_condition,
                 )
             }
         elif self.type == FunctionInfoType.FILE:
             mounts = {
-                self.file: _Mount(
-                    local_file=self.file,
-                    remote_dir=ROOT_DIR,
+                self.file: _Mount.local_file(
+                    self.file,
+                    remote_path=ROOT_DIR,
                 )
             }
         elif self.type == FunctionInfoType.NOTEBOOK:
@@ -193,9 +193,9 @@ class FunctionInfo:
                     ):
                         continue
                     remote_dir = ROOT_DIR / PurePosixPath(*m.__name__.split("."))
-                    mounts[path] = _Mount(
-                        local_dir=path,
-                        remote_dir=remote_dir,
+                    mounts[path] = _Mount.local_dir(
+                        path,
+                        remote_path=remote_dir,
                         condition=package_mount_condition,
                         recursive=True,
                     )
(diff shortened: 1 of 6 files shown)


erikbern:
This will cause a bunch of examples and integration tests to fail btw, so we'll have to update those separately

modal/mount.py
```diff
@@ -150,20 +155,23 @@ def __init__(
...
                 elif local_file:
                     remote_path = PurePosixPath(remote_dir) / Path(local_file).name
-                    self._entries = self.add_local_file(local_path=local_file, remote_path=remote_path)._entries
+                    self._entries = self.local_file(local_path=local_file, remote_path=remote_path)._entries
 
         self._is_local = True
         rep = f"Mount({self._entries})"
         super().__init__(self._load, rep)
 
+    def extend(self, *entries) -> "_Mount":
```
freider, line 172: Nice!

modal/_function_utils.py
```diff
@@ -144,18 +144,19 @@ def __init__(self, f, serialized=False, name_override: Optional[str] = None):
...
-                    remote_dir=self.remote_dir,
+                self.base_dir: _Mount.local_dir(
+                    self.base_dir,
+                    remote_path=self.remote_dir,
                     recursive=True,
                     condition=package_mount_condition,
                 )
             }
         elif self.type == FunctionInfoType.FILE:
+            remote_path = PurePosixPath(ROOT_DIR) / Path(self.file).name
```
freider, line 155: nit: ROOT_DIR is already a PurePosixPath, so no need to wrap it in another one

freider (approved):
Lgtm!

freider:
> A minor thing I realized is that `remote_path` in `add_local_file` is not the same as `remote_dir` in the `Mount` constructor – the former adds the basename of the file itself, the latter excludes it. Had to rewrite a few places, when changing code from `Mount(local_file=x, remote_dir=y)` to `Mount.local_file(x, remote_path=y)`
Yeah this was an intentional change I made in my patch, as the old way of single files didn't allow flexibility to rename the file which was kind of annoying sometimes. Figured it would be a good time to introduce the change when the API changes so that old code won't silently break because of it :)

> Open question: maybe the static factory methods should be called `from_local_dir` and `from_local_file` rather than `local_dir` and `local_file`? E.g.
Sure, that makes sense to me

modal/_function_utils.py
erikbern: Thanks, will fix!

erikbern:
> Yeah this was an intentional change I made in my patch, as the old way of single files didn't allow flexibility to rename the file which was kind of annoying sometimes. Figured it would be a good time to introduce the change when the API changes so that old code won't silently break because of it :)

sgtm – this will make it slightly less convenient to rewrite existing code that uses the `Mount` constructor, but nbd

> Sure, that makes sense to me

cool, will rename

PR #279, opened 2023-02-09
Adds protos for multipart blob uploads

freider:
Implementation and usage in followup PR

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -204,21 +204,21 @@ message BlobCreateRequest {
   // TODO(erikbern): how are these garbage collected?
   // Shouldn't they belong to an app?
   string content_md5 = 1;
+  string content_length_bytes = 2;
+}
+
+message MultiPartUpload {
+  repeated int64 part_lengths_bytes = 1;
+  repeated string upload_urls = 2;
+  string completion_url = 3;
 }
 
 message BlobCreateResponse {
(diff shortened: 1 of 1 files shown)


freider:
> Would it make more sense if it's the same method (`BlobCreate`) on the server, but the server returns a different response if multipart uploads are needed? We would have to send the file size to `BlobCreate` for this to make sense.

Perhaps, but a multipart upload requires *each part's* md5 checksum to be known ahead of time when creating the presigned urls so it would then require another roundtrip to the server with that data attached if we do it like that. Unless we feel checksumming of each individual part is unnecessary? We do still have checksumming for the full file, so perhaps that make sense?

freider:
I think you might be right that there are advantages to having the server determine the upload specification that outweigh the need of using per-part MD5 checksumming. Refactored a bit

erikbern (approved):
looks good!

modal_proto/api.proto
```diff
@@ -204,11 +204,22 @@ message BlobCreateRequest {
   // TODO(erikbern): how are these garbage collected?
   // Shouldn't they belong to an app?
   string content_md5 = 1;
+  string content_sha256_base64 = 2;
+  int64 content_length_bytes = 3;
```
erikbern, line 208: micro nit, but i think of "bytes" as a default unit for anything like this, so it could be omitted

PR #280, opened 2023-02-09
Fix mounted tests with pip -e

erikbern:
This makes the tests pass locally for me


aksh-at (approved):
Nice, not as bad as I thought it'd be!

erikbern:
> Nice, not as bad as I thought it'd be!

Nope it took probably 5 min in total

PR #283, opened 2023-02-09
Clean up some mount code for creating the client mounts

erikbern:
Minor drive-by cleanup opportunity that I ran into.


erikbern:
i take it back, this code was used by the `modal_base_images` package. will fix.

PR #285, opened 2023-02-10
Stub reduction: prep

erikbern:
`modal/stub.py` is nearly 900 lines and I'm thinking about merging it with the app which would make things even bigger.

However, turns out a lot of the code in stub.py does very little with the stub itself – it's a lot of methods that could really be functions instead. `run` and `serve` for instance do very little with the stub.

This PR itself is a bit of prep work that removes some places internally where we access stub properties when it's not needed.

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -110,7 +110,6 @@ class _Stub:
             self._description = name
         else:
             self._description = self._infer_app_desc()
-        self._app_id = None
         self._blueprint = blueprint
         self._client_mount = None
         self._function_mounts = {}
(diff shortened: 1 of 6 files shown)


modal/stub.py
```diff
@@ -347,22 +345,11 @@ async def serve(self, client=None, stdout=None, show_progress=None, timeout=None
...
-        if self._app_id is not None:
-            raise InvalidError(
-                f"Found existing app '{self._app_id}'. You may have nested stub.serve() inside a running app like this:\n"
-                'if __name__ == "__main__":\n'
-                "    with stub.run():\n"
-                "        stub.serve() # ❌\n\n"
-                "You might need to do something like this: \n"
-                'if __name__ == "__main__":\n'
-                "    stub.serve()\n"
+                "The stub already has an app running."
```
erikbern, line 350: I took the liberty of removing the previous exceptions. I think they were great in the past before we had the `modal serve` command. But it's probably better to promote `modal serve` today.

erikbern:
All of this is very easily revertible so let met know if there's any concerns!

PR #286, opened 2023-02-10
minor stub.serve cleanup

erikbern:
Not sure why we had all the `__anext__` stuff instead of just an `async for` loop – this seems to work.

Also removed all the `AppChange` stuff – afaict it was never consumed. Now `watch` is just a dummy generator yielding `None`

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -375,9 +375,10 @@ class _Stub:
             curr_proc = None
             try:
                 async for _ in watch(self._local_mounts, output_mgr, timeout):
-                    if sys.version_info.major == 3 and sys.version_info.minor <= 7:
-                        output_mgr.print_if_visible(
-                            "Live-reload skipped. This feature is unsupported below Python 3.8. Upgrade to Python 3.8+ to enable live-reloading."
+                    if sys.version_info <= (3, 7):
+                        warnings.warn(
+                            "Live-reload skipped. This feature is unsupported below Python 3.8."
+                            " Upgrade to Python 3.8+ to enable live-reloading."
                         )
                     else:
                         curr_proc = await restart_serve(
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -375,7 +375,9 @@ class _Stub:
                 curr_proc = None
                 try:
                     async for event in watch(self._local_mounts, output_mgr, timeout):
-                        if event in [AppChange.START, AppChange.TIMEOUT]:
+                        if event == AppChange.TIMEOUT:
+                            return
+                        elif event == AppChange.START:
                             continue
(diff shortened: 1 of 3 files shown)


modal/stub.py
```diff
@@ -367,36 +367,29 @@ async def serve(self, client=None, stdout=None, show_progress=None, timeout=None
...
-                    if sys.version_info.major == 3 and sys.version_info.minor <= 7:
-                        while event != AppChange.TIMEOUT:
+            async with self._run(client, output_mgr, None, mode=StubRunMode.SERVE) as app:
+                client.set_pre_stop(app.disconnect)
+                existing_app_id = app.app_id
+
+                curr_proc = None
+                try:
+                    async for event in watch(self._local_mounts, output_mgr, timeout):
+                        if event in [AppChange.START, AppChange.TIMEOUT]:
```
erikbern, line 377: think we can also remove all the `AppChange` stuff – afaict it's not used.

modal/stub.py
```diff
@@ -367,36 +367,27 @@ async def serve(self, client=None, stdout=None, show_progress=None, timeout=None
...
-                    event = await event_agen.__anext__()
-            finally:
-                if curr_proc:
-                    try:
-                        curr_proc.send_signal(signal.SIGINT)
-                    except ProcessLookupError:
-                        logger.warning("Could not interrupt app serve. Supervised process already terminated.")
-                await event_agen.aclose()
+                        else:
+                            curr_proc = await restart_serve(
```
thundergolfer, line 382: Because the restart happens inside the context manager, you'll get double logging here I think. the `__exit__` on the context manager closes the logs loop before the new one starts up. 

It's an complexity in the code that worth me commenting, in retrospect.

modal/stub.py
```diff
@@ -367,36 +367,27 @@ async def serve(self, client=None, stdout=None, show_progress=None, timeout=None
...
-                            event = await event_agen.__anext__()
-                        return
-
-                # live-reloading loop
-                while event != AppChange.TIMEOUT:
-                    curr_proc = await restart_serve(
-                        existing_app_id=app.app_id, prev_proc=curr_proc, output_mgr=output_mgr
-                    )
-                    event = await event_agen.__anext__()
+                        output_mgr.print_if_visible(
```
thundergolfer, line 378: Unfortunately this needs to be inside the context manager like it was otherwise it won't log at all. Same issue but in a different way; the logs loop is closed and in Python 3.7 there's no child started to create another.

erikbern: we're just talking about a warning printed for python 3.7 so is it ok not to use the output manager for this? just printing it?

thundergolfer (approved):

PR #289, opened 2023-02-10
Fix is_inside

erikbern:
The problem is that we now set `stub._app` locally too.

Updated to test to catch this as well, and fixing it.

### DIFF
--- a/client_test/container_app_test.py
+++ b/client_test/container_app_test.py
@@ -58,13 +58,13 @@ async def test_is_inside(servicer, unix_servicer, aio_client, aio_container_clie
 
     stub = get_stub()
 
-    # No container is running
-    assert not stub.is_inside()
-    assert not stub.is_inside(image_1)
-    assert not stub.is_inside(image_2)
-
     # Run container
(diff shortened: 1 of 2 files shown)


erikbern:
Not 100% sure if this resolves the integration test failure – something is weird with my integration tests. But I'm like 70% sure? Merging it.

client_test/container_app_test.py
```diff
@@ -58,13 +58,13 @@ def get_stub():
...
 
-    # No container is running
-    assert not stub.is_inside()
-    assert not stub.is_inside(image_1)
-    assert not stub.is_inside(image_2)
-
     # Run container
     async with stub.run(client=aio_client) as app:
+        # We're not inside the container (yet)
+        assert not stub.is_inside()
```
aksh-at, line 64: super minor but might as well keep both cases around 🤷

erikbern: yeah probably. my rationale was that "there is an app running on this stub" is a strict subset of "this stub exists" so it does increase the specificity. but maybe that's maybe making an implicit assumption.

PR #296, opened 2023-02-13
Process inputs one at a time

gongy:
Change to container scope - no more batching.

Repeatedly gets a single input, puts its output - all sequentially.

WIP.


mwaskom:
Doing some PR cleanup, feel free to re-open if still viable.

PR #297, opened 2023-02-14
Keep reference to shutdown task to prevent garbage collection

freider:
Got reminded of this behavior seeing this: https://news.ycombinator.com/item?id=34754276
Did a quick search of our code base and found this as the only place were we create tasks without using our `TaskContext` or otherwise saving a reference to the task

### DIFF
--- a/modal_utils/async_utils.py
+++ b/modal_utils/async_utils.py
@@ -301,6 +301,9 @@ def warn_if_generator_is_not_consumed(gen_f):
     return f_wrapped
 
 
+_shutdown_tasks = []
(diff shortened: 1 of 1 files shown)


freider:
I wonder if this could have caused the sporadic weird races/shutdown issues we have seen on client shutdown @thundergolfer ?

thundergolfer:
The shutdown hook was definitely being started and run, but it got interrupted by loop shutdown. I'm not sure whether keeping a reference to it would stop the shutdown until the task had finished.

PR #298, opened 2023-02-14
Assert for not naming the deployment modal.py

franz101:
If you name your stub modal.py it will cause a lot of errors.


erikbern:
Nice catch! Looks like there's a conflict but we can fix. I think it would be useful to have a test as well.

I'll leave this PR up for a bit until we get a chance to look at it.

Thanks!!

thundergolfer:
Closing as stale, but thanks for your contribution regardless!

PR #299, opened 2023-02-14
Fix DeprecationError from volume CLI commands

aksh-at:


modal/cli/volume.py
```diff
@@ -76,7 +75,9 @@ def create(name: str, cloud: str = typer.Option("aws", help="Cloud provider to c
 
 
 async def volume_from_name(deployment_name) -> _SharedVolumeHandle:
-    shared_volume = await aio_lookup(deployment_name)
+    from modal.aio import AioSharedVolume
```
erikbern, line 78: does this have to be in a local scope? think it could be at the top along with `from modal.shared_volume import AioSharedVolumeHandle, _SharedVolumeHandle`

PR #301, opened 2023-02-14
Paginated `SharedVolume` list

aksh-at:
See https://github.com/modal-labs/modal/pull/4921

### DIFF
--- a/modal/cli/volume.py
+++ b/modal/cli/volume.py
@@ -152,16 +152,19 @@ async def _glob_download(
 ):
     q: asyncio.Queue[Tuple[Optional[Path], Optional[api_pb2.SharedVolumeListFilesEntry]]] = asyncio.Queue()
 
-    for entry in await volume.listdir(remote_glob_path):
-        output_path = local_destination / entry.path
-        if output_path.exists():
-            if overwrite:
-                shutil.rmtree(output_path)
-            else:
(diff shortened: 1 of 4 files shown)


modal/cli/volume.py
```diff
@@ -181,9 +187,9 @@ async def consumer():
                 q.task_done()
 
     tasks = []
```
erikbern, line 189: probably a good use case for `async_utils.TaskContext`

modal/shared_volume.py
```diff
@@ -64,16 +64,29 @@ async def read_file(self, path: str) -> AsyncIterator[bytes]:
...
         * Passing a file path returns a list containing only that file's listing description
         * Passing a glob path (including at least one * or ** sequence) returns all files matching that glob path (using absolute paths)
         """
-        req = api_pb2.SharedVolumeListFilesRequest(shared_volume_id=self._object_id, path=path)
-        response = await retry_transient_errors(self._client.stub.SharedVolumeListFiles, req)
-        return list(response.entries)
+        entries = []
+        async for entry in self.iterdir(path):
+            entries.append(entry)
+        return entries
```
erikbern, line 89: you can use a list comprehension for async stuff too. can collapse these 4 lines into just `return [entry async for entry in self.iterdir(path)]`

PR #304, opened 2023-02-15
Increase max reties for MountPutFile requests

freider:
Backend will now (soon) reuse a previous transfer if one is in progress, so this effectively allows for ~15 min of processing time instead of ~3

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -26,7 +26,7 @@ from .exception import InvalidError, NotFoundError, deprecation_warning
 from .object import Handle, Provider
 
 
-MOUNT_PUT_FILE_TIMEOUT_SECONDS = 10 * 60  # 10 min max for transferring files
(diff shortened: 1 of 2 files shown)


erikbern:
doesn't this imply we need to increase the k8s pod draining time to 15 min too?

freider:
> doesn't this imply we need to increase the k8s pod draining time to 15 min too?

To be fully error-resilient yes but since we are running retries and have a 60s "heartbeat"/ttl on the in-progress transfer, it should restart the transfer on another server instance after at most 60s if it gets drained in the middle of a call. Also MountPutFiles taking longer than a minute should be rare to begin with. 
I'm a bit hesitant with increasing pod draining time since it makes deploys slower (and increases the window of possible version-instability), but we could do that if this turns out to be a problem?

freider:
Perhaps we should have an explicit retry signal instead, as suggested in other PR

freider:
I realized we could simply use the existing `exists==True` condition instead, as long as we make the backend return errors for old clients when retries are needed.

PR #305, opened 2023-02-15
SharedVolume.add_local_file and .add_local_dir

freider:


modal/shared_volume.py
```diff
@@ -64,6 +66,47 @@ async def read_file(self, path: str) -> AsyncIterator[bytes]:
...
+                relpath_str = subpath.relative_to(local_path).as_posix()
+                yield self.add_local_file(subpath, PurePosixPath(remote_path, relpath_str))
+
+        sem = asyncio.Semaphore(concurrency)
+
+        async def limited_task(coro):
+            async with sem:
+                return await coro
+
+        await asyncio.gather(*(limited_task(t) for t in gen_transfers()), return_exceptions=True)
```
erikbern, line 108: minor nit but i think it would be nice to break this out into a general purpose utility function to map with a pool

modal/shared_volume.py
```diff
@@ -51,7 +53,7 @@ async def write_file(self, remote_path: str, fp: BinaryIO):
         else:
             data = fp.read()
             req = api_pb2.SharedVolumePutFileRequest(shared_volume_id=self._object_id, path=remote_path, data=data)
-        await retry_transient_errors(self._client.stub.SharedVolumePutFile, req)
+        await retry_transient_errors(self._client.stub.SharedVolumePutFile, req, max_retries=20)
```
erikbern, line 56: 20 retries seems like a lot? or is this because of the transfer thing in the api server?

freider: Yeah, I was thinking of adding similar logic to the SharedVolumePutFile to allow for larger files. Perhaps we should have a smarter retry_transient_errors that has unbounded retries in case of some specific grpc status/error that indicates something is in progress? (Something like an http 202 - accepted)

erikbern: Yeah or maybe `SharedVolumePutFile` could just return after 55s with a payload that tells the client to retry? We use a similar pattern when we wait for images to be built.

modal_utils/async_utils.py
```diff
@@ -329,3 +330,15 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:
         return loop.run_in_executor(None, functools.partial(f, *args, **kwargs))
 
     return wrapper
+
+
+class ConcurrencyPool:
```
erikbern, line 335: nice! it's kind of weird asyncio doesn't have this built in

erikbern (approved):
great!

PR #306, opened 2023-02-15
Handle duplicate entries in base_dirs

aksh-at:
In this case the user had two copies of the same path in `base_dirs` so deduping is enough. But in the case where there are multiple entries, we know that they are all prefixes of each other since they contain `module_file`. So we can just take the shortest one.

### DIFF
--- a/modal/_function_utils.py
+++ b/modal/_function_utils.py
@@ -108,6 +108,7 @@ class FunctionInfo:
                 logger.info(f"Base dirs: {base_dirs}")
                 raise Exception("Wasn't able to find the package directory!")
             elif len(base_dirs) > 1:
+                # Base_dirs should all be prefixes of each other since they all contain `module_file`.
                 base_dirs.sort(key=lambda x: len(x))
 
             self.base_dir = base_dirs[0]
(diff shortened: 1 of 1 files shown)


modal/_function_utils.py
```diff
@@ -96,18 +96,22 @@ def __init__(self, f, serialized=False, name_override: Optional[str] = None):
...
-            if len(base_dirs) != 1:
+            if not base_dirs:
                 logger.info(f"Module files: {module_file}")
                 logger.info(f"Package paths: {package_paths}")
                 logger.info(f"Base dirs: {base_dirs}")
                 raise Exception("Wasn't able to find the package directory!")
-            (self.base_dir,) = base_dirs
+            elif len(base_dirs) > 1:
+                # Base_dirs should all be prefixes of each other since they all contain `module_file`.
+                base_dirs.sort(key=lambda x: len(x))
```
erikbern, line 112: `lambda x: len(x)` is equivalent to `len` so you can just do `.sort(key=len)` :)

aksh-at: Oops I was banging this out in a hurry.

erikbern: obvously super tiny nit, no need to fix

PR #307, opened 2023-02-15
Step 1 towards deploying standalone objects

erikbern:
This enables you to do `._deploy(label)` on a provider and it returns a handle. It's basically what we want in the future (but named `.deploy`).

However we want to get rid of the dummy apps later. For now, I'm just going to work with the existing structure and keep the dummy apps around for a bit. This PR does change one significant thing which is that it starts to send `object_entity` to the server. The fact that this is set on the app's deployment will let us change the index structure in the DB more easily. For old clients, we can infer the object entity by looking at the objects on the app.


erikbern:
Merging this since this is super non-intrusive (basically changes nothing, just adds). Happy to revise if there's any thoughts!

PR #309, opened 2023-02-16
Fixes bug where modal serve would not show logs or heartbeat MOD-878

freider:
Previously logs and heartbeats would only appear after the first change of a watched file. Now we instead init an empty app and let watch trigger an initial update of that app

Adds some test config/magic to make the test pick up hearbeats from the subprocess


erikbern:
my bad – i must have broken it during https://github.com/modal-labs/modal-client/pull/286

PR #323, opened 2023-02-19
Standalone resolver

erikbern:
This PR makes it possible to use the resolver without the app, which
1. Reduces some complexity of the `App` class
2. Should make it possible to clean up `Provider._deploy` so we don't have to use the app at all (makes it possible to deploy a single object without creating an app client-side)

### DIFF
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -12,7 +12,7 @@ from modal_utils.async_utils import synchronize_apis
 
 
 class StatusRow:
-    def __init__(self, progress):
+    def __init__(self, progress: Optional[Tree]):
         from ._output import step_progress  # Lazy import to only import `rich` when necessary.
 
         self._spinner = None
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -8,6 +8,8 @@ else:
     Spinner = TypeVar("Spinner")
     Tree = TypeVar("Tree")
 
+from modal_utils.async_utils import synchronize_apis
+
 
 class StatusRow:
     def __init__(self, progress):
@@ -33,7 +35,7 @@ class StatusRow:
             self._step_node.label = step_completed(message, is_substep=True)
 
 
-class Resolver:
+class _Resolver:
     # Unfortunately we can't use type annotations much in this file,
     # since that leads to circular dependencies
     _progress: Optional[Tree]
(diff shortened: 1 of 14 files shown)


modal/_resolver.py
```diff
@@ -8,50 +8,80 @@
...
-class Resolver:
+
+class StatusRow:
+    def __init__(self, progress):
+        from ._output import step_progress  # Lazy import to only import `rich` when necessary.
+
+        self._spinner = None
+        self._step_node = None
+        if progress is not None:
+            self._spinner = step_progress()
```
freider, line 21: Should this be?
```suggestion
            self._spinner = step_progress(progress)
```

modal/_resolver.py
```diff
@@ -8,50 +8,80 @@
     Spinner = TypeVar("Spinner")
     Tree = TypeVar("Tree")
 
+from modal_utils.async_utils import synchronize_apis
 
-class Resolver:
+
+class StatusRow:
+    def __init__(self, progress):
```
freider, line 15: Perhaps type this as an `Optional[Tree]` - I was confused at first thinking it was the initial message

erikbern: 👍

freider (approved):
lgtm!

modal/_resolver.py
```diff
@@ -8,50 +8,80 @@
...
-            if self._step_node is None:
-                self._spinner = step_progress()
-                self._step_node = self._progress.add(self._spinner)
-            step_progress_update(self._spinner, message)
+        self._client = client
+        self._app_id = app_id
+
+    @property
+    def app_id(self):
+        return self._app_id
```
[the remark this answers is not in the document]
erikbern: yeah i was actually planning to do that in the next PR, but i'll do it in this!

modal/app.py
```diff
@@ -71,58 +70,28 @@ def app_id(self) -> str:
...
         unindexed_object_ids = list(
-            set(obj.object_id for obj in self._local_uuid_to_object.values())
+            set(obj.object_id for obj in resolver.objects())
             - set(obj.object_id for obj in self._tag_to_object.values())
         )
         req_set = api_pb2.AppSetObjectsRequest(
             app_id=self._app_id,
             indexed_object_ids=indexed_object_ids,
             unindexed_object_ids=unindexed_object_ids,
-            new_app_state=new_app_state,  # type: ignore
```
freider, line 125: Ohhh no, I missed this removal in the review :O Was this intentional?

erikbern: Pretty sure it was accidental :(

PR #327, opened 2023-02-21
Adds protos for FunctionCallCancel

freider:
Also removes an unused field in FunctionGetInputsItem

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -417,7 +417,7 @@ message FunctionGetInputsItem {
   string input_id = 1;
   FunctionInput input = 2;
   bool kill_switch = 3;
-  string function_call_id = 4;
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -417,7 +417,7 @@ message FunctionGetInputsItem {
   string input_id = 1;
   FunctionInput input = 2;
   bool kill_switch = 3;
-  string function_call_id = 4;
+  reserved 4; // previously used
```
erikbern, line 420: 🤯

freider: Unrelated, but I noticed it wasn't being used except in a test

erikbern: didn't realize you can do this in protobuf!

PR #328, opened 2023-02-21
Changes ECR method name

luiscape:
Changes method name from `from_ecr` to `from_aws_ecr` as that is more user-friendly.


thundergolfer (approved):
🙏

PR #329, opened 2023-02-22
Prevent a bunch of internal tracebacks on intentional shutdown

freider:
This removes the long `RuntimeError: cannot schedule new futures after shutdown` traceback when a user Ctrl-C:s in the middle of executing a function.

However, it feels like there should be better way to detect these conditions by for example closing down any grpc channel/connection pools prior to shutting down the event loop.

The underlying error is a bit of a mystery to me - why would a co-routine in the event loop would still be executed *after* the ThreadPoolExecutor has been shut down (in what I would assume is the event loop teardown?).

Edit:
After some more investigation this seems to be a larger class of problems that we should catch on some other level... The test in this PR is still useful, but the solution needs to be more general. As it stands, Ctrl-C during a bunch of other states still lead to traceback hell

### DIFF
--- a/client_test/cli_test.py
+++ b/client_test/cli_test.py
@@ -16,6 +16,7 @@ import click.testing
 import pytest
 import pytest_asyncio
 
+from client_test.supports.skip import skip_windows
 from modal.cli.entry_point import entrypoint_cli
 from modal.client import Client
 from modal_proto import api_pb2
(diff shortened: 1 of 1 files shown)


freider:
All the exception spam turns out to be because of a deeper issue in synchronicity, which is being fixed in a PR there instead. Closing this for now

PR #331, opened 2023-02-22
Fix outputs lost when putting inputs returns slowly

gongy:
Pumping inputs awaits a gRPC call and then populates the map.
Receiving outputs will ignore inputs that are not in the map.
Existing bug involved ignoring outputs that were legitimate, but their inputs were not yet populated in the map.

**This PR adds a test that times out as a result of this bug, and fixes the bug.**

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -292,10 +292,8 @@ async def _map_invocation(
     have_all_inputs = False
     num_inputs = 0
     num_outputs = 0
-    # Map input_id -> next expected gen_index value, or -1 if complete
-    # Could contain keys which are not yet in the pending_inputs set
-    pending_outputs: Dict[str, int] = {}
-    completed_outputs: Set[str] = set()
+    pending_outputs: Dict[str, int] = {}  # Map input_id -> next expected gen_index value
+    completed_outputs: Set[str] = set()  # Set of input_ids whose outputs are complete (expecting no more values)
 
     input_queue: asyncio.Queue = asyncio.Queue()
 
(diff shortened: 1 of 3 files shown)


modal/functions.py
```diff
@@ -292,7 +292,10 @@ async def _map_invocation(
     have_all_inputs = False
     num_inputs = 0
     num_outputs = 0
-    pending_outputs = {}  # map input_id -> next expected gen_index value
+    # Map input_id -> next expected gen_index value, or -1 if complete
```
freider, line 295: Where is the code that sets it to -1 if complete? Earlier version of the PR?

modal/functions.py
```diff
@@ -292,7 +292,10 @@ async def _map_invocation(
...
+    # Could contain keys which are not yet in the pending_inputs set
+    pending_outputs: Dict[str, int] = {}
+    completed_outputs: Set[str] = set()
```
freider, line 298: I guess the potential downside is that for huge maps this will keep one entry for every input in memory until the map completes, but in practice that probably won't matter :)

freider (approved):
I am a bit tired, but I *think* it should work :) Nice catch with the outputs arriving before input response race condition!

modal/functions.py
gongy: Fixing comment! Thanks for catching

modal/functions.py
gongy: Ah, yes it is not ideal .. I can make this a TODO.

PR #334, opened 2023-02-23
Change `localhost` to `127.0.0.1` in test servicer

freider:
Since the former sometimes gives really slow performance on Windows.

See https://github.com/grpc/grpc/issues/23072


erikbern:
wow – looks like it took the test time from 12 min to 6 min!

freider:
yeah and most of that time is actually due to slow CI setup of the windows environment (not sure if something is wrong with the cached python setup) - the tests are now just as fast as on linux on my machine

erikbern:
🏎️

PR #336, opened 2023-02-24
Call function locally using __call__

erikbern:
This updates the calling convention so that `f(...)` now calls a function _locally_, i.e. basically like a Python function call.

Up until 2022-12-05, it would call the function _remotely_ but that was then deprecated in favor of `f.call` and the warning was increased to an error on 2023-01-23 in https://github.com/modal-labs/modal-client/pull/230

Also removed old deferred function calling methods `.enqueue` and `.submit`


freider (approved):
Makes sense! I'm a bit scared though that people will still accidentally call functions like this and now get frustrated with occasional weird errors when the function is executed in the "wrong" environment... In a way, disabling `__call__` and relying being explicit in using `.call()` or `.local_call()` would be "safer" :thinking:

luiscape:
This is a similar interface to Ray's, where `f(...)` is a local call and `f.remote(...)` returns a future.

This is clear to me. UX-wise, it is nice to do an additional operation (i.e. call `f.call(...)`) if I want things executed remotely. With that said, the verb `call` isn't clear. Calling it `f.remote(...)` would make it a copy of Ray's terminology ... but it does make sense?! Thoughts?

erikbern:
Yeah I tend to agree that maybe .remote would make more sense? There is a lot of discussion in https://github.com/modal-labs/modal-client/pull/97 about terminology. These things are possible to change (although require 2-3 months of deprecation). I'm fine opening up the discussion again. However let's keep that separate from this PR

PR #339, opened 2023-02-26
Simplify `modal serve`

erikbern:
Main changes:

1. Use `multiprocessing.Process` with a "spawn" context rather than a subprocess directly. This makes it a bit easier to create a child process. We need to spawn rather than fork in order to reload all modules.
2. Instead of `stub.serve` being a "recursive" function, the live-reload thing sits outside the stub, and the CLI calls it directly.
3. child processes just update the app – they don't run anything other than that. The main process controls heartbeats and logs

The main benefits are:
* shorter code (getting rid of a django utility we used was the majority of the reduction, but there was a bunch of other things too)
* simpler tests (got rid of all mocking)
* tests aren't mocking out the subprocess – they actually launch it properly, and we verify that it works
* added a basic test for py 3.7 and windows
* fixed an issue where tests in `cli_test.py` would fail locally when run in isolation (had to do with implicit fixture dependencies)
* code isn't reading env vars anymore (outside of `modal.config`)
* less code on the `Stub` class (which does too much)
* a more straightforward control flow on the `Stub` class
* only one process pulling logs (should mean no duplicated/missing lines etc)

The reason I started working on this was the last two points – in particular I wanted to refactor `stub._run`, but some complexity with serving made it hard.

(about 50 lines delta of this is a bunch of test skip decorator stuff that's fairly unrelated, but should be harmless)

### DIFF
--- a/client_test/conftest.py
+++ b/client_test/conftest.py
@@ -106,6 +106,8 @@ class MockClientServicer(api_grpc.ModalClientBase):
 
         self.enforce_object_entity = True
 
+        self.app_set_objects_count = 0
+
         @self.function_body
         def default_function_body(*args, **kwargs):
             return sum(arg**2 for arg in args) + sum(value**2 for key, value in kwargs.items())
@@ -155,6 +157,7 @@ class MockClientServicer(api_grpc.ModalClientBase):
     async def AppSetObjects(self, stream):
         request: api_pb2.AppSetObjectsRequest = await stream.recv_message()
         self.app_objects[request.app_id] = dict(request.indexed_object_ids)
+        self.app_set_objects_count += 1
         if request.new_app_state:
             self.app_state[request.app_id] = request.new_app_state
         await stream.send_message(Empty())
@@ -619,7 +622,7 @@ async def aio_container_client(unix_servicer):
 
 
 @pytest_asyncio.fixture(scope="function")
-async def server_url_env(servicer, monkeypatch, set_env_client):
+async def server_url_env(servicer, monkeypatch):
     monkeypatch.setenv("MODAL_SERVER_URL", servicer.remote_addr)
     yield
 
--- a/client_test/conftest.py
+++ b/client_test/conftest.py
@@ -106,6 +106,8 @@ class MockClientServicer(api_grpc.ModalClientBase):
 
         self.enforce_object_entity = True
 
+        self.app_client_disconnect_count = 0
+        self.app_get_logs_initial_count = 0
         self.app_set_objects_count = 0
 
         @self.function_body
@@ -137,11 +139,19 @@ class MockClientServicer(api_grpc.ModalClientBase):
         request: api_pb2.AppClientDisconnectRequest = await stream.recv_message()
         self.requests.append(request)
         self.done = True
+        self.app_client_disconnect_count += 1
         await stream.send_message(Empty())
 
     async def AppGetLogs(self, stream):
-        await stream.recv_message()
+        request: api_pb2.AppGetLogsRequest = await stream.recv_message()
+        if not request.last_entry_id:
+            # Just count initial requests
+            self.app_get_logs_initial_count += 1
+            last_entry_id = "1"
+        else:
+            last_entry_id = str(int(request.last_entry_id) + 1)
         await asyncio.sleep(0.1)
+        await stream.send_message(api_pb2.TaskLogsBatch(entry_id=last_entry_id))
         if self.done:
             await stream.send_message(api_pb2.TaskLogsBatch(app_done=True))
 
--- a/client_test/conftest.py
+++ b/client_test/conftest.py
@@ -627,6 +627,11 @@ async def server_url_env(servicer, monkeypatch):
     yield
 
 
+@pytest_asyncio.fixture(scope="function", autouse=True)
+async def reset_default_client():
+    AioClient.set_env_client(None)
+
+
 @pytest.fixture(name="mock_dir", scope="session")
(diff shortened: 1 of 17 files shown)


erikbern:
@freider do you mind checking if `modal serve` works on windows?

client_test/live_reload_test.py
```diff
@@ -1,75 +1,51 @@
...
-    pass
-
-
-class FakeProcess:
-    def send_signal(self, signal):
-        pass
-
-    def terminate(self):
-        pass
+@skip_windows("windows fails with codec issue (TODO: investigate)")
```
erikbern, line 9: i think because the subprocess is printing emojis in the terminal. super weird.

freider:
Basics seems to work on Windows. Reloads doesn't, but it didn't before either. And there seems to be something preventing Ctrl-C as well, but that also seems to be the case for the current main branch. Looking into it

freider:
Hmm, the test runs fine for me on Windows... 😬 Probably some encoding issue in the setup of the github actions windows env that's different from when I run it. Looking to see if I can repro

erikbern:
@freider was able to fix it by setting `PYTHONIOENCODING` to `utf-8` – tests are now running on Win again!

thundergolfer (approved):
If it passed the reload_providers integration test LGTM. The `ctx.Event` mechanism seems a good bit simpler than the supervising process + child mechanism.

modal/_live_reload.py
```diff
@@ -1,98 +1,94 @@
...
+    p = ctx.Process(target=_run_serve, args=(stub_ref, existing_app_id, is_ready))
+    p.start()
+    await asyncify(is_ready.wait)(timeout)
+    return p
+
+
+async def _terminate(proc: Optional[SpawnProcess], output_mgr: OutputManager, timeout: float = 5.0):
+    if proc is None:
+        return
+    proc.terminate()
```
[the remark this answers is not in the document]
erikbern: sure, i can do that

client_test/conftest.py
```diff
@@ -627,6 +627,11 @@ async def server_url_env(servicer, monkeypatch):
     yield
 
 
+@pytest_asyncio.fixture(scope="function", autouse=True)
+async def reset_default_client():
+    AioClient.set_env_client(None)
```
erikbern, line 632: not sure why it didn't cause issues earlier

pyproject.toml
```diff
@@ -28,6 +28,7 @@ filterwarnings = [
     "error::modal.exception.DeprecationError",
     "ignore::DeprecationWarning:pytest.*:",
     "module:Jupyter is migrating its paths to use standard platformdirs:DeprecationWarning",
+    "ignore:.*pkg_resources.*:DeprecationWarning::",
```
erikbern, line 31: unrelated but needed on my newest dev box

erikbern:
Ok the latest change moves around functionality a bit such that the supervisor process is responsible for logs & heartbeats, and all the child processes do is to live-patch the existing app then terminate.

This also deprecates the `stub.serve` function since it's no longer used by the child processes (they have a separate method that just patches an app then quits).

I think this should flow is more conceptually clean imho.

pyproject.toml
thundergolfer: I got the same today.

client_test/live_reload_test.py
```diff
@@ -1,75 +1,48 @@
...
+    run_serve_loop(stub_file, _watcher=fake_watch())
+    assert servicer.app_set_objects_count == 4  # 1 + number of file changes
 
 
 @pytest.mark.asyncio
-async def test_reloadable_serve_ignores_file_changes(client, monkeypatch, servicer, test_dir):
-    async def fake_watch(stub, output_mgr, timeout):
+async def test_no_change(test_dir, server_url_env, servicer):
+    async def fake_watch():
         # Iterator that never yields
```
thundergolfer, line 31:
```suggestion
        # Iterator that returns immediately, yielding nothing
```

modal/_live_reload.py
```diff
@@ -1,98 +1,101 @@
...
-        exe_entrypoint = py_script.with_suffix(".exe")
-        if exe_entrypoint.exists():
-            # Should be executed directly, ignoring sys.executable.
-            return [exe_entrypoint, *sys.argv[1:]]
-        script_entrypoint = py_script.with_name("%s-script.py" % py_script.name)
-        if script_entrypoint.exists():
-            # Should be executed as usual.
-            return [*args, script_entrypoint, *sys.argv[1:]]
-        raise RuntimeError("Script %s does not exist." % py_script)
+            output_mgr.print_if_visible(f"[red]Serve process {proc.pid} didn't terminate after {timeout}s, killing it")
```
thundergolfer, line 47:
```suggestion
            output_mgr.print_if_visible(f"[red]Serve process {proc.pid} didn't terminate after {timeout}s, killing it[/red]")
```

modal/stub.py
```diff
@@ -351,18 +350,34 @@ async def run(self, client=None, stdout=None, show_progress=None, detach=False)
...
-        **Note:** live-reloading is not supported on Python 3.7. Please upgrade to Python 3.8+.
-        """
-        from ._watcher import watch
-
+    async def _serve_update(
+        self,
+        existing_app_id: str,
+        is_ready: Event,
+    ) -> None:
+        # Used by child process to redeploy an app
```
thundergolfer, line 358:
```suggestion
        # Used by child process to reinitialize a served app
```

thundergolfer (approved):

PR #340, opened 2023-02-27
Remove rate_limit arg

aksh-at:
This argument is being fast-deprecated. Prior users have already been informed.


erikbern:
would have had a slight preference for keeping the argument but throwing a `deprecation_error` when used

PR #344, opened 2023-02-28
`.cancel()` method for FunctionCall(Handle)

freider:


pyproject.toml
```diff
@@ -27,6 +27,8 @@ filterwarnings = [
     "error::DeprecationWarning",
     "error::modal.exception.DeprecationError",
     "ignore::DeprecationWarning:pytest.*:",
+    "ignore::DeprecationWarning:pkg_resources.*:",
```
freider, line 30: Not sure why, but on a fresh dev box I got a bunch of new deprecationwarnings due to some pkg_resources usage in google.rpc... Had to had these lines to get the tests to work.

erikbern: makes three of us – jonathon and independently made the same change in other PRs!

erikbern (approved):

PR #348, opened 2023-02-28
fix: descriptions of apps created by modal CLI. crude impl

thundergolfer:
This afternoon may end up having been a waste of time, as this fix is pretty crude. 

The issue is that after moving away from `python3 app.py` invocation to usage of the `modal` CLI, unnamed ephemeral apps get a name which often includes CLI window dressing:

* `modal run dee.py...`
* `modal run --detached dee.py...`
* `modal serve dee.py...`

image

Some users apparently find the inclusion of program args in a stub's `description` useful, but I doubt any find the CLI stuff useful.

---

The implementation of the fix is quite ugly though. It uses a `global`. I think most of the complication comes from the fact that we don't create a new process via the CLI, it's the same process, unlike say `poetry run ...`. 

If anyone has a better implementation than this that'd be great.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -67,6 +67,16 @@ def _add_click_options(func, signature: inspect.Signature):
     return func
 
 
+def _get_clean_stub_description(func_ref: str) -> str:
+    # If possible, consider the 'ref' argument the start of the app's args. Everything
+    # before it Modal CLI cruft (eg. `modal run --detach`).
+    try:
+        func_ref_arg_idx = sys.argv.index(func_ref)
+        return " ".join(sys.argv[func_ref_arg_idx:])
+    except ValueError:
+        return " ".join(sys.argv)
+
+
 def _get_click_command_for_function(_stub, function_tag):
     blocking_stub = synchronizer._translate_out(_stub, Interface.BLOCKING)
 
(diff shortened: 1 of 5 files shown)


modal/cli/run.py
```diff
@@ -18,6 +19,9 @@
 
 run_cli = typer.Typer(name="run")
 
+# Used to set Modal cleaner app descriptions which don't include modal CLI parts, only
+# user code parts.
+modal_cli_app_desc: str = None
```
freider, line 24: A possible alternative to avoid using the global might be to let `_get_click_command_for_function` set the description on the stub (if it isn't already set) before running it? Then we can just pass the "default" description to _get_click_command_for_function below?
It would require moving the `_infer_app_desc` to some later point of execution though (like when accessing .description) instead of running it at stub __init__ like we do now though

freider (commented):
See inlined comment for idea

modal/cli/run.py
thundergolfer: Thanks for the idea! I'll give it a shot because adding globals is 🤮

thundergolfer: @freider alright I implemented something along the lines of your suggestion. It's not an equivalent implementation though, because the `global` mechanism allowed the stub to have the correct description as soon as it was imported. See https://github.com/modal-labs/modal-client/pull/348/files#diff-d1f9598cb3b725966536fc17038e0b6c9eaf84409994a573a248b3da948f74caR9.

modal/cli/run.py
```diff
@@ -63,6 +67,16 @@ def _add_click_options(func, signature: inspect.Signature):
     return func
 
 
+def _get_clean_stub_description(func_ref: str) -> str:
```
freider, line 70: Duplicate method from above?

modal/stub.py
```diff
@@ -123,7 +123,7 @@ def __init__(
         if name is not None:
             self._description = name
         else:
-            self._description = self._infer_app_desc()
+            self._description = None
```
freider, line 126: nit: this if/else is now the same as `self._description = name` right?

modal/cli/run.py
```diff
@@ -19,6 +20,9 @@
 
 run_cli = typer.Typer(name="run")
 
+# Used to set Modal cleaner app descriptions which don't include modal CLI parts, only
+# user code parts.
+modal_cli_app_desc: str = None
```
freider, line 25: This should be removable now right?

freider (commented):
see inline comments

modal/cli/run.py
thundergolfer: Yeh I just copied the fn rather than having it in one place and importing it. It's really simple so this seems fine? Rule of three and all that, I could centralize it then.

modal/cli/run.py
thundergolfer: Woops, sloppy.

freider (approved):

PR #350, opened 2023-03-01
Adds experimental support for cancelling individual inputs

freider:
Using sigusr1, an exception is thrown into the user(main) thread in case of sync code, and a CancelledError is thrown in case the user code is async

Also adds some new container tests that check that signal handling works as expected for sigusr1 (might be nice to add tests for sigint/sigterm as well

There is no code in the worker yet to actually send these signals (there are some non-trivial prerequisites there), so we could either merge this and have it be unused (albeit tested) functionality for a bit, or keep the PR until we have the necessary support on the worker?

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -76,6 +76,9 @@ def get_is_async(function):
         raise RuntimeError(f"Function {function} is a strange type {type(function)}")
 
 
+INPUT_CANCELLATION_MESSAGE = "skip-input"
+
+
 def run_with_signal_handler(coro):
     """Execute coro in an event loop, with a signal handler that cancels
     the task in the case of SIGINT or SIGTERM. Prevents stray cancellation errors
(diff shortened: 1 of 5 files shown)


modal/_container_entrypoint.py
```diff
@@ -514,6 +532,12 @@ def import_function(function_def: api_pb2.Function, ser_cls, ser_fun) -> tuple[A
         return obj, fun, is_async
 
 
+class InputCancellation(BaseException):
```
erikbern, line 535: maybe define it in `modal.exception` so that users can import it

erikbern:
Nice! Might make sense to ship this so that we have it present in the client even without the worker support. That way we have the optionality to implement the worker side of it later.

mwaskom:
Doing some PR cleanup; I'm guessing this has been superseded by the work you're currently doing to going to close.

PR #359, opened 2023-03-03
don't overload ImageJoinResponse

erikbern:
I caused some internal code to break in https://github.com/modal-labs/modal-client/pull/358


danielnorberg (approved):

danielnorberg:
Should we also add a new `ImageJoinStreamingRequest` and remove `last_entry_id` from `ImageJoinRequest`?

PR #360, opened 2023-03-03
do not overload ImageJoinRequest

danielnorberg:
See https://github.com/modal-labs/modal-client/pull/359, fix internal breakage due to https://github.com/modal-labs/modal-client/pull/358


freider:
looks good to me :)

PR #365, opened 2023-03-04
image snapshots progress bar

erikbern:
I realized the app logs were driving the image snapshot progress so I had to rewrite this (since I'm planning to not use app logs until the app is actually _running_)

This was a bit messier than I thought but I think it logically makes sense. Feels like we've been shoving all kinds of stuff into the app logs and at least this is one less item. This also makes it slightly easier to use progress bars for other things (e.g. mount uploads)


erikbern:
Since the revert in #366, this doesn't cleanly apply anymore. I would have to reapply reverted changes.

But I'm going to take this as an opportunity to approach this in a slightly different direction!

PR #369, opened 2023-03-06
ImageJoinStreaming: attempt 2

erikbern:
New attempt to do #363 and #365 but doing it in a slightly different way – this time just relying on the existing output code in the output manager.


erikbern:
This one surfaces a few janky things with the output code, which we probably need to clean up. I'll do some of it in later PRs, although this rabbit hole is starting to get pretty deep so we'll see.

PR #373, opened 2023-03-07
get deployment out of _run

erikbern:
`Stub.deploy` no longer uses `Stub._run` after this change.

This also resolves a minor issue where we currently print a spurious warning `Timed out waiting for logs. View logs at https://modal.com/logs/ap-Z28N8kYIalZGaZM6XJSWVN for remaining output.` (not sure if that's a recent regression or has been there before).

A nice improvement is we no longer rely on the cancellation of the logs task (during deploys) – we don't have to run a logs task since image build logs are separate now.

### DIFF
--- a/modal/_live_reload.py
+++ b/modal/_live_reload.py
@@ -86,14 +86,14 @@ async def _run_serve_loop(
     app = await _App._init_new(client, stub.description, detach=False, deploying=False)
 
     if unsupported_msg:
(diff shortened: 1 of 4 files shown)


erikbern:
Merging this – it should be a "pure" refactoring with zero behavioral change (other than fixing a very minor issue). I've also run a bunch of integration tests to verify it doesn't break anything.

PR #374, opened 2023-03-07
Only `Stub.run` can call `Stub._run_ephemeral`

erikbern:
This is a sub-step where next step is I'll merge `_run_ephemeral` into `run`. Breaking it up into two PRs just to keep the changes a bit smaller.


erikbern:
I can't find any client code other than `stub.serve` that uses the `serve_timeout` config. But `stub.serve` is deprecated. Not sure how the synthetic monitors work? at least the `badges` example seems to run `modal serve`

PR #376, opened 2023-03-08
`retry_transient_errors` coverage

aksh-at:
We can do this automatically (by always retrying unary methods in `grpc_utils.py`). But quick fix for now.


erikbern:
looks good. have these things been flaky though?

PR #377, opened 2023-03-08
Move run code into a separate module

erikbern:
stub.py has gotten very large and some of the functionality isn't really needed on the stub itself imo.

Since I'm planning to merge apps and stubs, it's helpful to get a bunch of stuff out of the Stub class. All the running-related code doesn't interact much with the stub other than getting a few objects, so I'm moving it to its own module.

I kept the docstrings on the Stub class for now, but that's another ~50 lines we could remove later if we deprecate the methods on the Stub class (which I think we probably should).

pure refactoring, zero behavioral change


erikbern:
stub.py is still brutally large, but this shrinks it from 835 to 703 lines, so that's a 16% reduction. I'll take it!

PR #379, opened 2023-03-09
fix: get_call_graph TypeError 

thundergolfer:
[modalbetatesters.slack.com/archives/C031Z7H15DG/p1678322711034919](https://modalbetatesters.slack.com/archives/C031Z7H15DG/p1678322711034919)

Looks like there's an unhandled state, TERMINATED, which is triggering `_missing_`, and that doesn't have the `@classmethod` annotation so it's not passing `cls` as first arg. 

I checked some relevant inputs and they do indeed have state = 3. I think that this state might be new for inputs, by way of input cancellation? cc @freider.


freider:
Yeah inputs that are cancelled by the user are marked as TERMINATED. Actually realizing now that we should probably use the call graph to automatically cancel any subcalls made by cancelled function calls as well!

PR #381, opened 2023-03-09
Fix modal shell

aksh-at:
Some of this logic broke in the OutputManager refactor. Will look into adding a simple test for it.

---

Resolves MOD-978


erikbern:
+1 to test!

PR #385, opened 2023-03-10
Handle stderr in CLI tests

erikbern:
I was hoping this would output the async traceback stuff, but I believe click doesn't run things in a subprocess so it doesn't properly capture the shutdown behavior. But this PR probably has some value for unrelated issues so I'm putting it up here.

### DIFF
--- a/client_test/cli_test.py
+++ b/client_test/cli_test.py
@@ -5,6 +5,7 @@ import traceback
 import unittest.mock
 from contextlib import asynccontextmanager
 from unittest import mock
+from typing import List, Optional
 
 import click
 import click.testing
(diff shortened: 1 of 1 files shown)


freider:
Oh I think I might have a branch locally that actually tests in a real sub process. Let me check!

freider:
I can't find it but I'm like 90% sure I wrote tests for that when I did the synchronicity patch that ensures to propagate the cancellations to the event loop... :thinking: Might have accidentally nuked it from my working tree :shrug:

freider (approved):

erikbern:
There's some stuff here that uses a subprocess: https://github.com/modal-labs/modal-client/blob/main/client_test/e2e_test.py

Will add something that runs `modal run`

erikbern:
Actually, I'm stupid. The error we're looking for is printed inside the containers, not by the client.

PR #386, opened 2023-03-10
Don't create FunctionInfo for local_entrypoint

aksh-at:
Quick fix for internal doc tests to pass. Creating a `FunctionInfo` from within doctests tries to serialize the entrypoint function, which fails because it refers to another Modal function. 

Longer-term fix is making serialized functions able to call each other, however in this case it seems unnecessary to construct the entire `FunctionInfo` object if all we need is the name.

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -422,8 +422,8 @@ class _Stub:
         information on usage.
 
         """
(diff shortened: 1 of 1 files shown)


freider:
:+1:

PR #396, opened 2023-03-14
Adds Modal user agent header data

luiscape:
This will appear as `modal-client/$VERSION`. Example:

```
modal-client/0.47.1573 (darwin; cpython/3.9.13)
```
Screenshot 2023-03-14 at 11 02 01


erikbern:
Could probably test this by inspecting the metadata in conftest.py but not super important. I can spend 3 min on it.

erikbern:
Turns out it was trivial: https://github.com/modal-labs/modal-client/pull/397

PR #398, opened 2023-03-14
Validation and error message for stub._blueprint objects

aksh-at:
Fixes the previously cryptic `AttributeError: 'int' object has no attribute 'local_uuid'` error.

Also sneaking in a better exception for CLI imports.

### DIFF
--- a/modal/cli/import_refs.py
+++ b/modal/cli/import_refs.py
@@ -33,6 +33,8 @@ class ImportRef:
 def parse_import_ref(object_ref: str) -> ImportRef:
     if object_ref.find("::") > 1:
         file_or_module, object_path = object_ref.split("::", 1)
+    elif object_ref.find(":") > 1:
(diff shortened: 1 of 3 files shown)


modal/cli/import_refs.py
```diff
@@ -71,6 +73,7 @@ def import_file_or_module(file_or_module: str):
 
         # Import the module - see https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
         spec = importlib.util.spec_from_file_location(module_name, file_or_module)
+        print("importing", spec, module_name, file_or_module)
```
erikbern, line 76: debug print?

erikbern:
Feels like we could automate a lot of the type checking – I mentioned it in https://linear.app/modal-labs/issue/MOD-763/runtime-type-checking-for-client-package-usage

modal/cli/import_refs.py
aksh-at: oops 🤦

PR #400, opened 2023-03-14
Class decorator v0

erikbern:
The main benefit is for container lifecycle management. Instead of the awkward (imo) syntax

```python
class FooModel:
    def __enter__(self):
        self.model = load_model()

    @stub.function(cpu=8)
    def predict(self, x):
        return self.model(x)
```

this PR enables you to write this:

```python
@stub.cls(cpu=8)
class FooModel:
    def __enter__(self):
        self.model = load_model()

    @stub.method()
    def predict(self, x):
        return self.model(x)
```

Slightly bigger example:

```python
import modal

stub = modal.Stub()

@stub.cls(cpu=8)
class Foo:
    @stub.method()
    def bar(self, x):
        print("bar!", x)
        return x ** 2

    @stub.method()
    def baz(self, x):
        return x ** 3


@stub.local_entrypoint
def run():
    foo = Foo()
    assert foo.bar(42) == 1764
    assert foo.baz(100) == 1000000
```

#  TESTS MISSING

* [X] Serialization
* [X] Lifecycle methods (container test)
* [X] Test that you can call one method from another method (both using `.call` and "internally")
* [X] Web endpoint tests

# Questions?

* Should we require some special syntax for the constructor
* ~Might not want to wrap special (dunder) methods~
* ~I just shoved all the decorator args for `stub.cls` into a `kwargs` but we should expand for better docs~
* ~Naming? "service" isn't maybe ideal~ – edit: going for `cls`
* Make sure the constructor takes no arguments

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -863,20 +863,20 @@ class _Stub:
             await wrapped_fn.call(cmd)
 
     @decorator_with_options_unsupported
-    def cls(self, cls=None, **kwargs):
+    def cls(self, user_cls=None, **kwargs):
         # TOOD(erikbern): include all the docstring from stub.function
 
-        # All methods to know what class they are on
-        kwargs.update(_cls=cls)
+        # All methods need to know what class they are on
+        kwargs.update(_cls=user_cls)
 
-        function_handles = {}
-        for k, v in cls.__dict__.items():
+        function_handles: Dict[str, _FunctionHandle] = {}
+        for k, v in user_cls.__dict__.items():
             if isinstance(v, (PartialFunction, AioPartialFunction)):
                 partial_function = synchronizer._translate_in(v)  # TODO: remove need for?
                 function_handles[k] = self.function(**kwargs)(partial_function)
 
-        cls._modal_function_handles = function_handles
-        return cls
+        _PartialFunction.initialize_cls(user_cls, function_handles)
+        return user_cls
 
 
 Stub, AioStub = synchronize_apis(_Stub)
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -26,7 +26,7 @@ from .app import _App, _container_app, is_local
 from .client import _Client
 from .config import logger
 from .exception import InvalidError, deprecation_warning
-from .functions import _Function, _FunctionHandle, class_decorator
+from .functions import _Function, _FunctionHandle, class_decorator, WebhookConfig
 from .gpu import GPU_T
 from .image import _Image, _ImageHandle
 from .mount import _get_client_mount, _Mount
@@ -51,12 +51,6 @@ class LocalEntrypoint:
 
     def __call__(self, *args, **kwargs):
         return self.raw_f(*args, **kwargs)
-
-
-class WebhookConfig:
-    def __init__(self, raw_f: Callable[..., Any], webhook_config: api_pb2.WebhookConfig):
-        self.raw_f = raw_f
-        self.webhook_config = webhook_config
 
 
 def check_sequence(items: typing.Sequence[typing.Any], item_type: typing.Type[typing.Any], error_msg: str):
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -856,7 +856,7 @@ class _Stub:
 
     @synchronizer.nowrap
     @decorator_with_options
-    def service(self, cls=None, **kwargs):
+    def cls(self, cls=None, **kwargs):
         # TOOD(erikbern): include all the docstring from stub.function
 
         # All methods to know what class they are on
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -862,7 +862,7 @@ class _Stub:
 
     @synchronizer.nowrap
     @decorator_with_options
-    def service(self, cls, **kwargs):
+    def service(self, cls=None, **kwargs):
         # TOOD(erikbern): include all the docstring from stub.function
 
         # All methods to know what class they are on
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -646,13 +646,6 @@ class _Stub:
         return function_handle
 
     @decorator_with_options_unsupported
-    def method(
-        self,
-        raw_f: Optional[Callable[..., Any]] = None,
-    ):  # TODO: return type!
-        return _PartialFunction(raw_f)
-
-    @decorator_with_options_unsupported
     @typechecked
     def web_endpoint(
         self,
@@ -663,6 +656,7 @@ class _Stub:
         ] = None,  # Label for created endpoint. Final subdomain will be <workspace>--<label>.modal.run.
         wait_for_response: bool = True,  # Whether requests should wait for and return the function response.
     ):  # TODO: return type!
+        # TODO(erikbern): let's move this out to not sit on the stub
         """Register a basic web endpoint with this application.
 
         This is the simple way to create a web endpoint on Modal. The function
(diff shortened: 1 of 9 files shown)


erikbern:
Running into an annoying issue with mypy for this. It doesn't handle class decorators and thinks the new class is the same as the un-decorated class. So none of the methods end up having a `.call` method on them, and this causes mypy to flag `my_obj.f.call()` etc.

This is maybe not a super big deal – I sort of doubt that many users will run mypy on user code. It only shows up in the tests for us. So maybe we could have some special decorator that fixes these issues later.

aksh-at (approved):
Nice, thanks for all the tests! I agree `@stub.service` is not amazing but can't think of anything definitively better. One idea is to just use `@stub.class` — it parallels `@stub.function`, and comes with less connotations about how it works.

The mypy thing is unfortunate. Could we make a [mypy plugin](https://mypy.readthedocs.io/en/stable/extending_mypy.html#configuring-mypy-to-use-plugins) at some point?

Also outside of scope of this, but wonder if we should override `__setattr__` for the class instance inside the container, so people don't try and mutate `self` and it becomes read-only? (One of the worries with the class-based syntax is that people don't realize the state is just local to the container)

modal/stub.py
```diff
@@ -857,5 +854,19 @@ async def interactive_shell(self, cmd=None, image=None, **kwargs):
         async with self.run():
             await wrapped_fn.call(cmd)
 
+    @synchronizer.nowrap
+    @decorator_with_options
```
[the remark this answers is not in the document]
erikbern: yeah for sure – thanks for noticing

erikbern:
> One idea is to just use `@stub.class`

unfortunately `class` is a reserved keyword in Python, so we can't use that (that's why variables denoting classes are usually written as `klass` or `cls`)

> The mypy thing is unfortunate. Could we make a [mypy plugin](https://mypy.readthedocs.io/en/stable/extending_mypy.html#configuring-mypy-to-use-plugins) at some point?

might work, haven't looked into it. another option is that we can have a `@method` decorator that doesn't do a whole lot except return the same `_FunctionHandle`

> we should override __setattr__ for the class instance

we want the state to be a bit mutable though? like, people definitely want to use it for initialization etc

aksh-at:
> unfortunately class is a reserved keyword in Python, so we can't use that (that's why variables denoting classes are usually written as klass or cls)

Good point, what about `@stub.cls` then? (and eventually `@app.cls`)

> we want the state to be a bit mutable though? like, people definitely want to use it for initialization etc

Yeah it would be nice if it's writable during `__enter__` but read-only for the actual functions. Maybe not worth it.

freider:
A few name suggestions: `runner`, `task`, `container` - all sort of pointing toward the same concept as `service` but in a more ephemeral way, and making it "intuitive" that this is where you can declare life cycle hooks etc.

Apart from that I would still lean toward making "modalified methods" explicit with an `@method`, both for the static typing it can provide and to make it less magic which methods are and which ones are not `(__init__, __enter__, __aenter__)`. Also makes it easy for users to create "non-modal" wrappers around modal methods on the same class (and the other way around).

As for mypy plugins, I did look at that at some point @aksh-at , but the biggest problem is that they would only be supported by a fraction of the python tooling that exists as far as I understand, since different tools use different type checkers (besides mypy there is also pyright, pylance (vscode) and intellijs own type checker, which don't support mypy plugins), so I think it's best to stay on vanilla typing + `typing_extensions` as far as possible.

erikbern:
> A few name suggestions: `runner`, `task`, `container` - all sort of pointing toward the same concept as `service` but in a more ephemeral way, and making it "intuitive" that this is where you can declare life cycle hooks etc.

What's confusing about those terms is that they indicate that the user is running a single container (like a stateful actor) but in reality it's a swarm of autoscaled containers together (more like an autoscale kubernetes server or similar).

My preference right now is probably `@stub.cls(...)`. Kind of minor but if we abbreviate `class` then I think maybe we should do the same with `stub.function` -> `stub.fun` and maybe also `meth` for methods although that's kind of lol

erikbern:
Planning to add a `@method` decorator before I merge this. Just thinking it through here:

* It would be nice if the class decorator doesn't have to create a new class, but could mutate the class in-place
* Ideally, it wouldn't have to mutate the class, but just the methods themselves
* However, that's sort of annoying, because that would require the methods to be `_FunctionHandles` that are updated in-place
* So, it's easier if the methods are some special `Method` class
* The class decorator would then traverse `__dict__` of the class and mutate it, replacing every `Method` with a `FunctionHandle`
* In order for types to propagate properly, `Method` would have to implement a similar interface to `_FunctionHandle`
* In fact, it would be nice if both `Method` and `FunctionHandle` inherit from the same base class, which may or may not derive from `Object`

freider:
One idea would be to make Method a very thin descriptor protocol that
returns FunctionHandles when attached to a class instance, using the cls
definition as it exist on the instance. Would be very light and need no
interface apart from the descriptor protocol I think, and no messing with
the dict. The actual function handles would be precreated for each method
in class creation (so the modal app knows about them)

fre 14 apr. 2023 kl. 22:42 skrev Erik Bernhardsson ***@***.***

erikbern:
Ok, this works now, but with some caveats:
* It still has the janky `.nowrap` thing
* The `cls` decorator passes a bunch of kwargs to the function
* It returns a new class instead of mutating the class in-place

I think I can potentially fix these things, and it might include the descriptor protocol :)

modal/stub.py
```diff
@@ -642,11 +639,19 @@ def function(
...
         )
 
         self._add_function(function, [*base_mounts, *mounts])
         return function_handle
 
+    @decorator_with_options_unsupported
+    def method(
+        self,
+        raw_f: Optional[Callable[..., Any]] = None,
+    ):  # TODO: return type!
```
erikbern, line 652: actually not sure if we can have a return type here, since the methods are "translated" from a `_PartialFunction` to a `_FunctionHandle` in runtime (using `_PartialFunction.__get__`). 

I think the only way to solve this is to have a shared base class that both derive from, and define that as the return type here.

erikbern:
Btw, minor preference for moving `method` and `web_endpoint` out of the stub (which it doesn't really need) just since the `Stub` class is so big. wdyt?

freider:
I like the idea of moving them out of the stub, if nothing else to get a bit more concise syntax by importing them in module scope and doing:
```python
class Foo:
     @method
     def foo(self):
        ...
```

modal/stub.py
freider: Static type checkers tend to support Descriptors nowadays, so you should be able to have `_PartialFunction` as the return type as long as the `__get__` method on the descriptor has `_FunctionHandle` as its return type. I experimented a bit with this a while ago and it seemed to work, at least in VSCode.

(Using `@typing.overload` on `__get__`, VSCode was even able to infer difference in types between `MyClass.my_descriptor` and `MyClass().my_descriptor`!)

modal/stub.py
```diff
@@ -572,29 +567,31 @@ def function(
         name: Optional[str] = None,  # Sets the Modal name of the function within the stub
         is_generator: Optional[bool] = None,  # If not set, it's inferred from the function signature
         cloud: Optional[str] = None,  # Cloud provider to run the function on. Possible values are aws, gcp, auto.
+        _cls: Optional[type] = None,  # Used for methods only
```
freider, line 570: Neat

freider (approved):
Lgtm! I'm in favor of moving out `method` as a standalone function before merging this (and move `web_endpoint` in a followup, since that has a deprecation path)

modal/stub.py
erikbern: Oh very interesting. Let me annotate `__get__` on `PartialFunction` then!

PR #401, opened 2023-03-15
Re-enable state management for AppSetObjects

freider:
Also adds a bunch of tests replicating state management from backend, to make sure we don't regress this behavior again


erikbern (approved):
nice!

PR #404, opened 2023-03-15
remove/clarify deprecated fields in protobuf

erikbern:


modal_proto/api.proto
```diff
@@ -210,7 +209,7 @@ message ImageRegistryConfig {
 message BaseImage {
   string image_id = 1;
   string docker_tag = 2;
-  // fields 3 and 4 are deprecated
+  reserved 4;
```
freider, line 212: I think you can even do `reserved 3-4;`

freider (approved):

modal_proto/api.proto
erikbern: yeah but i think it's implied that we already had field 3? feels like the main utility of reserving fields is when you deprecate the _last_ field in a message, to make sure it doesn't get reused

PR #406, opened 2023-03-15
fix: restore still-used MODAL_SERVE_TIMEOUT functionality

thundergolfer:
Found the reason our webhook benchmarks/demos were failing, which was partially hidden by _another_ client bug. We need this functionality in synmon still.

cc @erikbern unless you actually want to be rid of this setting.


erikbern:
this looks good. i was looking all over to figure out where it was used but couldn't find it – thought it was strange.

PR #407, opened 2023-03-15
fix: restore 'Initialized...' message in app runs

thundergolfer:


erikbern:
Nice! accidentally wiped it when I moved a bunch of code

erikbern:
Maybe good to add a test btw

thundergolfer:
Yeah agree, I just wanted to timebox this. Added a note to come back to it.

PR #408, opened 2023-03-16
Make description less spammy in containers

freider:
Is probably only used/emitted when using run_function build steps, but previously it looked like an error


erikbern:
also happens if you have multiple stubs in the same code base!

PR #409, opened 2023-03-16
Revert regression in .image assigment on stub

freider:
Maywe we should deprecate this syntax though?


erikbern:
yeah i'm open to other options

PR #417, opened 2023-03-17
`modal run --quiet`

aksh-at:


Flaque:
✨

client_test/cli_test.py
```diff
@@ -165,6 +165,13 @@ def test_run_detach(servicer, set_env_client, test_dir):
     assert servicer.app_state_history["ap-1"] == [api_pb2.APP_STATE_INITIALIZING, api_pb2.APP_STATE_DETACHED]
 
 
+def test_run_quiet(servicer, set_env_client, test_dir):
+    stub_file = test_dir / "supports" / "app_run_tests" / "default_stub.py"
+    # Just tests that the command runs without error for now (tests end up defaulting to `show_progress=False` anyway,
+    # without a TTY).
+    _run(["run", "--quiet", stub_file.as_posix()])
```
erikbern, line 172: you could maybe add an `expected_stdout` argument to `_run` and pass the empty string here

erikbern: oh i see, maybe it's kind of pointless then. I guess in order to truly test it, you would have to put it in e2e_test.py instead (where it runs stuff in a subprocess). but not important.

PR #420, opened 2023-03-21
use latest synchronicity

erikbern:
0.3.* isn't backwards compatible with 0.2.* but this PR makes it work.


erikbern:
rebased this on #422 which will cause this to fail because it includes a unit test for a regression

erikbern:
This passes all the integration tests too.

PR #424, opened 2023-03-22
Don't use typeguard in container clients

freider:
It adds a bunch of overhead on import by rewriting all of the annotated functions, so lets only use it on the client side where it is most likely to catch errors "early"


modal/_types.py
```diff
@@ -0,0 +1,17 @@
...
+from typing import TypeVar
+
+F = TypeVar("F")
+
+
+def typechecked(f: F) -> F:
+    # type checking function adds significant overhead when
+    # the typeguard.typechecked is applied, so only do it locally
+    # where it matters the most
+    is_local = not bool(os.environ.get("MODAL_IMAGE_ID", ""))
```
erikbern, line 11: i think you can use `config.get("image_id")` also

freider: nice!

PR #432, opened 2023-03-27
pass function call type (map/unary) to map request

gongy:


modal_proto/api.proto
```diff
@@ -473,6 +479,7 @@ message FunctionMapRequest {
   string function_id = 1;
   string parent_input_id = 2;
   bool return_exceptions = 3;
```
gongy, line 481: Would we rather have an extra
* `FUNCTION_CALL_TYPE_MAP_RETURN_EXCEPTIONS`
or distinguish with the boolean?

PR #434, opened 2023-03-27
web endpoint refactor

erikbern:
This breaks up the web decorators so you have to do

```python
@stub.function(cpu=42)
@stub.web_endpoint(method="POST")
def my_webhook(req):
    ...
```

Note that this is required even if without function arguments:

```python
@stub.function
@stub.web_endpoint(method="POST")
def my_webhook(req):
    ...
```

The new methods `web_endpoint`, `asgi_app`, and `wsgi_app` supersedes the old methods `webhook`, `asgi`, and `wsgi`, but I'm keeping them for now. I'll deprecate them shortly, but I want to update all integration tests first.

We could maybe make the `stub.web_endpoint` smart so that it doesn't require a bare `@stub.function` if there are no arguments, but I actually think that would make it more confusing – users are going to start out with `@stub.web_endpoint` and then it's going to be confusing where to put any function arguments. If we enforce a `@stub.function` everywhere then it's clear where to put these. That being said, I think we should _catch_ any time the user forgets a `@stub.function` and output an error message. I'll add that.

This supersedes #425. From that PR description: the benefits of this approach are:

* Less repetitive `Stub` class
* Better support in class decorators – see https://github.com/modal-labs/modal-client/pull/400
* Every webhook is now also a function you can call internally – unlike previously, where this wasn't possible

Note that this is already a reduction in code, and we'll be able to delete another ~100 lines once we remove support for the old methods.

TODO
====

- [X] Make sure there's a warning/exception if you apply the decorators in the wrong order
- [X] Make sure there's a warning/exception if the user only has a `web_endpoint` decorator but not a `function` decorator
- [X] Add tests for `asgi_app` and `wsgi_app` (we didn't have any so far!)

### DIFF
--- a/client_test/container_test.py
+++ b/client_test/container_test.py
@@ -259,18 +259,22 @@ def test_create_package_mounts_inside_container(unix_servicer, event_loop):
     assert items[0].result.data == serialize(0)
 
 
-@skip_windows_unix_socket
-def test_webhook(unix_servicer, event_loop):
+def _get_web_inputs(path="/"):
     scope = {
         "method": "GET",
         "type": "http",
-        "path": "/",
+        "path": path,
         "headers": {},
         "query_string": "arg=space",
         "http_version": "2",
     }
     body = b""
-    inputs = _get_inputs(([scope, body], {}))
+    return _get_inputs(([scope, body], {}))
+
+
+@skip_windows_unix_socket
+def test_webhook(unix_servicer, event_loop):
+    inputs = _get_web_inputs()
     client, items = _run_container(
         unix_servicer,
         "modal_test_support.functions",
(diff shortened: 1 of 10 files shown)


aksh-at:
Neat! What do you think of inverting the order of the decorators? This seems clearer to me:

```python
@stub.web_endpoint(method="POST")
@stub.function(cpu=42)
def f():
```

(I think because `web_endpoint` seems like an "optional add-on" to `function`, so it should go last)

erikbern:
Also want to point out that this PR doesn't close the door on supporting the reverse order later. It's just a bit annoying to implement right now. I think it's more important to get something out so that we can start to deprecate the old syntax. But supporting the reverse order would be 100% backwards compatible so we can always add that later.

erikbern:
Realized there's a few other tests I need to update, plus add tests for the deprecated syntax too

erikbern:
one more thing – stubs fail against prod right now with this:

```
Request failed with exception: InvalidError("Async (non-generator) function returned value of type <class 'async_generator'> You might need to use @stub.function(..., is_generator=True).") (execution time: 4.4 ms, total latency: 2347.7 ms)
```

will fix this

erikbern:
Rebased this on #443

erikbern:
rebased on #443

erikbern:
Passes all integration tests as well. Will merge early next week!

erikbern:
Rebased on all the type changes by @freider

This passes all integration tests – going to merge this then starting to deprecate the old syntax!

PR #435, opened 2023-03-28
Adjustments to allow type stub generation

freider:
Some adjustments to function signatures to make sure they can be used to create type stubs using new synchronicity patch.

Prerequisite: new synchronicity version supporting target module specification on create_blocking etc.

### DIFF
--- a/modal/object.py
+++ b/modal/object.py
@@ -321,5 +321,5 @@ class _Provider(Generic[H]):
 
 
 # Dumb but needed becauase it's in the hierarchy
-Generic, AioGeneric = synchronize_apis(Generic, __name__)  # erases base Generic type...
+synchronize_apis(Generic, __name__)  # erases base Generic type...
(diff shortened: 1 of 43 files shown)


modal/object.py
```diff
@@ -125,7 +125,8 @@ async def from_app(
         return handle
 
 
-synchronize_apis(Handle)
+synchronize_apis(Handle, target_module=__name__)
+AioHandle = Handle
```
erikbern, line 129: this doesn't seem right? `Handle` and `AioHandle` aren't the same classes

modal/object.py
```diff
@@ -319,4 +320,5 @@ async def _exists(
                 raise
 
 
-synchronize_apis(Provider)
+synchronize_apis(Provider, target_module=__name__)
+AioProvider = Provider
```
erikbern, line 324: also doesn't seem right

erikbern (approved):
lgtm except comments!

PR #436, opened 2023-03-28
Unwrap `UserCodeException` when return_exceptions=True

aksh-at:
Noticed we were propagating UserCodeExceptions from synchronicity without unwrapping them (these are meant to be internal wrappers only).

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -393,7 +393,10 @@ async def _map_invocation(
             output = await _process_result(item.result, client.stub, client)
         except Exception as e:
             if return_exceptions:
-                output = e
+                if isinstance(e, UserCodeException):
+                    output = e.exc
+                else:
+                    output = e
             else:
                 raise e
         return (item.idx, output)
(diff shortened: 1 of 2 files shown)


modal/functions.py
```diff
@@ -393,7 +393,10 @@ async def fetch_output(item):
             output = await _process_result(item.result, client.stub, client)
         except Exception as e:
             if return_exceptions:
-                output = e
+                if isinstance(e, UserCodeException):
```
erikbern, line 396: weird, this shouldn't be needed, since it indicates that synchronicity is leaking or something

aksh-at: Yeah I'm not sure if there's something deeper we can fix here. The problem is that synchronicity wraps it in `UserCodeException`, and relies on catching it back when it leaves the async interface. But here, within the async interface, we catch the exception and return it, so synchronicity doesn't have an opportunity to unwrap it. 

Let me know if you have any ideas though!

erikbern: what happens if you remove this unwrapping? are there actually cases where it's needed?

erikbern: do you have an example of a repro where this is needed? i can take a look

aksh-at: the test I fixed is a good example (that's how I noticed it)

PR #441, opened 2023-03-30
Add a MODAL_FUNCTION_RUNTIME override variable

ekzhang:
This environment variable is meant to let us test the gVisor runtime more easily. The short story is that we can do:

```
export MODAL_FUNCTION_RUNTIME=gvisor
modal run app.py
```

And it will create all functions in the app with a Function protobuf field set to use gVisor. It's meant for testing to force-activate gVisor. There is a Notion doc summarizing the gVisor API changes, please see it for details.


ekzhang:
Got a spot check so I'll merge it now :)

PR #443, opened 2023-03-31
Don't set function type to generator for webhooks

erikbern:
This is needed in order to support calling webhook functions locally in #434 . They are treated as generators by the server, but typically aren't defined as generators in the client code. This causes issues when you call it locally because the behavior is determined by `is_generator` defined on the function.

~After this change, the function type is determined by the signature of the function instead.~ EDIT: we actually now don't support generator webhooks anymore – we output an error message that they should use `StreamingResponse` instead.

However the container entrypoint code will override for all webhook types when it actually runs the function. This will work since the server never actually cares about the function type – it just cares about the outputs it actually gets back.


erikbern:
This passes all integration tests also.

erikbern:
I rolled #444 into this, along with fixes for those. Basically if the function is an async or sync generator, we auto-wrap it in a `StreamingResponse`

thundergolfer:
If I `serve` the current basic streaming example in `modal-examples` I get this error on request:

```python
✓ Created objects.
├── 🔨 Created fastapi_app => https://modal-labs--example-fastapi-streaming-fastapi-app-th-ecb0d3-dev.modal.run (label truncated)
└── 🔨 Created mount /home/ubuntu/modal-examples/07_webhooks/streaming.py
 ️️⚡️ Serving... hit Ctrl-C to stop!
└── Watching /home/ubuntu/modal-examples/07_webhooks.
  File "/pkg/modal/_container_entrypoint.py", line 324, in handle_input_exception
    yield
  File "/pkg/modal/_container_entrypoint.py", line 455, in call_function_async
    raise InvalidError(
modal.exception.InvalidError: Async (non-generator) function returned value of type <class 'async_generator'> You might need to use @stub.function(..., is_generator=True).
Request failed with exception: InvalidError("Async (non-generator) function returned value of type <class 'async_generator'> You might need to use @stub.function(..., is_generator=True).") (execution time: 1.2 ms, total latency: 1403.4 ms)
⠏ Running (0/1 containers active)...
```

modal/_asgi.py
```diff
@@ -50,12 +51,27 @@ def wsgi_app_wrapper(wsgi_app):
...
 
-def webhook_asgi_app(fn: Callable, method: str):
+def _stream_generator(fn: Callable) -> Callable:
+    # Wrap a generator function in a handler that returns a StreamingResponse
+    from fastapi.responses import StreamingResponse
+
+    @functools.wraps(fn)
+    async def wrapper(*args, **kwargs):
+        generator = fn(*args, **kwargs)
+        return StreamingResponse(generator)
```
thundergolfer, line 61: Need to set a `media_type` here otherwise webhook returns `content-type: None` and streaming response won't work in browsers at least (I think it still works with `curl --no-buffer`). 

Setting `media_type="text/html"` on behalf of the user is better than nothing, and gets the browser case to work, but ideally there'd be an easy way for users to set their content-type.

erikbern: Maybe it's better to not do the magic thing here then? We could catch it if the user function is a generator and tell them to use StreamingResponse. or wdyt?

erikbern: I'm ok with defaulting to text/html though. I guess that's what fastapi does if you just return a byte stream anyway. Just wondering if the case for streaming html is limited anyway. If 99% of the time involves some other content type then maybe it's better to force the user to be explicit. Not sure.

thundergolfer: Actually it isn't `content-type: none` with this snippet, it's JSON. Still not a streaming response though. 

```python
@stub.webhook()
def app():
    import time
    for i in range(10):
        yield f"frame {i}: hello world!".encode()
        time.sleep(1.0)
```

thundergolfer: Yeh less magic sounds good. Catch and offer a suggestion to return a `StreamingResponse` with appropriate `media_type` set. 

Currently the async test case snippet doesn't work if I run it locally with `MODAL_SYNC_ENTRYPOINT`. I get this error: 

```
ValueError: [TypeError("'async_generator' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]
```

PR #446, opened 2023-04-03
Add invoke task for creating type stubs

freider:
Just a small cleanup

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -2,6 +2,7 @@
 # Copyright (c) Modal Labs 2022
 
 import inspect
+import subprocess
 
 if not hasattr(inspect, "getargspec"):
     # Workaround until invoke supports Python 3.11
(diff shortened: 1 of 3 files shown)


tasks.py
```diff
@@ -79,3 +80,23 @@ def update_build_number(ctx, new_build_number):
     assert new_build_number > current_build_number
     with open("modal_version/_version_generated.py", "w") as f:
         f.write(f"{copyright_header_full}\nbuild_number = {new_build_number}\n")
+
+
+@task
+def type_stubs(ctx):
+    # we only generate type stubs for modules that contain synchronicity wrapped types
```
erikbern, line 87: couldn't this be inferred automatically if we just scan the whole package?

tasks.py
```diff
@@ -79,3 +80,23 @@ def update_build_number(ctx, new_build_number):
...
+        "modal.image",
+        "modal.mount",
+        "modal.object",
+        "modal.proxy",
+        "modal.queue",
+        "modal.secret",
+        "modal.shared_volume",
+        "modal.stub",
+    ]
+    subprocess.check_call(["python", "-m", "synchronicity.type_stubs", *modules])
```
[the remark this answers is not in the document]
freider: Should work most of the time, but for some advanced typing features (typing.overloads support) synchronicity needs to monkey patch some standard library stuff before modules are loaded, so it's safer to shell out to a new interpreter (to ensure we don't load modal modules in tasks.py first) 😬

tasks.py
freider: Yeah we could do that. It's a bit nicer to not generate the stubs for the packages where there are no wrapper classes, since in those cases editors would just use the implementation types instead and be able to infer more implicit types, but we could scan the package and check which modules contain/declare synchronicity wrappers I suppose?

PR #447, opened 2023-04-03
Remove the typeguard library entirely

freider:
Experimented with beartype, but it seems to perform much worse for actually finding errors while still doing code generation etc. on import... So for now, I'm disabling type checking altogether (falling back on "manual" checks for common mistakes)


thundergolfer:
👍

PR #449, opened 2023-04-04
Deprecate webhook et al

erikbern:
Mostly just fixing tests

### DIFF
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -134,6 +134,11 @@ async def deploy_stub(
     show_progress=None,
     object_entity="ap",
 ) -> _App:
+    if stub._loose_webhook_configs:
+        raise InvalidError(
+            f"Web endpoints {stub._loose_webhook_configs} need to be decorated with @stub.function too. Usage:\n\n"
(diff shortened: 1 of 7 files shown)


modal/runner.py
```diff
@@ -134,6 +134,11 @@ async def deploy_stub(
     show_progress=None,
     object_entity="ap",
 ) -> _App:
+    if stub._loose_webhook_configs:
+        raise InvalidError(
+            f"Web endpoints {stub._loose_webhook_configs} need to be decorated with @stub.function too. Usage:\n\n"
+            "@stub.function\n@stub.web_endpoint\ndef my_webhook():\n    ..."
```
freider, line 140: If we are deprecating the call-less decorator syntax for `@stub.function` (in favor of `@stub.function()`, we might want to use it in our examples like this, to avoid users having to migrate syntax again soon if they copy it. Same applies for the `@stub.asgi_app` decorator etc. below

freider (approved):
Minor comment

modal/runner.py
erikbern: yeah agree but this is everywhere – think we can keep it separate?

PR #456, opened 2023-04-05
fix: @warn_if_generator_is_not_consumed needs to allow next()

thundergolfer:
⚠️  Relies on https://github.com/modal-labs/synchronicity/pull/94.

Turns out we weren't properly fulfilling the _iterator protocol_ that is provided by a Python generator. So the decorator was changing the interface.

Will remove this gotcha affecting users: https://github.com/modal-labs/modal-examples/pull/267/files#diff-7d760443a54f6d0e66b1d1204a0cc9533d249f658501b557eea183cb76659635R57.

### DIFF
--- a/client_test/async_utils_test.py
+++ b/client_test/async_utils_test.py
@@ -171,6 +171,21 @@ async def test_warn_if_generator_is_not_consumed(caplog):
     assert "list" in caplog.text
 
 
+@pytest.mark.asyncio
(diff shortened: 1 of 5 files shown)


freider:
Just published latest main to synchronicity 0.4.2 on pypi

erikbern:
Nice!

PR #457, opened 2023-04-05
Deprecate the argument-less version of decorator_with_options

freider:
Emits a deprecation warning for now, and we can unsupport the feature in a month or so perhaps (given the very simple migration path)

### DIFF
--- a/modal_utils/decorator_utils.py
+++ b/modal_utils/decorator_utils.py
@@ -6,6 +6,15 @@ import inspect
 from modal.exception import deprecation_warning
 
 
+def pretty_name(qualname):
+    if "." in qualname:
+        # convert _Stub.function to stub.function
+        qualname = qualname.lstrip("_")
+        qualname = qualname[0].lower() + qualname[1:]
+
+    return qualname
+
+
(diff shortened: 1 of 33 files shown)


freider:
Corresponding change to examples to be merged first: https://github.com/modal-labs/modal-examples/pull/269

modal_utils/decorator_utils.py
```diff
@@ -1,7 +1,19 @@
 # Copyright Modal Labs 2022
+import datetime
 import functools
 import inspect
 
+from modal.exception import deprecation_warning
```
erikbern, line 6: i think we generally want to avoid cyclical dependencies between packages, but it's probably fine to have this for a short while

erikbern:
Nice! I recommend running all examples, integration tests, and docs against this branch and merge those changes before this change. We treat deprecation warnings as errors in CI/CD intentionally.

modal_utils/decorator_utils.py
freider: Ah, I didn't even realize this was in a separate package :O A good motiation to get rid of the deprecated stuff soon!

PR #459, opened 2023-04-06
Fix serve watching logic for directories

aksh-at:
This code is a little complicated. Mostly this is because watching individual files doesn't work (https://github.com/notify-rs/notify/issues/394) so we need to use directories + filtering to achieve the same effect.


erikbern (approved):

thundergolfer:
Thanks 🙏

PR #465, opened 2023-04-11
Support `force_build` for images

aksh-at:
Added it to every single `Image` function so it shows up in the docs (and type completion works)


aksh-at:
Actually `force_build` should also force rebuild everything below a step, if applied. Doing that now.

PR #466, opened 2023-04-12
Support serialized functions calling others

freider:
Main changes:
* Functions are serialized at *load* time, and not at decoration time. They were previously serialized twice ...
* Requires a "preload" phase when building stubs into apps, in order for the function id's of any referenced functions to be set before the functions are serialized as part of the load phase.
* Custom pickle hook needs to know both the object id and the synchronicity interface of a pickled _Handle/Handle/AioHandle, in order to unpickle it with the correct interface.
* Unpickler needs to have access to the loaded container app, in order to get the definitions of any referenced functions when unpickling references (NOTE: this currently excludes the possibility of referencing a globally unpacked `FunctionHandle.lookup()` from another app, I think, since those won't have a definition available. The definition is mostly used to know if the function is a generator or not, so not sure if it should be a hard requirement to have it...)

Backwards incompatibility:
* When generating serialized functions as closures around loop variables, the *last value* will now be closed over instead of the current value during the iteration. This can be worked around by adding a factory function, which is probably best practice anyway for creating closures within loops

### DIFF
--- a/client_test/function_serialization_test.py
+++ b/client_test/function_serialization_test.py
@@ -1,3 +1,4 @@
+# Copyright Modal Labs 2023
 import pytest
 
 from modal import Stub
--- a/client_test/function_serialization_test.py
+++ b/client_test/function_serialization_test.py
@@ -1,0 +1,34 @@
+import pytest
+
+from modal import Stub, App
+
+
+@pytest.mark.asyncio
+def test_serialize_deserialize_function(servicer, client):
+    stub = Stub()
+
+    @stub.function(serialized=True, name="foo")
+    def foo():
+        2 * bar.call()
+
+    @stub.function(serialized=True, name="bar")
+    def bar():
+        return 5
+
+    assert foo.object_id is None
+    assert bar.object_id is None
+
+    with stub.run(client=client) as app:
+        pass
+
+    assert foo.object_id is not None
+    assert bar.object_id is not None
+    assert {foo.object_id, bar.object_id} == servicer.reserved_functions
+
+    foo_def = servicer.app_functions[foo.object_id]
+    # kind of janky assertion - relies on cloudpickle not obfuscating the object id of the contained reference
+    assert bar.object_id.encode("ascii") in foo_def.function_serialized
+
+    app = App.init_container(client, app.app_id)
+    app.foo()  # intentional "local" call
+    assert len(servicer.client_calls) == 1
--- a/client_test/function_serialization_test.py
+++ b/client_test/function_serialization_test.py
@@ -1,10 +1,12 @@
 import pytest
 
-from modal import Stub, App
+from modal import Stub
+from modal.app import App
+from modal._serialization import deserialize
 
 
 @pytest.mark.asyncio
-def test_serialize_deserialize_function(servicer, client):
+async def test_serialize_deserialize_function(servicer, client):
     stub = Stub()
 
     @stub.function(serialized=True, name="foo")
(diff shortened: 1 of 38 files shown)


modal/functions.py
```diff
@@ -734,7 +735,6 @@ def __init__(
         is_generator=False,
         gpu: GPU_T = None,
         # TODO: maybe break this out into a separate decorator for notebooks.
-        serialized: bool = False,
```
freider, line 737: unused...

client_test/function_serialization_test.py
```diff
@@ -0,0 +1,42 @@
...
+    assert foo.object_id is not None
+    assert bar.object_id is not None
+    assert {foo.object_id, bar.object_id} == servicer.reserved_functions
+
+    foo_def = servicer.app_functions[foo.object_id]
+
+    assert len(servicer.client_calls) == 0
+    from modal_utils.async_utils import synchronizer
+
+    container_app = App.init_container(client, app.app_id)
```
freider, line 36: This is a kind of hacky way to test the deserialization logic, not sure how to best client test this without adding significant additional functionality to the mock servicer and actually running the container_entrypoint on the definition from here... We do have an integration test on the internal repo that tests this properly though

modal/app.py
```diff
@@ -39,6 +39,8 @@ class _App:
 
     _tag_to_object: Dict[str, _Handle]
     _tag_to_existing_id: Dict[str, str]
+    _object_id_to_object: Dict[str, _Handle]
```
erikbern, line 42: what's this needed for?

erikbern:
> Unpickler needs to have access to the loaded container app, in order to get the definitions of any referenced functions when unpickling references (NOTE: this currently excludes the possibility of referencing a globally unpacked FunctionHandle.lookup() from another app, I think, since those won't have a definition available. The definition is mostly used to know if the function is a generator or not, so not sure if it should be a hard requirement to have it...)

This is the thing that feels confusing to me. Would it make sense to just put the proto inside the serialization payload itself? I think we have the general problem with functions that object ids aren't enough to interact with functions, and the way `lookup` works right now is the server returns the function proto. So maybe we should think about serialization the same way?

modal/functions.py
```diff
@@ -844,9 +844,15 @@ def __init__(
 
         self._function_handle = function_handle
 
-        rep = r"Function({self._tag})"
+        rep = f"Function({self._tag})"
```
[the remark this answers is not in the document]
freider: No it creates a regular string, but parses the literal in "raw" mode which treats backslashes as literal backslashes rather than escape sequences (but it is mostly used for regex since that's often the behavior you want to make the regex readable)

modal/_serialization.py
```diff
@@ -15,23 +21,51 @@ def __init__(self, buf):
...
 
 
 class Unpickler(pickle.Unpickler):
-    def __init__(self, client, buf):
+    def __init__(self, client, buf, app=None):
         self.client = client
+        self.app = app
         super().__init__(buf)
 
+    def get_proto_for_object(self, object_id) -> Optional[google.protobuf.message.Message]:
```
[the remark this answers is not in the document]
freider: Mostly for organizing the code, as it was a bit too messy to inline into persistent_load imo

modal/app.py
freider: Deserialization - used when looking up referenced objects that are inlined as part of the serialized function "closure" (having only an object_id in the pickled reference)

erikbern:
> In general, it might make sense for the function (and other handles?) proto to be split across "api" proto (only the info needed to expose a handle to a user) and "execution" proto (used by modal to execute whatever the user wants to do)?

This makes sense to me. Afaik the only data we ever read is `is_generator` and I think _maybe_ `web_url`.

Would it be outside the scope of this to make this change? I'm not sure how big/small that task would be. Maybe it's a big refactoring?

erikbern (approved):
Nice – looks a lot cleaner now with the handle metadata work! Thanks for spending time refactoring it

PR #471, opened 2023-04-15
Link to app page URL from status spinner

aksh-at:
I find myself having to go to the apps page to look up a running app's URL too often. It is a bit hard to ctrl+click on it sometimes since it refreshes so often, but still better than nothing.

image

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -64,7 +64,7 @@ def step_progress(text: str = "") -> Spinner:
 
 
 def step_progress_update(spinner: Spinner, message: str):
-    spinner.update(text=Text(message, style="blue"))
+    spinner.update(text=message)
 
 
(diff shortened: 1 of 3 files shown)


erikbern:
Minor think but I find it a bit easier to display URLs in the terminal on their own line. The Terminal app in os x doesn't make them clickable, so you have to select the URL and copy it manually. But if the link is on its own line then you can just triple-click the line to select it.

aksh-at:
@erikbern you can cmd + double click on the links in Terminal (just confirmed). The URL in a line by itself doesn't look great unless we add a caption and padding around it, but then it's too much space.

erikbern:
ok, no strong feelings – I didn't know about command+doubleclick, but maybe other people don't know about tripleclicks.

PR #473, opened 2023-04-17
Introduce "handle metadata" for functions

freider:
Replaces the full function proto in the context of `_initialize_from_proto`. This lets us have a much lighter weight proto with only the fields necessary for users of the handle, rather than all the info needed to execute the function. It has two advantages:
* We can use it in serialization/deserialization of individual objects to get functioning Handles on deserialization, without any RPC calls or similar.
* We can "hydrate" an app (AppGetObjects) without having to get a bunch of backend data
* in #466 we can let the "preload" phase of functions return a handle proto, letting serialized functions encode references to other function handles, or even to themselves.

The concept is general for all `_Handle`s but only needed for FunctionHandles at this time. Will be used in a new version of #466 to avoid having to load an entire app to rehydrate a serialized handle.

As a followup backend PR, we'll add rpc support for returning FunctionHandles in AppGetObjects, AppLookupObject and FunctionCreate. Since the protos share data types and indices, we should be able to do this in a backwards compatible way

### DIFF
--- a/modal/_serialization.py
+++ b/modal/_serialization.py
@@ -1,15 +1,27 @@
 # Copyright Modal Labs 2022
 import io
 import pickle
+from typing import Optional
 
 import cloudpickle
 
 from modal_utils import async_utils
+from synchronicity import Interface
 from synchronicity.synchronizer import TARGET_INTERFACE_ATTR
 from .exception import InvalidError
 from .object import _Handle, Handle, AioHandle
 
 PICKLE_PROTOCOL = 4  # Support older Python versions.
+
+
+def get_synchronicity_interface(obj) -> Optional[Interface]:
+    return getattr(obj, TARGET_INTERFACE_ATTR, None)
+
+
+def restore_synchronicity_interface(raw_obj, target_interface: Optional[Interface]):
+    if target_interface:
+        return async_utils.synchronizer._translate_out(raw_obj, target_interface)
+    return raw_obj
 
 
 class Pickler(cloudpickle.Pickler):
@@ -21,7 +33,7 @@ class Pickler(cloudpickle.Pickler):
             return
         if not obj.object_id:
             raise InvalidError(f"Can't serialize object {obj} which hasn't been created.")
-        return (obj.object_id, getattr(obj, TARGET_INTERFACE_ATTR, None), obj._handle_proto())
+        return (obj.object_id, get_synchronicity_interface(obj), obj._handle_proto())
 
 
 class Unpickler(pickle.Unpickler):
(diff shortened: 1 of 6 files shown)


freider:
> Nice! I think this should make #466 a bit cleaner.
> 
> One thought about `FunctionHandleMetadata` is we could make it just a "dict" or similar to make it simpler. The benefit of that is that it's entity-neutral. But it's not obvious to me what's the best representation. Maybe that's an argument for keeping it the way you did it since it's fully backwards compatible – if we realize we want to support lots of attributes for other objects (or maybe in the future, make it possible for a function created in one language to be imported in another languate?) then I guess we can always change this later.

I thought about making it a dict but opted for proto mostly for ergonomics of nested entities and compatibility with the existing "full" proto making the transition simple. Easy to change later as you say

modal/_serialization.py
```diff
@@ -15,11 +17,11 @@ def __init__(self, buf):
         super().__init__(buf, protocol=PICKLE_PROTOCOL)
 
     def persistent_id(self, obj):
-        if not isinstance(obj, _Handle):
+        if not isinstance(obj, (_Handle, Handle, AioHandle)):
             return
         if not obj.object_id:
             raise InvalidError(f"Can't serialize object {obj} which hasn't been created.")
-        return obj.object_id
+        return (obj.object_id, getattr(obj, TARGET_INTERFACE_ATTR, None), obj._handle_proto())
```
erikbern, line 24: maybe put `getattr(obj, TARGET_INTERFACE_ATTR, None)` in a utility function? it took me a while to realize what it does

modal/_serialization.py
```diff
@@ -28,10 +30,11 @@ def __init__(self, client, buf):
...
 
     def persistent_load(self, pid):
-        object_id = pid
-        # TODO(erikbern): we should get the proto somehow,
-        # for functions
-        return _Handle._from_id(object_id, self.client, None)
+        (object_id, target_interface, handle_proto) = pid
+        raw_obj = _Handle._from_id(object_id, self.client, handle_proto)
+        if target_interface:
+            return async_utils.synchronizer._translate_out(raw_obj, target_interface)
```
erikbern, line 36: we talked about it on slack, but it would be nice to have just one interface and put the async stuff on top of the blocking stuff

erikbern:
You're changing the serialized data which might cause compatibility issues between clients of different versions – but I don't think it's a problem in practice

erikbern (approved):
Nice!

freider:
> You're changing the serialized data which might cause compatibility issues between clients of different versions – but I don't think it's a problem in practice

Yeah I think it should be relatively safe thanks to keeping the same proto types/indices as in the larger structure (and us not using any of the other fields anyways). One gotcha that I ran into though was that it's *not* safe to pass the wrong datatype into a proto constructor where the field is a message, even if all the field names and types line up - it segfaults the python interpreter due to a bug in protobuf :O (https://github.com/protocolbuffers/protobuf/issues/12373)

modal/_serialization.py
freider: Yeah, that would be nice for these sort of things (I guess we'd still need to know if we are serializing a wrapper or not, but not the type of wrapper)

modal/_serialization.py
freider: That utility function might belong in synchronicity though right? I'll add one here in the meanwhile to make the intent clear

PR #475, opened 2023-04-17
Type annotations for decorator stuff

erikbern:
We have a lot of decorators on the `Stub` class that themselves rely on another decorator to inject the function-to-be-decorated as the first argument. This is nice because we don't have to write nested functions. However the drawback is that mypy doesn't understand this properly, so our typing annotations end up having to hack around that (like making the function-to-be-decorated optional, even though it should always be provided).

Thanks to [PEP 612](https://peps.python.org/pep-0612/), we can support it without relying on those hacks.

### DIFF
--- a/.github/actions/setup-cached-python/action.yml
+++ b/.github/actions/setup-cached-python/action.yml
@@ -4,7 +4,7 @@ inputs:
   version:
     description: Which Python version to install
     required: true
-    default: "3.9"
+    default: "3.10"
 
 runs:
   using: composite
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -14,7 +14,7 @@ from modal._types import typechecked
 
 from modal_proto import api_pb2
 from modal_utils.async_utils import synchronize_apis
-from modal_utils.decorator_utils import decorator_with_options_unsupported, decorator_with_options
+from modal_utils.decorator_utils import decorator_simplify, decorator_with_options_unsupported, decorator_with_options
 from .retries import Retries
 
 from ._function_utils import FunctionInfo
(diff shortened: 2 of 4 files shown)


erikbern:
@freider does this approach roughly make sense? If it does, I can finish it, which I think requires updating to latest mypy and using https://pypi.org/project/typing-extensions/ for older Python versions

erikbern:
I think this is breaking because of lacking synchronicity support? Not sure

freider:
Yeah there are a couple of issues:
* The synchronicity-generated type stub for `web_endpoint()` is not smart enough to figure out the changed signature based on the Concatenate logic (nor will it include the `@decorator_simplify` in the type stub), and will emit the signature as is instead:

```python
def web_endpoint(self, raw_f: typing.Callable[..., typing.Any], method: str = 'GET', label: typing.Union[str, None] = None, wait_for_response: bool = True) -> WebhookConfig:
        ...
```
To fix this, we could potentially add some hook in synchronicity to detect certain "type modifying" decorators and include them verbatim in the emitted type stubs...


* Another issue, which should be easily fixable, is that the stub.function decorator has an incorrect return value for the f == None case, which doesn't allow WebhookConfig (only callables) as the input in the returned HOF. Not sure why this didn't error before this patch though? :thinking:

freider:
There also seems to be an issue with adding this decorator on *methods of classes*, since the decorated unbound function has another positional argument (self) before `raw_f`, so the Concatenate logic gets tripped up :( To fix that we'd need a separate decorator_simplify_method or something, but it feels a bit janky...

freider (commented):
Some comments regarding the failing mypy tests. Only some of them are due to synchronicity's type stubs, there are more general problems with the typing "meta programming" itself

freider:
Looked a bit at https://github.com/micheles/decorator which is a library for something similar (apart from the static type stuff) and they use the `__signature__` magic attribute to modify the signature of the returned function in order to get inspect.signature to return the right thing, so doing that could be an option in order to get type stub generation to work (since it uses inspect.signature under the hood). Still doesn't solve the decorator having to be different on methods vs functions though :/

erikbern:
The dumb solution for this is ofc that we just implement all decorators the old fashioned way, with two nested functions.

I.e. instead of 
```python
def function(raw_f: Callable[..., Any], image: Optional[_Image], ...):
```

we would do something like

```python
def function(image: Optional[_Image], ...) -> Callable[Callable[..., Any], _FunctionHandle]:
    def actual_function(raw_f, Callable[..., Any]) -> _FunctionHandle:
        ...

    return actual_function
```

it's just mildly annoying

PR #476, opened 2023-04-18
[MOD-1067] Don't allow mounting shared vols at /root

thundergolfer:
I've also removed the guard preventing client-side checking on Windows. 

This should be fine because we're doing pure path operations on strings provided by the user, and these strings should only ever represent POSIX paths because they declare mount points in a Unix container. (Don't have a Windows machine handy to tests this though)

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -1,13 +1,12 @@
 # Copyright Modal Labs 2022
 import asyncio
 import inspect
-import os
-import platform
+import posixpath
 import time
 import typing
 import warnings
 from dataclasses import dataclass
-from pathlib import Path
+from pathlib import PurePath
 from typing import Any, AsyncIterable, Callable, Collection, Dict, List, Optional, Set, Union
 
 import cloudpickle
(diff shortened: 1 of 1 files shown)


modal/functions.py
```diff
@@ -1,13 +1,12 @@
 # Copyright Modal Labs 2022
 import asyncio
 import inspect
-import os
-import platform
+import posixpath
```
freider, line 4: Huh, didn't know of this module

modal/functions.py
```diff
@@ -882,16 +881,17 @@ async def _load(self, resolver: Resolver, existing_object_id: str):
         shared_volume_mounts = []
         # Relies on dicts being ordered (true as of Python 3.6).
         for path, shared_volume in self._shared_volumes.items():
-            # TODO: check paths client-side on Windows as well.
-            path = Path(path).as_posix()
-            abs_path = os.path.abspath(path)
+            path = PurePath(path).as_posix()
+            abs_path = posixpath.abspath(path)
```
freider, line 885: nit: this will still return weird stuff on windows, e.g. `C:\\foo\bar\my/posix/path`... It will technically work since all we do is check if it is absolute or match some exact paths, but I think a "cleaner" way might be to use `PurePosixPath()` and then do `.is_absolute()` for the absoluteness check (As the input paths in the shared_volume dict should all be posix paths to begin with)

PR #474, opened 2023-04-17
tracing_hostname setting

danielnorberg:
Allow configuring the dd tracer hostname, e.g. by setting the env var `MODAL_TRACING_HOSTNAME=localhost`.

PR #478, opened 2023-04-18
tracing: respect datadog endpoint env vars

danielnorberg:
Allow configuring datadog agent endpoint using the `DD_TRACE_AGENT_URL` and `DD_AGENT_HOST` env vars.

Alternative to https://github.com/modal-labs/modal-client/pull/474.

With this we should be able to enable dd tracing for the client in tasks:

```
"DD_TRACE_ENABLED": "true",
"DD_AGENT_HOST": "127.0.0.1",
```

We could instead introduce `MODAL_TRACING_URL` and/or `MODAL_TRACING_HOSTNAME` but this seems like additional work for little gain.

Also respect `DD_SERVICE` if set.

### DIFF
--- a/modal/_tracing.py
+++ b/modal/_tracing.py
@@ -55,18 +55,21 @@ def inject_tracing_context(metadata: Dict[str, str]):
         logger.exception("Failed to inject tracing context")
 
 
+TRACING_SERVICE = None if os.environ.get("DD_SERVICE") else "modal-runtime-client"
+
+
 def wrap(*args, **kwargs):
     if not TRACING_ENABLED:
         return lambda f: f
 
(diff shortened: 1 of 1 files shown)


danielnorberg:
Example trace https://app.datadoghq.com/apm/traces?query=%40_top_level%3A1%20env%3Aprod%20service%3Abenchmark-stable-diffusion-preload&graphType=flamegraph&historicalData=false&shouldShowLegend=false&spanID=15536609373840989883&spanType=service-entry&timeHint=1681822714847&trace=447461830580988311015536609373840989883&traceID=4474618305809883110&viz=stream&start=1681822161463&end=1681823061463&paused=false

freider (approved):

PR #479, opened 2023-04-18
Move web_endpoint, asgi_app, and wsgi_app out of the stub

erikbern:
I'm trying to simplify the `Stub` class since it's very complex – it makes it harder to perform other refactorings (merging stubs and apps) and fundamentally makes it harder for users to understand what a stub "is". I want to make it a "carrier of other objects" but it's currently doing all kinds of other stuff.

This PR changes

```python
from modal import Stub

stub = Stub()

@stub.function(cpu=12)
@stub.web_endpoint(method="POST")
def my_f():
    ...
```

to

```python
from modal import Stub, web_endpoint

stub = Stub()

@stub.function(cpu=12)
@web_endpoint(method="POST")
def my_f():
    ...
```

Note that this is a temporary increase in LOC but that's just because there's a ton of deprecation warnings – those will go away in a few months.

~Currently failing on a unit test that detects web endpoints that were never decorated by a function. I'm going to rewrite that logic slightly.~ EDIT: done

### DIFF
--- a/client_test/function_test.py
+++ b/client_test/function_test.py
@@ -6,7 +6,7 @@ import time
 import cloudpickle
 from synchronicity.exceptions import UserCodeException
 
-from modal import Proxy, Stub, SharedVolume
+from modal import Proxy, Stub, SharedVolume, web_endpoint
 from modal.exception import InvalidError
 from modal.functions import Function, FunctionCall, gather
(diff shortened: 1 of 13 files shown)


freider (approved):
Nice!

PR #481, opened 2023-04-18
Fix is_persisted_ref for some Handlers

aksh-at:
Didn't realize there's another code path where `Handlers` can be initialized. Added a test for this + one for the original issue with persisted refs changing.

### DIFF
--- a/client_test/stub_test.py
+++ b/client_test/stub_test.py
@@ -284,3 +284,23 @@ def test_set_image_on_stub_as_attribute():
     custom_img = modal.Image.debian_slim().apt_install("emacs")
     stub.image = custom_img
     assert stub._get_default_image() == custom_img
+
+
+@pytest.mark.asyncio
+async def test_redeploy_persist(servicer, aio_client):
+    stub = AioStub()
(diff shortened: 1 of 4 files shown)


modal/_resolver.py
```diff
@@ -85,6 +85,7 @@ async def load(self, obj, existing_object_id: Optional[str] = None):
             #
             # Persisted refs are ignored because their life cycle is managed independently.
             # The same tag on an app can be pointed at different objects.
+            print(obj, type(obj))
```
erikbern, line 88: debug print

client_test/stub_test.py
```diff
@@ -283,3 +284,23 @@ def test_set_image_on_stub_as_attribute():
     custom_img = modal.Image.debian_slim().apt_install("emacs")
     stub.image = custom_img
     assert stub._get_default_image() == custom_img
+
+
+@pytest.mark.asyncio
+async def test_redeploy_persist(servicer, aio_client):
```
erikbern, line 290: thanks for adding a test!

PR #487, opened 2023-04-19
Remove type checking code

erikbern:
Don't think this is needed because of `@typechecked`?


freider (commented):
Sadly it is needed, temporarily, as we had to make typechecked into a noop on account of it adding high import time cost with the existing third party packages we found. My plan was to implement something simple using annotations myself but I haven't gotten to it, so for now I added back these "hand written" checks to not regress existing type check tests

PR #488, opened 2023-04-19
make _run_serve_loop a context manager

erikbern:
A user was asking for something that essentially requires running modal serve in a more programmatic way, and also get app info back.

I realize it's pretty tricky given the current client code, but with some minor refactoring it shouldn't be too hard. This also lets us unify the interface between running and serving to some extent.


erikbern:
This isn't the full set of changes, but given how complex this code is, I prefer to merge it piece by piece

PR #489, opened 2023-04-19
Protos changes for serialized functions calling others

freider:


modal_proto/api.proto
```diff
@@ -426,6 +426,19 @@ message FunctionCreateRequest {
   string existing_function_id = 7;
 }
 
+message FunctionPrecreateRequest {
+  string app_id = 1;
+  string function_name = 2;
+  string existing_function_id = 3;
+  Function.FunctionType function_type = 4;
+  WebhookConfig webhook_config = 5;
```
erikbern, line 434: Why is this needed?

freider: The webhook config? Its since the webhook url is part of the "function proto" we use when we initialize functions (due to us exposing it on the function handle api). I had a thought that maybe we should make that lazy load instead through an api call instead, but that felt like a larger change so could do that in a follow-up to get rid of some of this complexity

freider: Specifically, the webhook_config is used by FunctionCreate (and now FunctionPrecreate to set the webhook label for webhook functions, through some intricate logic...)

freider: We could, but it would lead to the FunctionHandle's .web_url attribute not working from *within* serialized functions I think. Probably not a big deal but I preferred to not break it in case people build stuff with to have a web endpoint returning the web_url of other functions etc.

PR #490, opened 2023-04-19
[MOD-979] Stop rooting relative Mount remote_path values at '/'

thundergolfer:
This is addressing a UX issue where relative `remote_path` values don't put files in the current working directory as users expect. (see Linear card)

⚠️  The server is already configured to handle relative paths, but this is a backwards incompatible, breaking client change. Any user code that currently expects `remote_path="foo.txt"` to be placed at `/foo.txt` will break.

Not sure yet how best to manage this incompatibility.

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -219,7 +219,7 @@ class _Mount(_Provider[_MountHandle]):
         local_path = Path(local_path)
         if remote_path is None:
             remote_path = local_path.name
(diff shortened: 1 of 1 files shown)


erikbern:
> Not sure yet how best to manage this incompatibility.

One way (but I'm not sure if it's great) would be to deprecate any such path for a short while (like a month) before adding it back

mwaskom:
Closing as stale but feel free to reopen!

PR #493, opened 2023-04-20
Use AppLookupObject to get handle_metadata in `.from_id()`

freider:
Also rename _initialize_from_proto to _initialize_from_handle_metadata (long name, but a bit clearer?)

### DIFF
--- a/modal_utils/grpc_utils.py
+++ b/modal_utils/grpc_utils.py
@@ -20,6 +20,7 @@ from typing import (
 
 import grpclib.events
 import grpclib.client
+from google.protobuf.message import Message
 from grpclib import GRPCError, Status
 from grpclib.exceptions import StreamTerminatedError
 from grpclib.protocol import H2Protocol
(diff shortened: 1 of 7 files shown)


erikbern:
Why is the `WhichOneOf` and `getattr` needed?

Great if you can add tests too!

freider:
> Why is the `WhichOneOf` and `getattr` needed?
> 
> Great if you can add tests too!

Mostly future proofing - the returned handle_metadata (previously object) can be any handle's proto object, but is currently a one_of that only has function as a valid field. When/if we add handle_metadata for any other object, this will make it automatically pick the "one_of" that is set. Not sure if there is a more ergonomic way to do it with the Python proto lib?

erikbern:
> Mostly future proofing - the returned handle_metadata (previously object) can be any handle's proto object, but is currently a one_of that only has function as a valid field. When/if we add handle_metadata for any other object, this will make it automatically pick the "one_of" that is set. Not sure if there is a more ergonomic way to do it with the Python proto lib?

Got it – I guess this is a pure protobuf thing that has nothing to do with objects/functions – maybe we should break it out into a utility function? I.e.

```python
def get_oneof(message: google.protobuf.Message, oneof_attr: str):
    attr: str = message.OneOf(oneof_attr)
    return getattr(message, attr)
```

modal_utils/grpc_utils.py
```diff
@@ -284,3 +285,11 @@ async def fallback(self, stream) -> None:
 
     cls.__abstractmethods__ = frozenset()
     return cls
+
+
+def get_proto_oneof(message: Message, oneof_group: str) -> Optional[Message]:
```
erikbern, line 290: awesome thanks!

PR #495, opened 2023-04-20
tag `get_inputs` span with `iteration`

danielnorberg:
Add `iteration` tag to `get_inputs` span so we can search for slow initial `get_inputs` spans in cold-start traces

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -186,7 +186,8 @@ class _FunctionIOManager:
 
             with trace("get_inputs"):
                 span = tracer.current_span()
-                span.set_tag("iteration", str(iteration))  # force this to be a tag string
+                if span:
+                    span.set_tag("iteration", str(iteration))  # force this to be a tag string
                 iteration += 1
                 response = await retry_transient_errors(self.client.stub.FunctionGetInputs, request)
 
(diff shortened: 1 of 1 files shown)


modal/_container_entrypoint.py
```diff
@@ -178,11 +179,15 @@ async def _generate_inputs(
     ) -> AsyncIterator[tuple[str, api_pb2.FunctionInput]]:
         request = api_pb2.FunctionGetInputsRequest(function_id=self.function_id)
         eof_received = False
+        iteration = 0
         while not eof_received:
             request.average_call_time = self.get_average_call_time()
             request.max_values = self.get_max_inputs_to_fetch()  # Deprecated; remove.
 
             with trace("get_inputs"):
+                span = tracer.current_span()
```
freider, line 188: Not sure how ddtrace works, but I assume this is safe to call even when the tracing is disabled?

danielnorberg: derp nej, inte i python 👎

danielnorberg:
tests are failing with 
> ModuleNotFoundError: No module named 'ddtrace'
🤔

danielnorberg:
oh, `ddtrace` might not be present if tracing is disabled

modal/_container_entrypoint.py
danielnorberg: ptal again

freider (approved):

PR #496, opened 2023-04-20
Deprecate old lifecycle syntax

erikbern:
Throws a warning that looks like this:

````
modal.exception.DeprecationError: 2023-04-20: @stub.function on methods is deprecated. Use the @stub.cls and @method decorators. Usage:

```
@stub.cls(cpu=8)
class MyCls:
    @method()
    def f(self):
        ...
```
````

There's a slightly more correct implementation that changes this in the backend so that we pass the class name as a part of the `Function` definition. But this is a bit more work to implement and it gets easier once remove support for deprecate the old syntax. So I'd rather do that first.

### DIFF
--- a/client_test/cli_imports_test.py
+++ b/client_test/cli_imports_test.py
@@ -29,8 +29,9 @@ other_stub = modal.Stub("BAR")
 @other_stub.function()
 def func():
     pass
+@stub.cls()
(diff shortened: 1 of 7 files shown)


erikbern:
This passes all integration tests btw

PR #498, opened 2023-04-21
Refactor some of the init/hydration code to have a bit more clear names

freider:
Functionality same as before, we now only have one code path for hydrating handles regardless if they have metadata or not.

`._hydrate(object_id, handle_metadata: Optional[Message])` now hydrates the object.
An object that uses handle_metadata (currently only functions) can implement their own `_hydrate_metadata()` function that is called by _hydrate with metadata if it's present, in order to bootstrap the handle


erikbern (approved):
Nice! I like the name hydrate

PR #500, opened 2023-04-21
modal serve: sleep inf

erikbern:
I think some platforms reject `time.sleep(1e9)`

Not sure if it's worth adding a test for this even though it's maybe doable.


freider (commented):
lol

PR #502, opened 2023-04-24
Determine stub to hydrate via imported function handle if possible

freider:
Use the imported function to infer the relevant stub to hydrate then use the container app to hydrate all functions of that stub.

Notable exceptions:
* Serialized functions are not imported from modules, so they don't have attached stubs. However, they don't need their global references to be hydrated in their module scope, since the serialized references are automatically hydrated on unpickling.
* Functions that are 'privately' decorated such that the imported function isn't a FunctionHandle but a regular python function will not have a linked stub. In those cases we try to pick up a single stub from the function's module scope if there is one. If there is more than one we use explicit stub names to try to determine the stub, and if that fails we print some warnings and don't hydrate any stub.

TODO:
* Run integration tests to see that nothing breaks outside of what is tested in the client

Followup PR:
* Refactor `stub.is_inside()` logic to also use the stub name (right now all the stubs pass the condition when they are not local...)

### DIFF
--- a/client_test/container_test.py
+++ b/client_test/container_test.py
@@ -36,11 +36,12 @@ def _get_inputs(args=((42,), {})) -> list[api_pb2.FunctionGetInputsResponse]:
 
 
 def _load_stub(servicer, module_name, stub_name):
-    # loads stub in another process to prevent leaking objects from client to container process
+    # loads stub from another interpreter to prevent leaking state from client into a container process (apart from what goes through the servicer)
     env = {"MODAL_SERVER_URL": servicer.remote_addr}
     lib_dir = pathlib.Path(__file__).parent.parent
+
     subprocess.check_call(
-        [sys.executable, "-c", f"import {module_name}\nwith {module_name}.{stub_name}.run():\n\tpass"],
+        [sys.executable, "-c", f"import {module_name}\n{module_name}.{stub_name}.deploy('deploy-name')"],
         cwd=lib_dir,
         env=env,
     )
@@ -587,9 +588,12 @@ def _run_e2e_function(
     *,
     stub_name=None,
     assert_result=api_pb2.GenericResult.GENERIC_STATUS_SUCCESS,
+    function_definition_type=api_pb2.Function.DEFINITION_TYPE_FILE,
 ):
     _load_stub(servicer, module_name, stub_var_name)
-    client, items = _run_container(servicer, module_name, function_name, stub_name=stub_name)
+    client, items = _run_container(
+        servicer, module_name, function_name, stub_name=stub_name, definition_type=function_definition_type
+    )
     assert items[0].result.status == assert_result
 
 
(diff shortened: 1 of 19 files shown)


modal/_container_entrypoint.py
```diff
@@ -494,9 +501,36 @@ def import_function(function_def: api_pb2.Function, ser_cls, ser_fun) -> Importe
...
+    active_stub = None
     if isinstance(fun, (FunctionHandle, AioFunctionHandle)):
         _function_proxy = synchronizer._translate_in(fun)
         fun = _function_proxy.get_raw_f()
+        active_stub = _function_proxy._stub
+    elif module is not None:
+        # This branch is reached in the special case that the imported function is 1) not serialized, and 2) isn't a FunctionHandle - i.e, not decorated at definition time
+        # Look for stubs in the same module as the function
+        # This is not necessarily enough, and the active stub might not even be loaded, but it's a best effort
+        # TODO(elias): Use some kind of global automatic "Stub registry" instead of looking at stubs in global scope of the module?
```
erikbern, line 513: yeah i think that would be a lot cleaner!

modal/_container_entrypoint.py
```diff
@@ -557,11 +590,13 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
         # Initialize the function
         with function_io_manager.handle_user_exception():
             imp_fun = import_function(container_args.function_def, ser_cls, ser_fun)
+            if imp_fun.stub:
+                _container_app = synchronizer._translate_in(container_app)
+                _client = synchronizer._translate_in(client)
```
erikbern, line 595: i don't think you need to translate in here? the `_hydrate_function_handles` is a synchronized function that will do the translation automatically

erikbern (approved):
looks good although i think your idea of using a global registry would be cleaner

modal/_container_entrypoint.py
freider: Done!

modal/_container_entrypoint.py
freider: `imp_fun.stub` is an unwrapped stub (to avoid having different interfaces when the user has an AioStub vs Stub). In that case it won't translate the inputs right?

client_test/container_test.py
```diff
@@ -589,10 +591,19 @@ def _run_e2e_function(
     stub_name=None,
     assert_result=api_pb2.GenericResult.GENERIC_STATUS_SUCCESS,
     function_definition_type=api_pb2.Function.DEFINITION_TYPE_FILE,
+    inputs=None,
+    is_builder_function: bool = False,
 ):
+    # TODO(elias): make this a bit more prod-like in how it connects the load and run parts by returning function definitions from _load_stub so we don't have to double specifiy things like definition type
```
erikbern, line 598: super minor but *specify

PR #503, opened 2023-04-24
Include stub name in function definition

freider:
When provided, this can be used to distinguish mutiple stubs from one another when executing a function in a container. Useful for improving `is_inside()` checks and function rehydration in some special cases


erikbern (approved):
It will be redundant with the app name in most cases, but it seems nicer to just send this one raw so we have it

PR #505, opened 2023-04-24
Minor object constructor cleanup

erikbern:
Was looking at merging Providers and Handles and it's fairly straightforward but will be easier if we simplify the Provider classes first. The main thing is this just turns `_preload` fully into an attribute rather than a base class method (afaict @freider had mostly structured the code that way already)

### DIFF
--- a/modal/object.py
+++ b/modal/object.py
@@ -162,6 +162,9 @@ _BLOCKING_P, _ASYNC_P = synchronize_apis(P)
 
 
 class _Provider(Generic[H]):
+    _load: Callable[[Resolver, str], Awaitable[H]]
+    _preload: Optional[Callable[[Resolver, str], Awaitable[H]]]
+
     def _init(
         self,
         load: Callable[[Resolver, str], Awaitable[H]],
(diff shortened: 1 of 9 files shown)


modal/object.py
```diff
@@ -162,40 +162,42 @@ async def _lookup(
...
-            # only sets _preload if one is provided, otherwise a noop is used
-            self._preload = preload
+        self._preload = preload
         self._rep = rep
         self.is_persisted_ref = is_persisted_ref
 
     def __init__(
         self,
-        load: Optional[Callable[[Resolver, str], Awaitable[H]]],
+        load: Optional[Callable[[Resolver, Optional[str]], Awaitable[H]]],
```
freider, line 183: Nit: Just noticed load is also Optional here, which it probably shouldn't be (as of now). I think I might have committed it by mistake when I was experimenting with making load optional just like preload.

freider (approved):
nice, it did feel a bit off with the preload being a mix of declared methods and overrides before. This is a lot nicer

modal/object.py
erikbern: yeah that's right, it shouldn't be optional – thanks!

PR #506, opened 2023-04-24
Fix the _strptime issue

erikbern:
This fixes a bunch of tests failing with weird ass `_strptime` failures.

Btw the way to reproduce this is to run: `pytest -v client_test/mount_test.py `

The absurd thing is that
1. Running the failing test in isolation works: `pytest -v client_test/mount_test.py -k test_create_package_mounts`
2. Running ALL client tests also works.

I think the problem might be that you can't `import ipykernel.iostream` in non-global scope or similar – it's very strange.

### DIFF
--- a/modal/_ipython.py
+++ b/modal/_ipython.py
@@ -1,13 +1,21 @@
 # Copyright Modal Labs 2022
 import sys
+import warnings
+
+ipy_outstream = None
+try:
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        import ipykernel.iostream
+
+    ipy_outstream = ipykernel.iostream.OutStream
+except ImportError:
+    pass
 
 
 def is_notebook(stdout=None):
+    if ipy_outstream is None:
+        return False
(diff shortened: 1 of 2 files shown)


modal/_ipython.py
```diff
@@ -1,13 +1,20 @@
...
+        return False
     if stdout is None:
         stdout = sys.stdout
-    try:
-        import ipykernel.iostream
-
-        return isinstance(stdout, ipykernel.iostream.OutStream)
-    except ImportError:
-        return False
+    return isinstance(stdout, ipykernel.iostream.OutStream)
```
erikbern, line 20: oops this should be `ipy_outstream`, 1 sec

freider (approved):
haha, no idea how this fixes something related to _strptime but I've also seen some weeeeird test inconsistencies from ipython-related stuff before (that time it was some issue with python entrypoint/plugin registration that sometimes triggered)

erikbern:
yeah it's absurd tbh.

PR #507, opened 2023-04-24
Remove some methods from Mount

erikbern:
* Getting rid of the constructor and using a factory function internally
* Making the `load` method more closure like by making it static and binding arguments to it

This is in preparation of merging `Provider` and `Handle` classes

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -131,42 +131,25 @@ class _Mount(_Provider[_MountHandle]):
 
     _entries: List[_MountEntry]
 
-    def __init__(
-        self,
-        # Mount path within the container.
-        remote_dir: Union[str, PurePosixPath] = None,
-        *,
-        # Local directory to mount.
-        local_dir: Optional[Union[str, Path]] = None,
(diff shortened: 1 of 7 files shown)


modal/mount.py
```diff
@@ -130,37 +131,25 @@ def f():
...
+    def _from_entries(*entries: _MountEntry) -> "_Mount":
+        rep = f"Mount({entries})"
+        load = functools.partial(_Mount._load_mount, entries)
+        obj = _Mount._from_loader(load, rep)
+        obj._entries = entries
+        obj._is_local = True
+        return obj
+
+    @property
+    def entries(self):
```
freider, line 151: Should `entries` be considered a public API? If not, we might want to underscore it to avoid users tripping over it in autocomplete

freider (approved):

modal/mount.py
erikbern: I think it's fine. That being said I think we need to figure out how to store "local" attributes on objects. If you import an existing Mount (either by looking it up, or from inside a container), calling `entries` on it will cause it to fail with an attribute error.

PR #508, opened 2023-04-26
[MOD 1176] [MOD 1204] Reject @web_endpoint et all, reject .call, .map on webhooks

thundergolfer:
* **MOD-1176** Actually supporting `.call` seems tricky, so I've not attempted it today. I've just provided a user-friendly error to the user.
* **MOD-1024** `@asgi_app` and the other two never actually worked, so it's OK to enforce `@asgi_app()` and not just deprecate. 

<kbd>
image
</kbd>


modal/functions.py
```diff
@@ -1247,6 +1258,12 @@ def _web_endpoint(
     * `wait_for_response=True` - tries to fulfill the request on the original URL, but returns a 302 redirect after ~150s to a result URL (original URL with an added `__modal_function_id=...` query parameter)
     * `wait_for_response=False` - immediately returns a 202 ACCEPTED response with a JSON payload: `{"result_url": "..."}` containing the result "redirect" URL from above (which in turn redirects to itself every ~150s)
     """
+    if not isinstance(method, str):
```
erikbern, line 1261: maybe another option is to make `method` a keyword-only argument?

erikbern (approved):
nice!

PR #513, opened 2023-05-01
Secret.from_dotenv

erikbern:
This adds a static factory method `Secret.from_dotenv()` and also changes the old `Secret({...})` constructor to `Secret.from_dict({...})`

Mostly just picked this up because I was going to refactor the `Secret` constructor anyway.

### DIFF
--- a/modal/secret.py
+++ b/modal/secret.py
@@ -45,7 +45,7 @@ class _Secret(_Provider[_SecretHandle]):
 
         Usage:
         ```python
-        @stub.function(secret=Secret.from_dict({"FOO": "bar"})
+        @stub.function(secret=modal.Secret.from_dict({"FOO": "bar"})
         def run():
             print(os.environ["FOO"])
         ```
@@ -85,7 +85,9 @@ class _Secret(_Provider[_SecretHandle]):
         If called with an argument, it will use that as a starting point for finding .env files.
         In particular, you can call it like this:
         ```python
-        Secret.from_dotenv(__file__)
+        @stub.function(secret=modal.Secret.from_dotenv(__file__))
+        def run():
+            print(os.environ["USERNAME"])  # Assumes USERNAME is defined in your .env file
         ```
         """
 
--- a/conftest.py
+++ b/conftest.py
@@ -12,4 +12,5 @@ def pytest_markdown_docs_globals():
         "web_endpoint": modal.web_endpoint,
         "asgi_app": modal.asgi_app,
         "wsgi_app": modal.wsgi_app,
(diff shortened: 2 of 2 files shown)


modal/secret.py
```diff
@@ -32,8 +33,8 @@ class _Secret(_Provider[_SecretHandle]):
     """
 
     @typechecked
-    def __init__(
-        self,
+    @staticmethod
+    def from_dict(
```
freider, line 37: Nice!

modal/secret.py
```diff
@@ -54,8 +55,40 @@ async def _load(resolver: Resolver, existing_object_id: Optional[str]) -> _Secre
...
+                from dotenv import find_dotenv, dotenv_values
+            except ImportError:
+                raise ImportError(
+                    "Need the `dotenv` package installed. You can install it by running `pip install python-dotenv`."
+                )
+
+            # TODO(erikbern): dotenv tries to locate .env files based on the location of the file in the stack frame.
+            # Since the modal code "intermediates" this, a .env file in the user's local directory won't be picked up.
+            # We avoid this by just using the cwd instead.
+            dotenv_path = find_dotenv(usecwd=True)
```
freider, line 80: It's better than using the stack frame (since it doesn't break modal run) but it somehow still feels suboptimal and potentially confusing to use cwd instead of the directory of the run file. Not sure if we can easily get it to use that dir somehow...
One way would be to allow an optional path argument to `from_dotenv`, either with the env file's explicit path, or a path to search upwards from e.g. `Secret.from_dotenv(__file__)` for convenience?

modal/secret.py
```diff
@@ -54,8 +55,40 @@ async def _load(resolver: Resolver, existing_object_id: Optional[str]) -> _Secre
...
+
+    def __init__(self, env_dict: Dict[str, str]):
+        """`Secret({...})` is deprecated. Please use `Secret.from_dict({...})` instead."""
+        deprecation_warning(date(2023, 5, 1), self.__init__.__doc__)
+        obj = _Secret.from_dict(env_dict)
+        self._init_from_other(obj)
+
+    @staticmethod
+    def from_dotenv():
+        async def _load(resolver: Resolver, existing_object_id: Optional[str]) -> _SecretHandle:
```
freider, line 69: Should we throw an explicit error here if we are on <py3.8? (given the dependency condition), or is the import error meant to indicate that issue?

freider (approved):

modal/secret.py
erikbern: yeah good point, let me add that as an argument

PR #514, opened 2023-05-02
Don't warn about not finding stub when using an image builder function

freider:
Followup to #502 

image builder functions aren't strictly linked to stubs, and the apps wouldn't necessarily be loadable anyways, so we shouldn't warn if we can't hydrate the stub when running one

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -408,6 +408,8 @@ message Function {
   string runtime = 30;
 
   string stub_name = 31;
+
+  bool is_builder_function = 32;
 }
 
 message FunctionHandleMetadata {
--- a/client_test/container_test.py
+++ b/client_test/container_test.py
@@ -7,7 +7,7 @@ import pytest
 import subprocess
 import sys
 import time
-from typing import List, Optional
+from typing import List, Optional, Tuple, Dict
 
 from grpclib.exceptions import GRPCError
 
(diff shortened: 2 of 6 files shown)


erikbern:
Random idea but would it make sense to unify builder functions with "normal" stub functions? Like maybe we should enforce that every builder function has a `stub.function` decorator too (only caveat is the image argument wouldn't make sense). Right now they have a different way of specifying their environment.

modal_proto/api.proto
```diff
@@ -408,6 +408,8 @@ message Function {
   string runtime = 30;
 
   string stub_name = 31;
+
+  bool is_builder_function = 32;
```
erikbern, line 412: maybe this should be some sort of function type? but we can revisit this once we have more types

freider: Yeah I was thinking of making it another enum value of the existing function type, but it felt semi-orthogonal to the existing values of function and generator. Agree we can revisit this later

freider:
> Random idea but would it make sense to unify builder functions with "normal" stub functions? Like maybe we should enforce that every builder function has a `stub.function` decorator too (only caveat is the image argument wouldn't make sense). Right now they have a different way of specifying their environment.

Yeah it would make sense to have them more unified, but we'd have to take some care to make sure new non-builder functionality wouldn't leak into the builder function's api.

There is a risk users would assume builder functions work the same as regular app functions (e.g. having access to other app assets, which they don't). Maybe that's a separate concern, and something we could solve with explicit object dependencies (class based functions?)

erikbern:
we should probably also eliminate the need for builder functions for many common use cases!

PR #524, opened 2023-05-04
Prevent double mount of "main" file

freider:
When `modal run`:ing a `.py` file, not using module syntax, the file was previously mounted twice due to FunctionInfo assigning the relative path instead of the absolute path to the file as the `file` attribute.

This fixes this, adds an "end to end" test for the number of mounts/files attached to a deployed stub, and and rearranges the automount code to be somewhat more readable.


aksh-at (approved):
nice, thanks for fixing! Has been bothering me too for a while

PR #527, opened 2023-05-05
[wip] basic setup for remote cls

aksh-at:
```python
from modal import Stub, method

stub = Stub()


@stub.cls()
class Model:
    def __init__(self, x: str, y: int) -> None:
        self.x = x
        self.y = y

    @method()
    def square(self, x):
        return x**2


@stub.local_entrypoint()
def main():
    m = Model.remote("hello", y=42)
    print(m.square.call(42))
```

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -579,6 +579,7 @@ class _Stub:
         ] = None,  # Set this to True if it's a non-generator function returning a [sync/async] generator object
         cloud: Optional[str] = None,  # Cloud provider to run the function on. Possible values are aws, gcp, auto.
         _cls: Optional[type] = None,  # Used for methods only
+        _bound_args: Optional[inspect.BoundArguments] = None,  # Used for parameterized functions only
     ) -> Union[
         _FunctionHandle, Callable[[Union[_PartialFunction, Callable[..., Any]]], _FunctionHandle]
     ]:  # Function object - callable as a regular function within a Modal app
@@ -679,6 +680,7 @@ class _Stub:
             cloud=cloud,
             webhook_config=webhook_config,
             _cls=_cls,
+            _bound_args=_bound_args,
         )
 
         self._add_function(function, [*base_mounts, *mounts])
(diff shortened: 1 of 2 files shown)


modal/stub.py
```diff
@@ -846,10 +848,30 @@ async def interactive_shell(self, cmd=None, image=None, **kwargs):
...
+            bound_args = sig.bind(*args, **kwargs)
+
+            function_handles: Dict[str, _FunctionHandle] = {}
+
+            for k, v in partial_functions.items():
+                function_handles[k] = self.function(**fn_kwargs, _bound_args=bound_args)(v)
+
+            _PartialFunction.initialize_cls(_self, function_handles)
+
+        return type(f"Remote{user_cls.__name__}", (user_cls,), {"__init__": __init__})
```
aksh-at, line 869: Feels weird to subclass the original class here. I think we should only copy over the `_PartialFunction` descriptors.

modal/stub.py
```diff
@@ -846,10 +848,30 @@ async def interactive_shell(self, cmd=None, image=None, **kwargs):
         async with self.run():
             await wrapped_fn.call(cmd)
 
+    def _make_remote_cls(
```
erikbern, line 851: minor preference for keeping this out of `Stub` (since it's a big class and does a lot of different things) but we can move it later since it's not a public method

modal/stub.py
```diff
@@ -870,32 +892,41 @@ def cls(
         cloud: Optional[str] = None,  # Cloud provider to run the function on. Possible values are aws, gcp, auto.
     ) -> type:
         function_handles: Dict[str, _FunctionHandle] = {}
+        partial_functions: Dict[str, _PartialFunction] = {}
+
+        if hasattr(user_cls, "remote"):
+            raise InvalidError(f"Class {user_cls} has an attribute named 'remote' [TODO: expand]")
+
+        kwargs = dict(
```
erikbern, line 900: instead of creating kwargs here, i think you can just call `self.function(... all the arguments ...)` and then get back a Callable[[Callable], Callable]` or something like that which you can use to create the function handles and also pass it to `_make_remote_cls`

just marginally nicer since you get some sort of type safety

erikbern: nbd if i'm being unclear – i can change this later

modal/stub.py
erikbern: yeah i think that makes sense

erikbern: one idea is to define a new base class for all `Remote` classes and inherit from that one. That way you can put the constructor on that class

modal/stub.py
```diff
@@ -870,32 +892,41 @@ def cls(
...
-                    interactive=interactive,
-                    keep_warm=keep_warm,
-                    cloud=cloud,
-                )(partial_function)
+                partial_functions[k] = synchronizer._translate_in(v)  # TODO: remove need for?
+                function_handles[k] = self.function(**kwargs)(partial_functions[k])
 
         _PartialFunction.initialize_cls(user_cls, function_handles)
+
+        user_cls.remote = self._make_remote_cls(user_cls, partial_functions, kwargs)
```
erikbern, line 929: Right now `user_cls.remote` is a class constructor. I think it would make more sense that it's just a factory function that returns a `Remote` instance instead. The benefit of doing that is it can be async and do I/O. So you can have `user_cls` do the RPC call that resolves the bound arguments to some sort of "class id". It seems nice to do that just once per class instantiation.

PR #529, opened 2023-05-06
v0 parametrized function support

aksh-at:
`@stub.cls`-decorated objects now have a `.remote()` function that forwards its arguments to the class constructor. This works by creating a "bound function" from the base that has these parameters registered:

```python
@stub.cls()
class Model:
    def __init__(self, x: str, y: int) -> None:
        self.x = x
        self.y = y

    @method()
    def f(self):
        print(self.x)
        return y**2


@stub.local_entrypoint()
def main():
    model = Model.remote("hello", y=42)
    model.f.call()
```

The pool of functions for each set of parameters scales independently. Also, a set of parameters uniquely determines the pools by value. So calling `Model.remote("hello")` from two different processes will still refer to the same "bound function ID".

### TODO

- [x] Validate that only primitive types are sent
- [x] Tests
- [x] Remove the need for `.call()`
- [x] Type hints?

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -3,6 +3,7 @@ import inspect
 from typing import Dict, Union
 from modal_utils.async_utils import synchronize_apis
 from .functions import _PartialFunction, PartialFunction, AioPartialFunction, _FunctionHandle
+from datetime import datetime
 
 
 class ClsMixin:
@@ -13,6 +14,9 @@ class ClsMixin:
     @staticmethod
     async def aio_remote(*args, **kwargs):
         ...
+
+
+ALLOWED_TYPES = (int, float, bool, str, bytes, type(None), datetime)
 
 
 def make_remote_cls_constructors(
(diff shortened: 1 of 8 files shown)


modal/cls.py
```diff
@@ -0,0 +1,34 @@
+import inspect
+from typing import Dict, Union
+from modal_utils.async_utils import synchronize_apis
+from .functions import _PartialFunction, PartialFunction, AioPartialFunction, _FunctionHandle
+
+
+class RemoteCls:
```
aksh-at, line 7: We don't really need this class for anything right now, might remove.

erikbern (approved):
Nice, looks quite clean and less complexity than I expected – great!

modal/cls.py
```diff
@@ -0,0 +1,33 @@
...
+from .functions import _PartialFunction, PartialFunction, AioPartialFunction, _FunctionHandle
+
+
+def make_remote_cls_constructors(
+    user_cls: type,
+    partial_functions: Dict[str, Union[PartialFunction, AioPartialFunction]],
+    function_handles: Dict[str, _FunctionHandle],
+):
+    original_sig = inspect.signature(user_cls.__init__)  # type: ignore
+    new_parameters = [param for name, param in original_sig.parameters.items() if name != "self"]
```
erikbern, line 14: technically speaking, `self` isn't a keyword in Python, so this would break if someone creates a constructor like 

```python
def __init__(xyz, self: int = 2):
    ...
```

but like, lol

aksh-at: Yeah I suppose I could check it's also the first argument 🤷

PR #532, opened 2023-05-08
Fix retries=0

aksh-at:
Broke a few days ago in refactoring

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -803,7 +803,9 @@ class _Function(_Provider[_FunctionHandle]):
         else:
             self._secrets = secrets
 
-        if isinstance(retries, int):
+        if retries is None or retries == 0:
+            self._retry_policy = None
+        elif isinstance(retries, int):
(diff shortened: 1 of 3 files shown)


modal/functions.py
```diff
@@ -803,16 +803,16 @@ def __init__(
         else:
             self._secrets = secrets
 
-        if isinstance(retries, int):
+        if retries is None or retries == 0:
```
erikbern, line 806: instead of changing this line, can you change this line: https://github.com/modal-labs/modal-client/blob/main/modal/retries.py#L64

erikbern:
my bad for breaking it!

modal/functions.py
aksh-at: I wasn't sure if that would break something on the server (since now we are sending a `RetryPolicy` vs `None` earlier). But it looks like it won't.

erikbern: just seems nicer to put all the validation in one place

PR #535, opened 2023-05-09
Make the `_Function` constructor a factory function

erikbern:
This is one step towards merging `Provider` and `Handle` classes. The reason this makes it a lot easier is that if all construction happens through factory functions, then it doesn't really matter where those are located – which means we can rename `_FunctionHandle` to `_Function` and move the factory function to the other class.

There is way too much state still left on the `_Function` object. Something like 90% of it is needed because of `interactive_shell`, and about 30% is needed for builder functions (there's some overlap). I think we can clean that up but it seems easier to do that in a subsequent PR.


freider (approved):
👍

PR #536, opened 2023-05-10
Move some stuff from stub.py to runner.py

erikbern:
The point is that the `Stub` class is humongous and it's going to be easier to merge `Stub` and `App` if we move a bunch of logic out of those classes. Now that pretty much all usage is through the CLI, we can deprecate the methods on the stub class and move them to `modal.runner` instead.


erikbern:
Reminder to myself: merging this will break a few integration tests, although they are super easy to fix. I'll put together a separate PR to fix those.

PR #537, opened 2023-05-10
Move mount determination logic to _load-time instead of decoration time

freider:
Previously we defined (but didn't load) all mounts for functions when the function was decorated. This included a bunch of file system operations and sometimes complex logic to be performed at function import time, unnecessarily, which also makes writing consistent tests for mounts tricky (can't control import order or monkey patch)

As a "bonus", also adds the package mounting of 'package functions' into the MODAL_AUTOMOUNT conditional.

Results in:
+ Faster imports
+ Less work in containers
+ Easier to test/mock


modal/stub.py
```diff
@@ -371,22 +368,15 @@ def _get_default_image(self):
...
-        # Create client mount
-        if self._client_mount is None:
-            self._client_mount = _get_client_mount()
-        mounts.append(self._client_mount)
+    def _get_local_mounts(self):
+        all_mounts = [
+            *self._mounts,
+        ]
+        for function_handle in self.registered_functions.values():
+            function = self[function_handle._info.get_tag()]
```
erikbern, line 376: this is pretty janky but we can fix it when we merge providers and handles

maybe add a comment that this is ugly though

freider: Yeah it's pretty horrible 😢 And it's only used by `modal serve` in order to determine which mounts to watch for file changes... Adding a comment

PR #540, opened 2023-05-12
[MOD-1252] fix: must set both logger's level and handler's level

thundergolfer:
Follow-up fix to https://github.com/modal-labs/modal-client/pull/470

----

This `logging` API 🤮. God it's confusing.

The key bit of the docs is: 

> Why are there two setLevel() methods? The level set in the logger determines which severity of messages it will pass to its handlers. The level set in each handler determines which messages that handler will send on. — https://docs.python.org/3/howto/logging.html#handlers

This doesn't set the level of the _root_ logger, which is still WARNING (30)

```
> /home/ubuntu/modal-client/modal/config.py(208)<module>()
-> def _store_user_config(new_settings, env=None):
(Pdb) logger.getEffectiveLevel()
10 # debug
(Pdb) logger.root.getEffectiveLevel()
30 # warning 
```


erikbern:
the `logging` package is super annoying – I never understood it

PR #541, opened 2023-05-14
remove support for decorator without parentheses

erikbern:
This takes the deprecation warning to a deprecation error if you use `@stub.function` et al without parentheses.

Once we remove the error, we can simplify this even further.

We started deprecating this on April 5 so it's just 5 weeks so far, but the remedy otoh is very simple – just add empty parentheses. So I think we can make it an error any day now.


freider (approved):
Let's do it

PR #544, opened 2023-05-15
Add logs eof to client proto

aksh-at:
We're going to send an EOF at the end of every task. Helps us properly terminate image logs.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -731,6 +731,7 @@ message ImageJoinStreamingResponse {
   GenericResult result = 1;
   repeated TaskLogs task_logs = 2;
   string entry_id = 3;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -979,6 +980,7 @@ message TaskLogsBatch {
   string function_id = 11;
   string input_id = 12;
   string image_id = 13;  // Used for image logs
+  bool eof = 14;
```
erikbern, line 983: is this one needed?

aksh-at: Yeah, the worker needs to send a logs batch with `eof` (see https://github.com/modal-labs/modal/pull/6040)

PR #553, opened 2023-05-22
Simplify SDK for adding local files to an image

erikbern:
todo: docstrings

question: should we support making the second argument optional? e.g. `modal.Image.debian_slim().add_local_file("xyz.txt")` would put it in `/root` or something?

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -314,7 +314,7 @@ class _Image(_Provider[_ImageHandle]):
         return _Image._from_args(base_images={"base": self}, **kwargs)
 
     @typechecked
-    def add_mount(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Image":
+    def copy_mount(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Image":
         """Copy the entire contents of a `modal.Mount` into an image.
         Useful when files only available locally are required during the image
         build process.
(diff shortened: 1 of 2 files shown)


modal/image.py
```diff
@@ -314,7 +314,7 @@ def extend(self, **kwargs) -> "_Image":
         return _Image._from_args(base_images={"base": self}, **kwargs)
 
     @typechecked
-    def copy(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Image":
+    def add_mount(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Image":
```
aksh-at, line 317: I think `os.PathLike` might be better than `Path` for what we accept here, since it's broader. Probably for a separate PR, since it was already this way.

modal/image.py
```diff
@@ -336,6 +336,24 @@ def copy(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Image":
...
 
+    def copy(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Image":
+        deprecation_warning(
+            date(2023, 5, 21),
+            "`Image.copy` is deprecated in favor of `Image.add_mount`, `Image.add_local_file`,"
+            " and `Image.add_local_dir`.",
+        )
+        return self.add_mount(mount, remote_path)
+
+    def add_local_file(self, local_path: Union[str, Path], remote_path: Union[str, Path]) -> "_Image":
```
aksh-at, line 347: `copy_local_file` (and similar for the others) might be slightly better so that people reaching for Dockerfile `COPY` find this

aksh-at (approved):
Nice! I think `copy_*` could be better, wdyt?

> question: should we support making the second argument optional? e.g. modal.Image.debian_slim().add_local_file("xyz.txt") would put it in /root or something?

Yeah that would make some cases easier (since it'll be in `cwd` in the container when we run). What if I have 
`modal.Image.debian_slim().add_local_file("subfolder/xyz.txt")`? I might expect it to have the same relative path inside the container maybe?

PR #555, opened 2023-05-22
Don't mount the origin package for each function more than once

freider:
Regardless if there are multiple functions in the stub. This patch simply reintroduces the _function_mounts stub-level mount cache that we had until last week.

It's a bit janky, but we should refactor this mount logic soon and get rid of a bunch of this. Preferably in a way that lets us upload all mount files in one common thread pool instead of using one for each mount sequentially

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -859,7 +859,7 @@ class _Function(_Provider[_FunctionHandle]):
 
         all_mounts = [
             _get_client_mount(),  # client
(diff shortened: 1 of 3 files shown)


erikbern (approved):
Seems good as a quick fix.

I think the nicer fix here is probably that we build up a dict of (module -> mount_id) as we add functions

thundergolfer:
@prbot approve

PR #557, opened 2023-05-23
Make 'no remote calls' into a real python warning, written to stderr

freider:
Had some issues when writing tests examining stdout due to this warning appearing there instead of stderr


aksh-at:
Oops yeah I should've used `warnings`, thanks for fixing!

PR #558, opened 2023-05-23
Load function sub-entities in parallel

freider:
Load all mounts, secrets, images and shared_volumes for a function in parallel


freider:
> This is a great speedup! My main concern is let's make sure it works if there are multiple objects depending on the same object – something like this has a risk of introducing race conditions.

I think this is resolved by the new "shared futures" logic in the Resolver (see the test_multi_resolve_concurrent_loads_once test I added as well). Basically "get or create" a new asyncio task for each load on the same object local uuid, so all dependents resolve by the same future and only load once. Not sure if I missed thinking about some aspect of it though 🤔

erikbern:
alright cool, if you feel confident then let's ship it! 🚢

freider:
> alright cool, if you feel confident then let's ship it! 🚢

Can haz ✅ ?

erikbern (approved):

erikbern:
@freider what's the status of this one?

freider:
There was some unrelated test-flake and then pip publishing issues that stalled the merge (wanted to be available to monitor after the merge in case something unexpected started breaking for new users).
Merging now :)

PR #560, opened 2023-05-23
Fix bump build number

aksh-at:
PR bot still runs into branch protection rules. Using `GITHUB_TOKEN` on the step itself seems unnecessary according to the docs, but trying it.


aksh-at:
Seems like I can't test this here since it pushes to this branch, not main. Will have to test live.

erikbern:
yeah there's already a `GITHUB_TOKEN` available

PR #564, opened 2023-05-24
Rename cli "envs" to "profiles"

freider:
Some quick fixes ahead of introducing a new concept of Environments which is different from the "env" we have been using for the CLI profiles


erikbern (approved):
Nice! Can probably remove support for the old stuff quite soon

PR #565, opened 2023-05-24
Fix markdown glitch

erikbern:
@freider shouldn't this have been caught by the doc tests? weird


freider:
The doc tests rely on the `markdown_it` parser to extract fence blocks, but it doesn't check for invalid/unclosed fences (maybe those are still valid Markdown?)

PR #567, opened 2023-05-25
Remove old usage of aio from tests, make deprecation non-pending

erikbern:
This basically just gets rid of all old usage of `modal.aio` from tests, and changes the deprecation warning so it's no longer pending.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -52,8 +52,8 @@ jobs:
         run: pytest -v -s client_test
 
       - name: Run docstring tests
-        run: pytest -s --markdown-docs -m markdown-docs modal
-       
+        run: pytest -s --markdown-docs -m markdown-docs modal --ignore modal/aio.py
+
 
   publish-base-images:
     name: |
--- a/client_test/aio_test.py
+++ b/client_test/aio_test.py
@@ -1,0 +1,24 @@
+import pytest
+
+from modal.exception import DeprecationError, InvalidError
+
+
+@pytest.mark.asyncio
+async def test_deprecated(servicer, client):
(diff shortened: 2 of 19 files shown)


erikbern:
@freider mypy is going nuts on some of this, any idea why?

erikbern:
This is also failing on doctests btw, but that should be easy to fix?

freider:
> @freider mypy is going nuts on some of this, any idea why?

Mainly two things.
**Missing type stubs for "private synchronicity wrapped functions"**
One is that .aio interfaces can't be easily communicated to mypy at runtime and we don't currently generate type stubs for the test files, so any wrapped classes that are created in the tests themselves, like in blob_test.py:

```
blob_upload, _ = synchronize_apis(blob_upload)
blob_download, _ = synchronize_apis(blob_download)
``` 

Won't let mypy infer the types of the returned values. Apparently, it's smart enough to figure out implicit types of wrapped functions as being functions, so this messes it up a bit. Either we could create type stubs for the offending test files, or we add some `#type: ignore` comments to fix this.

**Context managers are now intersection types that have both async and blocking context manager functionality**
This needs to be fixed in synchronicity so we properly set the type annotation on any wrapped context manager functions to *both* rather than only one type of context manager. Another option would be to move the async context manager to an `.aio.` attribute on the blocking one. I kind of like the first option, but it will take a little work and might not be super high prio, so perhaps we can just add some type: ignore for these cases as well?

freider:
Realized I could fix the context manager typing pretty easily after all, here is a PR in synchronicity:
https://github.com/modal-labs/synchronicity/pull/111

freider (approved):
Nice

PR #568, opened 2023-05-26
Turn deprecation_warning into deprecation_error

erikbern:
We had a few warnings since April 4, it's almost two months now.

I'll merge this next week probably.

Slowly reducing the complexity of the `Stub` class, which will make it a lot easier to merge it with `App`

### DIFF
--- a/client_test/stub_test.py
+++ b/client_test/stub_test.py
@@ -1,7 +1,6 @@
 # Copyright Modal Labs 2022
 import asyncio
 import logging
-import os
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


client_test/stub_test.py
```diff
@@ -164,48 +164,6 @@ def test_same_function_name(caplog):
     assert "square" in caplog.text
 
 
-skip_in_github = pytest.mark.skipif(
```
erikbern, line 167: don't think it was worth keeping these tests – `stub.serve` was kind of a dummy shim anyway, and these tests are skipped in github.

We have tests for live reloading anyway.

freider: :+1:

freider (approved):

PR #572, opened 2023-06-01
automate releases of alias package

erikbern:
This will create a corresponding package version of `modal` for every release of `modal-client`

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -79,7 +79,26 @@ def update_build_number(ctx, new_build_number):
 
     assert new_build_number > current_build_number
     with open("modal_version/_version_generated.py", "w") as f:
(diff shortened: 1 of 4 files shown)


tasks.py
```diff
@@ -79,7 +79,26 @@ def update_build_number(ctx, new_build_number):
...
+
+
+@task
+def create_alias_package(ctx):
+    from modal_version import __version__
+
+    with open("alias-package/setup.py", "w") as f:
+        f.write(
+            f"""\
+{copyright_header_full}
```
aksh-at, line 97: The other `setup.py` also adds cwd to the path, do we need that here?

aksh-at (approved):
nice!

PR #575, opened 2023-06-01
pin dependency of modal-client in alias package

erikbern:
Realized we weren't pinning the version to be the same, fixed that. Since I had to generate that file automatically too, I just got rid of the 3rd file and we're now generating all 3 project files dynamically.


erikbern:
@prbot approve

PR #576, opened 2023-06-02
Run ssort

erikbern:
Demo PR just showing the delta


erikbern:
will do this a few modules at a time instead

PR #578, opened 2023-06-05
ssort 1

erikbern:
Runs ssort on all code not in the `modal` package


erikbern:
not sure if it's possible to configure ssort's behavior so i'll close this for now :(

PR #583, opened 2023-06-10
Fix copy_local_file

erikbern:
This doesn't change the behavior of `copy_local_dir`, but I changed it so it also inlines the `.extend` call – so there's more symmetry with `copy_local_file`


erikbern:
Merging this since it's completely broken

PR #590, opened 2023-06-13
Raise an error if someone tries to start an app from a stub that is already starting or running

freider:
## Describe your changes

Previously, it could lead to unexpected behavior in case a user accidentally put a `stub.run()` block within their local entrypoint and then called the script using `modal run`. The second call would take over the output management and overwrite the original app run's output, making it look like the app hangs during setup. User ran into this in beta-testing slack

I don't love adding this extra stuff the already large stub class, but I don't see an easy way around that in this case?

### DIFF
--- a/client_test/cli_test.py
+++ b/client_test/cli_test.py
@@ -208,6 +208,15 @@ def test_run_local_entrypoint(servicer, set_env_client, test_dir):
     assert len(servicer.client_calls) == 4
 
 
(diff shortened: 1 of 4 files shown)


freider:
I did consider it but then didn't go that route for some reason. Can't see why now that I look at the code, so I'll change to using the already existing _set_app route (should be a super easy change)

PR #593, opened 2023-06-14
wait for inputs to be finished to avoid server errors

gongy:
The server rejects inputs if there are already 25000 inputs waiting to be finished.

This client change reduces the number of unnecessary API calls trying to append inputs, when it must wait for more inputs to be finished in order to continue uploading inputs.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -303,6 +303,7 @@ MAP_INVOCATION_CHUNK_SIZE = 100
 # Maximum number of unfinished inputs for maps
 MAP_INVOCATION_CONCURRENCY = 20000
 
+
 async def _map_invocation(
     function_id: str,
     input_stream: AsyncIterable[Any],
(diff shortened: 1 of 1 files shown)


modal/functions.py
```diff
@@ -355,6 +359,11 @@ async def pump_inputs():
             logger.debug(
                 f"Pushing {len(items)} inputs to server. Num queued inputs awaiting push is {input_queue.qsize()}."
             )
+
+            # Wait for inputs to be finished before continuing, as server will deny put inputs request
+            while len(pending_outputs) - len(completed_outputs) + len(items) > MAP_INVOCATION_CONCURRENCY:
+                await asyncio.sleep(0.5)
```
freider, line 365: This could potentially "lock up" the client without any feedback to the user in case nothing is consumed from the map in a long time (or at all). Perhaps add some logging at a certain time interval, saying something like "Too many queued inputs to the map - waiting for caller to consume outputs before adding additional inputs" to help the user understand what's happening?

gongy: I think this could be better integrated into the progress bar potentially.

A notion of completed inputs, uploaded inputs and total inputs?

PR #595, opened 2023-06-15
Blocked: Add client + cli support for environments

freider:
Environments changes for the CLI and other aspects of the client (`.lookup()` etc). Requires backend changes (https://github.com/modal-labs/modal/pull/6314) before this works in prod

* Basic CR(U)D operations for environments.
```
modal environment create dev
modal environment list
modal environment delete dev
```

* `--env` flags to most other CLI commands, allowing overriding of whatever the default environment is
```
modal run app.py --env=prod
modal deploy app.py --env=prod

modal app list --env=prod
modal secret create --env=staging
```

* MODAL_ENVIRONMENT (and the "environment" value of the modal.toml profile) allows setting off a default environment per workspace. Can be set via cli:
```
modal config set-environment dev  # sets default environment in current profile, client-side
```

* If no environment is specified in either `--env` or as a setting/env-var, an empty string will be sent (same as legacy clients). This will defer to the backend to assume there is a single environment and use that one, and throw an error otherwise. Since every workspace has an environment by default, that one will be used for all legacy calls, and when not specifying anything client-side.

TODO:
* Add some more client-side tests, esp. for "implied" environment inheritance when using the --env flag and programmatic deploys
* When user creates an environment, ask if it should be default? (if no default is set in the modal env, commands might start failing otherwise)
* When a command fails due to multiple environments existing - tell user/ask user how to set a default

TODO (backend):
* For special cases to work inside of containers (modal `.lookup()` without an explicit environment specified), and to allow users to introspect environment from within a container, we need to pass the "MODAL_ENVIRONMENT" env var to the container entrypoint from the worker as well.


erikbern (approved):
Lots of code – it looks fairly straightforward though. I think it will be helpful to play around with this once it's merge to really understand how it works

PR #598, opened 2023-06-16
Stub methods: turn deprecation warnings into errors

erikbern:
Also deleted a bunch of tests for these deprecated methods.

This should make it possible to delete a lot of the stub code soon


freider (approved):
feels good!

PR #599, opened 2023-06-21
Add more powerful mock servicer w/ interception

freider:
Taken as is from internal modal repo.
Adds a simple test demonstrating usage

### DIFF
--- a/modal_utils/grpc_testing.py
+++ b/modal_utils/grpc_testing.py
@@ -1,0 +1,125 @@
+import contextlib
+import inspect
+from collections import deque, defaultdict
+from typing import Any, Optional
+
+from grpclib import GRPCError, Status
+
+
+def patch_mock_servicer(cls):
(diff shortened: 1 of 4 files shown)


modal_utils/grpc_testing.py
```diff
@@ -0,0 +1,125 @@
+import contextlib
```
freider, line 1: This file was taken verbatim from our internal repo, and changed to use `typing.List` and `typing.Tuple` instead of builtin generics `list` and `tuple`

erikbern (approved):
Nice – makes sense! And this will definitely make tests a lot easier to write.

I generally don't like monkeypatching so my only comment is – is it possible to do this without it? (for instance by creating a subclass on the fly or something)

erikbern:
Ok I think maybe I was just hung up on the term "patch" – looking at the code now and I think I see what you're saying. Nice! Looking forward to seeing how we can use this in actual tests

PR #600, opened 2023-06-21
Fix cpu lower limit

rachelspark:
Raises InvalidError when attempting to request <0.25 CPU cores


erikbern:
nice – might want to add a unit test too of it's easy (not worth it if it's hard)

PR #601, opened 2023-06-22
Extend env tests + testing utils

freider:
Adding additional tests for use of `--env` flags with different CLI commands, making use of and extending the servicer interceptor functionality for test-local servicers


modal_utils/grpc_testing.py
```diff
@@ -76,45 +78,92 @@ async def intercepted_method(servicer_self, stream):
...
-    def add_response(self, method_name: str, custom_response: Optional[List[Any]] = None):
-        # adds one response to a queue of responses for method method_name
-        if custom_response is not None:
-            assert isinstance(
-                custom_response, list
-            ), "custom_response should be a list of messages sent back by the servicer"
-
-        if custom_response:
-            self.custom_responses[method_name].append(custom_response)
+    def add_response(
```
freider, line 96: Basically we add two new things to the `add_response` functionality:
1. A `request_filter` argument taking a filter function (in addition to the method name) so you can selectively respond when there are multiple requests to the same method
2. Automatic assertion at the end of the interception-block that all `add_response()` payloads have been consumed, so you don't accidentally pass a test even though the assigned value isn't even used

client_test/cli_test.py
```diff
@@ -361,6 +361,36 @@ def test_volume_get(set_env_client):
             assert f.read() == "foo bar baz"
 
 
-def test_environment_flag(set_env_client, test_dir, servicer):
-    stub_file = test_dir / "supports" / "app_run_tests" / "default_stub.py"
-    _run(["run", "--env=staging", str(stub_file)])
+@pytest.mark.parametrize("command", [["run"], ["deploy"], ["serve", "--timeout=1"], ["shell"]])
+@pytest.mark.usefixtures("set_env_client", "mock_shell_pty")
```
freider, line 365: Just found this: `pytest.mark.usefixtures` allows usage of fixtures without injection, for a bit cleaner test function signatures without unused variables...

erikbern (approved):

PR #608, opened 2023-06-26
Remove __init__ from Dict & Queue

erikbern:
This replaces `__init__` with a static factory method.

This will make it a bit easier to merge providers and handles, and it makes these constructors consistent with all other objects.

These objects aren't widely used so hopefully it won't make a big user impact.

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -1,4 +1,5 @@
 # Copyright Modal Labs 2022
+from datetime import date
 from typing import Any, Optional
 
 from modal_proto import api_pb2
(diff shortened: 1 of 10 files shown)


modal/dict.py
```diff
@@ -134,7 +137,9 @@ class _Dict(_Provider[_DictHandle]):
     ```
     """
 
-    def __init__(self, data={}):
+    @typechecked
+    @staticmethod
+    def new(data={}):
```
freider, line 142: Don't we use `Obj.from_x` naming for our other providers? Maybe we should do that here as well, since it takes an argument (albeit optionally, so I can see the point for `new` as well...)

freider (approved):
Cool. Maybe we should add a simple CHANGELOG file to the repo where we can add some short notice whenever we introduce new syntax, feels like it wouldn't be much overhead as long as it's in the same repo?

erikbern:
> Cool. Maybe we should add a simple CHANGELOG file to the repo where we can add some short notice whenever we introduce new syntax, feels like it wouldn't be much overhead as long as it's in the same repo?

Yeah makes sense, I think – what's the best way to maintain that do you think?

PR #611, opened 2023-06-27
[MOD-1371] Add experimental volumes API

gimaker:
This is API for working with volumes, which are durable storage units exposed via the file system. Unlike a networked file system changes to the volumes are only made visible when explcitly committed. Similarly, to get updates to the volume you need to explicitly reload.

Note that the backend support for volumes is not yet in place, so they are non-functional as is.

### DIFF
--- a/client_test/conftest.py
+++ b/client_test/conftest.py
@@ -129,6 +129,11 @@ class MockClientServicer(api_grpc.ModalClientBase):
 
         self.function2params = {}
 
+        self.volume_counter = 0
+        # Volume-id -> commit/reload count
+        self.volume_commits: Dict[str, int] = defaultdict(lambda: 0)
+        self.volume_reloads: Dict[str, int] = defaultdict(lambda: 0)
+
         @self.function_body
         def default_function_body(*args, **kwargs):
             return sum(arg**2 for arg in args) + sum(value**2 for key, value in kwargs.items())
@@ -625,6 +630,21 @@ class MockClientServicer(api_grpc.ModalClientBase):
                 token_secret="xyz",
             )
         )
+
+    async def VolumeCreate(self, stream):
+        await stream.recv_message()
+        self.volume_counter += 1
+        await stream.send_message(api_pb2.VolumeCreateResponse(volume_id=f"vo-{self.volume_counter}"))
+
+    async def VolumeCommit(self, stream):
+        req = await stream.recv_message()
+        self.volume_commits[req.volume_id] += 1
+        await stream.send_message(Empty())
+
+    async def VolumeReload(self, stream):
+        req = await stream.recv_message()
+        self.volume_reloads[req.volume_id] += 1
+        await stream.send_message(Empty())
 
 
 @pytest_asyncio.fixture
--- a/client_test/volume_test.py
+++ b/client_test/volume_test.py
@@ -1,0 +1,82 @@
+# Copyright Modal Labs 2023
+
+import pytest
+
+import modal
+from modal.exception import InvalidError
+from modal.volume import VolumeHandle
+
+from .supports.skip import skip_windows
+
+
+def dummy():
+    pass
+
+
+def test_volume_mount(client, servicer):
+    stub = modal.Stub()
+
+    dummy_modal = stub.function(
+        volumes={"/root/foo": modal.Volume()},
+    )(dummy)
+
+    with stub.run(client=client):
+        dummy_modal.call()
+
+
+@skip_windows("TODO: implement client-side path check on Windows.")
+def test_volume_bad_paths(client, servicer):
+    stub = modal.Stub()
+
+    dummy_modal = stub.function(volumes={"/root/../../foo": modal.Volume()})(dummy)
+    with pytest.raises(InvalidError):
+        with stub.run(client=client):
+            dummy_modal.call()
+
+    dummy_modal = stub.function(volumes={"/": modal.Volume()})(dummy)
+    with pytest.raises(InvalidError):
+        with stub.run(client=client):
+            dummy_modal.call()
+
+    dummy_modal = stub.function(volumes={"/tmp/": modal.Volume()})(dummy)
+    with pytest.raises(InvalidError):
+        with stub.run(client=client):
+            dummy_modal.call()
+
+
+def test_volume_duplicate_mount(client, servicer):
+    stub = modal.Stub()
+
+    volume = modal.Volume()
+    dummy_modal = stub.function(volumes={"/foo": volume, "/bar": volume})(dummy)
+    with pytest.raises(InvalidError):
+        with stub.run(client=client):
+            dummy_modal.call()
+
+
+def test_volume_commit(client, servicer):
+    stub = modal.Stub()
+    stub.vol = modal.Volume()
+
+    with stub.run(client=client) as app:
+        handle = app.vol
+        assert isinstance(handle, VolumeHandle)
+        # Note that in practice this will not work unless run in a task.
+        handle.commit()
+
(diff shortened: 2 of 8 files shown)


modal/volume.py
```diff
@@ -0,0 +1,106 @@
...
+from modal_utils.grpc_utils import retry_transient_errors
+
+from ._resolver import Resolver
+from .object import _Handle, _Provider
+
+
+class _VolumeHandle(_Handle, type_prefix="vo"):
+    """mdmd:hidden Handle to a `Volume` object."""
+
+    async def commit(self):
```
gimaker, line 15: commit and reload will only work when run inside a task.

I'm assuming users can call these outside of a task as well somehow? If so, we detect that to raise an appropriate error?

modal_proto/api.proto
```diff
@@ -1060,6 +1062,31 @@ message TokenFlowWaitResponse {
   bool timeout = 3;
 }
 
+message VolumeCreateRequest {
```
gimaker, line 1065: Proto changes could be a separate PR.

But I'm not sure there's a point to that - easier to review the proto changes with some context on how it will be used, I reckon.

modal/volume.py
```diff
@@ -0,0 +1,106 @@
...
+from modal_proto import api_pb2
+from modal_utils.async_utils import synchronize_api
+from modal_utils.grpc_utils import retry_transient_errors
+
+from ._resolver import Resolver
+from .object import _Handle, _Provider
+
+
+class _VolumeHandle(_Handle, type_prefix="vo"):
+    """mdmd:hidden Handle to a `Volume` object."""
```
gimaker, line 13: Didn't spend a huge amount of effort on documentation - figure we'll revisit that once this is no longer experimental and `mdmd:hidden` is removed.

client_test/volume_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+def dummy():
+    pass
+
+
+def test_volume_mount(client, servicer):
+    stub = modal.Stub()
+
+    dummy_modal = stub.function(
+        volumes={"/root/foo": modal.Volume()},
```
freider, line 20: I think @erikbern has mentioned him not liking the dict-style volume assigment of the current shared volumes and having some other idea of how to specify the mount point, and this might be a good time to introduce it since this is a new entity anyways.
Although I kind of like the "dict mounting" syntax :)

client_test/volume_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@skip_windows("TODO: implement client-side path check on Windows.")
+def test_volume_bad_paths(client, servicer):
+    stub = modal.Stub()
+
+    dummy_modal = stub.function(volumes={"/root/../../foo": modal.Volume()})(dummy)
+    with pytest.raises(InvalidError):
+        with stub.run(client=client):
+            dummy_modal.call()
```
freider, line 34: I think you should be able to skip the function calls and just do `pass` within the run blocks, as all the loading logic should reside within the setup of app/context manager anyways

modal/functions.py
```diff
@@ -976,6 +980,26 @@ async def _load(resolver: Resolver, existing_object_id: Optional[str]) -> _Funct
...
+            def _validate_mount_points(
+                display_name: str, volume_likes: typing.Dict[str, typing.Union[_Volume, _SharedVolume]]
+            ) -> typing.List[typing.Tuple[str, typing.Union[_Volume, _SharedVolume]]]:
+                validated = []
+                for path, vol in volume_likes.items():
+                    path = PurePath(path).as_posix()
+                    abs_path = posixpath.abspath(path)
+
+                    if path != abs_path:
+                        raise InvalidError(f"{display_name} {abs_path} must be a canonical, absolute path.")
```
freider, line 994: Displaying `path` instead of `abs_path` in the error makes things a bit clearer I think, since the abspath will 1. be an absolute path, and 2. contain a local prefix that the user didn't specify in the code in case they accidentally send in a relative path

modal/functions.py
```diff
@@ -1049,8 +1087,12 @@ async def shared_volume_loader():
             if stub and stub.name:
                 stub_name = stub.name
 
-            mount_ids, secret_ids, image_id, shared_volume_mounts = await asyncio.gather(
-                _load_ids(all_mounts), _load_ids(secrets), image_loader(), shared_volume_loader()
+            mount_ids, secret_ids, image_id, shared_volume_mounts, volume_mounts = await asyncio.gather(
```
freider, line 1091: 🚗 💨

client_test/volume_test.py
gimaker: I also don't love the dict-style - it's hard to extend in the future if you need to.

E.g. adding a `allow_cross_regional` flag per-shared volume, or adding a mount as `read-only` flag or such.

modal/volume.py
gimaker: > I guess the actual implemention will live in the worker server, so we should be able to just raise an error from another implementation of the extermal rpc call that we put in the actual backend server, which would be the one responding to any non-task client-side calls.

Good idea :+1: 

> Thinking a bit more about it we should also make sure that a user gets an error if they try to commit to a Volume which isn't attached to the currently running task (the task could technically access a "remote" volume)

They will get an error in that case, currently:

Screenshot from 2023-06-27 17-07-11

Should refine that a bit though, to hint that only volumes attached to the current task can be committed/reloaded.

modal/functions.py
gimaker: Makes sense, can do (using `abs_path` is old code, just moved around).

client_test/conftest.py
```diff
@@ -626,6 +631,21 @@ async def TokenFlowWait(self, stream):
             )
         )
 
+    async def VolumeCreate(self, stream):
```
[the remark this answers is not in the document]
gimaker: While nicer, as is it's more code and work to use it :upside_down_face: think we need to add some nice shared convenience methods to make for better ergonomics!

modal/volume.py
gimaker: https://github.com/modal-labs/modal/pull/6571/commits/2dc5141f95a4ffe51c07392a73cc7ff2dff75e43 - for returning errors when called outside of a function.

client_test/conftest.py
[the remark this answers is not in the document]
freider: Yeah we should add some more `ctx.assert_calls_**` type of assertions to it. Inspired by https://github.com/nock/nock

client_test/volume_test.py
freider: Yeah it might be that it's better to be able to assign some metadata about the function-volume relation through some other notation, e.g.
```python
vol = Volume()

@stub.function(volumes=[vol.mount("/my_vol", read_only=True)])
```

freider: Should be pretty easy to add as a client-only "private" type `AttachedVolume` which is returned by `vol.mount` and is the required type for the `volumes` parameter to functions (but the underlying Modal object type would still be Volume)

freider (approved):

client_test/volume_test.py
gimaker: Let's leave API improvements for a follow-up commit, if you're okay with that? We have time to change things around before this features goes public.

There's also the idea that I floated of unifying volume mounting for all different kinds of volumes, e.g.:

```python
vol1 = modal.SharedVolume()
vol2 = modal.NewFangledVolume()

@stub.function(volumes={"/vol1": vol1, "/vol2": vol2}) # <-- Shared volumes and "volumes" are attached to the function the same way
def f():
    ...
```

freider: 👍

PR #614, opened 2023-06-30
CLI for changin names and web suffixes of environments

freider:
Still a bit rough around the edges UI-wise, but this should at least unblock people who want to rename their "main" environment etc.

### DIFF
--- a/modal/cli/environment.py
+++ b/modal/cli/environment.py
@@ -2,6 +2,8 @@
 import json
 from typing import Optional
 
+from click import UsageError
+from google.protobuf.wrappers_pb2 import StringValue
 from typing_extensions import Annotated
 
 import typer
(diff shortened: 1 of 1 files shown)


modal/cli/environment.py
```diff
@@ -94,3 +96,38 @@ async def delete(
...
+
+    if set_name is not None:
+        if len(set_name) < 1:
+            raise UsageError("The environment name cannot be empty")
+
+        new_name = StringValue(value=set_name)
+    if set_web_suffix is not None:
+        new_web_suffix = StringValue(value=set_web_suffix)
+
+    update_payload = api_pb2.EnvironmentUpdateRequest(
```
erikbern, line 127: this is a more general note but i think it would be nice to keep actual "biz logic" out of the CLI code so that people can also call it from Python (this could be said about a lot of our code)

erikbern (approved):

modal/cli/environment.py
erikbern: i mean putting it in some module that's easier to import and call

freider: Good point, will fix for all the environment utilities in followup patch

PR #615, opened 2023-06-30
Prevent concurrent operations on a single volume

gimaker:
We don't support concurrent reload/commit requests for the same volume. Most of the time this will just result in an error but under certain (unlikely) circumstances this will cause problems.

Use a lock on the volume handle to prevent this. You can still create concurrent operations by instantiating several handles (e.g. via `Volume.lookup()`), but this should prevent the most common case. Good enough?

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -1,4 +1,5 @@
 # Copyright Modal Labs 2023
+import asyncio
 from typing import Optional
 
 from modal_proto import api_pb2
@@ -12,6 +13,12 @@ from .object import _Handle, _Provider
 class _VolumeHandle(_Handle, type_prefix="vo"):
     """mdmd:hidden Handle to a `Volume` object."""
 
+    _lock: asyncio.Lock
+
+    def _initialize_from_empty(self):
+        # To prevent multiple concurrent operations on the same volume.
+        self._lock = asyncio.Lock()
+
     async def commit(self):
         """Commit changes to the volume and fetch any other changes made to the volume by other tasks.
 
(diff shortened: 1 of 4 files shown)


modal/volume.py
```diff
@@ -35,8 +40,17 @@ async def reload(self):
 
         Reloading will fail if there are open files for the volume.
         """
-        req = api_pb2.VolumeReloadRequest(volume_id=self.object_id)
-        _ = await retry_transient_errors(self._client.stub.VolumeReload, req)
+        await self._do_reload()
+
+    async def _do_reload(self, lock=True):
+        if lock:
```
gimaker, line 46: Easy* conditional `async with`s with seems to be a Python 3.10+ feature :(

```
async with self._lock if lock else contextlib.nullcontext():
  ...
```

*: You could make your own async null context manager, but that remedy seems worse than the issue it solves.

modal/volume.py
```diff
@@ -12,6 +13,9 @@
 class _VolumeHandle(_Handle, type_prefix="vo"):
     """mdmd:hidden Handle to a `Volume` object."""
 
+    # To prevent multiple concurrent operations on the same volume.
+    _lock: asyncio.Lock = asyncio.Lock()
```
freider, line 17: I think this would actually create a lock so you can use *one of any* VolumeHandles on your running task, since the assignment is in class scope (this would not create an instance-specific variable in Python as might be expected in other languages, but a class variable). 

To set an instance variable you'd normally set it in your `__init__` method, but since our _Handle's are a bit special you'd set it by overloading _initialize_from_empty, e.g. 
```python
def _initialize_from_empty(self):
    self._lock = asyncio.Lock()
```

The type annotation can remain in the class scope however

modal/volume.py
freider: It's actually not too bad, and we have one from before in `client_test.cli_test` (`noop_async_context_manager`) - you could move it out into the modal_utils.async_utils module or something and reuse it in both places:

```
@asynccontextmanager
async def asyncnullcontext(*args, **kwargs):
    yield
```

modal/volume.py
gimaker: Doh, yeah...

gimaker: Was testing a class lock for this scenario btw:

```
    stub = modal.Stub()
    stub.vol = modal.Volume()

    deploy_stub(stub, "vol", client=client)

    f = Volume.lookup("vol", client=client)
    g = Volume.lookup("vol", client=client)
    assert f is g  # nope
```

freider (approved):

PR #618, opened 2023-07-01
Remove some old deprecated code

erikbern:
Some code that's raised deprecation errors for a long time


aksh-at (approved):
Should we bump up the minor version manually here? We haven't really been doing that, but slightly nicer to do for changes like this 🤷

erikbern:
yeah why not, we haven't been doing that much lately!

PR #621, opened 2023-07-03
fix bug where profile-default environment wasn't used for `modal run`

freider:
The environment set by environment variable or `.modal.toml` wasn't being used by `modal run` since it had the empty string as the default instead of None (empty string is a valid value used to communiate "use the only existing one", for compatibility with old clients)


aksh-at (approved):
Thanks for adding the test! Maybe we should fall back to the default when `env=""` as well? But I don't know why someone would do that intentionally

PR #625, opened 2023-07-03
Reduce implementation inheritance

erikbern:
I have a mild aversion towards implementation inheritance so let's try to avoid it


erikbern:
needs a bit more work, not worth it

PR #626, opened 2023-07-03
SharedVolume -> NetworkFileSystem

erikbern:
Mostly just as a PoC. This just renames everything but we also have to add back the old interface with deprecation warnings.


freider (approved):
I like it! But what about the deprecation path?

erikbern:
Yeah I didn't implement the deprecation path yet – shouldn't be super hard though:
1. A class `_SharedVolume` that inherits from `NetworkFileSystem` but raises deprecation warnings for all methods
2. Something for `modal volume` that tells people to use `modal nfs`
3. Tests that deprecation works

Can do that next if we're cool with names!

ekzhang (approved):

erikbern:
Made all warnings pending for now. That way we won't break integration tests and examples right away (hopefully).

Will fix those and then make it a non-pending warning.

PR #632, opened 2023-07-05
Update base images to use Debian 12 (Bookworm)

ekzhang:
Our current version, Bullseye, is now 2 years old. This causes problems for some users, for example, when they install `libpq-dev`, only to realize that they're getting an outdated version v13.10, but v14 was released in September 2022 and is required for some providers like Neon.

Bumping base image versions should only happen once every 2 years on the Debian release schedule and is a relatively painless process for most common workloads. The biggest change is that the supported glibc version increases, and newer software versions are available in the apt repository.

This change applies to not just debian_slim, but also conda and micromamba that are based on Debian.

**We plan to merge this not immediately** but batched with the next time we update the client requirements.txt file, to limit image rebuilds for users. I'm also updating the latest Python patch versions included in our base images.

### DIFF
--- a/client_test/image_test.py
+++ b/client_test/image_test.py
@@ -12,8 +12,7 @@ from modal_proto import api_pb2
 
 
 def test_python_version():
-    assert _dockerhub_python_version("3.9.1") == "3.9.15"
-    assert _dockerhub_python_version("3.9") == "3.9.15"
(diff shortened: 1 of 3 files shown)


luiscape (approved):

erikbern:
Nice!

ekzhang:
Superseded by @mwaskom's plan for client image builds / stack versions with CalVer

PR #636, opened 2023-07-05
Make it possible to set a default cloud from config or an env var

erikbern:
Primarily planning to use this for internal synthetic monitors, but this can also be useful for cases where people only want to use a particular cloud (and more importantly, a particular region in the future)


thundergolfer (commented):
LGTM

PR #637, opened 2023-07-06
Replace `create_package_mounts` with `Mount.from_local_python_packages`

freider:
Adds pending deprecation for create_package_mounts.

### DIFF
--- a/client_test/container_test.py
+++ b/client_test/container_test.py
@@ -234,14 +234,13 @@ def test_startup_failure(unix_servicer, event_loop):
 
 
 @skip_windows_unix_socket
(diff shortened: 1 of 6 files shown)


erikbern (approved):
Nice

Isn't there also some very similar functionality in `_function_utils.py`? We should try to merge that too

freider:
> Nice
> 
> Isn't there also some very similar functionality in `_function_utils.py`? We should try to merge that too

Yeah there is some overlap with the self-mounting/auto-mounting code, but a lot of that will be removed as part of the automount deprecation. Will see if I can clean it up a bit in a followup

PR #641, opened 2023-07-07
Support `modal shell` with no arguments

ekzhang:
Let me know if this is within scope, but I wanted to suggest that

```bash
modal shell
```

could create a new shell in the base Debian image by default. It seems like a more basic behavior that lets people interactively try out a shell in Modal without first having to create a file just to do so. Defaults matter here I think, and this just working the first time could be nice.

It's a quick utility and an example to show what Modal can do.

```bash
modal shell

modal shell --image python:3.11 --cpu 24 --memory 65536

modal shell --image pytorch/pytorch:2.0.1-cuda11.7-cudnn8-devel --gpu a10g:4
```

Resolves MOD-1426.

### DIFF
--- a/modal/cli/entry_point.py
+++ b/modal/cli/entry_point.py
@@ -89,8 +89,8 @@ entrypoint_cli_typer.add_typer(volume_cli)
 
 entrypoint_cli_typer.command("deploy", help="Deploy a Modal stub as an application.", no_args_is_help=True)(run.deploy)
 entrypoint_cli_typer.command("serve", no_args_is_help=True)(run.serve)
-entrypoint_cli_typer.command("shell", no_args_is_help=True)(run.shell)
 entrypoint_cli_typer.command("setup", help="Bootstrap Modal's configuration.")(setup)
+entrypoint_cli_typer.command("shell")(run.shell)
 
 entrypoint_cli = typer.main.get_command(entrypoint_cli_typer)
 entrypoint_cli.add_command(run.run, name="run")  # type: ignore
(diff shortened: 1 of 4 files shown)


luiscape:
(This reminds me of GCP's [Cloud Shell](https://cloud.google.com/shell) product.)

I like the idea. I wonder if we can provide arguments to make this a really nice experience. For example:

```shell
modal shell --image nvidia/cuda:11.7.0-devel-ubuntu20.04 --gpu a10g
```

What experience did you have in mind?

ekzhang:
> I wonder if we can provide arguments to make this a really nice experience

Yes, exactly! I think that would be really cool, personally, since command-line use is a lot more convenient for experimenting quickly with something. It's also just a nice "wow" introductory demo, especially if they run `htop` and see our machine with over a hundred CPU cores and hundreds of GB of RAM, haha.

ekzhang:
One slight glitch in this is that there's no easy way to specify

```python
    setup_dockerfile_commands=[
        "RUN apt-get update",
        "RUN apt-get install -y python3 python3-pip python-is-python3",
    ]
```

and we need to have Python installed in any Modal image

ekzhang:
Merged it into main, but now it throws this exception:
```
Exception in thread Thread-2:
  File "/usr/local/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/pkg/modal/_pty.py", line 122, in _read
    data_segments = queue.get_many(1024)
AttributeError: 'NoneType' object has no attribute 'get_many'
```

PR #644, opened 2023-07-10
Make all decorator arguments keyword-only

erikbern:
Currently if you forget the parantheses on `local_entrypoint`, you get this error:

```
erikbern@Eriks-MacBook-Air modal % modal run analytics/container_utilization.py
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ You need to specify a Modal function or local entrypoint to run, e.g.                                                                                                                            │
│                                                                                                                                                                                                  │
│ modal run app.py::my_function [...args]                                                                                                                                                          │
│                                                                                                                                                                                                  │
│ Registered functions and local entrypoints on the selected stub are:                                                                                                                             │
│ get_top_functions                                                                                                                                                                                │
│ get_usernames                                                                                                                                                                                    │
│                                                                                                                                                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

This change sort of catches it, although in an awkward way:

`TypeError: _Stub.local_entrypoint() takes 1 positional argument but 2 were given`

I think that's a very marginal improvement. But we probably should never allow positional arguments in the decorator either way because it makes it easier to make these mistakes. Making everything keyword only makes sense as a step in itself, but we should probably improve the error message when you forget the parantheses too

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -1367,6 +1367,7 @@ def _method(
 
 @typechecked
 def _web_endpoint(
+    *,
     method: str = "GET",  # REST method for the created endpoint.
     label: Optional[str] = None,  # Label for created endpoint. Final subdomain will be <workspace>--<label>.modal.run.
     wait_for_response: bool = True,  # Whether requests should wait for and return the function response.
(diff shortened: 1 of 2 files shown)


erikbern:
realizing this makes debugging a bit harder – we can fix it in other ways

PR #645, opened 2023-07-11
Automount refactor and deprecation

freider:
## Changes
* Package mounts have their own entity type in Mount, which basically just delegates entries to either DirEntry or FileEntry, but maintains info on which package is mounted
* Automounting is still enabled, but we diff the explicit mounts a user adds with the automounts and use a `PendingDeprecationWarning` if the user mounts miss any of the packages that would otherwise be automounted
* The mount deduplication code is more reliable now as it considers both the destination and local files added by mounts
* Package mounts (both explicit and implicit) now go into `/root` instead of `/pkg` in the containers. This was mostly done to avoid having the user entrypoint package/module (which goes into root) automatically get an extra implicit mount now that deduplication is doing better diffing
* Bonus: Descriptions for Python package mounts (except the main user entrypoint/stub file) are now status-printed as `PythonPackage:<name of package>` instead of showing the path to the directory/file. Do we want this? Otherwise easily reverted. Example:

```
├── 🔨 Created mount /Users/elias/code/client/modal_test_support/hello.py
├── 🔨 Created mount PythonPackage:client_test
```


erikbern:
> * Package mounts (both explicit and implicit) now go into `/root` instead of `/pkg` in the containers. This was mostly done to avoid having the user entrypoint package/module (which goes into root) automatically get an extra implicit mount now that deduplication is doing better diffing

I guess this could break things in theory if it overwrite something in `/root`? maybe not really an issue in practice

> 
> * Bonus: Descriptions for Python package mounts (except the main user entrypoint/stub file) are now status-printed as `PythonPackage:<name of package>` instead of showing the path to the directory/file. Do we want this? Otherwise easily reverted. Example:

I like it!

modal/_resolver.py
```diff
@@ -77,7 +77,11 @@ async def preload(self, obj, existing_object_id: Optional[str] = None):
             return await obj._preload(self, existing_object_id)
 
     async def load(self, obj, existing_object_id: Optional[str] = None):
-        cached_future = self._local_uuid_to_future.get(obj.local_uuid)
+        try:
+            local_uuid = obj.local_uuid
+        except AttributeError:
+            raise AttributeError(f"obj is {obj!r}, which is not a Modal object")
```
erikbern, line 83: minor preference for checking the type of `obj` explicitly

erikbern (approved):
Have you run integration tests with this? Might be good for extra safety.

Otherwise, 👍 !

freider:
After some experimentation with multi-file apps etc. my general feeling is that removing all kinds of auto-mounting would remove too much of the "magic" of the Modal client, so I'll try to rework this before merging into something that still auto-mounts essentials by default but gives the user the knobs needed to easily disable/tweak mounting when that's needed.

Most of the code in this PR is still useful, since it still makes handling of Mounts in general better

freider:
Superseded by #1230

PR #647, opened 2023-07-11
Remove usage of the deprecated web_url* fields

ekzhang:
Main change here is that `FunctionCreateResponse.web_url` and `FunctionCreateResponse.web_url_info` have been deprecated for a while, but I noticed we were actually still using them, so I'm removing their usage and moving up the deprecation comment to version 0.51.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -213,30 +213,30 @@ class _Client:
             client_type = api_pb2.CLIENT_TYPE_CLIENT
             credentials = None
 
-        if cls._client_from_env:
-            return cls._client_from_env
-
         if cls._client_from_env_lock is None:
             cls._client_from_env_lock = asyncio.Lock()
 
         async with cls._client_from_env_lock:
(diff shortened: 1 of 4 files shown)


modal/client.py
```diff
@@ -213,30 +213,30 @@ async def from_env(cls, _override_config=None) -> "_Client":
...
-                            "Token missing. Could not authenticate client. "
-                            "If you have token credentials, see modal.com/docs/reference/modal.config for setup help. "
-                            "If you are a new user, register an account at modal.com, then run `modal token new`."
-                        )
-                        raise AuthError(creds_missing_msg)
-                    else:
-                        raise
-                cls._client_from_env = client
-                return client
+            client = _Client(server_url, client_type, credentials)
```
erikbern, line 223: Is this an unrelated change?

ekzhang: Yeah it's unrelated, was moving things around while testing perf

erikbern (approved):

PR #648, opened 2023-07-11
Enable Ruff's isort rules

ekzhang:
When we removed isort and switched to Ruff, we didn't actually enable Ruff's isort rules, so our imports weren't being sorted.


ekzhang:
@prbot approve

erikbern:
👍

PR #654, opened 2023-07-18
Sandbox tasks

aksh-at:
Proto changes in https://github.com/modal-labs/modal-client/pull/666.

See outline in https://github.com/modal-labs/modal/pull/6869


modal/app.py
```diff
@@ -240,6 +245,26 @@ async def deploy(self, name: str, namespace, object_entity: str) -> str:
         deploy_response = await retry_transient_errors(self._client.stub.AppDeploy, deploy_req)
         return deploy_response.url
 
+    async def spawn_container(
+        self, program: str, *args: str, shell: bool = False, image: Optional["_Image"] = None, mounts=[]
```
erikbern, line 249: minor preference for expecting an array of strings here rather than using `*args` here. makes it more consistent with `subproces`: https://docs.python.org/3/library/subprocess.html

erikbern:
btw why is it a method on the app class (as opposed to its own global function) – how are containers connected to apps?

modal/app.py
aksh-at: I was using the [asyncio.subprocess API](https://docs.python.org/3/library/asyncio-subprocess.html#creating-subprocesses) as reference. Surprising that the two are inconsistent!

aksh-at:
> btw why is it a method on the app class (as opposed to its own global function)

It's because we need to dynamically instantiate objects ([images](https://github.com/modal-labs/modal-client/pull/654/files#diff-c0afbd862fcdf7415870a554cd8542f9d411d5cd6df011d61168000987df4becR256), mounts, and then the container). Doing that requires the app ID, client, etc. So it felt clean to have it be a method on the app to have access to those properties — usage would be something like `stub.app.spawn_container()`. 

Not sure if there's a better way to get the app ID where the user doesn't have to pass it in. `Client` exists as a singleton, but app does not.

modal/app.py
erikbern: ok yeah maybe `*args` is more common in recent code. alright seems fine then!

modal/app.py
```diff
@@ -240,6 +246,28 @@ async def deploy(self, name: str, namespace, object_entity: str) -> str:
         deploy_response = await retry_transient_errors(self._client.stub.AppDeploy, deploy_req)
         return deploy_response.url
 
+    async def spawn_sandbox(
```
erikbern, line 249: not super important since i'm refactoring this code anyway, but i think it's more consistent if you had a `Sandbox` class here which is a provider class, similar to all other objects. and then `spawn_sandbox` would mostly just call a static constructor factory on `Sandbox`. but we can fix that later once we've merged providers and handles

modal/sandbox.py
```diff
@@ -0,0 +1,94 @@
...
+
+from modal_proto import api_pb2
+from modal_utils.async_utils import synchronize_api
+from modal_utils.grpc_utils import RETRYABLE_GRPC_STATUS_CODES, retry_transient_errors, unary_stream
+
+from .client import _Client
+from .object import _Handle
+
+
+class _LogsReader:
```
erikbern, line 13: don't we already have very similar code for this elsewhere?

modal/app.py
aksh-at: It's probably because `subprocess` has the same command for both `exec` and `shell` forms, whereas they're different in `asyncio.subprocess`. Probably makes sense for us to do that too—something like `spawn_sandbox` and `spawn_sandbox_shell` maybe?

erikbern:
would be great to add at least one basic test for this, given that the client code might go through a few big refactorings soon

modal/sandbox.py
aksh-at: yeah mentioned in a comment a few lines below. We can combine later.

aksh-at: there's a bunch of small differences that make it not worth the effort of doing now

aksh-at:
> i realize this is prototype code to some extent, but it would be great to add at least one basic test for this, given that the client code might go through a few big refactorings soon

Ended up adding a mock servicer that uses `subprocess` directly, will add some more tests using this tomorrow.

client_test/conftest.py
```diff
@@ -574,6 +576,36 @@ async def QueueGet(self, stream):
         await stream.recv_message()
         await stream.send_message(api_pb2.QueueGetResponse(values=[self.queue.pop(0)]))
 
+    ### Sandbox
+
+    async def SandboxCreate(self, stream):
```
erikbern, line 581: Might be useful to use the new `servicer.intercept()` helper here – that way you can put the fixture logic in the test itself. See an example: https://github.com/modal-labs/modal-client/blob/main/client_test/cli_test.py#L378 (haven't used it myself yet, @freider built it)

aksh-at:
@prbot approve

PR #657, opened 2023-07-18
make 'modal config set-environment' visible

thundergolfer:
### Describe your changes

I think this may have been intended to be only hidden initially? 


freider:
Ah yeah I made all of the environment cli stuff hidden first - must have missed the revert on this one when I removed the flag on the others. Thanks for noticing and fixing!

PR #660, opened 2023-07-19
Precreate handles on providers

erikbern:
This is part of my work to merge providers and handles.

Step 1 is fairly silly which is that each provider just creates a handle in advance and returns it during load, hydrating the handle rather than constructing a new object on the fly.


erikbern:
@prbot approve

PR #662, opened 2023-07-20
Function and FunctionHandles 1:1

erikbern:
This mostly ties the Function and FunctionHandles 1:1 with the exception of the container app which I'll try to fix later but I don't think is consequential. Left a couple of TODOs that might be worth addressing in subsequent PRs

This code is a thicket... I think this change cleans things up a bit, but it needs a bit more combing.


aksh-at (approved):

erikbern:
Thanks. Next step is to move the handle to sit on the base `Provider` class

PR #663, opened 2023-07-20
remove _track_function_invocation

erikbern:
Afaict this is never used by the client library. It does get in the way of a later refactoring I want to make (removing the stub from the FunctionHandle class).


erikbern:
nvm i do see where this is used now. Let me rewrite it instead so it doesn't use the stub, should be doable

PR #664, opened 2023-07-20
[MOD-1454] Initial impl of 'modal vol ls'

thundergolfer:
Basically follows the structure of the `modal nfs ls` command. This works, but the RPC implementation wants a bit of optimization and edge-case tidying.


thundergolfer:
@prbot approve

PR #668, opened 2023-07-21
Unify all provider constructors

erikbern:
One step towards merging handles and providers is we want to construct "unhydrated" handle objects early on – at the point of the provider constructor. I already made this change for a lot of subclasses in #660 but not for some special constructors on the base class. This finishes the work.

The unhydrated handle objects are just scoped inside constructors for now – but the next PR will make it a core property of the base provider class, which will let us unify even more constructor logic. (The step after that is to move pretty much all handle logic to the providers instead, which will make the handles pointless)

A lot of the existing constructor code is quite messy and unfortunately this change makes it slightly _more_ messy in the short term, which is regrettable. But I think there's some short-term pain for long-term gain.

This PR also renames some internal constructors to have more consistent names. So for instance `_from_id` is now called `_new_hydrated` since it basically combines `_new` and `_hydrate`. Generally trying to use the verb "hydrate" throughout the code base to make it clear.

### DIFF
--- a/client_test/function_test.py
+++ b/client_test/function_test.py
@@ -457,8 +457,8 @@ def test_closure_valued_serialized_function(client, servicer):
     assert functions["ret_bar"]() == "bar"
 
 
-def test_from_id_internal(client, servicer):
-    obj = FunctionCall._from_id("fc-123", client, None)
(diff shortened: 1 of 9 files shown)


erikbern:
@prbot approve

PR #669, opened 2023-07-22
Construct a handle on the base provider constructor

erikbern:
This makes the `Provider` base class constructor an unhydrated `Handle` object as a part of it, which means every `Provider` always carries a handle (hydrated or not).

This ties the two classes even tighter together. Next step is to move a lot of logic from the handles to the providers, turning handles into a useless object eventually (that can be removed).


erikbern:
@prbot approve

PR #672, opened 2023-07-24
modal shell by not creating functions dynamically

erikbern:
This changes how `modal shell` works. Instead of creating a function dynamically from `exec_cmd`, keep the original function definition, but set a flag on it that tells the container entrypoint to run `exec_cmd` inside the container instead.

Note: broke out a minor part of this change into #674 to be deployed first – should make it easier to test this change against prod.


erikbern:
FYI ran all the integration tests and they pass (although we don't have any integration tests for shell, so maybe it's not that much of a guarantee)

erikbern:
@prbot approve

PR #675, opened 2023-07-24
Replicate handle methods on the provider

erikbern:
This lets you call "live" handle methods on the `Provider` class instead, which basically works by routing every method call to the underlying handle.

Only doing it for 2/5 handles in this PR:

* [X] `Queue`
* [X] `Dict`
* [ ] `Volume` (`commit`, `reload`, ...)
* [ ] `NetworkFileSystem` (`write_file`, `read_file`, ...)
* [ ] `Function` (all the function calling)

Replicating all methods is a fairly dumb workaround for now. But the idea is that once all methods are wrapped, we can return a `Provider` class instead of a `Handle` class in a bunch of places without breaking user code, thanks to the magic of duck typing. Once we return providers everywhere, we can move the biz logic from the handle classes the provider classes and get rid of the handle classes 🎉

### DIFF
--- a/client_test/conftest.py
+++ b/client_test/conftest.py
@@ -575,7 +575,11 @@ class MockClientServicer(api_grpc.ModalClientBase):
 
     async def QueueGet(self, stream):
         await stream.recv_message()
-        await stream.send_message(api_pb2.QueueGetResponse(values=[self.queue.pop(0)]))
+        if len(self.queue) > 0:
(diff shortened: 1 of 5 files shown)


erikbern:
@prbot approve

erikbern:
Will add the remaining methods in subsequent PRs and then start to remove handles

PR #678, opened 2023-07-25
modal shell: hacky workaround for old workers and newest client

erikbern:
It does look like #672 breaks with old workers (who don't know the `pty_type` field) and the newest client.

This is an incredibly dumb workaround that we might need for a few days.


erikbern:
won't hotfix this rn because it's kind of a edge case anyway, just in case it breaks something

i'll merge this first thing tomorrow morning

PR #679, opened 2023-07-25
Add max_concurrent_inputs to Function proto

gongy:


erikbern:
Minor thing but I find this name a bit suboptimal because
1. Setting a max to me generally means you're constraining something, whereas here we're enabling something
2. max per what?

something like `container_input_parallelism` would make more sense to me (although quite long)

PR #680, opened 2023-07-25
Catch issues fixed in #677

erikbern:
Adding unit tests that would have caught the issue in #677 

These unit tests have a lot of assertions besides the actual issue (which was that an exception was being thrown)


aksh-at (commented):
Nice!

gimaker (commented):
Post-merge reviewed and approved :heavy_check_mark:

PR #683, opened 2023-07-25
Update wsgi_app and asgi_app docs

thundergolfer:
The docs for `wsgi_app` in particular are a bit bare and annoying: 

<kdb>
image
</kbd>

You're told to click through to `asgi_app`, and then on that page to understand how to use the wsgi decorator you need to click _again_ to see the web endpoints guide.


ekzhang (approved):
Nice docs!

PR #687, opened 2023-07-27
Client supports concurrent inputs

gongy:
`function_io_manager` should be thread-safe as it executes on the `synchronicity` thread.

This PR contains:
- Unit tests in container_test.py
- An `asyncio.Semaphore` prevents the `n+1`th input from being fetched before in-progress ones are completed
- Spawns calls to sync or async functions in a separate task or thread if `allow_concurrent_inputs` is set

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -112,6 +112,7 @@ class _FunctionIOManager:
         self.total_user_time: float = 0
         self.current_input_id: Optional[str] = None
         self.current_input_started_at: Optional[float] = None
+        self._semaphore: Optional[asyncio.Semaphore] = None
         self._client = synchronizer._translate_in(self.client)  # make it a _Client object
         self._stub_name = self.function_def.stub_name
         self._container_app = None
@@ -189,10 +190,10 @@ class _FunctionIOManager:
         while not eof_received:
             request.average_call_time = self.get_average_call_time()
             request.max_values = self.get_max_inputs_to_fetch()  # Deprecated; remove.
-            request.input_concurrency = self.semaphore.maxsize
+            request.input_concurrency = self.input_concurrency
 
             # If number of active inputs is at max queue size, this will block.
-            await self._acquire_semaphore()
+            await self._semaphore.acquire()
             with trace("get_inputs"):
                 set_span_tag("iteration", str(iteration))  # force this to be a tag string
                 iteration += 1
@@ -228,7 +229,7 @@ class _FunctionIOManager:
                         break
 
             if not yielded:
-                await self._release_semaphore()
+                await self._semaphore.release()
 
     async def _send_outputs(self):
         """Background task that tries to drain output queue until it's empty,
@@ -246,18 +247,17 @@ class _FunctionIOManager:
             # TODO(erikbern): we'll get a RESOURCE_EXCHAUSTED if the buffer is full server-side.
             # It's possible we want to retry "harder" for this particular error.
 
-    async def _acquire_semaphore(self):
-        await self.semaphore.put(None)
-
-    async def _release_semaphore(self):
-        await self.semaphore.get()
-        self.semaphore.task_done()
 
     async def run_inputs_outputs(self, input_concurrency: int = 1):
         # This also makes sure to terminate the outputs
         self.output_queue: asyncio.Queue = asyncio.Queue()
+
         # Ensure we do not fetch new inputs when container is too busy
-        self.semaphore: asyncio.Queue[None] = asyncio.Queue(maxsize=input_concurrency)
+        # Before trying to fetch an input, acquire the semaphore.
+        # If no input is fetched, release the semaphore.
+        # When the output for the fetched input is enqueued, release the semaphore.
+        self.input_concurrency = input_concurrency
+        self._semaphore = asyncio.Semaphore(input_concurrency)
 
         async with TaskContext(grace=10) as tc:
             tc.create_task(self._send_outputs())
@@ -270,8 +270,9 @@ class _FunctionIOManager:
                     _set_current_input_id(None)
                     self.current_input_id, self.current_input_started_at = (None, None)
             finally:
-                # wait for all active inputs to place their outputs into the queue
-                await self.semaphore.join()
+                # collect all active input slots, meaning all outputs of outstanding inputs are enqueued
+                for _ in range(input_concurrency):
+                    await self._semaphore.acquire()
                 # send the eof to _send_outputs loop
                 await self.output_queue.put(None)
 
@@ -379,7 +380,7 @@ class _FunctionIOManager:
     async def complete_call(self, started_at):
         self.total_user_time += time.time() - started_at
         self.calls_completed += 1
-        await self._release_semaphore()
+        await self._semaphore.release()
 
     async def enqueue_output(self, input_id, started_at: float, output_index: int, data):
         await self._enqueue_output(
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -462,14 +462,9 @@ def call_function_sync(
                     function_io_manager.enqueue_output(input_id, started_at, output_index.value, res)
 
         if input_concurrency > 1:
-            executor = concurrent.futures.ThreadPoolExecutor()
-            futures = []
-
-            for input_id, args, kwargs in function_io_manager.run_inputs_outputs(input_concurrency):
-                futures.append(executor.submit(run_inputs, input_id, args, kwargs))
-
-            # This should complete instantly
-            concurrent.futures.wait(futures)
+            with concurrent.futures.ThreadPoolExecutor() as executor:
+                for input_id, args, kwargs in function_io_manager.run_inputs_outputs(input_concurrency):
+                    executor.submit(run_inputs, input_id, args, kwargs)
         else:
             for input_id, args, kwargs in function_io_manager.run_inputs_outputs(input_concurrency):
                 run_inputs(input_id, args, kwargs)
@@ -525,12 +520,9 @@ async def call_function_async(
                     await function_io_manager.enqueue_output.aio(input_id, started_at, output_index.value, value)
 
         if input_concurrency > 1:
-            tasks = []
-            async for input_id, args, kwargs in function_io_manager.run_inputs_outputs.aio(input_concurrency):
-                tasks.append(asyncio.create_task(run_input(input_id, args, kwargs)))
-
-            # This should complete instantly
-            await asyncio.gather(*tasks)
+            async with TaskContext() as execution_context:
+                async for input_id, args, kwargs in function_io_manager.run_inputs_outputs.aio(input_concurrency):
+                    execution_context.create_task(run_input(input_id, args, kwargs))
         else:
             async for input_id, args, kwargs in function_io_manager.run_inputs_outputs.aio(input_concurrency):
                 await run_input(input_id, args, kwargs)
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -113,6 +113,7 @@ class _FunctionIOManager:
         self.total_user_time: float = 0
         self.current_input_id: Optional[str] = None
         self.current_input_started_at: Optional[float] = None
+        self._input_concurrency: Optional[int] = None
         self._semaphore: Optional[asyncio.Semaphore] = None
         self._client = synchronizer._translate_in(self.client)  # make it a _Client object
         self._stub_name = self.function_def.stub_name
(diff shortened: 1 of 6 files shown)


erikbern:
what's the semaphore used for? i wonder if it's easier to just spawn n threads (if sync) or n asyncio tasks (if async)

gongy:
I could be wrong here but it feels cleaner to keep the fetch requests one-at-a-time.

gongy:
@prbot approve

modal/_container_entrypoint.py
```diff
@@ -236,9 +240,18 @@ async def _send_outputs(self):
             # TODO(erikbern): we'll get a RESOURCE_EXCHAUSTED if the buffer is full server-side.
             # It's possible we want to retry "harder" for this particular error.
 
-    async def run_inputs_outputs(self):
+    async def _acquire_semaphore(self):
```
erikbern, line 243: can you put this higher up? since there are currently callers above it

modal/_container_entrypoint.py
```diff
@@ -236,9 +240,18 @@ async def _send_outputs(self):
...
+
+    async def _release_semaphore(self):
+        await self.semaphore.get()
+        self.semaphore.task_done()
+
+    async def run_inputs_outputs(self, input_concurrency: int = 1):
         # This also makes sure to terminate the outputs
         self.output_queue: asyncio.Queue = asyncio.Queue()
+        # Ensure we do not fetch new inputs when container is too busy
+        self.semaphore: asyncio.Queue[None] = asyncio.Queue(maxsize=input_concurrency)
```
erikbern, line 254: also this should be even higher up since it's the place where it's being defined

erikbern: actually maybe `self.semaphore` should be initialized in the constructor?

modal/_container_entrypoint.py
```diff
@@ -176,49 +176,53 @@ def get_max_inputs_to_fetch(self):
...
         request = api_pb2.FunctionGetInputsRequest(function_id=self.function_id)
         eof_received = False
         iteration = 0
         while not eof_received:
             request.average_call_time = self.get_average_call_time()
             request.max_values = self.get_max_inputs_to_fetch()  # Deprecated; remove.
+            request.input_concurrency = self.semaphore.maxsize
 
+            # If number of active inputs is at max queue size, this will block.
+            await self._acquire_semaphore()
```
erikbern, line 189: let's use a context manager for this – the documentation actually even mentions that it's the preferred way: https://docs.python.org/3/library/asyncio-sync.html#semaphore

modal/_container_entrypoint.py
```diff
@@ -176,49 +176,53 @@ def get_max_inputs_to_fetch(self):
...
+                        eof_received = True
+                        break
+
+                    # If we got a pointer to a blob, download it from S3.
+                    if item.input.WhichOneof("args_oneof") == "args_blob_id":
+                        input_pb = await self.populate_input_blobs(item.input)
+                    else:
+                        input_pb = item.input
+
+                    # If yielded, allow semaphore to be released via enqueue_outputs
```
erikbern, line 216: why have two different paths to release the semaphore? seems complex vs having a single place that acquires and releases it (ideally using a context manager)

modal/_container_entrypoint.py
erikbern: wait why is self.semaphore a queue? I just assumed it was an `asyncio.Queue` object based on the name... let's rename the variable or use an actual semaphore

modal/stub.py
```diff
@@ -422,6 +422,7 @@ def function(
         proxy: Optional[_Proxy] = None,  # Reference to a Modal Proxy to use in front of this function.
         retries: Optional[Union[int, Retries]] = None,  # Number of times to retry each input in case of failure.
         concurrency_limit: Optional[int] = None,  # Limit for max concurrent containers running the function.
+        allow_concurrent_inputs: Optional[int] = None,  # Allow container to handle multiple inputs simultaneously.
```
erikbern, line 425: this description makes it sound like it's a boolean – prob worth pointing out what the number describes

modal/_container_entrypoint.py
gongy: ok i'll rename the variable. 

when we receive EOF, we need to wait for all the outputs to get enqueued. we use `await semaphore.join()` to know when that happens.

modal/_container_entrypoint.py
gongy: It's not possible to use a context manager without a huge refactor because the acquire and release happen in different functions.

erikbern: does it have to? it seems a bit risky to me that acquire and release are not 1:1

modal/_container_entrypoint.py
erikbern: this makes it feel like we should use a thread pool here?

modal/_container_entrypoint.py
gongy: The path separate occurs because we need to yield the input - and the input is run either in a separate task or thread. Then, when the input finishes, we enqueue the output - and that happens from that separate task (async case).

(same reason as why we can't use a context manager)

modal/_container_entrypoint.py
```diff
@@ -444,27 +476,34 @@ async def call_function_async(
...
                         raise InvalidError(
                             f"Async (non-generator) function returned value of type {type(res)}"
                             " You might need to use @stub.function(..., is_generator=True)."
                         )
                     value = await res
-                    await function_io_manager.enqueue_output.aio(input_id, output_index.value, value)
+                    await function_io_manager.enqueue_output.aio(input_id, started_at, output_index.value, value)
+
+        async for input_id, args, kwargs in function_io_manager.run_inputs_outputs.aio(input_concurrency):
+            asyncio.create_task(run_input(input_id, args, kwargs))
```
erikbern, line 506: this seems dangerous – should definitely use some sort of context manager to make sure it gets cleaned up and errors are handled properly

modal/_container_entrypoint.py
erikbern: why do you have to acquire and release the semaphore in those places? isn't it fine to just do it surrounding calling the user function? i.e. `res = fun(*args, **kwargs)`

modal/_container_entrypoint.py
gongy: Will do

modal/_container_entrypoint.py
gongy: The challenge is that the semaphore is acquired prior to the fetch, which happens in the generator. This will be included in the refactor proposal.

modal/_container_entrypoint.py
```diff
@@ -503,8 +503,11 @@ async def run_input(input_id, args, kwargs):
                     value = await res
                     await function_io_manager.enqueue_output.aio(input_id, started_at, output_index.value, value)
 
+        tasks = []
         async for input_id, args, kwargs in function_io_manager.run_inputs_outputs.aio(input_concurrency):
-            asyncio.create_task(run_input(input_id, args, kwargs))
+            tasks.append(asyncio.create_task(run_input(input_id, args, kwargs)))
+
+        await asyncio.gather(*tasks)
```
gongy, line 510: Is this a reasonable way to clean up the stray tasks?

We would expect this `await` to complete exactly instantly. An alternative would be asserting `task.done() for task in tasks`.

aksh-at: Yeah, asserting seems better. Also we should still use a `TaskContext` here, since it has a bunch of logic to clean up tasks when e.g. an exception happens.

modal/_container_entrypoint.py
gongy: I refactored to use a semaphore. The "wait for everything to be done" can be accomplished by collecting all the semaphore locks.

`for _ in range(n): await self._semaphore.acquire()`

gongy:
Ready for second review. Main changes are:
- Added unit tests in `container_test.py`
- Replaced `asyncio.Queue` with `asyncio.Semaphore`
- Collect completed tasks or threads at end of container life
- `self._semaphore` and `self._input_concurrency` are defined on the class but populated in `run_inputs_outputs`

The unit tests:
- Apply to both `sync` and `async` contexts
- Run 12 inputs that sleep for 1 second with a parallelism of 6
- Check each input takes 1 second
- Check first 6 inputs start together and the last 6 inputs start together 1 second later

modal/_container_entrypoint.py
gongy: Yeah you're right. That makes a lot of sense.

Makes me wonder if we should hold off on the synchronous implementation until we make an equivalent `ThreadContext`?

modal/_container_entrypoint.py
```diff
@@ -444,27 +487,39 @@ async def call_function_async(
...
                             " You might need to use @stub.function(..., is_generator=True)."
                         )
                     value = await res
-                    await function_io_manager.enqueue_output.aio(input_id, output_index.value, value)
+                    await function_io_manager.enqueue_output.aio(input_id, started_at, output_index.value, value)
+
+        if input_concurrency > 1:
+            async with TaskContext() as execution_context:
+                async for input_id, args, kwargs in function_io_manager.run_inputs_outputs.aio(input_concurrency):
+                    execution_context.create_task(run_input(input_id, args, kwargs))
```
erikbern, line 520: actually why do you need to create tasks here? can't you just run `asyncio.gather` on a bunch of coroutines?

modal/_container_entrypoint.py
```diff
@@ -444,27 +487,39 @@ async def call_function_async(
...
                         raise InvalidError(
                             f"Async (non-generator) function returned value of type {type(res)}"
                             " You might need to use @stub.function(..., is_generator=True)."
                         )
                     value = await res
-                    await function_io_manager.enqueue_output.aio(input_id, output_index.value, value)
+                    await function_io_manager.enqueue_output.aio(input_id, started_at, output_index.value, value)
+
+        if input_concurrency > 1:
+            async with TaskContext() as execution_context:
```
erikbern, line 517: actually i don't think we should use tasks here. `TaskContext` actually has an O(N^2) complexity while terminating iirc. but also asyncio tasks have a bunch of overhead and are annoying to debug. i think you probably want to create more like a pool and then join the coroutines in the end maybe? there's nothing built into asyncio unfortunately

modal/_container_entrypoint.py
```diff
@@ -587,10 +646,14 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
             imp_fun.fun = run_in_pty(imp_fun.fun, input_stream_blocking, pty_info)
 
         if not imp_fun.is_async:
-            call_function_sync(function_io_manager, imp_fun.obj, imp_fun.fun, imp_fun.is_generator)
+            call_function_sync(
```
erikbern, line 649: maybe worth changing the signature of this function to take an `ImportedFunction`

erikbern (commented):
looks a lot nicer – thanks for taking into account the feedback so far! left a couple of smaller comments

I'm still a bit nervous about the acquisition/release lifecycle – feels like there's a lot of things that can throw exceptions in weird ways that doesn't cause the semaphore to be released again. so my biggest outstanding concern is this process doesn't seem foolproof. but maybe it works well in practice, at least for the time being – sounds like we can shift this logic into the worker soon anyway

modal/_container_entrypoint.py
erikbern: this one might be good? https://github.com/gistart/asyncio-pool

otoh for your use case you can also just implement a small utility function and it's probably like 10 lines.

gongy: Happy to create a small utility class that is a subset of `TaskContext` - but maybe one of these points changes the picture?

- I think `TaskContext.wait` is `O(N^2)` but we don't use that here. Instead, `.stop()` is called on exit with no grace, meaning any in-progress tasks just get cancelled.

- Also, the overhead of coroutines should scale with the value of `allow_concurrent_inputs` set by the user. It would be potentially confusing if the user sets some `n` but only ever sees `<n` executing at one time.

erikbern: aren't you kicking off one task for every input?

erikbern: i think asyncio.Task has some minor overhead over just a straight up coroutine, so if you can use the latter only, that might be preferable

gongy: Do you think we can group this into the medium-sized refactor, which would involve:

- Refactor `run_inputs_outputs` to no longer be a generator, but instead expose `fetch_input` on `FunctionIOManager` API. `fetch_input` can return an EOF. `fetch_input` holds a lock in `FunctionIOManager` to prevent concurrent requests going out to the worker
- Sync/Async call-loops can now involve continuously spawning `run_inputs` into an asyncio pool, or `ThreadPoolExecutor` to manage the number of concurrent `run_inputs` calls
- `run inputs` involves: acquiring semaphore, calling `fetch_input`, running the user function and finally `enqueue_output`

erikbern: Yeah something like that makes sense!

PR #690, opened 2023-07-28
Image.workdir() cmd

aksh-at:


modal/image.py
```diff
@@ -1218,6 +1218,23 @@ def env(self, vars: Dict[str, str]) -> "_Image":
...
+        ```python
+        image = (
+            modal.Image.debian_slim()
+                .run_commands("git clone https://xyz app")
+                .workdir("/app")
+                .run_commands("yarn install")
+        )
+        ```
+        """
+        return self.extend(dockerfile_commands=["FROM base"] + [f"WORKDIR {shlex.quote(path)}"])
```
erikbern, line 1236: will this change the workdir for subsequent steps as well?

aksh-at: Yes, see https://github.com/modal-labs/modal/pull/6976

PR #691, opened 2023-07-28
Show logs from dynamically created images

aksh-at:
Simple fix for now, for log output from "dynamically created" images not showing up in the terminal. Medium term we might want to either tag the logs themselves with this value, or filter out specific image IDs that we know are being built statically?

---

Resolves MOD-1483

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -129,7 +129,7 @@ class OutputManager:
     _line_buffers: dict[int, LineBufferedOutput]
     _status_spinner: Spinner
     _app_page_url: Optional[str]
-    is_app_running: bool
+    _show_image_logs: bool
 
     def __init__(self, stdout: io.TextIOWrapper, show_progress: bool, status_spinner_text: str = "Running app..."):
         self.stdout = stdout or sys.stdout
@@ -145,7 +145,7 @@ class OutputManager:
         self._line_buffers = {}
         self._status_spinner = step_progress(status_spinner_text)
         self._app_page_url = None
-        self.is_app_running = False
+        self._show_image_logs = False
 
     def print_if_visible(self, renderable) -> None:
         if self._visible_progress:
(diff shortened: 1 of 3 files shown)


modal/_output.py
```diff
@@ -387,9 +392,13 @@ async def _get_logs():
                 logger.debug("App logs are done")
                 last_log_batch_entry_id = None
                 break
-            elif log_batch.image_id:
-                # Ignore image logs - these still exist for old clients
-                # New clients get them through ImageJoinStreaming (see image.py)
+            elif log_batch.image_id and not output_mgr.is_app_running:
```
erikbern, line 395: is this thing even working? i thought we don't write image logs to the app anymore

aksh-at: yeah we do. Right now, that's how the web interface shows logs for image builds for the app. So there's a difference in how the web UI fetches logs vs how the CLI does.

modal/_output.py
```diff
@@ -129,6 +129,7 @@ class OutputManager:
     _line_buffers: dict[int, LineBufferedOutput]
     _status_spinner: Spinner
     _app_page_url: Optional[str]
+    is_app_running: bool
```
erikbern, line 132: i'm not a big fan of this since
1. it couples the output mgr's behavior to the app – why does the output mgr need to know about the app's state? it would make more sense if it's just a flag `show_image_logs` or similar
2. mutating state, but i can't think of any way to avoid it so 🤷

modal/_output.py
erikbern: ok interesting, didn't know that

modal/_output.py
aksh-at: We could do a flag but then we'd have to launch the logs loop twice (once while creating the objects, and the other while running). I believe we used to do something like that?

erikbern: can we just rename it to `_show_image_logs` or similar? otherwise LGTM!!

erikbern (approved):
👍 except would prefer https://github.com/modal-labs/modal-client/pull/691#discussion_r1277761699

modal/_output.py
aksh-at: done

PR #692, opened 2023-07-30
Allow skipping modal client setup in registry images

aksh-at:
Adds a flag, `setup_modal_client` to Image static methods that load from a registry. This is so that images that don't need the Modal client (for `Sandbox` tasks) don't need Python installed.

```python
modal.Image.from_dockerhub("node:slim", setup_modal_client=False)
```

Honestly don't love this, open to ideas for a better name. Some other options:

```python
modal.Image.from_dockerhub("node:slim", for_sandbox=True)
```

```python
modal.Image.from_dockerhub("node:slim", install_modal_client_reqs=False)
```

```python
modal.Image.from_dockerhub("node:slim", no_python=True)
```

```python
modal.Image.from_dockerhub("node:slim", skip_modal_client=True)
```

### DIFF
--- a/client_test/image_test.py
+++ b/client_test/image_test.py
@@ -253,6 +253,22 @@ def test_dockerhub_install(servicer, client):
 
         assert any("FROM gisops/valhalla:latest" in cmd for cmd in layers[0].dockerfile_commands)
         assert any("RUN apt-get update" in cmd for cmd in layers[0].dockerfile_commands)
+        assert any("modal_requirements.txt" in cmd for cmd in layers[0].dockerfile_commands)
+
+
+def test_dockerhub_no_client_install(servicer, client):
+    stub = Stub(
+        image=Image.from_dockerhub(
+            "gisops/valhalla:latest", setup_dockerfile_commands=["RUN apt-get update"], setup_modal_client=False
+        )
+    )
+
+    with stub.run(client=client) as running_app:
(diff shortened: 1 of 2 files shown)


erikbern:
I wonder if it would make more sense to flip it around, so there's no modal package installed by default, and you have to run `.install_modal()` or something like that? Would need a deprecation path though.

ekzhang:
I thought this would be great for #641 but unfortunately we would still need Python in the image regardless, for the shell client code to run :(

aksh-at:
Ended up renaming to

```python
modal.Image.from_dockerhub("node:slim", no_python=True)
```

Don't love it, but it feels better than the others.

erikbern:
> We could, but it'll be pretty onerous to have to do that for every image for regular functions (not sandboxes). Seems like we can save our users quite a bit of time / boilerplate by having it be default.

I think boilerplate is sometimes fine if it's more conceptually clear what happens. But not a strong preference in this case.

Regarding the flag name, `no_python` feels a bit of a misnomer, because this is not about installing Python right? It's about installing Modal dependencies

ekzhang:
> Regarding the flag name, `no_python` feels a bit of a misnomer, because this is not about installing Python right? It's about installing Modal dependencies

Agree with this, I think `add_modal=False` would be better

mwaskom:
Closing as stale, we can revisit!

PR #685, opened 2023-07-25
Add handle proxies for Volume and NFS

erikbern:
Continuation of #675 

* [X] `Queue`
* [X] `Dict`
* [X] `Volume` (`commit`, `reload`, ...)
* [X] `NetworkFileSystem` (`write_file`, `read_file`, ...)
* [ ] `Function` (all the function calling)

~TODO: add some unit tests~ – edit: done

PR #694, opened 2023-07-31
expose all public FunctionHandle methods on Function

erikbern:
Continues work in #675 and #685:

* [X] `Queue`
* [X] `Dict`
* [X] `Volume` (`commit`, `reload`, ...)
* [X] `NetworkFileSystem` (`write_file`, `read_file`, ...)
* [X] `Function` (all the function calling)

At this point every handle method is now exposed on the provider classes too. This means that after this PR is merged, we can swap out handle classes for providers, so that the handle classes are internal only. Off the top of my head this is roughly:

* Any function created in global scope
* All objects created on the app object locally
* All objects created on the app object in the container

Once that's done, we should move all docs and implementation from the handle to the provider classes too

### DIFF
--- a/client_test/function_test.py
+++ b/client_test/function_test.py
@@ -37,6 +37,12 @@ def test_run_function(client, servicer):
         assert foo.call(2, 4) == 20
         assert len(servicer.cleared_function_calls) == 1
 
+        # Make sure we can also call the Function object
+        fun = stub.foo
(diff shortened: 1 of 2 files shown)


erikbern:
@prbot approve

aksh-at:
Nice!

PR #706, opened 2023-08-01
Fix minor bugs in dict.py

ekzhang:
- Reference to `SharedVolume`
- Broken `__contains__` implementation

related but does not resolve [this ticket](https://linear.app/modal-labs/issue/MOD-1492/add-better-reference-docstrings-for-dicts-and-queues)


erikbern (commented):
sorry my bad for introducing this!!

PR #708, opened 2023-08-01
Replace handle metaclass with `__init_subclass__`, add the same thing on the provider

erikbern:
Banging my head against the wall trying to merge providers and handles because things work 95% in a PR I have, but some of the remaining things are very deep subtle things that need a lot of refactoring. This is prep PR for some things I need to change for serialization to work properly. But we would have needed this eventually anyway.

update: ended up having to remove the `Generic[H]` stuff for `Provider`, but that seems fine since we're removing the `Handle` classes soon anyway! We can leave with slightly less type safety for a week or two

update 2: ended up removing the metaclass stuff in favor of `__init_subclass__`


aksh-at (approved):
Nice to see the metaclass go!

PR #709, opened 2023-08-01
Support (de)serializing providers

erikbern:
This PR adds a bunch of code in the short term but will be neutral in the long term once I remove handles later.


erikbern:
Merging since it's fairly low risk to break anything else – it's mostly new code

PR #712, opened 2023-08-02
stub.function now returns a Function

erikbern:
This is a fairly consequential change and finally brings us a big step towards merging providers and handles.

`@stub.function` now returns a `Function` rather than a `FunctionHandle`. This means all functions in global scope will be `Function`, and it means that most methods on `FunctionHandle` are no longer used.

This PR relies on duck typing which is a bit ugly but seems fine since this is just an interim solution.

Next step is to move all the code from the handle class to the provider class.

Notes:
* I've had this change ready for several days but there was a handful small unit tests that were failing – it put me on a wild goose chase and it turned out I needed some fairly substantive changes, eg #708 and #709
* Have run this against all integration tests and they are passing. Wanted to sanity check since this is a big change.
* In what places are we still exposing handles? Off the top of my head (1) lookups (2) looking up objects from the app (3) `_deploy` and maybe some other esoteric things. I'll change those in subsequent PRs.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -501,7 +501,7 @@ def import_function(function_def: api_pb2.Function, ser_cls, ser_fun, ser_params
     active_stub = None
     if isinstance(fun, Function):
         _function_proxy = synchronizer._translate_in(fun)
-        fun = _function_proxy._handle.get_raw_f()
+        fun = _function_proxy.get_raw_f()
         active_stub = _function_proxy._handle._stub
     elif module is not None and not function_def.is_builder_function:
         # This branch is reached in the special case that the imported function is 1) not serialized, and 2) isn't a FunctionHandle - i.e, not decorated at definition time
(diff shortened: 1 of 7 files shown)


modal/_container_entrypoint.py
```diff
@@ -499,10 +499,10 @@ def import_function(function_def: api_pb2.Function, ser_cls, ser_fun, ser_params
 
     # The decorator is typically in global scope, but may have been applied independently
     active_stub = None
-    if isinstance(fun, FunctionHandle):
+    if isinstance(fun, Function):
         _function_proxy = synchronizer._translate_in(fun)
-        fun = _function_proxy.get_raw_f()
-        active_stub = _function_proxy._stub
+        fun = _function_proxy._handle.get_raw_f()
```
aksh-at, line 504: Some users use `get_raw_f()` directly right now (for example to run local unit tests). Do we need to expose it directly on the proxy?

aksh-at (approved):
cool! Is it going to be hard to add typing to these objects later?

PR #717, opened 2023-08-03
container hydrates objects on the stub

erikbern:
This is a change that's been overdue for a long time but we need to do it to get rid of handles.

Instead of creating handles on the app when the container starts, use the stub to locate the providers. Then hydrate those.

This obviates the need for an ugly hack where functions created in the global scope has to "steal" their object ids from the container app.

This will break in some absurd edge cases, like if these things are all true:
* The container is running a serialized function (or a function decorated separate from its definition)
* There are multiple stubs with the same name
* The function needs to call some other function in global scope

In this case we'll fail to locate the stub, which will cause the global functions to not be hydrated properly. I'm mentioning this more as an intellectual curiosity than a real concern.

### DIFF
--- a/client_test/container_app_test.py
+++ b/client_test/container_app_test.py
@@ -21,29 +21,21 @@ def my_f_2(x):
 
 @skip_windows_unix_socket
 @pytest.mark.asyncio
-async def test_container_function_initialization(unix_servicer, container_client):
+async def test_container_function_lazily_imported(unix_servicer, container_client):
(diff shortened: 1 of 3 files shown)


erikbern:
@prbot approve

erikbern:
Would be great to have this reviewed async @aksh-at – there's a lot of tests so I feel pretty confident, but this is a fairly involved change

PR #718, opened 2023-08-03
Simplify object._deploy

erikbern:
I believe this is the last thing I need before I can make the change that causes handles to be internal (not user facing). But that's in a separate PR (should be small though).


erikbern:
@prbot approve

PR #720, opened 2023-08-03
App holds providers

erikbern:
This should be the last place we expose handles to users. Any access of objects on the app now returns the provider instead.

For some reason we had a ton of `isinstance` in tests, so most of the delta is just changing those.


erikbern:
This passes all integration tests as well

aksh-at (approved):

PR #723, opened 2023-08-06
Fix remote() for constructors with no args

aksh-at:
We had a short-circuit for `.remote()` without arguments, which isn't really needed.


erikbern:
Nice, this makes sense. It's possible my refactoring(s) broke this, since it seems like we didn't have any tests. Thanks for adding it!

PR #725, opened 2023-08-08
Move handle methods to providers: Queue and Dict

erikbern:
This moves all "live methods" of `Dict` and `Queue` to the providers (which has already proxied them for a few weeks)

I'm starting with the less frequently used objects to minimize risk (will do `Function` last)


erikbern:
Passes all integration tests btw

PR #727, opened 2023-08-08
Move handle methods to providers: NetworkFileSystem

erikbern:
Last major object before Function

PR #726, opened 2023-08-08
Move handle methods to providers: Volume

erikbern:
Continues work merging handles and providers

PR #728, opened 2023-08-08
Move handle methods to providers: Image

erikbern:
Same as #726 and #727 but images this time (which are fairly trivial since it's only one method)


erikbern:
Passes all integration tests

PR #736, opened 2023-08-09
Remove app object ownership: prep work

erikbern:
As a part of the provider/handle merge, I've realized it makes no sense that both the stub and the app holds objects. It's better if only the stub holds objects and the app just refers to those. Fixing this should also fix the `is_inside` problem.

This PR doesn't accomplish this! However, it contains a bunch of prep work for removing this ownership. I just wanted to roll a bunch of low-risk changes into a prequel PR so that the actual change itself is tinier.


erikbern:
@prbot approve

PR #742, opened 2023-08-09
Fix bug where app deployments don't delete prior objects

erikbern:
This currently unbreaks integration tests

I would like to add a client test to this so we can catch this issue sooner. But I'll put this PR up here for now in case we need to unbreak CI sooner.


erikbern:
edit: added client test for this as well

PR #743, opened 2023-08-10
deprecate app.obj and app[...]

erikbern:
This makes the `App` class a lot less useful, which means we can remove it later, which makes it possible to rename `Stub` to `App`


aksh-at (approved):
nice! Not having to do `stub.app` is immediately quite useful. We should update all our examples too

PR #749, opened 2023-08-14
Change to serve_stub

erikbern:
Needed because I want to change the signature of `import_stub` in a subsequent PR

PR #750, opened 2023-08-14
synchronize LocalEntrypoint

erikbern:
Doing this since it lets me simplify a lot of CLI code so none of it has to deal with synchronization etc. The problem is the `LocalEntrypoint` carries a `_Stub` class but it's actually better if it returns a `Stub` class.

Needed because I want to change the signature of import_stub in a subsequent PR similar to #749 (both PRs were broken out of a PR that was too big)


erikbern:
@prbot approve

PR #751, opened 2023-08-15
Clean up CLI

erikbern:
This gets rid of a bunch of CLI code that handles synchronization - there's no need to do it. This is mostly a leftover from when we had different types of interfaces, but it also took a few PRs (#748, #749, and #750) to prep for it.


erikbern:
@prbot approve

aksh-at (commented):
lgtm

PR #755, opened 2023-08-17
f(...) -> f.local(...), f.call(...) -> f.remote(...)

erikbern:
This changes the calling conventions of functions in the way we've discussed.

Will need a long deprecation window.

Also bumping the client version so that we can track it better.

I'm raising it as pending deprecation warnings right now so it doesn't break integration tests and examples. Will change it to real non-pending deprecation warnings once that's been addressed.


modal/cli/run.py
```diff
@@ -127,18 +126,11 @@ def f(ctx, *args, **kwargs):
...
             if isasync:
                 asyncio.run(func(*args, **kwargs))
             else:
                 func(*args, **kwargs)
-            if app.client.function_invocations == 0:
-                # TODO: better formatting for the warning message
-                warnings.warn(
-                    "Warning: no remote function calls were made.\n"
-                    "Note that Modal functions run locally when called directly (e.g. `f()`).\n"
-                    "In order to run a function remotely, you may use `f.call()`. (See https://modal.com/docs/reference/modal.Function for other options)."
```
erikbern, line 140: was trying to think what to do with this error message, but then i decided to remove the whole logic for it

client_test/cli_test.py
```diff
@@ -123,20 +123,6 @@ def test_run(servicer, set_env_client, test_dir):
...
-@pytest.mark.filterwarnings("error")  # any warnings that aren't caught will fail this test
-def test_local_entrypoint_yes_remote_calls(servicer, set_env_client, test_dir):
-    file = test_dir / "supports" / "app_run_tests" / "local_entrypoint.py"
-    res = _run(["run", file.as_posix()])
-    assert "Warning: no remote function calls were made" not in res.stderr
-
-
-@pytest.mark.filterwarnings("error")  # any warnings that aren't caught will fail this test
-def test_local_entrypoint_no_remote_calls(servicer, set_env_client, test_dir):
-    file = test_dir / "supports" / "app_run_tests" / "local_entrypoint_no_remote.py"
```
aksh-at, line 135: Can remove `"local_entrypoint_no_remote.py` too now

modal/functions.py
```diff
@@ -1155,10 +1154,37 @@ def _get_info(self):
...
     @synchronizer.nowrap
     def __call__(self, *args, **kwargs) -> Any:  # TODO: Generics/TypeVars
         if self._get_is_remote_cls_method():  # TODO(elias): change parametrization so this is isn't needed
-            return self.call(*args, **kwargs)
+            # TODO(erikbern): deprecate this soon too
+            return self.remote(*args, **kwargs)
+
+        deprecation_warning(
+            date(2018, 8, 16),
+            "Direct calling `f(...)` of local functions is deprecated. Use `f.local(...)` instead.",
```
aksh-at, line 1185: Since new users will run into this quite a bit, and we removed the "no remote calls" warning, maybe we should add some more text here to explain that function with name `x` is being called locally, and not remotely?

aksh-at (approved):
LGTM besides minor comments

modal/functions.py
erikbern: Updated the warning message. Here's an example output:

```
erikbern@Eriks-MacBook-Air modal-examples % modal run 01_getting_started/hello_world.py
✓ Initialized. View app at https://modal.com/apps/ap-gCRHpKu0LRI1fTyteqLpGi
✓ Created objects.
├── 🔨 Created f.
└── 🔨 Created mount /Users/erikbern/modal-examples/01_getting_started/hello_world.py
/Users/erikbern/modal-examples/01_getting_started/hello_world.py:59: PendingDeprecationError: 2018-08-16: Calling Modal functions like `f(...)` is deprecated. Use `f.local(...)` if you want 
to call the function in the same Python process. Use `f.remote(...)` if you want to call the function in a Modal container in the cloud
  print(f(1000))
hello 1000
1000000
/Users/erikbern/modal-examples/01_getting_started/hello_world.py:62: PendingDeprecationError: 2018-08-16: `f.call(...)` is deprecated. It has been renamed to `f.remote(...)`
  print(f.call(1000))
hello 1000
1000000
hello 0
world 1
hello 2
world 3
hello 4
world 5
hello 6
world 7
hello 8
world 9
hello 10
world 11
hello 12
world 13
hello 14
world 15
hello 16
world 17
hello 18
world 19
2470
✓ App completed.
```

PR #756, opened 2023-08-17
Remove legacy GPU field

ekzhang:
Was deprecated in #122, time to remove.

Server changes to make this not break tests are in PR 7204 of our monorepo

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -996,7 +996,6 @@ message GPUConfig {
 }
 
 message Resources {
-  reserved 1; // Legacy boolean "gpu" field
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -996,7 +996,7 @@ message GPUConfig {
 }
 
 message Resources {
-  bool gpu = 1; // Legacy GPU toggle.
+  reserved 1; // Legacy boolean "gpu" field
```
erikbern, line 999: you can remove this line, i think – it's implied that the field has been removed unless it's the last one

PR #757, opened 2023-08-18
Remove the custom repr stuff from objects

erikbern:
I personally find custom reprs pretty annoying and I'm not sure why it was there – I think maybe from a long time ago when we hadn't figured out the object hierarchy properly.


erikbern:
Pushed a new commit. This is failing because of the `build_def` stuff for `run_function`. Not sure how to fix that in any obvious way but will think about it.

PR #758, opened 2023-08-18
Fix modal vol ls

gimaker:


ekzhang:
Why is it `modal vol ls` and not `modal volume ls`?

gimaker:
> Why is it modal vol ls and not modal volume ls?

@thundergolfer ping ^

PR #759, opened 2023-08-19
Fix async local entrypoints

erikbern:
This was broken because of an annoying quirk with synchronicity. If a synchronized function returns a callable, synchronicity tries to wrap the return value as well. The point of this is to handle decorators.

However in the case of `LocalEntrypoint.raw_f`, we just want to return a `Callable` but not have it wrapped (because it's just a function). I solved it in a stupid way by returning a `FunctionInfo` object instead. But I'm not super happy with it.

### DIFF
--- a/client_test/cli_test.py
+++ b/client_test/cli_test.py
@@ -123,6 +123,16 @@ def test_run(servicer, set_env_client, test_dir):
     _run(["run", file_with_entrypoint.as_posix() + "::stub.main"])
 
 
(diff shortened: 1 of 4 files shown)


erikbern:
@prbot approve

thundergolfer (commented):
LGTM. Thanks for the fix and explanation.

erikbern:
I think we should probably add a feature to synchronicity where you have to mark decorators explicitly. Will think about it.

PR #760, opened 2023-08-19
Send globals along with `run_function` step definition

aksh-at:
Allows us to fix the issue where changing a global variable referenced in a function doesn't rebuild the function:

```python
BASE_MODEL = "foo"

def download_models():
    from transformers import LlamaForCausalLM, LlamaTokenizer

    LlamaForCausalLM.from_pretrained(BASE_MODEL)
    LlamaTokenizer.from_pretrained(BASE_MODEL)
```

We're filtering out any functions referenced here because they can't be serialized in a "stable" way, but there might be other types we should filter out. 

An alternative would be letting users specify parameters explicitly:

```python
def download_models(model):
    from transformers import LlamaForCausalLM, LlamaTokenizer

    LlamaForCausalLM.from_pretrained(model)
    LlamaTokenizer.from_pretrained(model)

image = modal.Image.debian_slim().run_function(download_models, "foo")
```

This isn't great because it's easy for users to miss, and then leads to the same confusion as now when changing a variable doesn't change the image.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -960,14 +960,6 @@ class _Function(_Provider, type_prefix="fu"):
         """mdmd:hidden"""
         return f"{inspect.getsource(self._raw_f)}\n{repr(self._build_args)}"
 
-    def get_globals(self):
-        from cloudpickle.cloudpickle import _extract_code_globals
-
-        func = self._raw_f
-        f_globals_ref = _extract_code_globals(func.__code__)
-        f_globals = {k: func.__globals__[k] for k in f_globals_ref if k in func.__globals__}
-        return f_globals
-
     def _bind_obj(self, obj, objtype):
         # This is needed to bind "self" to methods for direct __call__
         # TODO(erikbern): we're mutating self directly here, as opposed to returning a different _FunctionHandle
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -960,6 +960,14 @@ class _Function(_Provider, type_prefix="fu"):
         """mdmd:hidden"""
         return f"{inspect.getsource(self._raw_f)}\n{repr(self._build_args)}"
 
+    def get_globals(self):
+        from cloudpickle.cloudpickle import _extract_code_globals
+
+        func = self._raw_f
+        f_globals_ref = _extract_code_globals(func.__code__)
+        f_globals = {k: func.__globals__[k] for k in f_globals_ref if k in func.__globals__}
+        return f_globals
+
     def _bind_obj(self, obj, objtype):
         # This is needed to bind "self" to methods for direct __call__
         # TODO(erikbern): we're mutating self directly here, as opposed to returning a different _FunctionHandle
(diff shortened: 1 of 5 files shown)


aksh-at:
@prbot approve

ekzhang:
Can look on Monday?

ekzhang:
Not sure if you want another pair of eyes or just want to merge immediately

aksh-at:
No rush. I decided to hold off to think about this a bit more.

modal/functions.py
```diff
@@ -960,6 +960,14 @@ def get_build_def(self) -> str:
         """mdmd:hidden"""
         return f"{inspect.getsource(self._raw_f)}\n{repr(self._build_args)}"
 
+    def get_globals(self):
+        from cloudpickle.cloudpickle import _extract_code_globals
+
+        func = self._raw_f
+        f_globals_ref = _extract_code_globals(func.__code__)
```
erikbern, line 967: Is this basically the same as inspect.getclosurevars?

Minor preference for putting this in function_utils.py maybe?

modal_proto/api.proto
```diff
@@ -855,10 +855,12 @@ message Image {
   // Note: field 13 is getting replaced field 16. The client now sends both fields,
   // but the server still only reads field 13.
   bool gpu = 13;
-  string build_function_def = 14;
   string context_mount_id = 15;
   GPUConfig gpu_config = 16;
   ImageRegistryConfig image_registry_config = 17;
+
+  string build_function_def = 14;
+  bytes build_function_globals = 18;
```
erikbern, line 863: Can we just serialize this into the previous field? Since I think we already do that client side with other arguments

erikbern (approved):

modal/functions.py
aksh-at: `inspect.getclosurevars` has some weird behavior, e.g. see this issue https://github.com/python/cpython/issues/80878:

> We are taking names from func.__code__.co_names, which also contains names of object's attributes we visited. As a result, we may get more names unexpectedly in 'globals', 'builtins', and even 'unbound'.

Makes sense to move it to function_utils.py, moved

modal_proto/api.proto
aksh-at: Was going to do that earlier but I thought it'd be nice to be able to "turn off" using `build_function_globals` in the image hash if it turned out to have issues. Also think we should separate out the other arguments at some point into a separate PB as we discussed

modal/image.py
```diff
@@ -197,7 +197,7 @@ async def _load(resolver: Resolver, existing_object_id: Optional[str], handle: _
                 build_function_def = build_function.get_build_def()
                 build_function_id = (await resolver.load(build_function)).object_id
 
-                globals = build_function.get_globals()
+                globals = build_function._get_info().get_globals()
```
erikbern, line 200: Minor thing (because this is some cruft I introduced) but we should probably replace `_get_info` with a property instead. In general let's avoid accessing internals of other objects.

But I can fix separately since it's mostly my fault that the `Function` interface is a bit janky

PR #766, opened 2023-08-22
Change the signature of load and preload to operate on Provider not Handle

erikbern:
Doing this in order to move more functionality out of the `Handle` class.

This has the additional benefit of making it possible to use normal methods as loaders (i.e. we don't need anonymous closure necessarily) but that's a separate change.


erikbern:
@prbot approve

PR #769, opened 2023-08-22
Move _hydrate_from_app from Handle to Provider

erikbern:
Last large method. The rest is a bunch of tiny stuff that should be easy to remove.


erikbern:
@prbot approve

PR #775, opened 2023-08-22
Get rid of almost all handle logic

erikbern:
Fairly big PR but it was hard to do this one incrementally.

This one basically moves all handle logic to the provider, making the handle object purely ceremonial. Will remove it soon.


erikbern:
Ran integration tests with this and everything passes

erikbern:
@prbot approve

PR #781, opened 2023-08-23
parametrized functions: misc cleanup

erikbern:
A couple of changes:

1. No need to inspect the signature of the constructor, since it's only used for type errors messages (rare?) and we can do almost the same thing without
2. Instead of constructing a new class _per parametrized function_, we construct one class, and then use that for all parameters. This is conceptually more aligned since it means `cls` creates a new class (rather than a family of classes)

Should be a noop other than error messages


erikbern:
@prbot approve

PR #782, opened 2023-08-23
Deprecate `modal vol` for `modal volume`

ekzhang:
Currently, we have these commands:

```shell
modal nfs  # edits network file systems, previously "shared volumes"
modal volume  # deprecated, tells the user to use `modal nfs` instead
modal vol # edits volumes
```

We deprecated `modal vol` in #626 on July 5. Currently it's August 23. I think after a 1.5 months, it might be worth renaming `modal vol` to `modal volume` finally, and people using the CLI will hopefully no longer be confused with the switch.

There is a note in the `modal volume` help string that tells users to use `modal nfs` instead, and we can leave that there for a few more months.


erikbern:
Nice! What do you think about starting with just a hard-deprecation of `modal volume`? Just to reduce the risk of people running the "wrong" command. But not something I feel strongly about.

erikbern:
Another thought is that `modal vol` maybe is nicer than `modal vol` because it's shorter? But I have no preference either way, it's just a thought.

ekzhang:
> Nice! What do you think about starting with just a hard-deprecation of `modal volume`? Just to reduce the risk of people running the "wrong" command. But not something I feel strongly about.

Hm, what do you mean by that? I think currently `modal volume` already gives a hard deprecation, with exit code 1.

erikbern (approved):
Seems good to merge! The comments above are very minor and I'm happy with whatever route here

ekzhang:
> Another thought is that `modal vol` maybe is nicer than `modal vol` because it's shorter? But I have no preference either way, it's just a thought.

That makes sense, I think it was just a bit confusing since we call them `modal.Volume` everywhere and but there's a `vol` in the CLI. Maybe will let it sit for a few days.

erikbern:
> I think currently `modal volume` already gives a hard deprecation, with exit code 1.

Great – I thought we had just soft-deprecated it (that it would output a warning). If it's already hard-deprecated (with a failure) then ignore my previous comment

erikbern:
> That makes sense, I think it was just a bit confusing since we call them `modal.Volume` everywhere and but there's a `vol` in the CLI. Maybe will let it sit for a few days.

Yeah true, and it's not _that_ hard to type "volume". I'm fine with either.

PR #788, opened 2023-08-24
Separate method for generators

erikbern:
This breaks out `.remote` into `.remote_gen` for generators. The reason is it makes it easier to define the function signature, which will make it easier to support decorators for live methods, which is required for lazy loading functions later.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -675,7 +675,7 @@ class _Function(_Object, type_prefix="fu"):
                     validated.append((path, vol))
                 return validated
 
-            async def _load_ids(providers) -> typing.List[str]:
+            async def _load_ids(providers) -> List[str]:
                 loaded_handles = await asyncio.gather(*[resolver.load(provider) for provider in providers])
                 return [handle.object_id for handle in loaded_handles]
 
@@ -1073,7 +1073,7 @@ class _Function(_Object, type_prefix="fu"):
     @warn_if_generator_is_not_consumed
     async def starmap(
         self, input_iterator, kwargs={}, order_outputs=None, return_exceptions=False
-    ) -> AsyncGenerator[typing.Any, None]:
+    ) -> AsyncGenerator[Any, None]:
         """Like `map` but spreads arguments over multiple function arguments
 
         Assumes every input is a sequence (e.g. a tuple).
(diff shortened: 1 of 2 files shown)


modal/functions.py
```diff
@@ -1106,20 +1106,46 @@ def remote(self, *args, **kwargs) -> Awaitable[Any]:  # TODO: Generics/TypeVars
...
         else:
-            return self._call_function(args, kwargs)
+            deprecation_warning(
+                date(2018, 8, 16), "`f.call(...)` is deprecated. It has been renamed to `f.remote_gen(...)`"
+            )
+            return self.remote(*args, **kwargs)
+
+    async def shell(self, *args, **kwargs):
+        if self._is_generator:
+            # TOOD(erikbern): would be easy to add support, but it seems like a niche case tbh
```
aksh-at, line 1146: Would be nice to add support for this if it's easy. Since LLM streaming has been popular recently, people do define generator functions more often, and when that happens, it's useful to `modal shell` to debug that function's environment.

modal/functions.py
```diff
@@ -1106,20 +1106,46 @@ def remote(self, *args, **kwargs) -> Awaitable[Any]:  # TODO: Generics/TypeVars
...
+                "A non-generator function cannot be called with `.remote_gen(...)`. Use `.remote(...)` instead."
+            )
+        async for item in self._call_generator(args, kwargs):  # type: ignore
+            yield item
+
+    def call(self, *args, **kwargs) -> Awaitable[Any]:  # TODO: Generics/TypeVars
         if self._is_generator:
-            return self._call_generator(args, kwargs)  # type: ignore
+            deprecation_warning(
+                date(2018, 8, 16), "`f.call(...)` is deprecated. It has been renamed to `f.remote(...)`"
```
aksh-at, line 1135: What's up with the date?

aksh-at (commented):
Nice, I think this will be nicer for users ultimately

modal/functions.py
erikbern: ok let me see what i can do

modal/functions.py
erikbern: This is the existing deprecation warning split in two – I think the git diff isn't matching it well

aksh-at: You're right. Weird that the year is 2018.

modal/functions.py
erikbern: Ok I think this should work now

modal/functions.py
erikbern: lol yeah that's obviously wrong – will fix.

sorry was confused since i just copy pasted

aksh-at (approved):

PR #790, opened 2023-08-24
@live_method decorator

erikbern:
This annotates all "live" methods with a decorator. Only used for functions so far. Currently all the decorator does is to check that the object is hydrated, but a later point I'm planning to use it for "lazy" hydration of certain objects, in particular parametrized functions.

There's a failing test right now because we don't "unhydrate" when an app finishes running – will add that and rebase before merging this.

I'll also add this decorator to all other object types at a later point – images, volumes etc.


erikbern:
@prbot approve

PR #794, opened 2023-08-25
Deprecate from_dockerhub in favor of from_registry

ekzhang:
See MOD-1535 if that makes sense — many of the uses of these are not actually from DockerHub

```python
Image.from_registry("nvcr.io/nvidia/pytorch:22.12-py3")
Image.from_registry("python:3.11")
Image.from_registry("ghcr.io/modal-labs/cadre:latest")
Image.from_registry("public.ecr.aws/docker/library/python:slim-bookworm")
```

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -21,7 +21,7 @@ from ._resolver import Resolver
 from ._serialization import serialize
 from .app import is_local
 from .config import config, logger
-from .exception import InvalidError, NotFoundError, RemoteError, deprecation_error
+from .exception import InvalidError, NotFoundError, RemoteError, deprecation_error, deprecation_warning
 from .gpu import GPU_T, parse_gpu_config
 from .mount import _Mount
 from .network_file_system import _NetworkFileSystem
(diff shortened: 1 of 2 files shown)


ekzhang:
Marking as draft until consensus

modal/image.py
```diff
@@ -882,6 +882,19 @@ def from_dockerhub(
         setup_dockerfile_commands: List[str] = [],
         force_build: bool = False,
         **kwargs,
+    ) -> "_Image":
+        deprecation_warning(
+            date(2023, 8, 25), "`Image.from_dockerhub` is deprecated. Use `Image.from_registry` instead."
+        )
+        return _Image.from_registry(tag, setup_dockerfile_commands, force_build, **kwargs)
```
erikbern, line 889: super minor nit (not blocking) but we generally use "newspaper style" function ordering in Modal, i.e. callee is defined before caller

erikbern (approved):
looks good!

modal/image.py
ekzhang: Got it!

PR #802, opened 2023-08-29
Fix MODAL_PROFILE bug in set-environment CLI

ekzhang:
If you set the profile via `MODAL_PROFILE`, this would display `default` instead of the currently-passed profile.

For example:
```sh-session
$ MODAL_PROFILE=modal_labs modal config set-environment prod
New default environment for profile default: prod
```

Even though it updates the `modal_labs` profile

### DIFF
--- a/modal/cli/config.py
+++ b/modal/cli/config.py
@@ -3,7 +3,7 @@ import pprint
 
 import typer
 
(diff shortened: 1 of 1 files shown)


ekzhang:
@prbot approve

PR #807, opened 2023-08-30
Add st_mode to MountFile and unify internal_api

ekzhang:
Adding a mode flag to the `MountFile` message, but not handling it in the server yet.

This change is fully backwards-compatible, just adding more information to the client MountBuild message protobuf.

### DIFF
--- a/modal/_blob_utils.py
+++ b/modal/_blob_utils.py
@@ -322,13 +322,14 @@ class FileUploadSpec:
     use_blob: bool
     content: Optional[bytes]  # typically None if using blob, required otherwise
     sha256_hex: str
+    mode: int  # file permission bits (last 12 bits of st_mode)
     size: int
 
 
 def get_file_upload_spec(filename: Path, mount_filename: str) -> FileUploadSpec:
     # Somewhat CPU intensive, so we run it in a thread/process
-    size = os.path.getsize(filename)
-    if size >= LARGE_FILE_LIMIT:
+    stat = os.stat(filename)
+    if stat.st_size >= LARGE_FILE_LIMIT:
         use_blob = True
         content = None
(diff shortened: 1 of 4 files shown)


modal/_blob_utils.py
```diff
@@ -322,13 +322,14 @@ class FileUploadSpec:
...
     sha256_hex: str
+    mode: int  # file permission bits (last 12 bits of st_mode)
     size: int
 
 
 def get_file_upload_spec(filename: Path, mount_filename: str) -> FileUploadSpec:
     # Somewhat CPU intensive, so we run it in a thread/process
-    size = os.path.getsize(filename)
-    if size >= LARGE_FILE_LIMIT:
+    stat = os.stat(filename)
```
thundergolfer, line 331: Does this work on Windows?

ekzhang: Yeah, it does appear to, as of at least Python 3.3.4. https://stackoverflow.com/questions/5275731/os-stat-on-windows

I can add a client test to make sure also

ekzhang: Thanks for pointing it out! I ended up doing this, which should be fully backwards-compatible.

```python
        # Python appears to give files 0o666 bits on Windows (equal for user, group, and global),
        # so we mask those out to 0o755 for compatibility with POSIX-based permissions.
        mode=stat.st_mode & (0o7777 if platform.system() != "Windows" else 0o7755),
```

By the way, the mode flag needs to be included in the mount content hash. But we need to do this in a backwards-compatible way on the client mount to not trigger a bunch of image rebuilds, so I'm just skipping that field in the current JSON-based hash format if the mode is equal to `0o644`.

I just checked our client mount, and there are no files with executable bit set, luckily.

PR #811, opened 2023-08-30
Fix deprecation warning in modal shell

erikbern:
It's pretty unfortunate that this code path is untested, but it's a bit annoying to fix and I think this needs to be rewritten anyway.


ekzhang:
Thanks!

PR #813, opened 2023-08-30
Do not allow large serialized functions

ekzhang:
Modal breaks with a cryptic "unknown 413" gRPC error message when a function in a Jupyter Notebook closes over a global variable in scope that is over 100 MiB when serialized. This PR warns users if their function closes over 256 KiB of data (a conservative limit), then emits a full error if the size is over 16 MiB.

Also added a test for this behavior in the client.


aksh-at (approved):
nice, thanks!

PR #816, opened 2023-08-30
Use plain objects for Modal classes

erikbern:
The rough idea is that `@stub.cls` doesn't create a new Python class, it returns an instance of `modal.Cls` instead. That in turn has a `__call__` method which creates a `modal.Obj` class. That in turn implements `__getattr__` which returns `Function` objects. So for instance to support `FooCls(1, 2, 3).method("xyz")` then

* `FooCls` is an _object_ (not a class) of type `Cls`
* `FooCls(1, 2, 3)` returns an object of type `Obj`
* `FooCls(1, 2, 3).method` is a `Function` which comes from `Obj.__getattr__`

The main benefit of this is that we can support `modal.Cls.lookup("my-object")` quite easily, since `modal.Cls` will be a Modal object (with an object id etc) later – I haven't implemented the backend part of this, but this should be quite doable.

This PR is not big but somewhat complex mostly because of a lot of subtleties (for instance – _inside_ function methods, you need `self` to refer to the original class instance, but you need `self.method` to still return `Function` objects). I think there's some room to simplify things over time.

This supersedes #805 since it's a completely different approach. It does add the constructor as an alternative to `.remote` but I haven't implemented the deprecations etc. Will do that in a separate PR.

### DIFF
--- a/client_test/cls_test.py
+++ b/client_test/cls_test.py
@@ -5,7 +5,6 @@ import pytest
 from modal import Stub, method
 from modal._serialization import deserialize
 from modal.cls import ClsMixin
-from modal.functions import Function
 from modal_proto import api_pb2
 
 stub = Stub()
@@ -59,6 +58,11 @@ def test_call_cls_remote_sync(client):
         foo_remote = FooRemote.remote(3, "hello")
         # Mock servicer just squares the argument
         # This means remote function call is taking place.
+        assert foo_remote.bar.remote(8) == 64
+        assert foo_remote.bar(8) == 64
+
+        # Check new syntax
+        foo_remote = FooRemote(3, "hello")
         assert foo_remote.bar.remote(8) == 64
         assert foo_remote.bar(8) == 64
 
@@ -118,7 +122,6 @@ def test_run_class_serialized(client, servicer):
     obj = cls()
     meth = fun.__get__(obj, cls)
 
-    assert isinstance(obj.bar, Function)
     # Make sure it's callable
     assert meth(100) == 1000000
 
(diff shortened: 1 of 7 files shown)


erikbern:
Some notes on how to implement the `cls` modal object later:
* On the backend it's just an id mapping to a dict of base functions by name
* The cls would have a load method that creates this
* When a user does a lookup, it returns the dict of base functions as metadata
* During object construction (when you try to instantiate eg `MyClass(1, "foo")`, it just returns and `Obj` just like now (non-blocking non-async function just creating a thin reference)
* Lazy construction of functions just like now

client_test/cls_test.py
```diff
@@ -62,6 +61,11 @@ def test_call_cls_remote_sync(client):
         assert foo_remote.bar.remote(8) == 64
         assert foo_remote.bar(8) == 64
 
+        # Check new syntax
```
erikbern, line 64: in the next PR, I'll deprecate the "old" remote syntax

erikbern:
This passes all integration tests

aksh-at (approved):
LGTM. As discussed on Slack, might be nice to check if typing still works

erikbern:
@aksh-at looks like types work well in the new code – i added a bunch of assertions

aksh-at:
> @aksh-at looks like types work well in the new code – i added a bunch of assertions

That's awesome! That means it works much better than the old `.remote()` construction

erikbern:
> That's awesome! That means it works much better than the old `.remote()` construction

Yeah kind of by luck – I wasn't trying to :)

PR #817, opened 2023-08-30
add internal worker_id param for functions

aksh-at:
See https://github.com/modal-labs/modal/pull/7425

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -794,6 +794,7 @@ class _Function(_Object, type_prefix="fu"):
                 stub_name=stub_name,
                 is_builder_function=is_builder_function,
                 allow_concurrent_inputs=allow_concurrent_inputs,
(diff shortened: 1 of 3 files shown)


modal/functions.py
```diff
@@ -794,6 +795,7 @@ async def volume_loader():
                 stub_name=stub_name,
                 is_builder_function=is_builder_function,
                 allow_concurrent_inputs=allow_concurrent_inputs,
+                worker_id=_worker_id,
```
erikbern, line 798: wdyt about just reading this from an environment variable here? like `config.get(...)`

aksh-at: yeah I was thinking that. Only reason against it was that it could be nice in theory to pin multiple functions to different workers for debugging (e.g. CPU and GPU). But maybe that's too much of a stretch.

PR #819, opened 2023-08-30
Standalone Python builds

ekzhang:
## Problem

Our current way of setting up images from Docker Hub is mechanical and doesn’t work for different Linux distributions.

```jsx
Image.from_registry(
    "nvidia/cuda:12.2.0-devel-ubuntu22.04",
    setup_dockerfile_commands=[
        "RUN apt-get update",
        "RUN apt-get install -y python3 python3-pip python-is-python3",
    ],
)
```

For example, CentOS images would require you to use `dnf`, and alpine would be `apk`. The user is responsible for figuring out these commands, which is difficult unless you’re familiar with Linux packaging.

It’s also not easy to install newer Python versions. For instance, Ubuntu 20.04 ships with Python 3.8, and the best way to get a newer Python would be to add `deadsnakes/ppa`.

```docker
RUN add-apt-repository ppa:deadsnakes/ppa
RUN apt install python3.9 python3.9-pip
RUN ln -s /usr/local/bin/python3.9 /usr/local/bin/python
RUN ln -s /usr/local/bin/pip3.9 /usr/local/bin/pip
```

This is pretty complex and error-prone to set up.

For most images created originally from Modal, we just base off `debian_slim`, so it’s pretty easy to assume that apt is present, and we provide helpers like `apt_install()`. However, for images pulled from Docker Hub, there is likely considerably more diversity, so adding parameters to help install Python for specific distributions might be brittle.

## Summary

Adds a parameter `add_python` to `Image.from_registry()`, which installs a specific minor version of Python inside the image.

```python
Image.from_registry("nvidia/cuda:12.2.0-devel-...", add_python="3.10")
Image.from_registry("ubuntu:20.04", add_python="3.11")
Image.from_registry("opensuse/leap:15.05", add_python="3.9")
```

This would eliminate all of our current documented use of `setup_dockerfile_commands`. We can still support that parameter, just run it after adding Python into the image.

This parameter runs **after** pulling the image rootfs, but before installing the Modal package.


modal_base_images/base_mounts.py
```diff
@@ -0,0 +1,69 @@
+# Copyright Modal Labs 2022
```
erikbern, line 1: 2023 :)

PR #821, opened 2023-09-01
NFS support for sandboxes

aksh-at:
Can use both "ephemeral" and persisted sandboxes. For example:

```python
@stub.local_entrypoint()
def main():
    nfs = modal.NetworkFileSystem.new()
    sb = stub.app.spawn_sandbox(
        "bash",
        "-c",
        "echo foo > /cache/a.txt",
        network_file_systems={"/cache": nfs},
    )
    sb.wait()
    for x in nfs.read_file("/a.txt"):
        print(x)

```

### DIFF
--- a/client_test/sandbox_test.py
+++ b/client_test/sandbox_test.py
@@ -6,7 +6,8 @@ import pytest
 import time
 from pathlib import Path
 
-from modal import Image, Mount, Secret, Stub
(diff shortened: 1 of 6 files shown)


aksh-at:
@prbot approve

erikbern (commented):
LGTM!

PR #822, opened 2023-09-01
Deprecate old syntax for parametrized functions

erikbern:
This adds deprecation warnings for the old syntax (using `.remote` on the class, NOT using `.remote` on the method) as well as deprecating the mixin (since it's no longer needed)


erikbern:
@prbot approve

PR #824, opened 2023-09-03
broken: add type hints for decorated function arguments

erikbern:
This should in theory work, in that the type arguments of a modalized function should carry through so that you can call it using `.remote` with the same parameter types, and it will know what those are.

But I believe synchronicity mangles ParamSpec right now. Looks like `Callable[P, Any]` gets translated into `Callable[[P], Any]` or something similar. Not sure what's up.

### DIFF
--- a/client_test/function_test.py
+++ b/client_test/function_test.py
@@ -3,7 +3,7 @@ import asyncio
 import inspect
 import pytest
 import time
-import typing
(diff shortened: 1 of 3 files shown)


mwaskom:
Closing as stale for now but can revisit, agree this would be nice to have

PR #826, opened 2023-09-04
[mod-1323] modal shell platform-specific fixes

thundergolfer:
I've identified at least 3 issues with gvisor's terminal handling

1. The exit of the replica/slave side (e.g. the `sh` or `bash` session) does not exit the parent process.
2. Issuing <kbd>CTRL</kbd> + <kbd>d</kbd> (EOF) on an empty buffer does not send EOF to slave, so session doesn't exit
3. Can't set raw mode with `pty.setraw(pty.STDIN_FILENO)` so echo doesn't work.

These PR has workarounds for 1 and 3. Fixing 2 will require a fix to gvisor as I can't think of a simple workaround. 

These changes are also compatible with `runc`.

### DIFF
--- a/modal/_pty.py
+++ b/modal/_pty.py
@@ -7,7 +7,6 @@ import platform
 import select
 import signal
 import sys
-import termios
 import traceback
 from typing import Optional, Tuple, no_type_check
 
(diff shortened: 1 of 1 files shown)


erikbern:
nice!

PR #827, opened 2023-09-05
Make dockerfile_commands() consistent

ekzhang:
Other methods like `run_commands()` take variable arguments here, except for dockerfile_commands.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -644,7 +644,7 @@ class _Image(_Object, type_prefix="im"):
     @typechecked
     def dockerfile_commands(
         self,
-        dockerfile_commands: Union[str, List[str]],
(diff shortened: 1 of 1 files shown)


modal/image.py
```diff
@@ -644,7 +644,7 @@ def poetry_install_from_file(
     @typechecked
     def dockerfile_commands(
         self,
-        dockerfile_commands: Union[str, List[str]],
+        *dockerfile_commands: Union[str, List[str]],
         context_files: Dict[str, str] = {},
```
erikbern, line 648: i guess this could break old code that passes `context_files` as a positional argument. maybe fine?

erikbern: Good idea to add `*`. This method isn't super widely used so I think it's fine to jump straight to `*args`

PR #834, opened 2023-09-05
Only setup Rich tracebacks when running CLI

ekzhang:
This means that `import modal`, as well as Modal in Jupyter notebooks and `python` directly with `app.run()`, will not register global traceback handlers.

## Describe your changes

Resolves MOD-1579.


aksh-at (approved):

erikbern:
Awesome – have wanted this for a while!

Not within the scope of this change, but it reminds me there are other things done in global scope in config.py with potential side effects (logging, warnings) and maybe we should change those in a similar way.

PR #839, opened 2023-09-05
Extend PYTHONPATH instead of overriding it

howard36:
Fixes (MOD-1572)

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1229,9 +1229,17 @@ class _Image(_Object, type_prefix="im"):
         )
         ```
         """
-        return self.extend(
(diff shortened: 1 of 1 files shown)


aksh-at (changes requested):
Can discuss in Slack, but I don't think it works because we override `PYTHONPATH` in the runtime. I think the change needs to be in Rust. Does it work when you test it?

howard36:
Oh, my bad. I thought this was the only place where environment variables were being set. Can you point to which part of the Rust code overrides it? (I'm still not familar with the codebase)

Haven't tested this yet, but will test future changes (is there an easy way to test this?)

howard36:
Closing this, see new PR in the other repo.

PR #840, opened 2023-09-06
Hotfix: Deprecated 172.19.0.1 bridge mentions

ekzhang:


erikbern:
What is this change for? What about old clients?

PR #842, opened 2023-09-06
Create classes on the backend, support class lookups

erikbern:
This enables code such as

```python
MyClass = Cls.lookup("my-cls-app", "Foo")  # lookup class
my_object = MyClass("foo", 123)  # instantiate class
res = my_object.method.remote(1, 2, 3, "xyz")  # call method on object
```

Besides test fixtures and tests (most of this PR), the actual logic is not advanced – just plumbing to hydrate from metadata properly.


aksh-at (approved):
nice!

PR #850, opened 2023-09-08
Handle unserializable values in image `run_function` globals

aksh-at:
I don't love this because it serializes each global value twice. But I wanted to keep the serialization format the same as now. `dill` has a `.pickles()` check, but don't want to add another dep either.


aksh-at:
@prbot approve

ekzhang:
> add another dep

*shudder*

PR #853, opened 2023-09-10
Warnings if you forget parentheses for `function`, `cls`, and `local_entrypoint`

erikbern:
We've had really bad errors for these in the past, e.g. `Modal stub has no registered functions. Nothing to run.` for this code:

```python
import modal

stub = modal.Stub()

@stub.function
def f():
    pass
```

This now outputs this instead: 
```
InvalidError: Did you forget parentheses? Suggestion: `@stub.function()`.
```

### DIFF
--- a/client_test/stub_test.py
+++ b/client_test/stub_test.py
@@ -312,7 +312,7 @@ def test_keyboard_interrupt(servicer, client):
         raise KeyboardInterrupt()
 
 
-def test_local_entrypoint_without_parantheses():
+def test_local_entrypoint_forgot_parentheses():
     stub = Stub()
 
     with pytest.raises(InvalidError) as excinfo:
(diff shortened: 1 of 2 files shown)


aksh-at (approved):
Nice, that's clever! Tiny downside is if someone does something like

```python
@stub.function(image)
```

(i.e. uses an arg without the parameter), they will get this warning which could be confusing. If we want to guard against that we could also check if the `_warn*` arg is a function? Not super important

erikbern:
> (i.e. uses an arg and not a kwarg), they will get this warning which could be confusing. If we want to guard against that we could also check if the `_warn*` arg is a function? Not super important

Yeah I thought about that too. I think it's a bit confusing to allow positional arguments for `image` for this one. But I can add an extra check for this specific case.

aksh-at:
> I think it's a bit confusing to allow positional arguments for image for this one

Not saying we should allow that, but people will try to do it, and then get even confused by this warning

PR #855, opened 2023-09-11
Handle `web_endpoint` without parens, unify decorator check between `web_endpoint`, `asgi_app`, and `wsgi_app`

erikbern:
I forgot to handle `web_endpoint` in https://github.com/modal-labs/modal-client/pull/853 – adding it

EDIT: also adding `asgi_app` and `wsgi_app`

Turns out we already had a check but I'm changing it slightly to not allow method as a positional arg. I doubt people pass `method` as a positional argument? But if they do for whatever reason, they will get an error about it:
```
Positional arguments to `@web_endpoint` are not allowed. Suggestion: `@web_endpoint(method="GET")`
```

EDIT: turns out we also had a check for `asgi_app` and `wsgi_app` that did allow for `label` to be provided positionally.

### DIFF
--- a/client_test/decorator_test.py
+++ b/client_test/decorator_test.py
@@ -8,51 +8,43 @@ from modal.exception import InvalidError
 def test_local_entrypoint_forgot_parentheses():
     stub = Stub()
 
-    with pytest.raises(InvalidError) as excinfo:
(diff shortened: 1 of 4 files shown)


aksh-at (approved):
nice, yeah makes sense to add the "position arguments are not allowed" to everything just in case.

PR #857, opened 2023-09-11
Move spawn_sandbox to the stub

erikbern:
Instead of `stub.app.spawn_sandbox`, users should use `stub.spawn_sandbox`

There is a bit of irony in this PR:
1. We're moving methods from the app to the stub, in order to deprecate the app, so that we can rename the stub to the app
2. Part of why stubs and apps should be merged is that there's already annoying mutual dependencies between those two apps. This PR exacerbates these dependencies. But that seems fine since the end goal is to merge those and get rid of the dependencies altogether.

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -14,6 +14,7 @@ from modal_utils.async_utils import synchronize_api, synchronizer
 from ._function_utils import FunctionInfo
 from ._ipython import is_notebook
 from ._output import OutputManager
+from ._resolver import Resolver
 from .app import _App, _container_app, is_local
 from .client import _Client
 from .cls import _Cls
@@ -29,6 +30,7 @@ from .proxy import _Proxy
 from .queue import _Queue
 from .retries import Retries
 from .runner import _run_stub
+from .sandbox import _Sandbox
 from .schedule import Schedule
 from .secret import _Secret
 from .volume import _Volume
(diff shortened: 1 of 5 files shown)


erikbern:
This code is a bit janky because of the mutual relationship between app and stub. Which is exactly the problem I'm trying to solve by merging the app and stub. But this PR makes this slightly more janky in the short term.

erikbern:
@prbot approve

modal/stub.py
```diff
@@ -613,5 +616,49 @@ def _get_deduplicated_function_mounts(self, mounts: Dict[str, _Mount]):
...
+        Refer to the [docs](/docs/guide/sandbox) on how to spawn and use sandboxes.
+        """
+        from .sandbox import _Sandbox
+        from .stub import _default_image
+
+        if self._app is None:
+            raise InvalidError("`stub.spawn_sandbox` requires a running app.")
+
+        # TODO(erikbern): pulling a lot of app internals here, let's clean up shortly
+        resolver = Resolver(self._app._client, environment_name=self._app._environment_name, app_id=self._app._app_id)
```
aksh-at, line 646: yeah this was the main reason for putting this on `App` earlier, but makes sense to do this now temporarily.

erikbern: Might make sense to merge the resolver and app classes later

PR #862, opened 2023-09-12
[mod 1599] Volume.read_file support

thundergolfer:
### Describe your changes

Proto msgs and `Volume.read_file` method for volumes. 

- MOD-1599

   - Requires servers to be deployed with implementations of RPCs

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1445,7 +1445,7 @@ service ModalClient {
   // Volumes
   rpc VolumeCreate(VolumeCreateRequest) returns (VolumeCreateResponse);
   rpc VolumeCommit(VolumeCommitRequest) returns (google.protobuf.Empty);
-  rpc VolumeGetFile(VolumeGetFileRequest) returns (VolumeGetFileResponse);
+  rpc VolumeGetFile(VolumeGetFileRequest) returns (stream google.protobuf.BytesValue);
   rpc VolumeList(VolumeListRequest) returns (VolumeListResponse);
   rpc VolumeListFiles(VolumeListFilesRequest) returns (stream VolumeListFilesResponse);
   rpc VolumeReload(VolumeReloadRequest) returns (google.protobuf.Empty);
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1262,6 +1262,13 @@ message VolumeGetFileRequest {
   bytes path = 2;
 }
 
+message VolumeGetFileResponse {
+  oneof data_oneof {
+    bytes data = 1;
+    string data_blob_id = 2;
+  }
+}
+
 message VolumeListFilesEntry {
   enum FileType {
     UNSPECIFIED = 0;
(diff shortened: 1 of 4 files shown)


erikbern:
maybe wip but it would be good to add a client test

modal_proto/api.proto
```diff
@@ -1257,6 +1257,11 @@ message VolumeCommitRequest {
   string volume_id = 1;
 }
 
+message VolumeGetFileRequest {
+  string volume_id = 1;
+  bytes path = 2;
```
gimaker, line 1262: Why `bytes` as opposed to `string`?

We use `string` pretty much everywhere for paths, including for listing entries in a volume:
```
message SharedVolumeListFilesEntry {
  enum FileType {
    UNSPECIFIED = 0;
    FILE = 1;
    DIRECTORY = 2;
  }
  string path = 1;
  FileType type = 2;
}
```

modal_proto/api.proto
```diff
@@ -1433,6 +1438,7 @@ service ModalClient {
   // Volumes
   rpc VolumeCreate(VolumeCreateRequest) returns (VolumeCreateResponse);
   rpc VolumeCommit(VolumeCommitRequest) returns (google.protobuf.Empty);
+  rpc VolumeGetFile(VolumeGetFileRequest) returns (stream google.protobuf.BytesValue);
```
gimaker, line 1441: Do you foresee any issues with streaming GiB or TiB worth of data here? Just curious if there's a reason we didn't take the same approach for `SharedVolumeGetFile` :thinking: 

For really large files the stream is liable to be interrupted before the download completes? E.g. we deploy a change and churn server on the backend while the download is in progress.

luiscape: Tangentially related: maybe we can have this return indexed byte chunks so we can implement resume-ability?

modal_proto/api.proto
thundergolfer: Filepaths in the volume are `bytes` so it seemed incorrect to only accept UTF-8 strings. It would make certain files in a volume inaccessible by the RPC. 

SharedVolume RPCs indeed use `string` and I initially did that, but switched because I thought it was a bug.

modal_proto/api.proto
thundergolfer: Yeh I definitely did 😅. I set a limit on the server-side PR, but it's a hand wavy limit. 

> For really large files the stream is liable to be interrupted before the download completes?

Yep it's a vulnerability for sure. The `SharedVolumeGetFile` also has a long blocking upload to S3 which can also get interrupted. So I think the difference is in how much data is pushed through the external server. The `SharedVolumeGetFile` mechanism adds an upload step to shift the long-running stream from external server to S3.

thundergolfer: Actually I think in the medium-term future we'd like to just directly return an pre-signed download link from store. We can't do this currently but will be able to soon-ish. 

This makes the `SharedVolumeGetFileResponse` API more appropriate for sure.

modal_proto/api.proto
gimaker: Fair enough! Makes sense.

modal_proto/api.proto
gimaker: > has a long blocking upload to S3 which can also get interrupted. So I think the difference is in how much data is pushed through the external server.

True that. At least we have have lots of bandwidth available internally in our backend, which might not be the case for clients. But yeah, downloading really large files will be painful medium-term :|

gimaker (approved):
:+1:

PR #864, opened 2023-09-12
WIP: Modal method calling a local function

erikbern:
Currently just reproduces a problem reported by a user. Will work on a fix next.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -1128,6 +1128,9 @@ class _Function(_Object, type_prefix="fu"):
         return self._info
 
     def _get_self_obj(self):
+        # TODO(erikbern): See https://github.com/modal-labs/modal-client/pull/864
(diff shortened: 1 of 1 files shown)


danielnorberg:
Afaiu my repro was not parameterized, not sure if that matters.

erikbern:
Yeah you're right, it has to do with the class, not whether it uses parameters. Let me rename the PR.

erikbern:
I think the fix should work. That being said, I'm not super happy with this fix – I think it would be better if a function "knows" that it's a method and if you call `.local()` when an object isn't present, it should fail. Will see if I can fix.

erikbern:
I suspect there are some weird cases that might still cause issues, e.g. if the `@cls` decorator isn't in global scope but applied independently. But tbh this feels like a pretty fringe use case. We can deal with it later.

PR #869, opened 2023-09-15
Patch progress bars for methods in `_Cls`

gongy:
### Brief

Save `OutputManager` object on `_Cls`. Then, when building methods, set their output manager from the `_Cls`.

After (both class function f and regular function go)

https://github.com/modal-labs/modal-client/assets/8001209/79176919-838c-43ab-9a99-2c0ab5b0684f

Before (only regular function go)

https://github.com/modal-labs/modal-client/assets/8001209/d5390cdc-6b92-4406-b217-487dbfd9d853

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -9,6 +9,7 @@ from google.protobuf.message import Message
 from modal_proto import api_pb2
 from modal_utils.async_utils import synchronize_api
 
+from ._output import OutputManager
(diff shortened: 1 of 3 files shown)


erikbern:
LGTM! although let's clean up the output mgr code separately – i think it would be better as a constructor argument. but that's an existing sin

erikbern (approved):

PR #870, opened 2023-09-16
stub.run() and run_stub() now returns the stub itself

erikbern:
Another step towards making the `App` class purely internal. `stub.run()` and `run_stub` now returns the stub itself, since it implements basically the same interface (leveraging duck typing, basically).

This shouldn't break any user code unless they are doing something very hacky with Modal internals.


erikbern:
@prbot approve

PR #871, opened 2023-09-17
Add @live_method to a bunch of live methods

erikbern:
Should be completely harmless. Adding this so that I can support lazy loading of `from_name` (which will make it possible to merge it with `lookup`)


erikbern:
@prbot approve

danielnorberg:
👍

PR #874, opened 2023-09-18
Token flow: run a local http server validating the request

erikbern:
This isn't used yet, but the frontend could in theory hit localhost:xyz and validate the token flow now rather than asking the user to click a button.

Took 10 min to implement so now we have it. We can add support for it in the frontend later.

Regarding security, I think there are only two threats that are easily mitigated:
* If the validation URL is an arbitrary URL, an attacker can run their own validation server on the public internet. To prevent this, we restrict it to `localhost` only – we only store a port number
* We don't want the validation to be something too dumb (eg relying on just 200 OK), since then an attacker could potentially leverage some http server that's commonly running on localhost. So we make sure to return the token flow id as an arbitrary identifier which the frontend can sanity check.

### DIFF
--- a/client_test/conftest.py
+++ b/client_test/conftest.py
@@ -129,6 +129,8 @@ class MockClientServicer(api_grpc.ModalClientBase):
         self.sandbox_defs = []
         self.sandbox: subprocess.Popen = None
 
+        self.token_flow_localhost_port = None
+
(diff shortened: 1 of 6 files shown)


erikbern:
@prbot approve

PR #877, opened 2023-09-18
Implement relay tunneling support for the client

ekzhang:
This is an experimental feature. Relays allow you to forward ports within a running Modal container.

```python
@modal.function()
def my_function():
    with modal.forward(8000, mode="tls") as tunnel:
        print(tunnel.url)
        some_web_server.listen(8000)
```

This gives you a unique, transient HTTPS URL. The tunnel acts on the transport layer (TCP), and it automatically terminates TLS for all incoming connections as well, so data is encrypted at all stages in transit.

This can be used to proxy anything over TCP/TLS, including web servers, Jupyter notebooks, VS Code, web interfaces like Gradio, SSH, and so on. Even an SMTP server, if you really wanted to.

Tunnels support multiple distributed connection to multiple server relays that we can run in all Modal regions, and they automatically detect which instance has the lowest latency to the current container.

While we're developing, I didn't expose this publicly yet. You must import the `tunnel` function from `modal._relay_client`.

See the associated PR on the server side for implementation / ops details.

### DIFF
--- a/modal/_relay_client.py
+++ b/modal/_relay_client.py
@@ -1,0 +1,172 @@
+"""Client for Modal relay servers, allowing users to expose TLS.
+
+Based on the Rust implementation in our internal modal-relay crate. That one is
+end-to-end tested and is the source of truth for our client protocol.
+"""
+
+import asyncio
+import contextlib
+import json
+from dataclasses import dataclass
+from typing import Any, AsyncIterator, Tuple
+
+from google.protobuf.empty_pb2 import Empty
+
+from modal_proto import api_pb2
+from modal_utils.async_utils import TaskContext, synchronize_api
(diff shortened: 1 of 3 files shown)


erikbern:
Interesting!

Would it make sense to do the forwarding in the worker instead? Might be a performance improvement. I.e. the worker copies data to the container port itself (I guess there will have to be some network setup on container initialization so the container port is accessible to the worker)

Also, would it make sense to make the forwarding a property of the caller instead? So basically you call a function but you get back an endpoint URL to that container instead. I think that could be pretty useful if you want to spawn a "session backend" programmatically for instance.

ekzhang:
Merging now, still private API. Time to iterate internally before we release.

```python
from modal._relay_client import tunnel
```

PR #880, opened 2023-09-18
run `__enter__` on local class construction

erikbern:
This makes the behavior more consistent with remote calls.

~Not sure if this causes it to run on the synchronicity thread – will see if I can figure it out.~ EDIT: verifying this using a bunch of assertions


erikbern:
@prbot approve

ekzhang:
Looks good

PR #883, opened 2023-09-19
Revert run `__enter__` on local class construction

gimaker:
Reverts modal-labs/modal-client#880

Reverting because it appears to have some unintended side effects - `__enter__` now runs twice in the container:

```python
import os
from modal import Stub, method

stub = Stub()

@stub.cls()
class Test:
  def __init__(self):
    print(f"__init__, pid={os.getpid()}, is_inside={stub.is_inside()}")

  def __enter__(self):
    print(f"__enter__, pid={os.getpid()}, is_inside={stub.is_inside()}")

  @method()
  def f(self):
    print(f"f, pid={os.getpid()}, is_inside={stub.is_inside()}")


@stub.local_entrypoint()
def main():
    print(f"local pid={os.getpid()}")
    t = Test()
    t.f.remote()
```

-->

```
local pid=628711
__init__, pid=7, is_inside=True
__enter__, pid=7, is_inside=True
__enter__, pid=7, is_inside=True
f, pid=7, is_inside=True
```

Prior to this change the output was:

```
local pid=629011
__init__, pid=7, is_inside=True
__enter__, pid=7, is_inside=True
f, pid=7, is_inside=True
```

### DIFF
--- a/client_test/cls_test.py
+++ b/client_test/cls_test.py
@@ -1,7 +1,6 @@
 # Copyright Modal Labs 2022
 import inspect
 import pytest
-import threading
 from typing import TYPE_CHECKING
 
 from typing_extensions import assert_type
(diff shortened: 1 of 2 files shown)


gimaker:
Screenshot from 2023-09-19 11-09-21

:facepalm:

danielnorberg (approved):

PR #885, opened 2023-09-19
Annotate .modal.toml with metadata

erikbern:
This change makes it so that
1. We preserve any comments in `~/.modal.toml`
2. When a token is created, it adds a comment such as `created on 2023-09-19 for workspace modal-labs`

E.g. I have a million tokens and it's a bit annoying to keep track of them:

```
[personal]
token_id = "ak-xxx"
token_secret = "as-yyy"
sentry_dsn = ""

[dev]
server_url = "http://localhost:8888"
sentry_dsn = ""
token_id = "ak-xxx"
token_secret = "as-yyy"

[bot]
token_id = "ak-xxx"
token_secret = "as-yyy"

[modal_labs] # created on 2023-09-19 for workspace modal-labs
token_id = "ak-xxx"
token_secret = "as-yyy"
environment = "main"
active = true

[_test]
token_id = "abc"
token_secret = "xyz"

[localhost-test]
token_id = "ak-xxx"
token_secret = "as-yyy"
```

ANOTHER OPTION: We could potentially write it into the config as real payload data:

```
[modal_labs]
token_id = "ak-xxx"
token_secret = "as-yyy"
environment = "main"
active = true
created_on = 2023-09-19
workspace_username = "modal-labs"
```

the last two would be dummy config in the sense that it's not actually used

### DIFF
--- a/modal/cli/token.py
+++ b/modal/cli/token.py
@@ -1,4 +1,5 @@
 # Copyright Modal Labs 2022
+import datetime
 import getpass
 import itertools
 import webbrowser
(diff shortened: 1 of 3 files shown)


thundergolfer (approved):
Are you holding off on this because it'll cause an image rebuild?

thundergolfer:
Actually it won't cause an image rebuild because it's not a change to the container `requirements.txt` 👍

erikbern:
> Actually it won't cause an image rebuild because it's not a change to the container `requirements.txt` 👍

good point! ok let me maybe fix this!

mwaskom:
Could be nice but I think partially superseded by now showing the workspace in `modal profile list` and a little stale; going to close for now.

PR #888, opened 2023-09-19
Implement demo `modal jupyter lab` CLI

ekzhang:
This starts a Jupyter Lab on Modal and opens it in your browser.

The command is _very_ janky right now since it builds an image on the first run, and it also shells out to `modal run`. Just a demo. We can remove it later.


ekzhang:
@prbot approve

Only for demo purposes.

PR #894, opened 2023-09-19
Allow specifying multiple GPUs as `type:count`.

ekzhang:
For example, `a10g:2` asks for 2 A10G GPUs.

This can be used right now in `modal launch` and in the future with `modal shell`.


erikbern (approved):
nice – do you mind adding a tiny test for this also?

erikbern (commented):
thanks for the test!

PR #897, opened 2023-09-20
Raise InvalidError in more places

ekzhang:
Previously, deploys of secrets that were invalid or conflicted with another secret would raise an GRPCError from the client, or an internal server error on the web.

This translates them to the correct client error type.

### Before (web UI)

image

### After

image

Resolves MOD-1612.


thundergolfer:
:100

PR #900, opened 2023-09-20
Re-add running `__enter__` on local class construction

erikbern:
Reverts modal-labs/modal-client#883 which was in itself a revert of #880 

Added a test for the underlying problem and also a fix. The code is a bit janky imo but at least somewhat well-tested now.

### DIFF
--- a/client_test/cls_test.py
+++ b/client_test/cls_test.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2022
 import inspect
 import pytest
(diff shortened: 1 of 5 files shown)


gimaker (approved):
Feeling fairly lost in this part of the code but the fix makes sense :+1:

erikbern:
this code is a bit gnarly tbh, but at least there's a bunch of tests now

PR #901, opened 2023-09-20
Remove interactive function selection

erikbern:
We have an interactive function picker when you run `modal shell` (but not `modal run`).

A user just reported a bug that is caused because functions from other apps don't set this property.

In general I want to harmonize object construction so that objects constructed from different methods end up having the same properties. But this will take a fair bit of work – for it to work on remote objects, we have to make sure all properties are a part of the metadata returned from the server.

Until then, fixing this is unfortunately somewhat annoying, other than fixing it in some hacky ways. More generally, maintaining the panel code has caused me considerable cost already during various refactorings. I see the value of the interactive picker, but it feels pretty limited? I think removing it is best route forward - it fixes the immediate user problem and should reduce overall maintenance cost a bit.

This is also one out of two code paths in the client that relies on string representation of objects (the other one being `run_function` so this gets me halfway towards my goal of removing those as well. See #757. That would let us get rid of another thing that has a bunch of maintenance cost for not a ton of benefit (imo).

### DIFF
--- a/client_test/function_test.py
+++ b/client_test/function_test.py
@@ -534,15 +534,6 @@ def test_from_id(client, servicer):
     assert rehydrated_function_call.object_id == function_call.object_id
 
 
-def test_panel(client, servicer):
-    stub = Stub()
-    stub.function()(dummy)
-    function = stub["dummy"]
-    assert isinstance(function, Function)
-    image = stub._get_default_image()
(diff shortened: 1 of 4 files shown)


erikbern:
@prbot approve (got Akshat's blessing over Slack)

PR #903, opened 2023-09-21
Better support for `cls` inheritance

aksh-at:
This now works:

```python
class Base:
    def __enter__(self):
        self.x = 2

    @modal.method()
    def run(self, y):
        return self.x * y

@stub.cls()
class Foo(Base):
    pass

@stub.local_entrypoint()
def main():
    print(Foo().run.remote(3)) # == 6
```

### DIFF
--- a/modal/_function_utils.py
+++ b/modal/_function_utils.py
@@ -89,13 +89,13 @@ class FunctionInfo:
 
         if name_override is not None:
             self.function_name = name_override
-        elif f.__qualname__ != f.__name__:
+        elif f.__qualname__ != f.__name__ and not serialized:
             # Class function.
-            if len(f.__qualname__.split(".")) > 2 and not serialized:
+            if len(f.__qualname__.split(".")) > 2:
                 raise InvalidError("@stub.cls classes must be defined in global scope")
             self.function_name = f"{cls.__name__}.{f.__name__}"
         else:
-            self.function_name = f.__name__
+            self.function_name = f.__qualname__
 
         self.signature = inspect.signature(f)
         module = inspect.getmodule(f)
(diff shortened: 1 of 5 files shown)


modal/_function_utils.py
```diff
@@ -86,8 +86,17 @@ class FunctionInfo:
...
         self.raw_f = f
         self.cls = cls
-        # TODO(erikbern): if f.__qualname__ != f.__name__,  we should infer the class name instead
-        self.function_name = name_override if name_override is not None else f.__qualname__
+
+        if name_override is not None:
+            self.function_name = name_override
+        elif f.__qualname__ != f.__name__:
+            # Class function.
+            if len(f.__qualname__.split(".")) > 2:
```
aksh-at, line 94: maybe we should fallback to `__qualname__` here if len > 2? I don't think anyone was legitimately using that though

modal/stub.py
```diff
@@ -609,11 +609,14 @@ def wrapper(user_cls: CLS_T) -> _Cls:
             partial_functions: Dict[str, PartialFunction] = {}
             functions: Dict[str, _Function] = {}
 
-            for k, v in user_cls.__dict__.items():
-                if isinstance(v, PartialFunction):
-                    partial_functions[k] = v
-                    partial_function = synchronizer._translate_in(v)  # TODO: remove need for?
-                    functions[k] = decorator(partial_function, user_cls)
+            for parent_cls in user_cls.mro():
```
erikbern, line 612: neat, didn't know you can do this

erikbern:
we should add a test to `container_test.py` too – i'm happy to do that if it's not straightforward. that makes sure it works on the container side (importing functions and executing them)

erikbern (approved):
great with container tests too!

PR #905, opened 2023-09-21
Serialize the ASGI protocol directly with Modal functions

ekzhang:
First step to making Python pickle-independent web endpoints.

This adds a DataFormat to function inputs and outputs, and fills in Pickle by default. We then use the new ASGI-based serialization format (via protobuf) for all web endpoint outputs.

This also adds support on the client for taking ASGI-encoded inputs, which will be useful in the future, since `data_format` has been filled in for `FunctionInput` as well.


ekzhang:
Also after our discussion today and some Googling, I saw there's an open issue for WebTransport support in ASGI:

https://github.com/django/asgiref/issues/280

Not something we want to implement right now, but that gives me more confidence that we can probably just rely on pseudo-ASGI-over-protobuf as a internal format for Modal's web endpoint handling, regardless of language in the future.

modal_proto/api.proto
```diff
@@ -90,6 +97,68 @@ enum WebhookAsyncMode {
   WEBHOOK_ASYNC_MODE_AUTO = 4; // redirect to polling endpoint if execution time nears the http timeout
 }
 
+// A web endpoint connection-related message.
```
erikbern, line 100: would it make sense to put this in a separate proto file?

modal/_container_entrypoint.py
```diff
@@ -380,27 +395,31 @@ async def complete_call(self, started_at):
...
-    async def enqueue_output(self, input_id, started_at: float, output_index: int, data):
+    async def enqueue_output(self, input_id, started_at: float, output_index: int, data: Any, data_format: int) -> None:
         await self._enqueue_output(
             input_id,
             started_at=started_at,
             gen_index=output_index,
+            data_format=data_format,
             status=api_pb2.GenericResult.GENERIC_STATUS_SUCCESS,
-            data=self.serialize(data),
+            data=self.serialize_data_format(data, data_format),
```
erikbern, line 405: this would start writing ASGI serialized format before we added support for it on the other side? i'm not quite following

erikbern:
I was confused about the deployment strategy since this current PR starts to write ASGI already but I just saw you mentioned it in the server-side PR that this will be merged in two parts. That makes more sense! I think it would make it a lot easier to review also. Let me know once this PR is broken up in two and I can take a deeper look since it's lots of code.

modal/_container_entrypoint.py
erikbern: (nvm, just saw the comment that this will be broken up)

ekzhang:
> Is the plan to break out the proto changes only or will this be broken up into multiple parts? Just thinking what's easiest to review

It's reviewable right now in full, with the server changes. There's some data dependencies.

The deployment plan is in https://github.com/modal-labs/modal/pull/7745 — first just split out the `api.proto` and `_serialization.py` changes of this PR first. Then the server changes are merged, then we cycle worker versions, then we merge this PR.

modal_proto/api.proto
ekzhang: Hmm I don't have strong opinions either way. I think it's possible, but also everything is contained in a single message right now.

modal/_asgi.py
```diff
@@ -23,8 +23,6 @@ async def receive():
 
         # Run the ASGI app, while draining the send message queue at the same time,
         # and yielding results.
-        # TODO(gongy): we currently create one ASGI instance per concurrent input,
-        # but it would be sufficient to share this background ASGI context.
```
ekzhang, line 27: I removed this comment because I think it's outdated — we only call the function once in `import_function()` to produce the ASGI app right now even for multiple inputs and regardless of input concurrency. Then this gets passed into `asgi_app_wrapper`. (cc @gongy)

modal/_container_entrypoint.py
```diff
@@ -161,6 +161,12 @@ def serialize(self, obj: Any) -> bytes:
     def deserialize(self, data: bytes) -> Any:
         return deserialize(data, self._client)
 
+    def serialize_data_format(self, obj: Any, data_format: int) -> bytes:
+        return serialize_data_format(obj, data_format)
```
ekzhang, line 165: This file contains the main client change in the PR. We call `serialize_data_format` with FunctionPutOutputs, which adds `data_format=api_pb2.DATA_FORMAT_PICKLE` to all `FunctionPutOutputsRequest` from the client.

Except for web endpoints, which we serialize with `api_pb2.DATA_FORMAT_ASGI`.

The reason why we need to cycle worker versions with the new protobuf changes before we can deploy this is that workers proxy the FunctionPutOutputs RPC, and prost deletes any unknown Protobuf fields.

modal/_container_entrypoint.py
ekzhang: Yep! The rest of this file is just plumbing for `serialize_data_format`

modal/_serialization.py
```diff
@@ -42,13 +46,142 @@ def persistent_load(self, pid):
...
     return buf.getvalue()
 
 
-def deserialize(s: bytes, client):
+def deserialize(s: bytes, client) -> Any:
     """Deserializes object and replaces all client placeholders by self."""
     return Unpickler(client, io.BytesIO(s)).load()
+
+
+def _serialize_asgi(obj: Any) -> api_pb2.Asgi:
```
ekzhang, line 61: The changes in this file add code to serialize and deserialize ASGI messages. This makes an assumption that they are well-formed. It can throw an exception if the `obj` is an invalid ASGI message, but at that point, it's a bug with whatever web framework the user is using.

modal/_serialization.py
```diff
@@ -42,13 +46,142 @@ def persistent_load(self, pid):
...
+        return api_pb2.Asgi(
+            http_response_trailers=api_pb2.Asgi.HttpResponseTrailers(
+                headers=flatten_headers(obj["headers"]),
+                more_trailers=obj.get("more_trailers"),
+            )
+        )
+    elif msg_type == "http.disconnect":
+        return api_pb2.Asgi(http_disconnect=api_pb2.Asgi.HttpDisconnect())
+    else:
+        logger.debug("skipping serialization of unknown ASGI message type %r", msg_type)
```
ekzhang, line 111: If we get a message like a WebSocket response, or an unknown ASGI extension which somehow got enabled despite us not advertising support for it, we skip the serialization.

modal/_serialization.py
```diff
@@ -42,13 +46,142 @@ def persistent_load(self, pid):
...
+    elif data_format == api_pb2.DATA_FORMAT_ASGI:
+        return _serialize_asgi(obj).SerializeToString()
+    else:
+        raise InvalidError(f"Unknown data format {data_format!r}")
+
+
+def deserialize_data_format(s: bytes, data_format: int, client) -> Any:
+    if data_format == api_pb2.DATA_FORMAT_UNSPECIFIED:
+        # TODO: Remove this after Modal client version 0.52, when the data_format field is always set.
+        return deserialize(s, client)
```
ekzhang, line 179: We need to keep this for compatibility with clients before today, since they won't be aware to set data_format. I made sure to set `data_format` in all locations and tested the code.

modal/functions.py
```diff
@@ -140,7 +141,7 @@ async def _process_result(result, stub, client=None):
         raise RemoteError(result.exception)
 
     try:
-        return deserialize(data, client)
+        return deserialize_data_format(data, data_format, client)
```
ekzhang, line 144: This is internal to the `_Invocation` class, which I updated to be data format-aware. So FunctionGetOutputs can read ASGI-serialized messages too.

ekzhang (commented):
Description of changes

modal/_serialization.py
erikbern: would it be better to raise an exception? this will just ignore it effectively (people won't see the debug log unless they override the verbosity level)

modal/_serialization.py
```diff
@@ -42,13 +46,142 @@ def persistent_load(self, pid):
...
     """Deserializes object and replaces all client placeholders by self."""
     return Unpickler(client, io.BytesIO(s)).load()
+
+
+def _serialize_asgi(obj: Any) -> api_pb2.Asgi:
+    def flatten_headers(obj):
+        return [s for k, v in obj for s in (k, v)]
+
+    msg_type = obj.get("type")
+    if msg_type == "http":
```
erikbern, line 66: are all of these covered by tests? presumably the container_test code covers most of these, but it's hard to know

erikbern (approved):
Looks good! I've been trying to think of ways this could break but there's nothing I can think of.

erikbern:
> It's hard to just review one section of the code since they depend on each other. The `api.proto` / `_serialization.py` changes won't make sense without looking at how they're used.

Not worth changing anything at this point, but would it have made sense hypothetically to merge the ASGI serialization on its own with tests checking that this part works? Not sure if it's hard to have a test that basically launches an FastAPI server and makes sure the protocol works

modal/_container_entrypoint.py
```diff
@@ -161,6 +161,12 @@ def serialize(self, obj: Any) -> bytes:
     def deserialize(self, data: bytes) -> Any:
         return deserialize(data, self._client)
 
+    def serialize_data_format(self, obj: Any, data_format: int) -> bytes:
```
aksh-at, line 164: tiny nit, should this be
```suggestion
    def serialize_data_format(self, obj: Any, data_format: api_pb2.DataFormat.ValueType) -> bytes:
```

modal/_serialization.py
aksh-at: I think there's some kinds of messages we want to ignore as unimplemented without throwing (like the [lifespan protocol](https://asgi.readthedocs.io/en/latest/specs/lifespan.html)). But yeah for `websocket.*` events might be nice to throw explicitly

aksh-at (approved):

modal/_serialization.py
ekzhang: I think we should ignore it, because there are many messages that we can't predict. The idea is "accept as many servers as possible, but require compliance on our side by users."

WebSockets just don't work on Modal, and that's fine documented as-is I think. We don't need to crash the user's app after a WebSocket request.

modal/_serialization.py
ekzhang: Okay sounds good! I'll round-trip a few requests through!

modal/_serialization.py
ekzhang: Hmm fair enough — on second thought I don't think we actually forward WebSocket connection events on our side though, so it's a bit of a moot point?

modal/_container_entrypoint.py
ekzhang: Honestly don't know how this affects mypy, can I defer it until later?

erikbern: I've had a bunch of mypy issues with `ValueType`

PR #906, opened 2023-09-21
token flow: add unique client secret

erikbern:
Since `TokenFlowCreate` and `TokenFlowWait` are unprotected endpoints, in theory a malicious user with access to the `token_flow_id` could hijack a victim's tokens if they do it quickly. This can especially happen if a token flow is successful but abandoned for some reason, in conjunction with a user leaking their `token_flow_id` (maybe sharing a link to it by mistake).

Adding a secret to the client that's used to correlate `Create` and `Wait` requests. Unlike the token flow id, the secret is never displayed (and never part of an URL) so there should be no risk of a leak (it would require network snooping, but in that case the attacker can get access in other ways).

### DIFF
--- a/client_test/token_flow_test.py
+++ b/client_test/token_flow_test.py
@@ -9,7 +9,7 @@ from modal.token_flow import TokenFlow
 @pytest.mark.asyncio
 async def test_token_flow_server(servicer, client):
     tf = TokenFlow(client)
-    async with tf.start() as (token_flow_id, _):
+    async with tf.start() as (token_flow_id, _, _):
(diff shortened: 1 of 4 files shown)


erikbern:
Also adding support for a challenge code that's the displayed client-side and needs to be entered on the frontend. This prevents a set of attacks that's even harder to exploit but roughly the vuln is – a bad person starts a token flow, then tricks a victim into visiting the token flow link and clicking "approve". Entering a code makes this harder.

Neither of these things are supported on the server side or on the frontend yet, but I'll add support for it shortly. We'll have to respect old clients for a while though.

PR #907, opened 2023-09-22
[MOD-1518] Add CLI for uploading files to volumes

gimaker:
Example usage:
```
$ MODAL_PROFILE=local modal volume put voltest large.file /foo/bar/a.file
⠼ Uploading file 'large.file' to '/foo/bar/a.file'...
✓ Uploaded file 'large.file' to '/foo/bar/a.file'
```

Which then shows up in the volume:
```
$ MODAL_PROFILE=local modal shell voltest.py
...
Spawning /bin/bash
root@modal:~# ls -l /volume/foo/bar/
total 204800
-rw-rw-r-- 1 root root 209715200 Jan  1  1970 a.file
root@modal:~#
```

(As a follow-up, we should probably set ctime based on the upload time)

Like for mounts, uploads are lazy - we don't reupload if a file with the same contents already exists:

```
$ fallocate -l 200M large.file
$ time MODAL_PROFILE=local modal volume put voltest large.file
⠧ Uploading file 'large.file' to '/large.file'...
✓ Uploaded file 'large.file' to '/large.file'

real	0m20,541s
user	0m1,417s
sys	0m0,517s
$ time MODAL_PROFILE=local modal volume put voltest large.file
⠼ Uploading file 'large.file' to '/large.file'...
✓ Uploaded file 'large.file' to '/large.file'

real	0m1,095s
user	0m0,510s
sys	0m0,098s
```

**Note that this will happily overwrite existing files and directories with impunity!** We should fix that. It's pretty easy to shoot yourself in the foot as is:

```
$ modal volume put some-dir /some-dir
$ modal volume put some-file.txt /some-dir   # note the lack of a trailing slash
# Oops! /some-dir was now wiped and replaced by a file :scream: 
```

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -1,13 +1,17 @@
 # Copyright Modal Labs 2023
 import asyncio
+import time
+from pathlib import Path, PurePosixPath
 from typing import AsyncIterator, List, Optional, Union
 
 from modal_proto import api_pb2
-from modal_utils.async_utils import asyncnullcontext, synchronize_api
+from modal_utils.async_utils import ConcurrencyPool, asyncnullcontext, synchronize_api
 from modal_utils.grpc_utils import retry_transient_errors, unary_stream
 
-from ._blob_utils import blob_iter
+from ._blob_utils import blob_iter, blob_upload_file, get_file_upload_spec
 from ._resolver import Resolver
+from .config import logger
+from .mount import MOUNT_PUT_FILE_CLIENT_TIMEOUT
 from .object import _Object, live_method, live_method_gen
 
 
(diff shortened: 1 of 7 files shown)


gimaker:
Will look into why this broke tests! Volume tests all work locally :thinking:

modal/volume.py
```diff
@@ -192,5 +196,95 @@ async def read_file(self, path: Union[str, bytes]) -> AsyncIterator[bytes]:
...
+        assert _local_path.is_dir()
+
+        def gen_transfers():
+            for subpath in _local_path.rglob("*"):
+                if subpath.is_dir():
+                    continue
+                relpath_str = subpath.relative_to(_local_path).as_posix()
+                yield self._upload_file(subpath, PurePosixPath(remote_path, relpath_str))
+
+        files = await ConcurrencyPool(20).run_coros(gen_transfers(), return_exceptions=True)
```
thundergolfer, line 246: I think we want to check if there's any exceptions in the returned list and raise an exception if there are.

thundergolfer (approved):
LGTM, with one note about error handling.

modal/volume.py
gimaker: We want to abort early on exceptions, so `return_exception=False` - except that `asyncio.gather()` seems a bit wonky, in that it doesn't cancel other the other coros on exception:

```
    async def f(x):
        await asyncio.sleep(x)
        print(x)
        if x == 2:
            raise RuntimeError("bad")

    def gen():
        for i in range(0,5):
            yield f(i)

    try:
        g = asyncio.gather(*gen())
        await g
    except Exception as e:
        g.cancel()
        print("got exception")
        await asyncio.sleep(5)
```
-->
```
1
2
got exception
3
4
```

:exploding_head: 

Upon further inspection, `ConcurrencyPool` also seems completely broken - it doesn't limit concurrency at all.

thundergolfer: > it doesn't limit concurrency at all.

😖 good to know. guess that's a yak to shave

gimaker: Fixed here https://github.com/modal-labs/modal-client/pull/907/commits/f4b10817b4fd8e0340340cffc0c5cd30fd888751

modal_utils/async_utils.py
```diff
@@ -337,10 +337,28 @@ def __init__(self, concurrency_limit: int):
...
+                    self.semaphore.release()
+                raise e
+
+        # asyncio.gather() is weird - it doesn't cancel outstanding awaitables on exceptions when
+        # return_exceptions=False --> wrap the coros in tasks are cancel them explicitly on exception.
+        tasks = [asyncio.create_task(blocking_wrapper(coro)) for coro in coros]
+        g = asyncio.gather(*tasks, return_exceptions=return_exceptions)
+        try:
+            return await g
+        except Exception as e:
```
erikbern, line 358: Might want to catch BaseException here for extra safety

gimaker:
Realized over the weekend that the API here is bad:
```
vol.add_local_file("foo.txt")
vol.add_local_file("bar.txt")
```
should really be something like:

```
vol.upload()
  .add_local_file("foo.txt")
  .add_local_file("bar.txt")
  .commit()
```

^ leads to fewer RPCs if you upload a diverse set of files, and more importantly minimizes the number of commits.

Making `add_local_file` and `add_local_dir` protected and addressing in a follow-up PR.

PR #909, opened 2023-09-22
Adds options to protobuf methods

luiscape:
We want to display human readable rpc method names and descriptions as documentation for the events tracked. These new options allow for the annotation of methods, which we can then use to generate audit log documentation.


gimaker (commented):
:confetti_ball:

PR #910, opened 2023-09-22
Annotates messages for audit logging

luiscape:
Annotates gRPC messages that need to collected by our audit logs system. I picked messages based on the mutability of objects in Modal: if an RPC mutates a Modal object (e.g. creating an App or a Volume) then we create an audit log entry.


luiscape:
@prbot approve

PR #912, opened 2023-09-22
Fix derived class inheritance when base class is in separate file

aksh-at:
Will add a test for this separately later


erikbern:
would be nice with tests but no rush. mostly because a lot of the code handling all the special cases could use some refactoring and it would be nice to have tests to make sure we're not breaking things!

PR #924, opened 2023-09-26
Bump max function input size and mount file limit size

ekzhang:
This doubles the threshold before we upload chunks to S3 for inputs / outputs, which adds quite a lot of latency right now for files that aren't that big (<2 MiB, even).

LARGE_FILE_LIMIT hasn't been updated since May 2022 (https://github.com/modal-labs/modal/pull/1610) — and I see no systems reason why it should be this low, since it's a transient transfer that we we don't store in-database on the server anyway.

If the user has a mount that has _a lot of files_ between 1 MiB and 4 MiB, this might make them use more memory when starting their mount process, but that's something we should resolve on the client side if and when it happens, with a semaphore.

### DIFF
--- a/modal/_blob_utils.py
+++ b/modal/_blob_utils.py
@@ -22,11 +22,11 @@ from modal_utils.http_utils import http_client_with_tls
 from modal_utils.logger import logger
 
 # Max size for function inputs and outputs.
-MAX_OBJECT_SIZE_BYTES = 1024 * 1024  # 1 MiB
+MAX_OBJECT_SIZE_BYTES = 2 * 1024 * 1024  # 2 MiB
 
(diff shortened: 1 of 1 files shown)


erikbern:
We might have a constraint on the server side that rejects inputs/outputs larger than 1MB. I did some quick grepping and didn't see anything, but it might be worth doing an extra check.

ekzhang:
> We might have a constraint on the server side that rejects inputs/outputs larger than 1MB. I did some quick grepping and didn't see anything, but it might be worth doing an extra check. (Or just run a script against prod to see if something breaks)

There is actually a place where we send `map` invocations, and that is hardcoded to 100 inputs at a time right now. It might exceed gRPC request size limit of 100 MiB if we bumped to 2 MiB here.

ekzhang:
I'm going to drop down `MAP_INVOCATION_CHUNK_SIZE` from 100 to 49?

PR #926, opened 2023-09-26
Revert #922

erikbern:
Had to revert #925 as a part of this as well

I'll try to add a breaking test to catch regressions as the next step. Will give it another shot after that.


erikbern:
@prbot approve

PR #928, opened 2023-09-26
Chunk web bodies to fit under output item limit

ekzhang:
This resolves MOD-1652.

### Performance Gain in Prod

For this web endpoint:

```python
from modal import Stub, web_endpoint


stub = Stub("bad7")


@stub.function(cloud="oci")
@web_endpoint()
def f(size: int = 32):
    return "a" * size
```

We have the following results before (with MAX_OBJECT_SIZE_BYTES = 1 MiB):

image

And after:

image

So returning a 1.2 MB response is almost twice as fast.

### DIFF
--- a/modal/_asgi.py
+++ b/modal/_asgi.py
@@ -23,7 +23,7 @@ def asgi_app_wrapper(asgi_app):
             if msg["type"] == "http.response.body":
                 body_chunk_size = MAX_OBJECT_SIZE_BYTES - 1024  # reserve 1 KiB for framing
                 size = len(msg.get("body", b""))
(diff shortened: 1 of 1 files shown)


erikbern:
nice, that's clever!

erikbern:
maybe we shouldn't chunk if the body is larger than some very large limit, like 100MB? in that case it's probably more efficient to go through S3

ekzhang:
> maybe we shouldn't chunk if the body is larger than some very large limit, like 100MB? in that case it's probably more efficient to go through S3

Hmm makes sense!

PR #931, opened 2023-09-27
use `_run_e2e_function` for cls tests

erikbern:
This catches the issue introduced in #922 

The actual issue is somewhat complex, but roughly what happens is that some app/stub initialization code caused `Function` objects defined locally to be overwritten with `Function` objects that are pure remote functions and don't have a reference to the local function definition. This happens because the hydration for classes overwrite any existing functions.

The reason the tests didn't previously catch this is that `_run_container` doesn't set any app objects, so the stub's objects wouldn't be overwritten. However `_run_e2e_function` uses the `deploy_stub_externally` helper, which creates all the app's objects with the servicer first. This causes the container to overwrite the methods just like in prod.

Simple way to reproduce it in prod btw:

```python
import modal
stub = modal.Stub()

@stub.cls()
class Cls:
    @modal.method()
    def g(self):
        return 42

@stub.local_entrypoint()
def run():
    return Cls().g.remote()
```

I don't love this fix, because it feels like something very foundational broke in prod and it shouldn't be so hard to test it in the client. Maybe we need a better way to run tests e2e from the client. I'm toying with the idea of maybe having a few simple "prod tests" that basically run against prod and check that basic stuff works.

### DIFF
--- a/client_test/container_test.py
+++ b/client_test/container_test.py
@@ -767,3 +767,14 @@ def test_call_function_that_calls_function(unix_servicer, event_loop):
         unix_servicer, "modal_test_support.functions", "stub", "cube", inputs=_get_inputs(((42,), {}))
     )
     assert result == 42**3
+
+
+@skip_windows_unix_socket
+def test_call_function_that_calls_method(unix_servicer, event_loop):
+    _run_e2e_function(
+        unix_servicer,
+        "modal_test_support.functions",
+        "stub",
+        "function_calling_method",
+        inputs=_get_inputs(((42, "abc", 123), {})),
+    )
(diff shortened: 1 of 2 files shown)


client_test/container_test.py
```diff
@@ -767,3 +767,14 @@ def test_call_function_that_calls_function(unix_servicer, event_loop):
         unix_servicer, "modal_test_support.functions", "stub", "cube", inputs=_get_inputs(((42,), {}))
     )
     assert result == 42**3
+
+
+@skip_windows_unix_socket
+def test_call_function_that_calls_method(unix_servicer, event_loop):
```
erikbern, line 771: this one is a test i initially thought would reproduce the error, but it doesn't. however seems useful to have.

erikbern:
@prbot approve

erikbern:
merging this since it's pure test code but @aksh-at i'm still open to any comments!

PR #932, opened 2023-09-28
Print warning about paths if `modal` is not found

erikbern:
This checks whether the `modal` binary is present on the path. If not, it outputs a warning and a link to more information (in the future we could also try to resolve the problem automatically).

This assumes the user is able to run the script using `python -m modal setup`

You can try this by deleting the `modal` binary after installation.

image

### DIFF
--- a/modal/cli/entry_point.py
+++ b/modal/cli/entry_point.py
@@ -45,20 +45,32 @@ def modal(
     pass
 
 
-def setup():
-    # Sanity check the path
+def check_path():
+    """Checks whether the `modal` executable is on the path and usable."""
+    console = Console()
+    url = "https://modal.com/docs/guide/troubleshooting#command-not-found-errors"
     try:
         subprocess.run(["modal", "--help"], capture_output=True)
         # TODO(erikbern): check returncode?
+        return
     except FileNotFoundError:
-        console = Console()
-        url = "https://modal.com/docs/guide/troubleshooting#command-not-found-errors"
         console.print(
             "[red]The `[white]modal[/white]` command was not found on your path!\n"
-            "You may need to add it to your path or use `[white]python -m modal[/white]` as a workaround.\n"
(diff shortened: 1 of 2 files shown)


modal/cli/entry_point.py
```diff
@@ -42,6 +45,25 @@ def modal(
     pass
 
 
+def setup():
+    # Sanity check the path
+    try:
+        subprocess.run(["modal", "--help"], capture_output=True)
+        # TODO(erikbern): check returncode?
```
thundergolfer, line 52: Can also do `except PermissionError: ...` to handle case where file is present but not executable.

thundergolfer (approved):

modal/cli/entry_point.py
erikbern: sure good idea

ekzhang:
Can we insert a divider before "If your browser didn't open"?

erikbern:
> Can we insert a divider before "If the web browser didn't open"?

Yeah sounds good, it's a bit confusing.

erikbern:
latest:

image

erikbern:
this isn't perfect but seems like a clear pareto improvement – but let's keep iterating

modal/cli/entry_point.py
```diff
@@ -42,6 +46,36 @@ def modal(
...
+        )
+    except PermissionError:
+        text = (
+            "[red]The `[white]modal[/white]` command is not executable!\n"
+            "You may need to give it permissions or use `[white]python -m modal[/white]` as a workaround.[/red]\n"
+        )
+    text += "See more information here:\n\n" f"[link={url}]{url}[/link]\n"
+    console = Console()
+    console.print(text)
+    console.print(Rule(style="white"))
```
ekzhang, line 69: nice!

PR #936, opened 2023-09-28
remove/simplify load_function_from_module

erikbern:
minor simplification of this code

1. it was written as a for-loop but we really only support the case where n=1 or n=2 and every other case should just throw an exception
2. it was only called from `_container_entrypoint` so i think it's easier to keep all code there

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -14,7 +14,7 @@ import sys
 import time
 import traceback
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, AsyncGenerator, AsyncIterator, Callable, Optional
+from typing import TYPE_CHECKING, Any, AsyncGenerator, AsyncIterator, Callable, Optional, Type
 
 from grpclib import Status
 from synchronicity.interface import Interface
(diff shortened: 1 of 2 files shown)


modal/_container_entrypoint.py
```diff
@@ -579,13 +579,29 @@ def import_function(function_def: api_pb2.Function, ser_cls, ser_fun, ser_params
...
+            raise LocalFunctionError("Attempted to load a function defined in a function scope")
+
+        parts = qual_name.split(".")
+        if len(parts) == 1:
+            cls = None
+            fun = getattr(module, qual_name)
+        elif len(parts) == 2:
+            cls_name, fun_name = parts
+            cls = getattr(module, cls_name)
+            fun = getattr(cls, fun_name)
```
erikbern, line 602: in particular this line can be rewritten quite a bit – `cls` should generally be a `Cls` object, so we can look up the method directly.

(this isn't always true in weird cases like if `@stub.cls` was not applied in global scope, to be strict)

erikbern:
All integration tests are passing btw

modal/_container_entrypoint.py
aksh-at: I thought `@stub.cls` not in global scope shouldn't work anyway because the function qualname will have 3 parts

aksh-at (approved):
Lgtm. `FunctionInfo` already raises on the client-side when parts > 2 so don't have to do anything there.

modal/_container_entrypoint.py
erikbern: I was talking about the case where the class is _defined_ in global scope, but wrapped in local scope, i.e.

```python
import modal

stub = modal.Stub()

class Cls:
    @modal.method()
    def f(self):
        return 123


if __name__ == "__main__":
    Cls = stub.cls()(Cls)
    with stub.run():
        print(Cls().f.remote())
```

This actually fails right now but I don't think for reasons that have to do with qualnames.

PR #937, opened 2023-09-29
glob support for 'modal volume get'

thundergolfer:


modal/cli/_download.py
```diff
@@ -0,0 +1,88 @@
...
+                            b += fp.write(chunk)
+
+                    print(f"Wrote {b} bytes to {output_path}", file=sys.stderr)
+            finally:
+                q.task_done()
+
+    tasks = []
+    tasks.append(asyncio.create_task(producer()))
+    for _ in range(10):
+        tasks.append(asyncio.create_task(consumer()))
```
erikbern, line 86: i don't think you need to use `asyncio.create_task` here right? coroutines should be enough.

if you need to use tasks, consider using `TaskContext` which makes sure all tasks are collected

thundergolfer: I didn't write this code I just extracted it out of NFS and repurposed it. 

But I'll look at it now and address your suggestions 👍

thundergolfer:
@prbot approve

PR #938, opened 2023-09-29
implementation of an inside context manager

erikbern:
~Zero tests so far~, but it took me 5 min to implement this.

Update: there are now tests for this

Usage:

```python
import modal

stub = modal.Stub("xyz")
stub.image = modal.Image.debian_slim().pip_install("scipy")


with stub.image.run_inside():
    import scipy
    print("Inside!")


@stub.function(image=stub.image)
def f():
    print("Hi from f!")


@stub.local_entrypoint()
def main():
    print("Hi!")
    f.remote()
```

Unfortunately this still requires the image to be attached to the stub. This requires a larger refactoring to how we map object ids. But eventually I want to enable this:

```python
import modal

stub = modal.Stub("xyz")
image = modal.Image.debian_slim().pip_install("scipy")


with image.run_inside():
    import scipy
    print("Inside!")


@stub.function(image=image)
def f():
    print("Hi from f!")


@stub.local_entrypoint()
def main():
    print("Hi!")
    f.remote()
```

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1,4 +1,5 @@
 # Copyright Modal Labs 2022
+import contextlib
 import os
 import shlex
 import sys
@@ -9,6 +10,7 @@ from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 import toml
+from google.protobuf.message import Message
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 
 from modal._types import typechecked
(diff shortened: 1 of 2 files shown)


erikbern:
Thinking more about this I think it's also quite easy to get rid of the requirement that the image is assigned to the stub. We can simply hydrate the function's image from the container entrypoint code. This feels like a minor QoL!

erikbern:
This is low risk since it doesn't touch any existing code. Merging this so I can write some integration tests. We don't have to move any examples over at this point – can let it simmer.

PR #941, opened 2023-09-29
handle metadata plumbing

erikbern:
Big delta but this is just a lot of plumbing (mostly updating test fixtures) – it just makes sure we populate the handle metadata on a bunch of server calls properly, and the client now _expects_ those fields to be set (rather than before when they were optional).


erikbern:
this is 95% plumbing so it should be safe to merge

erikbern:
@prbot approve

PR #942, opened 2023-09-29
Don't launch Lynx for the token flow

erikbern:
Irfan found a hilarious problem that `modal setup` would launch Lynx on his devbox. It's easy to replicate if you just `apt install lynx`. When you run `modal setup` this is what it looks like:

image

After this fix:

image

### DIFF
--- a/modal/cli/token.py
+++ b/modal/cli/token.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2022
 import getpass
 import itertools
+import os
 import webbrowser
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


erikbern:
This uses a bit of `webbrowser` internals that may not be stable across Python versions – let me see if I can fix

erikbern:
MOD-1674

erikbern:
Side note but we should probably have some sort of failure page on Modal for super old browsers

irfansharif (approved):
Thanks!

PR #943, opened 2023-10-02
split stub's app into local vs container

erikbern:
Going to try a bit of divide & conquer with the rest of the app – I think one step towards removing the `App` class is to split it up into a class for the local app (the launcher process) and one for the container. Starting with the stub code.


erikbern:
@prbot approve

PR #950, opened 2023-10-04
Amend body chunking for bodies >300 MiB

ekzhang:
Previously, #928 would re-chunk chunked bodies to 2 MiB sections, but _only_ if the chunk was less than 200 MiB. Otherwise they would get uploaded to S3 in a single big Protobuf object.

I stress tested the new web endpoint handler, and while it's better than the old one at this, our containers get a lot of RAM, so they can easily send over a big 1 GiB blob. This then takes memory to unwrap in the protobuf and so on. Not great. I tested that it's okay with a single big 1 GiB blob, but 2 GiB was too much :/

Instead, I'm going to set a separate chunk size of 300 MiB and limit body chunks to that size, even if they go through S3.


ekzhang:
Tested with `MODAL_SYNC_ENTRYPOINT=1`, seems to work.

PR #951, opened 2023-10-04
remove `_App._tag_to_object`

erikbern:
This is #922 all over again but this time after lots of test and fixing.

The point of this is to remove as much state as possible from the `_App` class so that we can remove it later (probably by splitting it first).


erikbern:
This is failing integration tests, which is concerning. Will try to add a client test that reproduces the failure and merge it separately before I attempt this again.

PR #953, opened 2023-10-04
Remove `_App._tag_to_object` (3rd attempt)

erikbern:
Basically #922 rebased, similar to #951, but contains a bugfix for an issue introduced in #951 (see #952 for a new unit test catching this).

This passes all integration tests as well.

Background for this PR: I want to reduce the role of the existing app class step by step (possibly by splitting it next). This PR lets us remove the app class, which lets us rename `Stub` to `App`.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -216,9 +216,13 @@ class _App:
             raise AttributeError(f"No such attribute `{tag}`")  # Dumb workaround for doc thing
         deprecation_error(date(2023, 8, 10), "`app.obj` is no longer supported. Use the stub to get objects instead.")
 
-    def _get_object(self, tag: str) -> Optional[_Object]:
-        # TODO(erikbern): remove objects from apps soon
-        return self._tag_to_object.get(tag)
+    def _has_object(self, tag: str) -> bool:
+        return tag in self._tag_to_object_id
+
+    def _hydrate_object(self, obj, tag: str):
+        object_id: str = self._tag_to_object_id[tag]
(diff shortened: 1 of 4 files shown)


erikbern:
Btw just a note on why this was so frustrating: the container flow for how objects are initialized is quite hard to test properly because it happens in global scope on construction of objects. We have better test coverage for it since this debacle. But fundamentally I think this code could benefit from simplification a lot, because it's clearly quite brittle.

The root cause of almost all of the complexity stems from the fact that we need to support `is_inside` for images in global scope. Replacing that with a context manager (see #938) will let us defer object initialization until the modules have been fully loaded. This will make it a lot easier to test, and will remove a bunch of edge cases. I want to work towards this, although it will take 3-6 months due to the complex deprecation path.

aksh-at (approved):
Lgtm but so did the last one, and it had an insidious bug. What's the main difference between the two?

erikbern:
> Lgtm but so did the last one, and it had an insidious bug. What's the main difference between the two?

1. This passes integration tests (very sloppy of me to not run it last time)
2. Added several unit tests that reproduces integration test issues: #931 and #935 in particular, as well as #952 (honorable mention for #930 which I also added but doesn't repro it)

aksh-at:
Yup that makes sense. I was just wondering what the "fix" for the bug was, because the two PRs look so similar at a glance.

erikbern:
The main fix was part of #953 – this commit: https://github.com/modal-labs/modal-client/pull/935/commits/d6084632a5530177cb620b6019b3fc4a6b92775e

PR #957, opened 2023-10-05
Disallow getting the app from the stub

erikbern:
This turns a warning into an error, which lets me remove the `Stub._local_app` member next.

EDIT: also ended up removing the `spawn_sandbox` method from the `App` class since it's near pointless with `stub.app` (you _could_ call it from the global `container_app`, but I doubt it's common)

EDIT 2: passes all integration tests

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -13,14 +13,12 @@ from ._output import OutputManager
 from ._resolver import Resolver
 from .client import _Client
 from .config import logger
-from .exception import InvalidError, deprecation_error, deprecation_warning
+from .exception import InvalidError, deprecation_error
 from .object import _Object
 
 if TYPE_CHECKING:
     from rich.tree import Tree
 
-    import modal.image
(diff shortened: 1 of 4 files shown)


aksh-at (approved):
Does this warrant a minor version bump?

erikbern:
> Does this warrant a minor version bump?

Yeah why not – there's been a lot of deprecation changes in the last few weeks and it might be good to keep track of old client versions. I have a lot more deprecation changes coming up, but most of them just introduce warnings, which is probably less of a reason for a version bump. I think warning -> error is a bigger reason for bumps

erikbern:
Thanks @aksh-at !

aksh-at:
> I think warning -> error is a bigger reason for bumps

Yeah exactly, so if someone pins to ~=0.53.xyz for example they won't get hit with the deprecation errors automatically.

PR #958, opened 2023-10-05
remove _App._output_mgr and _App._resolver

erikbern:
Removes two fields on the `_App` class, brings me one step closer towards removing the `_App` class


erikbern:
@prbot approve

PR #959, opened 2023-10-05
Split the App class

erikbern:
This takes the `App` class and splits it into `ContainerApp` and `LocalApp`.

There's a bit of overlapping code between the two that I simply put in both classes – that's why this PR is like +50 lines.

We should be able to optimize each class quite a bit though, basically removing both long term and merging them into the container code and the runner code. A lot of it can also go to the stub class. And a lot of the overlapping code is just deprecated code throwing the same errors. So there's plenty of opportunity to make this a negative LOC change over time.

This PR also frees up the name "App" to be used by the Stub, which is great. We should hold off making that change for a while, because of the confusion it might cause for people upgrading client versions. But it's great to get this change out as soon as possible so that clock starts ticking. I think we could probably rename Stub to App 1-2 months later.

EDIT: Passes all integration tests.

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -1,7 +1,7 @@
 # Copyright Modal Labs 2022
 from modal_version import __version__
 
-from .app import App, container_app, is_local
+from .app import container_app, is_local
 from .client import Client
 from .cls import Cls
 from .dict import Dict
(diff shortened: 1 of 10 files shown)


erikbern:
Passes all integration tests.

There's a lot of bloated code (indirection, duplication etc) already and this change exacerbates it by a fair bit. We should be able to clean up a lot of it but it might take a bit of time because of (a) supporting older clients (b) some of it depends on future refactorings i'm planning to make

aksh-at (approved):

PR #961, opened 2023-10-06
Remove ASGI scope client field if not present

ekzhang:
This is a workaround for https://github.com/django/asgiref/issues/410


ekzhang:
By the way, this means that we can't push `DATA_FORMAT_ASGI` connection scopes to web endpoints with client version before this PR was merged, since it might break WSGI apps.

PR #964, opened 2023-10-09
Fix modal shell

erikbern:
Very frustrating that this code has zero tests but let me rewrite it a bit so we can actually test it.

It works when I run this manually though.


aksh-at (approved):

thundergolfer (commented):
🙏

PR #966, opened 2023-10-11
Reduce lower threshold for S3 web responses to 40 MiB

ekzhang:
If a single chunk of body data exceeds 40 MiB, it will be forwarded via S3 instead of Redis now.

Previously, the limit was 200 MiB.


aksh-at (approved):
makes sense!

PR #967, opened 2023-10-11
local enter fixes

erikbern:
This solves two issues:

* We call `__enter__` very eagerly, in particular when the user tries to access an attribute that doesn't exist on a class – this leads to confusing errors such as no GPU available
* We don't support `__aenter__` on local function calls

It solves these by moving the responsibility of calling `__enter__` and `__aenter__` to the local call instead, i.e. unless you call a function, we don't run those locally (however, we still call the constructor).

The fix is more complex than I anticipated. Not sure how to simplify it though. We might want to unify some code for calling local functions with the container entrypoint code at some point.


erikbern:
@prbot approve

aksh-at (commented):
It seems like we're calling `__enter__` on every local call now, is that right? Do we want to instead only do that the first time it's called? Could argue that should be the expected behavior because the local `obj` is the same

PR #972, opened 2023-10-12
Refactor lookups

erikbern:
This one unifies `lookup` and `from_name` by making `lookup` a wrapper around `from_name`. In particular `lookup` was always "eager" but this PR changes the behavior such that `lookup` is now lazy. This introduces subtle changes, e.g.

1. If the object doesn't exist, the exception isn't raised until you call a method on the object (which forces the hydration)
2. Attributes such as `object_id` aren't written until later

You can call `.resolve()` on the object to force it to hydrate though.

Note that `from_name` was always lazy and wasn't possible to use previously outside of app initialization – the only use case was basically things like `stub.xyz = Dict.from_name("foo")` and then use `stub.xyz` to refer to the dict once the app was hydrated.

Why are we doing this? This is a precursor to a larger change we deprecate both `lookup` and `from_name` and replace them with two other methods. If we don't merge `lookup` and `from_name`, then we will end up with 2*2 different methods. So I think it's preferable to merge `lookup` and `from_name` even though they are slightly different – the similarities are far greater imo.

### DIFF
--- a/modal/object.py
+++ b/modal/object.py
@@ -52,6 +52,7 @@ class _Object:
         is_persisted_ref: bool = False,
         preload: Optional[Callable[[O, Resolver, Optional[str]], Awaitable[None]]] = None,
         hydrate_lazily: bool = False,
+        client: Optional[_Client] = None,
     ):
         self._local_uuid = str(uuid.uuid4())
         self._load = load
@@ -61,8 +62,8 @@ class _Object:
         self._hydrate_lazily = hydrate_lazily
 
         self._object_id = None
-        self._client = None
         self._is_hydrated = False
+        self._client = client
 
         self._initialize_from_empty()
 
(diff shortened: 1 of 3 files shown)


modal/object.py
```diff
@@ -215,7 +217,7 @@ async def resolve(self):
                 " wasn't defined in global scope."
             )
         else:
-            resolver = Resolver()
+            resolver = Resolver(client=self._client)
```
aksh-at, line 220: Looking at the args for `Resolver`, should we also supply `environment_name` here? Not 100% sure how that works

aksh-at (approved):
LGTM besides comment

modal/object.py
erikbern: Hm good question. It might not matter but let me think it through.

erikbern:
This passes all integration tests, merging it.

PR #974, opened 2023-10-13
Raise better warning when function isn't hydrated

aksh-at:
Without this you get a really confusing `NoneType has no object stub` warning. Easy to accidentally get if you import a second Modal function inside the first one, but not in global scope.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -860,6 +860,7 @@ class _Function(_Object, type_prefix="fu"):
 
         obj._raw_f = raw_f
         obj._info = info
+        obj._function_name = info.function_name
         obj._tag = tag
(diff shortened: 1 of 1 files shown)


modal/functions.py
```diff
@@ -177,6 +177,13 @@ class _OutputValue:
     value: Any
 
 
+def _raise_fn_not_hydrated(function_name: str):
+    raise InvalidError(
+        f"Function {function_name} not initialized.\n"
+        + "This can happen if you forget to import the function or cls in global scope."
```
aksh-at, line 183: Not sure if there's something better we can say here

erikbern:
It's possible https://github.com/modal-labs/modal-client/pull/975 might cover this case too. Let me add a unit test for it

PR #975, opened 2023-10-13
`_try_hydrate` -> `resolve`

erikbern:
Making this part of the public interface since we need it in #972

### DIFF
--- a/modal/object.py
+++ b/modal/object.py
@@ -211,8 +211,8 @@ class _Object:
             raise ExecutionError(
                 "Object has not been hydrated and doesn't support lazy hydration."
                 " This might happen if an object is defined on a different stub,"
-                " or if it's on the same stub but it didn't get created because it was"
-                " in a different module."
+                " or if it's on the same stub but it didn't get created because it"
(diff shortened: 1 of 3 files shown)


modal/object.py
```diff
@@ -202,13 +202,20 @@ def is_hydrated(self) -> bool:
...
-        if not self._is_hydrated and self._hydrate_lazily:
+    async def resolve(self):
+        if self._is_hydrated:
+            return
+        elif not self._hydrate_lazily:
+            raise ExecutionError(
+                "Object has not been hydrated and doesn't support lazy hydration."
+                " This might happen if an object is defined on a different stub,"
+                " or if it's on the same stub but it didn't get created because it was"
+                " in a different module."
```
aksh-at, line 213: I don't think it's immediately clear what "different module" means here. Wdyt of explicitly mentioning "if it's not imported in global scope"?

aksh-at (approved):
Nice!

modal/object.py
erikbern: sure

PR #977, opened 2023-10-14
Make is_hydrated a property

erikbern:
Broke this out of something else because it's pretty unrelated.

Made this for consistency


erikbern:
Haha just realized I started with #972, from which I broke out #974, from which I broke out #976, from which I broke out #977. So I'm 4 layers deep into this now!

PR #979, opened 2023-10-16
Error out for functions that try to create 20GB A100s

irfansharif:
This was deprecated a while ago, and was using 40GB A100s behind the scenes. Simply error out going forward.


irfansharif:
@prbot approve

PR #981, opened 2023-10-17
Run tunnel via the worker in Rust

ekzhang:
Having reliability issues with the Python implementation of the client, and it also might be nice to not depend on Python-specific implementation.

This replaces the relay client with a pair of intercepted worker RPCs: `TunnelStart` and `TunnelStop`.


ekzhang:
Tested and working.

ekzhang:
@prbot approve

thundergolfer (commented):
LGTM

PR #982, opened 2023-10-18
Expose "modal.forward" and fix reference docs

ekzhang:
This is the generated doc:

---

# modal.forward

```python
@contextlib.contextmanager
def forward(port: int) -> Iterator[Tunnel]:
```

Expose a port publicly from inside a running Modal container, with TLS.

This is an EXPERIMENTAL API and may change in the future.

**Usage:**

```python
from flask import Flask


app = Flask(__name__)

@app.route("/")
def hello_world():
    return "Hello, World!"


@stub.function()
def run_app():
    # Start a web server inside the container at port 8000. `modal.forward(8000)` lets us
    # expose that port to the world at a random HTTPS URL.
    with modal.forward(8000) as tunnel:
        print("Server listening at", tunnel.url)
        app.run("0.0.0.0", 8000)
```

### DIFF
--- a/modal/_tunnel.py
+++ b/modal/_tunnel.py
@@ -51,9 +51,9 @@ async def _forward(port: int) -> AsyncIterator[Tunnel]:
 
     @stub.function()
     def run_app():
-        # Suppose we run a web server inside the container at port 8080. `modal.forward(8080)` lets us
+        # Start a web server inside the container at port 8000. `modal.forward(8000)` lets us
         # expose that port to the world at a random HTTPS URL.
-        with modal.forward(8080) as tunnel:
+        with modal.forward(800) as tunnel:
             print("Server listening at", tunnel.url)
             app.run("0.0.0.0", 8000)
     ```
--- a/modal/_tunnel.py
+++ b/modal/_tunnel.py
@@ -53,7 +53,7 @@ async def _forward(port: int) -> AsyncIterator[Tunnel]:
     def run_app():
         # Start a web server inside the container at port 8000. `modal.forward(8000)` lets us
         # expose that port to the world at a random HTTPS URL.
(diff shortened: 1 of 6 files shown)


ekzhang:
@prbot approve

erikbern:
What's this `FunctionWithAio` thing? Why do you need it?

ekzhang:
The return type of synchronize_api on a context manager from synchronicity is FunctionWithAio, and it wasn't showing up properly in the docs since our docs generator didn't detect it as a function

I had to edit our docs generator to show the correct documentation page for this new function.

To be honest synchronicity is black box to me, was hoping you knew what FunctionWithAio was :)

erikbern:
> To be honest synchronicity is black box to me, was hoping you knew what FunctionWithAio was :)

lol ok – synchronicity has become a black box to me as well. @freider i think understands it better at this point

I was confused why this was needed since we have a lot of methods that are context managers. Not sure why that case was working but not functions in the global scope. But 🤷

modal/_tunnel.py
```diff
@@ -17,16 +17,47 @@
...
     @property
     def url(self) -> str:
+        """Get the public HTTPS URL of the forwarded port."""
         return f"https://{self.host}"
 
 
 @contextlib.asynccontextmanager
 async def _forward(port: int) -> AsyncIterator[Tunnel]:
-    """Forward a port from within a running Modal container."""
+    """Expose a port publicly from inside a running Modal container, with TLS.
```
erikbern, line 35: Btw it would be good if `_forward` can take an `Optional[_Client]` as an argument (and default to `from_env` if it's `None`) – that way it should be pretty easy to write a test for it, since you can inject a reference to the test service client. Not super important since the RPC interface is fairly trivial right now.

PR #985, opened 2023-10-19
Revert "Refactor lookups"

erikbern:
This is causing issues with classes. Reverting it for now. Will add a unit test for the bug tomorrow and try to re-apply this change.


erikbern:
@prbot approve

PR #989, opened 2023-10-19
Docs fixes

thundergolfer:
* Improve docstring for `.map`
* Introduce change to fail `inv docs` if "TODO" strings leak into our docs. I've been manually cleaning them up when I spot them and it's time to make it a build-time check. Should have very few false positives.


thundergolfer:
@prbot approve

erikbern:
LGTM

PR #990, opened 2023-10-20
hide 'resolve' method from docs

thundergolfer:


erikbern:
At some point this should be exposed to users, to force eager hydration when otherwise it would be lazy. But I can add some more documentation at that point!

PR #994, opened 2023-10-20
Support keep_warm on method and warn if it's on the cls with 2+ methods

erikbern:
Fixing a counterintuitive thing – previously if you had code like this:

```python
@stub.cls(keep_warm=2)
class Cls:
    @method()
    def f(self):
        ...

    @method()
    def g(self):
        ...
```

Then you'd end up with 4 idle containers. This change moves `keep_warm` to sit on the method instead (with a deprecation warning if you use it on the cls:

```python
@stub.cls()
class Cls:
    @method(keep_warm=2)
    def f(self):
        ...

    @method()  # this one will not have idle containers
    def g(self):
        ...
```

There are obviously many other function arguments but most of them are about the _environment_ the function is running in, which doesn't make sense to have on a method level – i think it makes it more confusing where `__enter__` is running.

However these are the ones that modify the _function_ behavior that we could potentially move too:

* `retries`
* `concurrency_limit`
* `allow_concurrent_inputs`
* `container_idle_timeout`
* `timeout`
* `interactive`

I don't have super strong feelings although some of them seem like they could make sense in either place.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -1377,10 +1377,12 @@ class _PartialFunction:
         raw_f: Callable[..., Any],
         webhook_config: Optional[api_pb2.WebhookConfig] = None,
         is_generator: Optional[bool] = None,
+        keep_warm: Optional[int] = None,
     ):
         self.raw_f = raw_f
         self.webhook_config = webhook_config
         self.is_generator = is_generator
+        self.keep_warm = keep_warm
         self.wrapped = False  # Make sure that this was converted into a FunctionHandle
 
     def __get__(self, obj, objtype=None) -> _Function:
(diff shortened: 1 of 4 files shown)


aksh-at (approved):
Nice, makes sense!

gongy (approved):
Looks good! Only a minor concern it's unergonomic for a normal, one-method class to need to set `keep_warm` on the method (makes `function` -> `cls` migration one step longer). Maybe we can only show the warning for multi-method classes?

erikbern:
Updated it so it only warns if there's 2+ methods

Can keep this as a warning for now and revisit later – it's possible we want to run all methods by the same class in the same container at some point for instance

PR #997, opened 2023-10-22
Fix function decorator in different file

aksh-at:
We already get the correct module for the function / cls, so we should use it for getting the file as well.

This is needed because if a decorator is defined in a separate file, `functools.wraps` will update `__module__` correctly, but it can't patch the `__code__` object, which is what `inspect.getfile()` uses for figuring out the file for the function.


aksh-at:
A bit annoying to add tests for this, since none of our tests cover the "run as file" branch in `FunctionInfo`. Will make a separate PR with tests for that file.

PR #998, opened 2023-10-23
`from_name` simplification

erikbern:
Should be a noop that just reduces some lines. It's a part of refactoring single-object apps.

1. Make `lookup` call `from_name`
2. Move the implementation of `_hydrate_from_app` into `from_name`


erikbern:
Passes all integration tests too

erikbern:
@prbot approve

PR #999, opened 2023-10-23
_container_entrypoint simplificaiton

erikbern:
Don't think there is any point translating clients back and forth


erikbern:
Holding off on merging this since it's breaking something @luiscape is working on

erikbern:
Luis gave me green light

PR #1000, opened 2023-10-23
pre-deprecated fields in AppCreate

erikbern:
~These are no longer needed since 0.48 is gone.~

Actually deprecated them – will remove them in a subsequent PR so they aren't part of the RPC


erikbern:
Turns out the server does rely on these fields (despite multiple comments we don't use them). Annoying.

Will see what I can do to make the best out of this situation.

PR #1006, opened 2023-10-24
inline 4 app methods into `_deploy_single_app`

erikbern:
We currently have the same code paths for deploying an app as for deploying a single object.

I'm working on getting the latter out into its own set of RPCs so that we can refactor the storage on the backend later. However it's hard to refactor the code because there's two code paths that depend on the same set of helpers.

So in order to make it easier to work with the single object case, I'm duplicating all the code so that the "regular app" case is completely separate. So this PR is noop refactor that just rewrites `_deploy_single_app` so it doesn't depend on a bunch of app helper methods:
* `_init_from_name`
* `_init_new`
* `_init_existing`
* `deploy`

It just duplicates the code and inlines it for now, but it's a starting point. This way I can rewrite the RPC flow gradually for single apps without affecting how we deploy "regular" apps.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -233,6 +233,8 @@ class _LocalApp:
         existing_object_id: Optional[str] = app._tag_to_object_id.get("_object")
         resolver = Resolver(app._client, environment_name=environment_name, app_id=app.app_id)
         await resolver.load(obj, existing_object_id)
+        if existing_object_id is not None:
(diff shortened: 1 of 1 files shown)


erikbern:
Ran all integration tests out of paranoia and they are passing. Merging this for now. It should be very low risk.

erikbern:
@prbot approve

PR #1007, opened 2023-10-25
Adds checkpoint / restore logic to container entrypoint

luiscape:
A checkpointing task will execute a function and, after its import sequence, it will call the gRPC method `ContainerCheckpoint` and enter a [busy-wait loop](https://en.wikipedia.org/wiki/Busy_waiting) waiting for the environment variable `MODAL_FUNCTION_RESTORED` to be populated. We execute the target function (or a shell) after the env var is populated.

The call to `ContainerCheckpoint` triggers the checkpointing of the function. The function is checkpointed while busy waiting, so it breaks from that loop when restored.


test/conftest.py
```diff
@@ -278,6 +279,14 @@ async def AppHeartbeat(self, stream):
         self.app_heartbeats[request.app_id] += 1
         await stream.send_message(Empty())
 
+    ### Checkpoint
```
erikbern, line 282: the methods are sorted in alphabetical order

erikbern: also the section header should be "container" probably

modal_proto/api.proto
```diff
@@ -639,8 +635,8 @@ message Function {
   // TODO: combine into enum?
   bool is_builder_function = 32;
   bool is_auto_snapshot = 38;
-
   bool is_method = 39;
+  bool is_checkpointing_function = 40;
```
erikbern, line 639: why is this a property of the function and not on `ContainerArgs`?

erikbern:
Looks good! A bit hard for me to understand the whole flow but generally code is clean and looks well tested

luiscape:
@prbot approve

test/container_test.py
```diff
@@ -92,8 +96,16 @@ def _run_container(
             # This is really only an issue for tests.
             sys.modules.pop(module_name)
 
+        env = os.environ.copy()
+        if is_checkpointing_function:
+            # Environment variable is set to allow restore from a checkpoint.
+            # Override server URL to reproduce restore behavior.
+            env["MODAL_FUNCTION_RESTORED"] = "1"
```
thundergolfer, line 103: Could be `_MODAL_FUNCTION_RESTORED` to indicate private.

luiscape: I'll drop this in favor of a state file mounted in the restore task.

PR #1008, opened 2023-10-25
Use two RPCs to deploy single-object apps

erikbern:
Unfortunately all object creation relies on an app to be present. So in order to deploy an app we actually need two phases

1. Create app, or get the app if it already exists, along with the object id
2. Create the object
3. Update the app (only needed if the app didn't exist in step 1)

Adding RPC methods for both phases. I'll implement them on the server shortly as just thin wrappers around app operations.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -195,15 +195,25 @@ message AppDeployResponse {
   string url = 1;
 }
 
+message AppPrepareSingleObjectRequest {
+  string name = 1;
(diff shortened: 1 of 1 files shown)


erikbern:
In a few months I think we basically need to shim this on the server side and send something else back instead of app ids:

1. Pseudo-app (some "made up" app id that doesn't exist)
2. Send back some other sort of deployment id using the app_id field
3. Something else

PR #1009, opened 2023-10-25
single object app proto changes

erikbern:
I completely de-scoped this project. Reverted the previous PR. Just going to focus on

1. Getting rid of the `_object` stuff from the client
2. Making it possible to have apps with the same name but different type (requires us to send the type information during lookups)

This should be quite easy to do, and just a few lines code changes on the server.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -185,7 +185,7 @@ class _LocalApp:
     ):
         # Look up any existing deployment
         app_req = api_pb2.AppGetByDeploymentNameRequest(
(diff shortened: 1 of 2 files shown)


erikbern:
Server changes needed for this:

1. In `AppLookup`, return `app_id` as well
2. In `AppSetObjects`, if `single_object_id` is present, use that for `{"_object": ...}`

Should be trivial. A minor later change we can do soon is to migrate the `{"_object": ...}` thing to its own db column

PR #1012, opened 2023-10-25
Revert "Use AppLookup for deploying objects"

erikbern:
Reverts modal-labs/modal-client#1010

This is breaking an internal test. Reverting for now.


erikbern:
@prbot approve

PR #1013, opened 2023-10-25
Remove CLIENT_TYPE_WEB_SERVER

ekzhang:
This is here for legacy internal reasons. See the SecretMake PR on the monorepo.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -33,10 +33,10 @@ enum AppState {
 enum ClientType {
   CLIENT_TYPE_UNSPECIFIED = 0;
   CLIENT_TYPE_CLIENT = 1;
-  CLIENT_TYPE_WORKER = 2; // deprecated
+  CLIENT_TYPE_WORKER = 2 [deprecated=true];
   CLIENT_TYPE_CONTAINER = 3;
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -36,7 +36,7 @@ enum ClientType {
   CLIENT_TYPE_WORKER = 2; // deprecated
   CLIENT_TYPE_CONTAINER = 3;
   CLIENT_TYPE_SERVER = 4; // deprecated
-  CLIENT_TYPE_WEB_SERVER = 5;
+  CLIENT_TYPE_WEB_SERVER = 5; // deprecated
```
erikbern, line 39: i think we can maybe use `[deprecated=true]` for these

ekzhang: Nice I didn't realize that works!

luiscape:
`CLIENT_TYPE_WEB_SERVER` is used in audit logs to indicate that a request originated from the web UI (cc @irfansharif). Are we replacing this with something else?

image

ekzhang:
Ah I thought it was just used for `service_client()`. That's my bad then.

ekzhang:
Note that the IP address that we're listing is an AWS IP address though, it just comes from the EC2 address of modal-web.

PR #1017, opened 2023-10-26
Reinstate CLIENT_TYPE_WEB_SERVER

ekzhang:


erikbern:
Why did you reinstate it? I don't think it's used right?

PR #1019, opened 2023-10-26
Break out `_deploy_single_object` to a separate module

erikbern:
No functional change, just moving code. This reduces the size of `app.py` and avoids a circular import.


erikbern:
@prbot approve

PR #1020, opened 2023-10-26
Deprecate @stub.function(secret=)

ekzhang:
Right now we have both secret= and secrets= as arguments.


ekzhang:
Hmm I don't really have the energy to deprecate this right now

PR #1022, opened 2023-10-27
Don't fail if object_id for an imported object fails

erikbern:
Resolves MOD-1801

Basically generalize `is_persisted_ref` to `is_another_app` – turns out there's multiple ways objects can be imported from other apps, and this check pertains to any such case.


aksh-at (approved):
Nice, thanks for adding the test!

PR #1023, opened 2023-10-28
Break up volume initialization for functions

erikbern:
Noop refactor other than moving validation to happen a bit earlier (on function creation not function loading). Just simplifies volume initialization a bit. Minor benefit is we can test it on Windows too.

I need to do this in order to rewrite object dependency mgmt, which is a bit of an experiment to see if I can remove stub assignments. But either way this seems like a simplification.


erikbern:
@prbot approve

gimaker (commented):
:heavy_check_mark:

PR #1025, opened 2023-10-30
Simplify object dependencies in Function and Sandbox

erikbern:
A step towards breaking out object dependencies to a separate function.

This is a continuation of #1023 and #1024 and is also a net reduction in complexity. In this PR we take advantage of how the `object_id` properties are now mutated onto the object.


erikbern:
@prbot approve

PR #1026, opened 2023-10-30
Remove the ref argument to _Image

erikbern:
Can't find anywhere this is actually used?


erikbern:
ran the integration tests out of paranoia against this – they all pass too.

merging it!

PR #1028, opened 2023-10-30
Move resolver.load out of all load closures

erikbern:
This moves the dependency management to sit inside the resolver instead (which I think we can clean up separately).

Every object that has dependencies (functions, images, sandboxes, classes) now pass a list of dependencies to `_from_loader`.


erikbern:
@prbot approve

PR #1030, opened 2023-10-31
Updates restore hook

luiscape:
A restored function will have config loaded into memory that is stale. We need to update config data in order for this function to connect to our backend. We had planned on doing this with the update of an environment variable. That approach fails because a process needs to be restarted in order for its env vars to be updated, which is undesirable. Instead, we are now creating a restore state file. A simple JSON file containing a key-value mapping for all config values that need to be updated by a restored container. 

For example:

```json
{
    "task_id": "new-task-id",
}
```

will instruct the restored container to replace the `task_id` variable from the config.


luiscape:
@prbot approve

PR #1034, opened 2023-11-02
Reliability: Backpressure the container output queue

ekzhang:
These queues were not size-limited in the past, which means they could go up to infinity.

We need to rework the request input/output lifecycle and our ad-hoc pipelining implementation to not queue up outputs infinitely, if we want to support high-performance in the future. But this is a step towards it for our current system.

### DIFF
--- a/modal/_asgi.py
+++ b/modal/_asgi.py
@@ -11,10 +11,10 @@ from ._blob_utils import MAX_OBJECT_SIZE_BYTES
 
 def asgi_app_wrapper(asgi_app):
     async def fn(scope, body=None):
-        messages_from_app: asyncio.Queue[Dict[str, Any]] = asyncio.Queue()
+        messages_from_app: asyncio.Queue[Dict[str, Any]] = asyncio.Queue(1)
 
         # TODO: send disconnect at some point.
-        messages_to_app: asyncio.Queue[Dict[str, Any]] = asyncio.Queue()
+        messages_to_app: asyncio.Queue[Dict[str, Any]] = asyncio.Queue(1)
         await messages_to_app.put({"type": "http.request", "body": body})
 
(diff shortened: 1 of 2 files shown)


erikbern:
What's the impact of this change?

ekzhang:
- `_asgi.py` — this queue had no backpressure before, and I changed the max size to 1. It doesn't need to be any higher than 1
- `_container_entrypoint.py` — the relevant change is adding 49 as the maximum size for the put outputs loop ->

Currently, if a container starts a function that executes very fast, it will create outputs very fast, and the container does not wait for its outputs to be sent to the worker before retrieving and executing more inputs. This means that we will build a backlog of infinitely many outputs on the container, slowing down requests and increasing memory usage to infinity.

This PR adds backpressure so that outputs cannot be queued infinitely. It will improve system reliability against users who run very fast maps on large amounts of inputs with short execution duration that we currently cannot handle.

ekzhang:
Also worth noting the companion PR: https://github.com/modal-labs/modal/pull/8535

This PR being merged, by itself, doesn't actually help because the inputs will then go onto another internal queue on the worker, which also has zero backpressure. So that PR fixes that issue too.

erikbern:
Thanks for the explanation. Maybe it's better to move some of this logic to the worker in the long run but I'd imagine that's a lot more work

ekzhang:
> Thanks for the explanation. Maybe it's better to move some of this logic to the worker in the long run but I'd imagine that's a lot more work

Makes sense! We should be able to remove either the container's `_enqueue_outputs` in a later refactor since the worker already has an output queue, or the worker's output queue because vice versa.

gongy (approved):
Discussed in person and this looks good to me.

PR #1037, opened 2023-11-04
Don't override patch version if specified

aksh-at:
Allow users to specify granular patch versions, such as `python_version=3.10.11`. 

Restores the default versions we had before #1036, so that all users aren't forced to rebuild their images when they update their client.


erikbern:
Thanks!

PR #1038, opened 2023-11-05
Only include explicit mounts in function deps

erikbern:
While working on getting rid of stub assignments, I ran into a problem which is that the function dependencies change between local and container. This is because a bunch of the mount stuff is inferred automagically.

Rewriting the object code a bit so that `deps` is a method not a property. This lets me add a boolean flag in the case of functions to alter the behavior to use only explicit mounts. That way I can use the same logic container-side later to hydrate the exact same objects.


erikbern:
@prbot approve

PR #1042, opened 2023-11-06
container_test.py cleanup

erikbern:
A bit of a rabbit hole from #1041. I realized it's a bit annoying to work with the container tests so I started cleaning things up. I might merge another PR in a sec too.

PR #1043, opened 2023-11-06
Remove _run_e2e_function

erikbern:
Follow up to #1042 and similar to it something that I broke out of #1041 to make it easier to fix tests. Note that this would have been even more negative delta if it wasn't for a bunch of assertions I added.

Removing this helper method makes it a bit easier to work with tests since it's more composable. Also turns out we don't need most `deploy_stub_externally`


erikbern:
@prbot appropve

erikbern:
pure test change, merging this

thundergolfer (commented):
LGTM

PR #1044, opened 2023-11-06
Remove output queue from the container

gongy:
Write outputs directly without queue indirection.

Tested on a 100k map over sleep(0), sleep(0.1) and sleep(5).

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -105,7 +105,6 @@ class _FunctionIOManager:
         self._stub_name = self.function_def.stub_name
         self._input_concurrency: Optional[int] = None
         self._semaphore: Optional[asyncio.Semaphore] = None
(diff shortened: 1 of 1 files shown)


modal/_container_entrypoint.py
```diff
@@ -489,21 +465,19 @@ def run_inputs(input_id, args, kwargs):
                         raise InvalidError(f"Generator function returned value of type {type(res)}")
 
                     for value in res:
-                        function_io_manager.enqueue_generator_value(
+                        function_io_manager.push_generator_value(
                             input_id, started_at, output_index.value, value, imp_fun.data_format
                         )
                         output_index.increase()
```
ekzhang, line 471: This part could probably be slightly more efficient as a batched request if it were a single RPC over UDS, but that's okay. We can always optimize later if it's actually an issue!

ekzhang (approved):
LGTM

PR #1047, opened 2023-11-07
Token flow: don't send redundant platform & node to server

erikbern:
We're already sending these as a part of the metadata since a while ago, so this is redundant.

I just updated the server to use the metadata rather than the payload.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1347,8 +1347,6 @@ message TaskResultRequest {
 }
 
 message TokenFlowCreateRequest {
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -1347,8 +1347,6 @@ message TaskResultRequest {
 }
 
 message TokenFlowCreateRequest {
-  string node_name = 1 [ (modal.options.audit_target_attr) = true ];
```
erikbern, line 1350: @luiscape ok to remove this?

luiscape: Yes

PR #1048, opened 2023-11-08
Adds CPU brand enums

luiscape:
Useful for determining checkpoint compatibility.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -458,6 +458,8 @@ class _FunctionIOManager:
         self._client = await _Client.from_env()
         self._waiting_for_checkpoint = False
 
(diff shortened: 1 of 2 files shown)


erikbern:
This seems like something we should do server-side? A function isn't necessarily tied to a platform, a checkpoint is. Maybe I'm confused and we can discuss more!

luiscape:
We'll populate the `Function` definition with information about available checkpoints per platform / CPU brand, i.e. this mapping

```
map<int32, CheckpointInfo> checkpoint_info = 44;
```

Roughly translates to:

```json
{
    "amd": { "status": "ready", "checksum": "..."},
    "intel": { "status": "ready", "checksum": "..."}
}
```

so the worker uses the compatible checkpoint image.

PR #1049, opened 2023-11-09
Deprecate `Stub.is_inside(...)` in favor of `Image.run_inside()`

erikbern:
I've mentioned it before but a lot of complexity in the client comes from the fact that `is_inside` needs to run in global scope – this makes object hydration much messier than it otherwise would be. `run_inside` also doesn't require images to be attached to the stub, which means we can remove stub assignments later.


aksh-at (approved):
LGTM!

PR #1054, opened 2023-11-15
All pip_install_* Image commands have flags

aksh-at:
A bit annoying that we have to replicate these keyword arguments on all functions, but it's needed so they show up in docs.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -102,6 +102,25 @@ def _flatten_str_args(function_name: str, arg_name: str, args: Tuple[Union[str, 
     return ret
 
 
+def _make_pip_install_args(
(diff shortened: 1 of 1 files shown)


aksh-at:
@prbot approve

erikbern:
:+1:

maybe we can stick those flags into a separate `options: dict[str, str | bool]` argument? and we could translate them to command line arguments mechanically by just replacing underscores with dashes etc

aksh-at:
yeah I guess if we have a mechanical way of doing it, we don't have to explain every option separately in the docs? 

It's not a huge maintenance burden to do it the current way though, and it makes it nice since users can see each option in the reference docs separately.

PR #1055, opened 2023-11-16
Add fields to AppListRequest to support filter+sorting

RohanArepally:
## Describe your changes

https://linear.app/modal-labs/issue/MOD-1864/update-apiapp-endpoint-such-that-we-can-filter-by-app-state

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -261,8 +261,17 @@ message AppStats {
   string environment_name = 11;
 }
 
+enum AppSortType {
+  DEFAULT = 0;
+  RECENT_CREATED = 1;
+  NAME = 2;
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -261,8 +261,17 @@ message AppStats {
   string environment_name = 11;
 }
 
+enum AppSortType {
```
aksh-at, line 264: We try to follow this [style guide](https://buf.build/docs/best-practices/style-guide#enums) for protobuf naming, and so according to it should leave `0` reserved for "UNSPECIFIED" and also prefix all the fields with `APP_SORT_`.

aksh-at (approved):
LGTM besides comment!

PR #1063, opened 2023-11-22
Allow switching to background-commit enabled volumes

thundergolfer:
This is a bit ugly, but expedient. 

It's not easy right now to make volume mounts configurable. I tried a few ways, most seriously in commit 1e4b593bc26a548c6b409668ed7e0b763359e708 via a `modal.VolumeMount` struct:

```python
vol_mount = modal.VolumeMount(
    "/var/data",
    volume_handle,
    allow_background_commits=True,
)

@stub.function(
    volumes={vol_mount.path: vol_mount},
    timeout=900,
)
def hello():
    subprocess.run("ls -la /var/data/", shell=True, check=True)
    return
```

But this isn't great API-wise, and it runs into issues anyway because the current `volumes` dict only wants volume objects.

In the near future it would be nice to have an API that supports mount options on the volume-level not function-level. Background commits will soon become supported by default and thus no per-volume config is needed, but we may want to make the background commit _interval_ configurable on-mount.

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -18,7 +18,7 @@ from .schedule import Cron, Period
 from .secret import Secret
 from .shared_volume import SharedVolume
 from .stub import Stub
-from .volume import Volume, VolumeMount
+from .volume import Volume
 
 __all__ = [
     "__version__",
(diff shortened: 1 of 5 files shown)


gimaker (approved):
Yeah the dict-style API is very limiting, and I'm all for refactoring it to be less limiting but that seems like a later problem.

PR #1064, opened 2023-11-22
Add client support for unencrypted tunnels

ekzhang:
Allows tunnels to forward raw TCP sockets from containers. This won't work immediately. After we merge this, I also need to merge in the server implementation.

See companion PR internally which implements this on `modal-relay` servers.


ekzhang:
@prbot approve

erikbern (approved):

PR #1069, opened 2023-11-27
Catch cloudpickle version mismatch

aksh-at:
A user ran into a cryptic issue because they had `python_version='3.10'` inside their image, and 3.11 locally. We use different and incompatible cloudpickle versions pre- and post-Python 3.11, so this causes issues. Unfortunately we have to guess that this is happening based on the exception message.


erikbern:
does cloudpickle have some sort of serialization version number? i don't think so but it would be nice if they had

aksh-at:
> does cloudpickle have some sort of serialization version number? i don't think so but it would be nice if they had

yup I mentioned this in the comment but it doesn't look like they do

PR #1073, opened 2023-11-28
Reduce flickering on green completed substeps

ekzhang:
During an application run or image build, where logs are coming dynamically, sometimes Rich has a bug where it flickers `[green]` as an output. This styling doesn't appear anyway on our substep completed character, so we can just remove it and reduce the flickering.


danielnorberg (commented):
:star:

danielnorberg (approved):

PR #1076, opened 2023-11-29
Add support for direct private registry authentication

aksh-at:
The registry [spec](https://docs.docker.com/registry/) allows you to authenticate with credentials directly or using a bearer token. So far we supported fetching these credentials/tokens from a third-party (e.g. AWS or GCP) or accessing public registries, but didn't support accessing a private registry with credentials directly.

This adds support for that and clarifies the internal enum to reflect what's being sent. Also fixes a previous issue where we had `DOCKERHUB` mapped to `0` in the protobuf enum, which is against convention. 

In the future we could consider collapsing `from_registry`, `from_aws_ecr` and `from_gcp_artifact_registry` into a single function, because we can determine the auth type based on the secrets provided. But maybe it's nice to have the separation because it makes it easier to find the docs for what you're looking for?


aksh-at:
@prbot approve since this is pretty reversible

PR #1077, opened 2023-11-30
[Mount] Adding support for .modalignore files

philipk19238:
## Describe your changes
- Adding support for .modalignore files
- Will work similar to .gitignore, where nested modalignores will stack and the syntax resembles the "wildmatch" pattern used by git
- Reference: https://python-path-specification.readthedocs.io/en/latest/readme.html

- [ x] Client+Server: this change is compatible with old servers

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -5,11 +5,12 @@ import concurrent.futures
 import dataclasses
 import functools
 import os
+import pathspec
 import time
 import typing
 from datetime import date
 from pathlib import Path, PurePosixPath
-from typing import AsyncGenerator, Callable, List, Optional, Sequence, Tuple, Union
+from typing import AsyncGenerator, Callable, Dict, Generator, List, Optional, Sequence, Tuple, Union
 
 import aiostream
 from google.protobuf.message import Message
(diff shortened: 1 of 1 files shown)


modal/mount.py
```diff
@@ -121,15 +122,38 @@ def get_files_to_upload(self):
...
-            if self.condition(local_filename):
+            if self.condition(local_filename) and self._apply_modal_ignore_rules(local_filename, ignore_rule_map):
                 local_relpath = Path(local_filename).relative_to(local_dir)
                 mount_path = self.remote_path / local_relpath.as_posix()
                 yield local_filename, mount_path.as_posix()
 
     def watch_entry(self):
         return self.local_dir, None
 
+    def _create_modal_ignore_rule_map(self, path_gen: Generator[str, None, None]) -> Dict[str, pathspec.PathSpec]:
```
aksh-at, line 136: Both these functions can go in `_mount_utils.py`

aksh-at (changes requested):
Thanks for contributing this! I think it broadly makes sense, some implementation comments.

Also would be nice to add at least a simple test. See `test/mount_test.py` for how you can do this.

mwaskom:
Closing as stale but ping if you'd like to keep working on this

PR #1089, opened 2023-12-04
[MOD-1925] modal volumes: commit on exit

thundergolfer:
### Describe your changes

Details in [MOD-1925](https://linear.app/modal-labs/issue/MOD-1925/add-commit-on-exit-functionality-for-volumefs-volumes)

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -431,6 +431,28 @@ class _FunctionIOManager:
         self._client = await _Client.from_env()
         self._waiting_for_checkpoint = False
 
+    async def volume_commit(self, volume_ids: list[str]) -> None:
+        """
+        Perform volume commit for given `volume_ids`.
+        Only used on container exit to persist uncommitted changes on behalf of user.
+        """
+        if not volume_ids:
+            return
+        results = await asyncio.gather(
+            *[
+                retry_transient_errors(
+                    self._client.stub.VolumeCommit, api_pb2.VolumeCommitRequest(volume_id=v_id), max_retries=10
+                )
+                for v_id in volume_ids
+            ],
+            return_exceptions=True,
+        )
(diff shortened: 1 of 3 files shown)


modal/_container_entrypoint.py
```diff
@@ -740,6 +762,12 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
         else:
             run_with_signal_handler(call_function_async(function_io_manager, imp_fun))
 
+        # Commit on exit to catch uncommitted volume changes and surface background
+        # commit errors.
+        function_io_manager.volume_commit(
```
gimaker, line 767: Is it possible that the user code keeps open handles to files on the volume when this code is called?

Say, if a user spawns a thread that keeps an open file handle in write mode. There's nothing preventing that, right? As is, we will not include that file in the commit (unless you explicitly fsync it first)

gimaker: To get around that, we could extend `VolumeCommitRequest` with a flag to get include everything in the snapshot though, including files with active writers.

gimaker (approved):
Not very familiar with the container entrypoint code but LGTM

modal/_container_entrypoint.py
thundergolfer: Yes that is possible. I think implementing this in such a way that no files can be left open and all dirty pages are synced involves the more challenging 'on shutdown' implementation in the worker.

The quickest fix is https://docs.python.org/3/library/os.html#os.sync 

Seems to work on gvisor. These are produced in the fs on `os.sync()` for two open files:

<kbd>
image
</kbd>

gimaker: We get `fsync`s after `os.sync()`? I did not expect that - I don't see that without gVisor. Interesting.

PR #1090, opened 2023-12-04
Fix current ID functions on concurrent inputs

ekzhang:
This is done using a context variable, which works in both sync and async contexts. Now current_input_id() and current_function_call_id() should reliably return the correct values regardless of where they are called, even when allow_concurrent_inputs is set to a value greater than 1.

TODO: Add a client test for this behavior.


gongy:
This looks good to me (very clean)

PR #1092, opened 2023-12-04
Delete modal/extensions/pymc.py

ekzhang:
This is pretty old code and has been giving me mypy errors when running locally, for a while now.

Removing because it's not used anywhere or officially supported.


erikbern:
yeah seems like the right move

ekzhang:
@prbot approve

PR #1095, opened 2023-12-04
Raise an error on Python 3.12

erikbern:
A bunch of users have run into weird tracebacks on Python 3.12. Let's just add an exception for this.

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -1,4 +1,13 @@
 # Copyright Modal Labs 2022
+import sys
+
+if sys.version_info >= (3, 12):
+    raise RuntimeError(
+        "Modal is not supported on Python 3.12 yet. See https://github.com/modal-labs/modal-client/issues/1056"
+    )
+elif sys.version_info < (3, 7):
+    raise RuntimeError("Modal requires Python 3.7 or higher")
+
 from modal_version import __version__
 
 try:


modal/__init__.py
```diff
@@ -1,4 +1,13 @@
 # Copyright Modal Labs 2022
+import sys
+
+if sys.version_info >= (3, 12):
+    raise RuntimeError(
+        "Modal is not supported on Python 3.12 yet. See https://github.com/modal-labs/modal-client/issues/1056"
+    )
+elif sys.version_info < (3, 7):
+    raise RuntimeError("Modal requires Python 3.7 or higher")
```
ekzhang, line 9: We have already set [python_requires](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires) to only allow installations on Python 3.7+, so I think this code can never execute?

ekzhang (commented):
Do you think we should just add `<3.12` to the `setup.cfg` so they can't install it instead?

erikbern:
> Do you think we should just add `<3.12` to the `setup.cfg` so they can't install it instead?

yeah maybe that's a better approach, i can do that

jcano-sl:
I'm gettingh this error on Windows 10  (Python 3.11.5)  : TypeError: bad operand type for unary -: 'NoneType'

When executing:

import datetime
import json
import psycopg2

from jsonschema2db import JSONSchemaToPostgres

schema = json.load(open('test_schema.json'))
translator = JSONSchemaToPostgres(
    schema,
    postgres_schema='schm',
    item_col_name='loan_file_id',
    item_col_type='string',
    abbreviations={
        'AbbreviateThisReallyLongColumn': 'AbbTRLC',
    }
)

con = psycopg2.connect('host=localhost dbname=postgres user=postgres password=test1234')
translator.create_tables(con)

any hint? Thanks

erikbern:
@jcano-sl I think that was a production issue that we solved and not related to 3.12

PR #1097, opened 2023-12-04
Client changes for data_in web endpoints

ekzhang:
Here are all remaining client changes needed to support streaming request bodies and WebSockets (*) on the client, at the moment. The server PR needs to be merged first, which takes into account the client version when deciding how to invoke a web endpoint.

This PR bumps the client version to **0.56**.

(*): Caveat with WebSocket support is that it won't be very efficient until we make generator outputs go through `data_out` rather than the function output system.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -49,6 +49,7 @@ class MockClientServicer(api_grpc.ModalClientBase):
     # TODO(erikbern): add more annotations
     container_inputs: list[api_pb2.FunctionGetInputsResponse]
     container_outputs: list[api_pb2.FunctionPutOutputsRequest]
+    fc_data_in: dict[str, asyncio.Queue[api_pb2.DataChunk]]
 
     def __init__(self, blob_host, blobs):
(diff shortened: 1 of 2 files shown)


modal/_asgi.py
```diff
@@ -4,18 +4,43 @@
...
+                            continue
+                        last_index = chunk.index
+                        if chunk.data_blob_id:
+                            message_bytes = await blob_download(chunk.data_blob_id, client.stub)
+                        else:
+                            message_bytes = chunk.data
+                        message = deserialize_data_format(message_bytes, chunk.data_format, client)
+                        await messages_to_app.put(message)
+                except Exception:  # TODO: Catch specific exceptions versus transient errors.
+                    await asyncio.sleep(0.1)
```
ekzhang, line 43: This can probably be moved into a helper function in the future with shared logic if, e.g., expose the data channel for users?

ekzhang:
@prbot approve

PR #1101, opened 2023-12-05
volumes: Use os.sync before committing on exit

thundergolfer:
### Describe your changes

Follow up to https://github.com/modal-labs/modal-client/pull/1089#discussion_r1414511679

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -9,6 +9,7 @@ import importlib
 import inspect
 import json
 import math
+import os
 import pickle
 import signal
 import sys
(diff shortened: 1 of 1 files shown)


modal/_container_entrypoint.py
```diff
@@ -438,6 +440,7 @@ async def volume_commit(self, volume_ids: list[str]) -> None:
         """
         if not volume_ids:
             return
+        await asyncify(os.sync)()
```
gimaker, line 443: We should do the same for explicit commits

gimaker (approved):

PR #1105, opened 2023-12-07
Add 'modal volume rm' command

thundergolfer:
## Describe your changes

Ad `modal volume rm`


danielnorberg:
👍

PR #1110, opened 2023-12-07
Client support for WebSockets in web endpoints

ekzhang:
These are all client changes needed to support WebSockets. We already have the protocol set up (in #1097), so it's just a matter of adding new enum variants.


erikbern (commented):
Nice!

PR #1112, opened 2023-12-08
perf: intro --compile option to poetry installs

thundergolfer:
## Describe your changes

Feb 27th 2023 a new Poetry version was release that **turned off** `.pyc` compilation on install: https://python-poetry.org/blog/announcing-poetry-1.4.0.

This speeds up installation a bit, and slows down the application's first run. Very bad tradeoff in our case.

This re-enables `.pyc` compilation, bringing cold start `import` perf back in line with non-poetry images. 

`sentence_transformers` package now imports in 5.1s, not 9.8s. 

**Aside:** I also introduce relaxed pinning of the Poetry version, as having not version range selection seems imprudent. 

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -659,7 +659,7 @@ class _Image(_Object, type_prefix="im"):
 
         dockerfile_commands = [
             "FROM base",
-            "RUN python -m pip install poetry",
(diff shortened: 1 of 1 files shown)


aksh-at (approved):
great find, thanks!

PR #1116, opened 2023-12-10
Track and hydrate object dependencies of functions using closurevars

erikbern:
This PR:
1. Adds a utility helper that takes a callable (a user function) and returns a list of Modal objects in global scope that it refers.
2. Tracks all such dependencies of user functions, which will cause those objects to be hydrated in the container as well

The point is to obviate stub assignments since this takes care of most/all cases where it's still needed. We don't yet support a bunch of weird cases (wrapping objects in data structures etc) but neither does the current stub assignment logic so I think it's fine.

It lets us get rid of remaining needs for stub assignments, which at this point is really just weird ones:
1. Ephemeral Dict/Queues (for other ephemeral objects, they are already referred to by functions, so there's no need for stub assignments)
4. References to objects in other apps.
5. Quasi-deployed objects by tying volumes to an app (vs using a separate single-object app). I did some analytics and this is very rare.

In case 1 and 2, this code will make it work without the assignment i.e. you can just refer to the object itself from inside the function and it will by hydrated.

For case 3, we should encourage people to use persisted objects instead.

### DIFF
--- a/modal/_function_utils.py
+++ b/modal/_function_utils.py
@@ -310,7 +310,7 @@ def get_referred_objects(f: Callable) -> List[Object]:
     for obj in inspect.getclosurevars(f).globals.values():
         if isinstance(obj, Object):
             ret.append(obj)
-        elif callable(obj):
+        elif inspect.isfunction(obj):
             ret += get_referred_objects(obj)
 
     return ret
--- a/modal/_function_utils.py
+++ b/modal/_function_utils.py
@@ -5,9 +5,10 @@ import site
 import sys
 import sysconfig
 import typing
+from collections import deque
 from enum import Enum
 from pathlib import Path, PurePosixPath
-from typing import Callable, Dict, List, Optional, Type
+from typing import Callable, Dict, List, Optional, Set, Type
 
 from modal_proto import api_pb2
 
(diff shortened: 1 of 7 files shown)


erikbern:
Passes all integration tests

modal/_function_utils.py
```diff
@@ -297,3 +298,25 @@ def get_tag(self):
...
+
+    ret: List[Object] = []
+    for obj in inspect.getclosurevars(f).globals.values():
+        if isinstance(obj, (Function, Cls)):
+            # These are always attached to stubs, so we shouldn't do anything
+            pass
+        elif isinstance(obj, Object):
+            ret.append(obj)
+        elif inspect.isfunction(obj):
+            ret += get_referred_objects(obj)
```
aksh-at, line 320: can this infinite loop with two functions that call each other?

modal/functions.py
```diff
@@ -679,6 +679,12 @@ def _deps(only_explicit_mounts=False) -> List[_Object]:
                 deps.append(nfs)
             for _, vol in validated_volumes:
                 deps.append(vol)
+
+            # Add implicit dependencies from the function's code
+            objs: list[Object] = get_referred_objects(info.raw_f)
+            _objs: list[_Object] = synchronizer._translate_in(objs)
+            deps += _objs
```
aksh-at, line 686: do these need to be deduped anywhere?

aksh-at (approved):
nice, lgtm besides comments

modal/_function_utils.py
erikbern: haha shit you're right – let me fix (and add a test)

modal/functions.py
erikbern: it's fine, as long as the list is the same locally and in the container

modal/_function_utils.py
erikbern: Fixed this by not using recursion, and keeping track of all objects seen so far.

PR #1120, opened 2023-12-11
Exposes memory checkpointing flag

luiscape:
Makes the function and class flag `checkpointing_enabled` public.


erikbern:
Nice! I think we should make it opt-out soon though, not opt-in. But maybe it's good to keep it opt-in for a period.

PR #1122, opened 2023-12-12
Increase buffer on messages_to_app

ekzhang:
This evens out RPC request load on the data channel.


aksh-at (approved):
nice, thanks!

ekzhang:
Closing because we already set the HTTP/2 initial window size on the inbound gRPC proxy, so this shouldn't really be necessary.

PR #1126, opened 2023-12-13
Turn `is_inside` into deprecation errors

erikbern:
This is part 1 of the cleanup. It will also let us clean up a bunch of code in `_container_entrypoint.py` but I'll do that separately.


erikbern:
@prbot approve

erikbern:
@aksh-at and @freider have raised some valid concern with this but supporting `is_inside` has caused me a large amount of pain and I don't think it's worth it. Removing it will help us simplify a lot of the code and get rid of stub assignments. We can revisit this in the future and discuss other potential approaches that could also work.

PR #1128, opened 2023-12-13
Deprecate gpu=True

erikbern:
We stopped supporting `gpu=True` for functions a long time ago, but we've been supporting it for `run_function` since we didn't have the ability to target a gpu anyway.

Now we have, so let's turn it into a deprecation warning. In a few months we can remove support for it altogether.

Note that we still support `gpu="any"` so we'll just tell people to change to that if they don't care about the GPU type.

### DIFF
--- a/modal/gpu.py
+++ b/modal/gpu.py
@@ -5,7 +5,7 @@ from typing import Optional, Union
 
 from modal_proto import api_pb2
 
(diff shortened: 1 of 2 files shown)


aksh-at (approved):
thanks, lgtm!

PR #1130, opened 2023-12-13
blueprint -> indexed_objects

erikbern:
This just unifies the terminology between the server and the client (and the RPC) by referring to these objects as "indexed_objects" consistently.

Zero behavioral change.

Will start to clean up crap on the stub class next.


erikbern:
@prbot approve

PR #1131, opened 2023-12-13
Deprecate kwargs to stubs

erikbern:
I think this is a fairly obscure feature that we can raise errors on quite soon


erikbern:
@prbot approve

PR #1133, opened 2023-12-14
[MOD-1980] Add skip_reload flag to VolumeCommit RPC response

gimaker:
Reloading on commit for volumes where background commits are enabled makes no sense. This allows us to avoid the reload-on-commit in those cases.

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -136,9 +136,10 @@ class _Volume(_Object, type_prefix="vo"):
             req = api_pb2.VolumeCommitRequest(volume_id=self.object_id)
             try:
                 # TODO(gongy): only apply indefinite retries on 504 status.
-                _ = await retry_transient_errors(self._client.stub.VolumeCommit, req, max_retries=90)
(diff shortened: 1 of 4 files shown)


modal/volume.py
```diff
@@ -136,9 +136,10 @@ async def commit(self):
             req = api_pb2.VolumeCommitRequest(volume_id=self.object_id)
             try:
                 # TODO(gongy): only apply indefinite retries on 504 status.
-                _ = await retry_transient_errors(self._client.stub.VolumeCommit, req, max_retries=90)
-                # Reload changes on successful commit.
-                await self._do_reload(lock=False)
+                resp = await retry_transient_errors(self._client.stub.VolumeCommit, req, max_retries=90)
+                if not resp.omit_reload:
```
thundergolfer, line 140: nit: `skip_reload`, merely because "omit" is a rarer verb in these contexts.

thundergolfer (approved):

PR #1134, opened 2023-12-15
Rename `run_inside` to `imports`

erikbern:
The reason we make this change is that it's confusing with behavior like this:

```python
with image.run_inside():
    print("hello")  # this will be printed regardless!
```

Renaming it `imports()` makes it harder to mis-use the function and run into unexpected behavior.

This adds a _pending_ deprecation warning for any usage of `run_inside` since I want to clean up all uses of `run_inside` from integration tests and examples first. After that's done, I'll go back and remove `pending=True`

As a part of the rename, I also made a minor change so that there's a warning if the block triggers any non-`ImportError` exception. We still swallow it, but I think it's helpful to warn. We probably don't want to re-raise it since sometimes import problem manifest themselves as non-`ImportError` exceptions – I have encountered that in third party libraries.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -24,7 +24,7 @@ from ._resolver import Resolver
 from ._serialization import serialize
 from .app import is_local
 from .config import config, logger
-from .exception import InvalidError, NotFoundError, RemoteError, deprecation_error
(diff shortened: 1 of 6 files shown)


aksh-at (approved):
Nice, thanks! Warning for non-`ImportError` is a good idea

PR #1139, opened 2023-12-20
Returns default value when using `Modal.Dict.get()`

luiscape:
Mimics the behavior of Python's [**dict**](https://docs.python.org/3/library/stdtypes.html?highlight=dict#dict.get). 

> Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError).

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -112,15 +112,15 @@ class _Dict(_Object, type_prefix="di"):
         await retry_transient_errors(self._client.stub.DictClear, req)
 
     @live_method
-    async def get(self, key: Any) -> Any:
+    async def get(self, key: Any, default: Optional[Any] = None) -> Any:
         """Get the value associated with a key.
 
-        Raises `KeyError` if the key does not exist.
+        Returns `default` if key does not exist. 
         """
         req = api_pb2.DictGetRequest(dict_id=self.object_id, key=serialize(key))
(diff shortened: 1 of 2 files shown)


modal/dict.py
```diff
@@ -143,7 +143,11 @@ async def __getitem__(self, key: Any) -> Any:
 
         This function only works in a synchronous context.
         """
-        return await self.get(key)
+        value = await self.get(key)
```
erikbern, line 146: this is a bit of an edge case but the user might store `None` in the dictionary and we shouldn't raise a `KeyError` on that

```
>>> d = {"foo": None}
>>> print(d["foo"])
None
```

You might want to define a custom sentinel e.g. `NOT_FOUND = object()` in the global scope and do this:

```python
value = await self.get(key, NOT_FOUND)
if value is NOT_FOUND:
    raise KeyError(...)
```

erikbern (approved):
Approved assuming we fix the behavior with `None` as a value

test/dict_test.py
```diff
@@ -19,3 +19,5 @@ def test_dict(servicer, client):
             _ = stub.d["foo"]
 
         assert stub.d.get("foo", default=True)
+        stub.d["foo"] = None
+        assert stub.d["foo"] is None
```
erikbern, line 23: 👍

PR #1149, opened 2023-12-29
Empty commit to test 2FA PyPI publishing

howard36:
Another empty commit, to verify that Github Actions can still automatically publish to PyPI, now that we have 2FA enabled.


howard36:
Resolves MOD-1852, together with #1148

PR #1162, opened 2024-01-04
Support entrypoint parameters annotated as optional

mwaskom:
Previously, we would fail to build a CLI for entrypoints with parameters annotated as `Optional[T]`. This PR improves the logic of the parsing from type annotations -> click options to handle both `Optional[T]`, and `T | None` spellings for optional arguments.

In implementing this change, I also modified some of the existing approach for handling pre/post evaluation annotations and slightly improved the error message that is generated when an annotation cannot be parsed.

- MOD-1251

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -126,7 +126,7 @@ def _get_click_command_for_function(stub: Stub, function_tag):
     if function.is_generator:
         raise InvalidError("`modal run` is not supported for generator functions")
 
-    signature: Dict[str, inspect.Parameter]
+    signature: Dict[str, ParameterMetadata]
     if function.info.cls is not None:
         cls_signature = _get_signature(function.info.cls)
         fun_signature = _get_signature(function.info.raw_f, is_method=True)
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -5,11 +5,12 @@ import inspect
 import re
 import sys
 import time
-from typing import Any, Callable, Dict, Optional, TypedDict, get_type_hints
+from typing import Any, Callable, Dict, Optional, get_type_hints
 
 import click
 import typer
 from rich.console import Console
+from typing_extensions import TypedDict
 
 from ..config import config
 from ..environments import ensure_env
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -59,6 +59,9 @@ def _get_param_type_as_str(annot: Any) -> str:
     m = re.match(r"typing.Optional\[([\w.]+)\]", annot_str)
     if m is not None:
         return m.group(1)
+    m = re.match(r"typing.Union\[([\w.]+), NoneType\]", annot_str)
+    if m is not None:
+        return m.group(1)
     m = re.match(r"([\w.]+) \| None", annot_str)
     if m is not None:
         return m.group(1)
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -3,6 +3,7 @@ import asyncio
 import datetime
 import functools
 import inspect
+import re
 import sys
 import time
 from typing import Any, Callable, Dict, Optional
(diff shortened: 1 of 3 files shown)


modal/cli/run.py
```diff
@@ -58,20 +51,39 @@ def _get_signature(f: Callable, is_method: bool = False) -> Dict[str, inspect.Pa
     return {param.name: param for param in inspect.signature(f).parameters.values()}
 
 
+def _get_param_type_as_str(annot: Any) -> str:
+    """Return annotation as a string, handling various spellings for optional types."""
+    if annot is inspect.Signature.empty:
+        return "Any"
+    annot_str = str(annot)
+    annot_patterns = [
+        r"typing.Optional\[([\w.]+)\]",
```
aksh-at, line 60: Is there ever a case where this is just `Optional`? (no `typing.`?)

aksh-at (approved):
nice, thanks for adding tests! The string matching is obviously a bit annoying, and it would be nice if we could work with the real types instead. I guess there's no way to convert string annotations to real types?

modal/cli/run.py
mwaskom: Hm good point, yeah I think an annotation with postponed evaluation where `Optional` was imported directly will end up without the `typing` prefix (whereas an evaluated annotation will have the full name regardless of how it was spelled in the code). Simple enough to make it optional (heh) in the regex.

erikbern (approved):
just a random thought but would it make sense to fix this in click directly?

but otherwise, lgtm!

mwaskom:
> The string matching is obviously a bit annoying, and it would be nice if we could work with the real types instead. I guess there's no way to convert string annotations to real types?

Yeah, agreed, but I don't think so. Even if we want to get into `eval`-ing string-typed annotations (I doubt that we do?), the object that you end up with for an `Optional` annotation is actually an instance of `typing._UnionGenericAlias` (`Optional` is just some sugar for creating a union with `None`), not an actual type, and AFAICT this class has no public API for unpacking its elements. I might be missing something (I haven't looked closely into how libraries that use annotations for validation handle optionals) but I think we're basically left working with strings.

mwaskom:
OK, reworked things a little bit further so that we are using the type hints as returned by `typing.get_type_hints` rather than the raw annotation metadata that `inspect.signature` retrieves. The upside is that `get_type_hints` evaluates the annotations and will handle spellings like `import typing as t; t.Optional[int]`. I don't think it helps us avoid string parsing altogether, but it should make it more robust.

> would it make sense to fix this in click directly?

My sense is that there isn't really anything to "fix" in click; click wants a `type`, but type hints do not themselves have to be literal types. I am curious about how libraries like Pydantic handle this because that feels like a similar usecase, but I don't my way around that codebase and it's not obvious where to find the relevant logic.

mwaskom:
Hm except the CI failures for older Python versions causes me to realize that this introduces a new potential issue: users who have `from __future__ import annotations` and then use a type hint spelling that is not supported at runtime by their Python (e.g. `X | Y`) will get an exception when we run `get_type_hints`.

It seems like our choices are:

1. Live with that edge case (maybe with a nicer error message that encourages users to write type hints that are supported by their Python runtime)
2. Fall back to the raw annotations in the case where we encounter an exception when getting type hints
3. Revert the `get_type_hints` change altogether and go back to always working with the annotation metadata, hence supporting only relatively common type hint spellings

Thoughts?

aksh-at:
That's an annoying edge case, but I think you're right it's going to be pretty rare. Right now the types people can use for entrypoint functions are far more limited so this PR is a strict improvement anyway. Both option #1 or #2 seem reasonable to deal with this.

modal/cli/run.py
mwaskom: Handled this better by evaluating the type hints

aksh-at (approved):

aksh-at:
Fixes MOD-1251

PR #1164, opened 2024-01-04
Push `from_name` and `lookup` down into two intermediary temporary classes and change behavior slightly

erikbern:
This splits the way we do lookups into two ways based on the class

1. Functions and classes – these always require an object "tag"
2. All other objects – we assume these are always single-object apps

The intermediary classes are temporary – we're giong to push everything down into each specific subclass shortly. This is just some prep work that makes it easier.

It relies on some analytics I did which indicates that there's a strict correspondence between object types and lookup methods. If you analyze all calls to `AppLookupObject`, function & class lookups _always_ have the "tag" set, but all other entities _never_ have the "tag" set. This basically ends up _enforcing_ this, and is a step towards specializing the implementation of each class (which later will allow me to remove the single-object app code from the client)

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -18,7 +18,7 @@ from .functions import (
     _Function,
     _PartialFunctionFlags,
 )
-from .object import _Object
+from .object import _AppObject
 
 T = TypeVar("T")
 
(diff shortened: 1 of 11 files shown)


erikbern:
Passes all integration tests as well.

erikbern:
@prbot approve

erikbern:
Will revert this if it's causing issues. I'll start working on the server-side implementation of `ClassGet` and `FunctionGet` next.

PR #1167, opened 2024-01-05
Attach decorator-level options to parameterized functions

gongy:
Look up and execute a parametrized function that is published by another workspace.

Syntax:

```python
modal.Cls.lookup(
  workspace="modal-labs", "vllm-inference", "Model"
).with_options(
  volumes={"/data": Volume.lookup("my-llama-weights")},
  allow_concurrent_inputs=10,
  concurrency_limit=3,
) # Attachs a api_pb2.FunctionOptions object
```

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -274,10 +274,6 @@ class _Cls(_Object, type_prefix="cs"):
         keep_warm: Optional[int] = None,
         allow_background_volume_commits: bool = False,
     ) -> "_Cls":
-        # TODO(gongy): figure out a cleaner way to clone a _Cls
-        cls = _Cls._from_other(self)
-        cls._initialize_from_other(self)
-
         retry_policy = _parse_retries(retries)
         resources = api_pb2.Resources(gpu_config=parse_gpu_config(gpu)) if gpu else None
 
@@ -291,6 +287,7 @@ class _Cls(_Object, type_prefix="cs"):
         ]
         replace_volume_mounts = len(volume_mounts) > 0
 
+        cls = _Cls._from_other(self)
         cls._options = api_pb2.FunctionOptions(
             replace_secret_ids=bool(secrets),
             secret_ids=[secret.object_id for secret in secrets],
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -287,7 +287,7 @@ class _Cls(_Object, type_prefix="cs"):
         ]
         replace_volume_mounts = len(volume_mounts) > 0
 
-        cls = _Cls._from_other(self)
+        cls = self.clone()
         cls._options = api_pb2.FunctionOptions(
             replace_secret_ids=bool(secrets),
             secret_ids=[secret.object_id for secret in secrets],
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -1,7 +1,8 @@
 # Copyright Modal Labs 2022
+import os
 import pickle
 from datetime import date
-from typing import Any, Callable, Dict, List, Optional, Type, TypeVar
+from typing import Any, Callable, Collection, Dict, List, Optional, Type, TypeVar, Union
 
 from google.protobuf.message import Message
 from grpclib import GRPCError, Status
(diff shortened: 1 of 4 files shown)


modal/cls.py
```diff
@@ -160,6 +169,11 @@ def _get_metadata(self) -> api_pb2.ClassHandleMetadata:
             )
         return class_handle_metadata
 
+    def with_options(self, **kwargs: Any) -> "_Cls":
+        self._options = _parse_function_options(**kwargs)
```
erikbern, line 173: might be better for documentation to not use **kwargs here

modal/cls.py
```diff
@@ -160,6 +169,11 @@ def _get_metadata(self) -> api_pb2.ClassHandleMetadata:
             )
         return class_handle_metadata
 
+    def with_options(self, **kwargs: Any) -> "_Cls":
```
erikbern, line 172: might be better for documentation to not use **kwargs here

modal/cls.py
erikbern, line 173: you don't want to mutate the object here right? you want to return a new object with this set

modal_proto/api.proto
```diff
@@ -685,6 +685,19 @@ message CheckpointInfo {
   string checkpoint_id = 3;
 }
 
+message FunctionOptions {
```
erikbern, line 688: can you use the same `message Function` for this?

modal/cls.py
gongy: i'm worried if I have the wrong approach

```python

    
def _initialize_from_other(self, other: "_Cls"):
    self._user_cls = other._user_cls
    self._functions = other._functions
    self._options = other._options
    self._callables = other._callables
    self._output_mgr: Optional[OutputManager] = other._output_mgr
    
 # and something like this
 
 
@classmethod
def from_existing(existing_cls):
    """Produces a new Cls from an existing Cls"""

    cls = _Cls._from_loader(
        existing_cls,
        existing_cls._load,
        existing_cls._rep,
        existing_cls._is_another_app,
        existing_cls._preload,
        existing_cls._hydrate_lazily,
        existing_cls._deps,
    )
    cls._initialize_from_existing(existing_cls)
        
 
 

```

erikbern: not following – i'm saying `with_options` should not mutate `self` but return a new object instead

luiscape:
What do you think of adopting the same syntax Docker Hub, Hugging Face Hub et al use with the namespace being determined by a identifier before a `/`?

```
```

gongy:
Definitely want to go in that direction @luiscape. Aiming to add this interface in an open-ended way that lets us move to a cleaner format later.

modal/cls.py
```diff
@@ -227,6 +252,52 @@ async def _load_remote(obj: _Object, resolver: Resolver, existing_object_id: Opt
...
+        retries: Optional[Union[int, Retries]] = None,
+        timeout: Optional[int] = None,
+        concurrency_limit: Optional[int] = None,
+        allow_concurrent_inputs: Optional[int] = None,
+        container_idle_timeout: Optional[int] = None,
+        keep_warm: Optional[int] = None,
+        allow_background_volume_commits: bool = False,
+    ) -> "_Cls":
+        # TODO(gongy): figure out a cleaner way to clone a _Cls
+        cls = _Cls._from_loader(self, self._load, self._rep)
```
gongy, line 270: This clone needs review

erikbern: yeah this seems a bit weird. I think you're assuming that `cls` is hydrated so you could just skip `_from_loader`.

You could probably add a simple clone method on the `_Object` base class that does `_Object.__new__(cls)` and then calls `_initialize_from_other`

modal/object.py
```diff
@@ -117,6 +121,17 @@ def _init_from_other(self, other: O):
         # Transient use case, see Dict, Queue, and SharedVolume
         self._init(other._rep, other._load, other._is_another_app, other._preload)
 
+    @classmethod
+    def _from_other(cls: Type[O], other: "_Object"):
```
erikbern, line 125: minor thing but you might want to add `-> O` to the signature

erikbern: actually why is this a classmethod? it probably makes more sense as an instnace method:

```python
def clone(self) -> O:
   ...

gongy:
@prbot approve

erikbern (approved):
🚢

gongy:
It inherits the options set when the class was originally wrapped.

PR #1168, opened 2024-01-05
use `ClassGet` and `FunctionGet` in the client

erikbern:
This switches class and function lookups away from `AppLookupObject` to `ClassGet` and `FunctionGet`

Not super happy with how bloated this PR is, since a lot of the code is quite repeated. I think the biggest offender is the duplication between `lookup` and `from_name`. I hope to be able to revisit that later. Unfortunately it's going to get worse in the short term since I need to add `lookup` and `from_name` to all the stateful object classes next.


erikbern:
Passes all integration tests as well

erikbern:
Btw a benefit of this is it will be easier to have specialized lookup methods, e.g. for `Cls` we might actually want a bunch of custom code that instantiates a slightly different class than when the class definition is present locally. So that's an unexpected upside that will simplify my life a bit.

erikbern:
@prbot approve

PR #1170, opened 2024-01-06
Add FunctionOptions proto message

gongy:
It will be used to modify function definitions when deriving one function from another through `with_options`.


erikbern:
Can we reuse the existing `Function` message for this?

gongy:
Not sure if @aksh-at has opinion but this message will be stored in a separate column in the Postgres function table. It's cleaner to have a stricter type, that can later be integrated into the existing `Function` message.

aksh-at:
Yeah I think it makes sense to have a separate message since this is a subset of the `Function` message. At some point we could have `Function` have this nested.

modal_proto/api.proto
```diff
@@ -369,6 +369,9 @@ message AppLookupObjectRequest {
   string object_id = 5;
   string object_entity = 6;
   string environment_name = 7;
+
+  bool lookup_published = 8; // Lookup object on app published by another workspace
```
gongy, line 373: The reason we use a bool instead of breaking into a new `NAMESPACE` is that currently we still wish apps to be deployed into a single namespace i.e. `DEPLOYMENT_NAMESPACE_WORKSPACE`. As such, we cannot simply add a `DEPLOYMENT_NAMESPACE_PUBLISHED`, for example, without weird logic around intra/cross-namespace de-duplication.

PR #1172, opened 2024-01-06
Support Sandbox.from_id

aksh-at:
Depends on https://github.com/modal-labs/modal-client/pull/1171 and https://github.com/modal-labs/modal/pull/9555

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -2,6 +2,7 @@
 import os
 from typing import Dict, List, Optional, Sequence, Union
 
+from google.protobuf.message import Message
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 
 from modal.exception import InvalidError, SandboxTerminatedError, SandboxTimeoutError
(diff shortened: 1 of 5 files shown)


aksh-at:
@prbot approve

modal/sandbox.py
```diff
@@ -158,11 +159,18 @@ async def _load(provider: _Sandbox, resolver: Resolver, _existing_object_id: Opt
 
             sandbox_id = create_resp.sandbox_id
             provider._hydrate(sandbox_id, resolver.client, None)
-            provider._stdout = LogsReader(api_pb2.FILE_DESCRIPTOR_STDOUT, sandbox_id, resolver.client)
-            provider._stderr = LogsReader(api_pb2.FILE_DESCRIPTOR_STDERR, sandbox_id, resolver.client)
 
         return _Sandbox._from_loader(_load, "Sandbox()", deps=_deps)
 
+    def _hydrate_metadata(self, handle_metadata: Optional[Message]):
```
erikbern, line 165: could be wrong but i think `Message` is always set now, so you don't have to make it `Optional` and check that it's not `None`

Not sure why it's still `Optional` in the base class. But looking at `_Function` and `_Cls` they both assume it's set

I can fix this separately

PR #1176, opened 2024-01-07
Break out partial functions into their own module

erikbern:
`functions.py` was approaching 2k lines. Broke out `PartialFunction` into a separate module to make it a bit shorter. Even after this, `functions.py` is 1,471 lines, so maybe we need to do even more!


erikbern:
Btw unrelated but I really don't like the name `PartialFunction` – let me know if you can think of anything better!

mwaskom (approved):
I’ll think a little bit about the naming question, it’s a good point. Approving the PR since we can do some renaming separately.

PR #1184, opened 2024-01-09
[mod 2017] modal shell loops spawn sh indefinitely with ASGI app functions

thundergolfer:
### Describe your changes

- MOD-2017

Another bug that can be fixed in a follow-up: having `async def` with `@asgi_app` is invalid, but it's not caught by the client. 


thundergolfer:
@prbot approve

PR #1185, opened 2024-01-09
Rename restrict_network to block_network

thecodingwizard:
Question: Should I bump the `api.proto` number, or is this not necessary?


aksh-at (approved):
> Question: Should I bump the api.proto number, or is this not necessary?

Don't have to bump it since it's just a rename

PR #1186, opened 2024-01-09
Add a changelog to the client library

ekzhang:
I'm tracking a lightweight changelog. It starts with a version compatibility table, which we update whenever we bump the minimum supported client version.

The rest of the changes are grouped by minor version. We create a new section whenever we bump the latest minor version. Each section starts with a list of paragraphs with major new features, breaking changes, or deprecations, followed by a list of individual PRs that are less important.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,81 @@
+# Changelog
+
+The Modal client library is still in pre-1.0 development. Patch releases are made on every change.
+
+We will update this changelog as we release new _minor_ versions.
+
+Sometimes breaking changes are necessary, but we will try to minimize them and provide migration guides and deprecation warnings several versions in advance of removing functionality. This gradually takes place over several months.
+
+We appreciate your patience while we speedily work towards a stable release of the client.
+
+**Historical support matrix:**
+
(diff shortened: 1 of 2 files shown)


erikbern:
I thought about this a few days ago and my gut feeling was to use the Releases feature in Github – any thoughts on that one?

@mwaskom might have some opionns on this also – he mentioned we should add a changelog the other day :)

ekzhang:
Ongoing thread with Michael [here](https://modal-com.slack.com/archives/C056CGAANRM/p1704837141010949).

> my gut feeling was to use the Releases feature in Github

Releases allow us to apply a tag and description to a Modal client version. But we would create a lot of releases. Also there's a feature to automatically generate release notes from commit messages, it's just a list of PRs though.

Putting release notes in Markdown file is simpler I think. Also don't want to write instructions on what buttons to press in GitHub UI. :')

ekzhang:
Closing in favor of proposal to use a PR template and update CHANGELOG.md for every patch release in the `publish-client` workflow: https://github.com/modal-labs/modal-client/blob/569281a2dbf5d3f408b49bd1e3cde65515c9100a/.github/workflows/ci-cd.yml#L92

PR #1188, opened 2024-01-10
Support `modal container exec`

thecodingwizard:
MOD-2064

See https://github.com/modal-labs/modal/pull/9600

## Changelog

You can now execute commands in running containers with `modal container exec [container-id] [command]`.

### DIFF
--- a/modal/cli/container.py
+++ b/modal/cli/container.py
@@ -1,10 +1,11 @@
 # Copyright Modal Labs 2022
 
 import asyncio
+import contextlib
 import os
 import select
 import sys
-from typing import List, Union
+from typing import List, Optional, Union
 
 import typer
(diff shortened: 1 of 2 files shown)


modal/cli/container.py
```diff
@@ -33,3 +43,112 @@ async def list():
...
+            return None
+        return sys.stdin.buffer.read()
+
+    async def _write():
+        while True:
+            data = await _read_stdin()
+            if data is None:
+                return
+
+            await client.stub.ContainerExecPutInput(
```
gongy, line 90: Talked about this in person but we can add network resilience here later!

gongy (approved):
Looks good to me! Nice work :)

PR #1191, opened 2024-01-10
Setup a changelog that pulls from PR description

mwaskom:
## Describe your changes

This sets up the basic infrastructure to start recording changes in the client on a per-version basis. Once this is merged, the PR template will include a "Changelog" section. For user-facing changes (new features, bug fixes, deprecations, etc.), we can add a short description in this section. Once the PR is merged, the action that triggers on a push to main will fetch the PR description and insert it into `CHANGELOG.md`. The section is optional, and the action should abort if you delete the "Changelog" section or simply leave it empty.

The changelog is currently organized with h3 headers (containing the modal version and date) for each micro release. Because the action simply prepends new information, we can revise the existing content or add new content manually if necessary (provided that we are careful to preserve the token that marks where new content should be inserted!)

Please review carefully: I developed this in a test repo and everything looked good, but we won't be able to exercise it in the modal-client repo until it's actually merged, and it could break the automatic release process if it fails on an edge case.

Resolves MOD-2054

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
-This changelog documents user-facing changes (features, fixes, and deprecations) in `modal`.
+This changelog documents user-facing updates (features, enhancements, fixes, and deprecations) to the `modal` client library. Patch releases are made on every change.
+
+The client library is still in pre-1.0 development, and sometimes breaking changes are necessary. We try to minimize them and publish deprecation warnings / migration guides in advance, typically providing a transition window of several months.
+
+We appreciate your patience while we speedily work towards a stable release of the client.
 
 <!-- NEW CONTENT GENERATED BELOW. PLEASE PRESERVE THIS COMMENT. -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+This changelog documents user-facing changes (features, fixes, and deprecations) in `modal`.
+
+<!-- NEW CONTENT GENERATED BELOW. PLEASE PRESERVE THIS COMMENT. -->
--- a/tasks.py
+++ b/tasks.py
@@ -209,7 +209,7 @@ def update_changelog(ctx):
     # Build the new changelog and write it out
     from modal_version import __version__
 
-    date = datetime.now().strftime("%Y-%m-%d")
+    date = datetime.datetime.now().strftime("%Y-%m-%d")
     new_section = f"### {__version__} ({date})\n\n{update}"
     final_content = f"{header}\n\n{new_section}\n\n{previous_changelog}"
     with open("CHANGELOG.md", "w") as fid:
--- a/tasks.py
+++ b/tasks.py
@@ -2,6 +2,7 @@
 # Copyright (c) Modal Labs 2022
 
 import inspect
+import re
 import subprocess
 
 if not hasattr(inspect, "getargspec"):
@@ -15,6 +16,7 @@ import sys
 from pathlib import Path
 from typing import Optional
 
+import requests
 from invoke import task
 
 year = datetime.date.today().year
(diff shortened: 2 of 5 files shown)


ekzhang:
We can add the changelog via entering into a box if we use a YAML issue template https://github.com/google/gvisor/blob/master/.github/ISSUE_TEMPLATE/bug_report.yml

tasks.py
```diff
@@ -15,6 +16,7 @@
 from pathlib import Path
 from typing import Optional
 
+import requests
```
ekzhang, line 19: Is requests installed?

tasks.py
```diff
@@ -155,3 +157,60 @@ def type_stubs(ctx):
...
+        break_idx = m.span()[1]
+        header = content[:break_idx]
+        previous_changelog = content[break_idx:]
+    else:
+        print("Aborting: Could not find token in existing changelog to mark insertion spot")
+
+    # Build the new changelog and write it out
+    from modal_version import __version__
+
+    date = datetime.now().strftime("%Y-%m-%d")
```
ekzhang, line 212: Maybe `%B %-d, %Y` for

## 0.55.5069 (January 10, 2024)

CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
+
+This changelog documents user-facing changes (features, fixes, and deprecations) in `modal`.
```
ekzhang, line 3: > in _the Modal client library_.

How do you feel about the copy from https://github.com/modal-labs/modal-client/blob/387410ac6c5b860cde32f3cd8f96a9a16d77646d/CHANGELOG.md — users often ask us about our version support policy so it would be nice to include a sentence?

ekzhang (approved):
nice!

tasks.py
erikbern: more of an iso 8601 fan

erikbern:
Looks good! What's the flow here, is the intent that we manually edit these changelogs as well? Or just rely on fully automated ones

CHANGELOG.md
mwaskom: Yeah good thinking! Where are you pulling the **Historical support matrix** from here? I think that @howard36 has been doing some work on defining support windows as well, I wonder how that intersects with this.

tasks.py
mwaskom: I think so, on the basis of the `setup-cached-python` step that runs before this. It worked when I copy-pasted that action source into my test repo. That said, it looks like requests is just a transitive dependency and not in on `requirements` file. Let me know if you think we need to add it there or if my understanding of the setup flow is otherwise off.

tasks.py
```diff
@@ -155,3 +157,60 @@
...
+    url = f"https://api.github.com/repos/modal-labs/modal-client/pulls/{pull_number}"
+    headers = {"Authorization": f"Bearer {os.environ['GITHUB_TOKEN']}", "Accept": "application/vnd.github.v3+json"}
+    response = requests.get(url, headers=headers).json()
+    pr_description = response.get("body")
+    if pr_description is None:
+        print("Aborting: No PR description in response from GitHub API")
+        return
+
+    # Parse the PR description to get a changelog update
+    comment_pattern = r"<!--.+?-->"
```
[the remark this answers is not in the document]
mwaskom: I think this is fine as it's just stripping out the instructions we add to the PR template but I guess an edge case would be if someone expands those in the future using a multi-line comment.

tasks.py
mwaskom: 🧑‍⚖️

mwaskom:
> Looks good! What's the flow here, is the intent that we manually edit these changelogs as well? Or just rely on fully automated ones

I think we'll want to see how things go maybe. In theory I don't think we'd need to make manual edits, but a virtue of this approach is that we _can_ if we need to (i.e. it doesn't get auto-generated top-to-bottom every release). One use-case for manual editing that I can imagine is writing up some notes for minor (0.YY) releases to aggregate the key new features.

tasks.py
mwaskom: That said I'm not totally sure how to interpret the failing mypy step, which pertains to `requests`...

mwaskom: Added `requests` and `types-requests` to the dev requirements.

CHANGELOG.md
mwaskom: Copied some of your header text over!

We might need to think a little more about the best way to programmatically sync the deprecation schedule that @howard36 implemented with the changelog, since it currently lives as a config in the `modal` repo. I've left that out for now. (I guess there's also a subtle distinction between "when will functionality be removed from the client" and "when will the server no longer support older clients").

ekzhang:
Re: syncing the deprecation schedule, it's probably fine to just manually update it. We only make changes every few months, and changing more often than that would confuse users!

PR #1196, opened 2024-01-11
add modal volume cp cli command

shreya-51:
adds modal volume cp cli command

### DIFF
--- a/modal/cli/volume.py
+++ b/modal/cli/volume.py
@@ -6,7 +6,7 @@ from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import Optional
+from typing import List, Optional
 
 from click import UsageError
 from grpclib import GRPCError, Status
@@ -270,13 +270,12 @@ async def rm(
 @synchronizer.create_blocking
 async def cp(
     volume_name: str,
-    src_path: str,
-    dst_path: str,
+    paths: List[str],  # accepts multiple paths
     env: Optional[str] = ENV_OPTION,
 ):
     ensure_env(env)
     volume = await _Volume.lookup(volume_name, environment_name=env)
     if not isinstance(volume, _Volume):
         raise UsageError("The specified app entity is not a modal.Volume")
-
-    await volume._copy_file(src_path, dst_path)
+    *src_paths, dst_path = paths
+    await volume.copy_files(src_paths, dst_path)
--- a/modal/cli/volume.py
+++ b/modal/cli/volume.py
@@ -266,11 +266,13 @@ async def rm(
         raise
 
 
-@volume_cli.command(name="cp", help="Copy a file within a volume.")
+@volume_cli.command(
+    name="cp", help="Copy source file to destination file or multiple source files to destination directory."
+)
 @synchronizer.create_blocking
 async def cp(
     volume_name: str,
-    paths: List[str],  # accepts multiple paths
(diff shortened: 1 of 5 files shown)


modal/cli/volume.py
```diff
@@ -264,3 +264,18 @@ async def rm(
         if exc.status in (Status.NOT_FOUND, Status.INVALID_ARGUMENT):
             raise UsageError(exc.message)
         raise
+
+
+@volume_cli.command(name="cp", help="Copy a file within a volume.")
```
thundergolfer, line 269: The `help` could use more detail I reckon. We also support more than just copying a single file. `man cp` overview is a good starting guide:

```
NAME
       cp - copy files and directories

SYNOPSIS
       cp [OPTION]... [-T] SOURCE DEST
       cp [OPTION]... SOURCE... DIRECTORY
       cp [OPTION]... -t DIRECTORY SOURCE...

DESCRIPTION
       Copy SOURCE to DEST, or multiple SOURCE(s) to DIRECTORY.
```

modal/cli/volume.py
```diff
@@ -264,3 +264,18 @@ async def rm(
...
+@synchronizer.create_blocking
+async def cp(
+    volume_name: str,
+    paths: List[str],  # accepts multiple paths
```
thundergolfer, line 273: The comment should include the fact that the last path is treated as the destination. That's not knowable from just the type signature and the parameter name.

modal/volume.py
```diff
@@ -341,5 +341,17 @@ async def _upload_file(
             mode=file_spec.mode,
         )
 
+    @live_method
+    async def copy_files(self, src_paths: Sequence[Union[str, bytes]], dst_path: Union[str, bytes]) -> None:
+        """
+        Copy files within the volume from src_paths to dst_path.
```
thundergolfer, line 347: As we've found, the behavior of this method's underlying RPC is non-trivial, so I reckon we should provide more info about this method's behavior in the docstring. 

We could simplify things by just including a line like "The semantics of the copy operation follow those of UNIX `cp` command."

modal/volume.py
```diff
@@ -341,5 +341,17 @@ async def _upload_file(
...
+        """
+        src_paths = [path.encode("utf-8") for path in src_paths if isinstance(path, str)]
+        if isinstance(dst_path, str):
+            dst_path = dst_path.encode("utf-8")
+
+        request = api_pb2.VolumeCopyFilesRequest(volume_id=self.object_id, src_paths=src_paths, dst_path=dst_path)
+        await retry_transient_errors(self._client.stub.VolumeCopyFiles, request, base_delay=1)
```
thundergolfer, line 354: Just thought of the idempotency of this RPC. Pretty sure it is idempotent, but we don't have a backend test for that property.

test/volume_test.py
```diff
@@ -225,3 +226,56 @@ def mount_put_file(_request):
...
+    with stub.run(client=client):
+        for file_path in file_paths:
+            local_file_path = tmp_path / file_path
+            local_file_path.write_text("test copy")
+            await stub.vol._add_local_file.aio(local_file_path, file_path)
+            object_id = stub.vol.object_id
+
+        stub.vol.copy_files(file_paths, "test_dir")
+
+    print(servicer.volume_files[object_id].keys())
```
thundergolfer, line 268: stay `print` statement

thundergolfer (approved):
LGTM after addressing comments 👍

PR #1201, opened 2024-01-12
Allow JSON log format for client logger

aksh-at:
`MODAL_LOG_FORMAT=JSON` configures the client loggers to use a JSON formatter instead of the string one.

This fixes an issue with log handling in our internal server logs (since we use `modal_utils` there).


thundergolfer (approved):
LGTM, thanks for fixing.

PR #1202, opened 2024-01-13
Add debug logging showing which files trigger a serve reload

freider:
## Changelog

`modal serve`: Setting MODAL_LOGLEVEL=DEBUG now displays which files cause an app reload during serve


aksh-at (approved):
Nice, lgtm

PR #1203, opened 2024-01-14
Unit-consistency: don't use 'memory' to mean an non-adjustable amount of GiB

thundergolfer:
~`modal.gpu.A100(memory=40)` → `modal.gpu.A100(memory_gb="40")`~

`modal.gpu.A100(memory=40)` → `modal.gpu.A100(size="40GB")`

### Describe your changes

Putting this up as a result of some discussion in `#client` Slack channel. 

Currently a `memory: int` parameter is dominantly used to provide way to specify a integer amount of MiB in some allowed range. But in the `gpu.py` module `memory: int` is used to provide a way to specify _only two possible_ integer values of _GiB_, not MiB.

This change proposes that: 

1. when measurement parameters have names that do not include the unit (e.g. `memory`, `timeout`), we should consistently use only one unit for that measurement. (And where possible, sue the SI base unit, which is seconds in the case of time.)
2. A `int` parameter shouldn't be used when the set of valid integers is very small.
    - But what should be used? A `Enum` is a bit clunky, but most correct. 

2 is more contentious than 1. 

## Changelog

- `gpu.A100` class now supports specifying GiB memory configuration using a `size: str` parameter. The `memory: int` parameter is deprecated.

### DIFF
--- a/test/gpu_test.py
+++ b/test/gpu_test.py
@@ -141,7 +141,7 @@ def test_a100_20gb_gpu_unsupported():
 
     with pytest.raises(ValueError) as err:
         stub.function(gpu=modal.gpu.A100(memory=20))(dummy)
-    assert err.value.args[0].startswith("A100 20GB is unsupported, consider A10 or A100 40GB instead")
+    assert err.value.args[0].startswith("A100 20GB is unsupported, consider")
 
 
 A100_GPU_COUNT_MAPPING = {1: api_pb2.GPU_TYPE_A100, **{i: api_pb2.GPU_TYPE_A100 for i in range(2, 5)}}
--- a/test/gpu_test.py
+++ b/test/gpu_test.py
@@ -102,26 +102,26 @@ def test_cloud_provider_selection(client, servicer):
         stub.function(cloud="foo")(dummy)
 
 
-A100_GPU_MEMORY_MAPPING = {
-    0: api_pb2.GPU_TYPE_A100,
-    40: api_pb2.GPU_TYPE_A100,
-    80: api_pb2.GPU_TYPE_A100_80GB,
-    "40": api_pb2.GPU_TYPE_A100,
-    "80": api_pb2.GPU_TYPE_A100_80GB,
-}
-
-
-@pytest.mark.parametrize("memory,gpu_type", A100_GPU_MEMORY_MAPPING.items())
-def test_memory_selection_gpu_variant(client, servicer, memory, gpu_type):
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


thundergolfer:
Gonna remove 3.7 CI/CD checks because 3.7 doesn't have https://docs.python.org/3.8/library/typing.html#typing.Literal

mwaskom:
> (Also this should arguably memory_gib, as the 40GiB card has 42.9GB)

Given that Nvidia extensively refers to these cards as "40GB" and "80GB" I think that following them is fine and that deviating would be annoying.

In fact since that's a pretty strong convention, how would you feel about just using a string type with those tokens? That could be `A100(memory="40GB")` or if you don't want to overload `memory` too much, something like `A100(size="40GB")`.

Sorry, I know we already went around on this once or twice on slack and don't feel super strongly here; but something feels awkward to me about having the unit in the parameter name. It also doesn't really feel like it effectively solves the "this isn't like other memory parameters in that it's not a request within a continuous range" issue.

luiscape:
I agree with @mwaskom's comment with preference for:

```python
A100(memory="40GB")
```

This is effectively a variant, but calling it `A100(variant="40GB")` doesn't sound right. NVIDIA refers to this as "[GPU Memory](https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/a100/pdf/a100-80gb-datasheet-update-nvidia-us-1521051-r2-web.pdf)", maybe that's an option? I would prefer one of the following two options over `A100(memory="40")`:

```python
A100(memory="40GB")
A100(gpu_memory="40GB")
```

thundergolfer:
**TODO:** Update guide once decision reached https://modal-com.slack.com/archives/C056CGAANRM/p1705441543149609?thread_ts=1705089867.066649&cid=C056CGAANRM

thundergolfer:
Lambda Cloud's terminology provides another suggestion: `A100(vram="40GB")` 😄

mwaskom (approved):

test/gpu_test.py
```diff
@@ -129,7 +140,7 @@ def test_a100_20gb_gpu_unsupported():
 
     with pytest.raises(ValueError) as err:
         stub.function(gpu=modal.gpu.A100(memory=20))(dummy)
-    assert err.value.args == ("A100 20GB is unsupported, consider A10 or A100 40GB instead",)
+    assert err.value.args[0].startswith("A100 20GB is unsupported, consider")
```
mwaskom, line 143: FYI in case you didn't know, can also pass an expected error message regex to the `match` parameter of `pytest.raises`.

PR #1204, opened 2024-01-14
Drop Python 3.7 testing from Github Actions workflows

thundergolfer:
* Python 3.7 is EOL: https://devguide.python.org/versions/
* Motivated by me hitting an annoying 3.7 incompatibility in https://github.com/modal-labs/modal-client/pull/1203


aksh-at:
Should we explicitly drop support for it in our `setup.py` too then?

thundergolfer:
Oh yes. There's also support for it in `image.py`.

PR #1205, opened 2024-01-16
Adds `volumes` argument to Stub

freider:
## Describe your changes

Similarly to `mounts` and `secrets`, this sets a baseline of volumes for all of the stub's functions

Minor incompatibility for apps using the name `volumes` as a general Modal object assigned through a stub constructor (and such assignments has been deprecated for a while afaik)

## Changelog

* modal.Stub now takes a `volumes` argument for setting the default volumes of all the stub's functions, similarly to the `mounts` and `secrets` argument.

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -1,9 +1,9 @@
 # Copyright Modal Labs 2022
 import inspect
-import os
 import typing
 import warnings
 from datetime import date
+from pathlib import PurePosixPath
 from typing import Any, AsyncGenerator, Callable, ClassVar, Dict, List, Optional, Sequence, Tuple, Union
 
 from synchronicity.async_wrap import asynccontextmanager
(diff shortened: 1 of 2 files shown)


modal/stub.py
```diff
@@ -69,6 +69,15 @@ def check_sequence(items: typing.Sequence[typing.Any], item_type: typing.Type[ty
         raise InvalidError(error_msg)
 
 
+def check_path_dict(
+    dct: typing.Dict[typing.Union[str, os.PathLike], typing.Any], item_type: typing.Type, error_msg: str
```
gimaker, line 73: Looks like this is my bad but for extra credit if you want to fix it: we should probably work with `PurePosixPath` here rather than `PathLike`. We don't want Windows paths :upside_down_face:

freider: Good point, changing this in a bunch of places. Changing all of our Pathlike usages in similar places to PurePosixPath now

PR #1206, opened 2024-01-16
Remove "--cloud" parameter from "modal nfs create"

gimaker:
It does nothing anymore (all NFS instances are created in AWS).

Also deprecates the `cloud` argument for `NetworkFileSystem.new()`.


modal/network_file_system.py
```diff
@@ -95,10 +96,11 @@ async def _load(provider: _NetworkFileSystem, resolver: Resolver, existing_objec
                 provider._hydrate(existing_object_id, resolver.client, None)
                 return
 
-            cloud_provider = parse_cloud_provider(cloud) if cloud else None
+            if cloud:
+                deprecation_warning(date(2024, 1, 13), "Argument `cloud` is deprecated (has no effect).")
```
freider, line 100: nit: Today is the 16th ;) 

Remember to double check that we aren't using the cloud argument in any integration tests before mering, otherwise they will fail due to the deprecation

freider (approved):

modal/network_file_system.py
gimaker: Updated! (The cloud argument isn't used in any integration tests on the backend)

PR #1208, opened 2024-01-16
add block_network flag to functions

thecodingwizard:
see https://github.com/modal-labs/modal/pull/9674


gongy:
Could we add it to `cls` too

gongy (approved):

PR #1216, opened 2024-01-17
[MOD 2035] automatically handle resource exhausted input rate limiting

thundergolfer:
### Describe your changes

- MOD-2035

This turned into a bit of a rabbit hole. We previously did actually retry `RESOURCE_EXHAUSTED` status codes, until Richard used that code for the "Workspace billing limit" issue (legit) and in that case `RESOURCE_EXHAUSTED` is _not_ retryable. So he removed the retrying.

The crux is that some `RESOURCE_EXHAUSTED` errors are retryable and some aren't. The client needs some way to differentiate them. The best way to do that appears to be the `details` part of a gRPCError and the accompanying `RetryInfo` message.

Here are some examples of others discussing or dealing with this issue. 

- https://groups.google.com/g/remote-execution-apis/c/drAQ6N1pNbQ (Bazel devs discussed how to tell when to retry `RESOURCE_EXHAUSTED`)
- https://github.com/open-telemetry/opentelemetry-collector/pull/5147 (Using `RetryInfo` with `RESOURCE_EXHAUSTED`
- https://grpclib.readthedocs.io/en/latest/errors.html#error-details

This PR adds the client-side changes to make `RESOURCE_EXHAUSTED` errors conditionally retryable on the presence of `RetryInfo`. 

There's also some annoying `grpclib` -> `grpcio` data munging required on the server to get this working. Please check out that PR as well (PR #9700)

    - yes, the client will get no error details and just error immediately like it currently does.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -3,6 +3,7 @@ syntax = "proto3";
 package modal.client;
 
 import "modal_proto/options.proto";
+import "google/protobuf/duration.proto";
 import "google/protobuf/empty.proto";
 import "google/protobuf/wrappers.proto";
 
@@ -1267,6 +1268,14 @@ message Resources {
   uint32 memory_mb = 2;
   uint32 milli_cpu = 3;
   GPUConfig gpu_config = 4;
+}
+
+// Used in RESOURCE_EXHAUSTED errors to differentiate
+// retryable and non-retryable version of that error status.
+// Idea taken from googleapis:
+// github.com/googleapis/googleapis/blob/8781c9414dc1543ae0198f727b4d15020eceabf6/google/rpc/error_details.proto#L91C9-L91C18
+message RetryInfo {
+  google.protobuf.Duration retry_delay = 1;
 }
 
 message Sandbox {
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1272,8 +1272,8 @@ message Resources {
 
 // Used in RESOURCE_EXHAUSTED errors to differentiate
 // retryable and non-retryable version of that error status.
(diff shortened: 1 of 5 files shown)


modal_proto/api.proto
```diff
@@ -1269,6 +1270,14 @@ message Resources {
   GPUConfig gpu_config = 4;
 }
 
+// Used in RESOURCE_EXHAUSTED errors to differentiate
+// retryable and non-retryable version of that error status.
+// Lifted from googleapis as an alternative to using the google-common-protos PyPi package.
+// ref: github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto
+message RetryInfo {
```
thundergolfer, line 1277: Added in our proto as it felt overkill to depend on `google-common-protos` just for this message.

modal_proto/api.proto
```diff
@@ -1269,6 +1270,14 @@ message Resources {
   GPUConfig gpu_config = 4;
 }
 
+// Used in RESOURCE_EXHAUSTED errors to differentiate
+// retryable and non-retryable version of that error status.
+// Lifted from googleapis as an alternative to using the google-common-protos PyPi package.
+// ref: github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto
```
luiscape, line 1276: We've added `googleapis-common-protos` to `setup.cfg` in this PR. Is that intended? If so, can you update comment?

ekzhang:
Are you sure we should automatically retry input rate limit errors? I think it's worth surfacing them to the user.

thundergolfer:
Closing as stale

PR #1217, opened 2024-01-18
Change import + mounting logic to be more similar to `python` and `python -m`

freider:
This isn't exactly 100% backwards compatible but makes the `modal` cli behave more similar to how `python` works in regard to imports and what is put on the sys.path and not.

1. When calling `modal run a/b/c.py` we now automatically put a/b on the import path, similar to how it works with `python a/b/c.py`.

2. When using MODAL_AUTOMOUNT=0 together with "module notation" of the called Modal script, e.g. `modal run pkg_a.some_module`, the mounted entrypoint file `some_module.py` will now be placed at `/root/pkg_a/some_module.py` inside the container instead of `/root/some_module.py` which was the case before. This is important since if the user then "explicitly" mounts the same package using `Mount.from_local_python_package()` it will use the same path structure. 

In addition, a bunch of unit tests were previously only testing some of the mounting code and as such misrepresented what the end result of the client mounting actually was. These tests were updated to use the MountPutFile and MountBuild rpc calls for tracking which mounts were created instead, and are now representative of the actual behavior of the client

Fixes: MOD-2104

## Changelog

* The `modal` cli now works more like the `python` cli in regard to script/module loading:
    - Running `modal my_dir/my_script.py` now puts `my_dir` on the PYTHONPATH.
    - `modal my_package.my_module` will now mount to /root/my_package/my_module.py in your Modal container, regardless if using automounting or not (and any intermediary `__init__.py` files will also be mounted)


test/mounted_files_test.py
```diff
@@ -105,18 +105,13 @@ def test_mounted_files_serialized(supports_dir, env_mount_files):
...
-    stdout = p.stdout.decode("utf-8")
-    stderr = p.stderr.decode("utf-8")
-    print("stdout: ", stdout)
-    print("stderr: ", stderr)
-    files = set(stdout.splitlines()).difference(env_mount_files)
 
+    files = set(servicer.files_name2sha.keys()) - set(env_mount_files)
     # Assert we include everything from `pkg_a` and `pkg_b` but not `pkg_c`:
     assert files == {
-        "/root/package.py",
```
freider, line 119: This file wasn't actually being mounted like this in production, probably due to some deduplication, but appeared in this test due to the test only using some parts of the mounting code. Refactoring the test to actually run the mount code e2e and checking what's being sent over the wire let me remove this incorrect assertion.

test/mounted_files_test.py
```diff
@@ -131,6 +126,17 @@ def test_mounted_files_package(supports_dir, env_mount_files):
     }
 
 
+def test_mounted_files_package_no_automount(supports_dir, env_mount_files, servicer, server_url_env, monkeypatch):
```
freider, line 129: This tests the new behavior. Previously disabling automount would put the entrypoint file at the root directory instead, regardless how it was being triggered, but now it's being placed at the correct package location (only automounting any __init__.py files on the path to the entrypoint).

This has the advantage that if the user disables automounting and then adds the root package manually, the entrypoint file will not be duplicated in a different location and risk botched imports.

test/cli_imports_test.py
```diff
@@ -102,18 +113,29 @@ def test_import_object(dir_structure, ref, expected_object_type, mock_dir):
         assert isinstance(_translated_obj, expected_object_type)
 
 
-def test_import_package_properly():
-    # https://modalbetatesters.slack.com/archives/C031Z7H15DG/p1664063245553579
-    # if importing pkg/mod.py, it should be imported as pkg.mod,
```
freider, line 107: The old comment behavior changes with this PR, as it wasn't in line with how `python pkg/mod.py` works.
If users want to keep getting the fully qualified python path of the imported module, they should use `modal run pkg.mod` instead.

freider: Sadly the betatesters link here doesn't work anymore, so I'm not sure about how this change afffects the mentioned use case

modal/cli/import_refs.py
```diff
@@ -47,27 +47,15 @@ class NoSuchObject(modal.exception.NotFoundError):
...
-        # before importing, in case of imports etc. of other modules in that package
-        # are needed
+        # when using a script path, that scripts directory should also be on the path as it is with `python some/script.py`
+        sys.path.insert(0, str(Path(file_or_module).resolve().parent))
 
-        # Let's first assume this is not part of any package
         module_name = inspect.getmodulename(file_or_module)
-
-        # Look for any __init__.py in a parent directory and maybe change the module name
-        directory = Path(file_or_module).parent
```
aksh-at, line 61: nice that all this complexity is gone!

modal/_function_utils.py
```diff
@@ -53,6 +53,21 @@ def package_mount_condition(filename):
     return os.path.splitext(filename)[1] in [".py"]
 
 
+def entrypoint_only_package_mount_condition(entrypoint_file):
```
aksh-at, line 56: If the `__init__.py` refers to other files/modules that we haven't mounted, does that cause problems?

aksh-at (approved):
Nice! Thanks for fixing the tests too

PR #1223, opened 2024-01-19
Change Mount.from_local_python_packages remote path to /root

freider:
## Describe your changes
Was previously /pkg, which differed from where auto-mounted packages were put (/root). This makes it consistent, so that explicit mounting is put at the same location as auto-mounting, which is probably what users want most of the time, e.g. when manually adding static content to packages that isn't being picked up by the auto mounting logic.

Fixes MOD-2102

If users mount packages and then refer to their location using `/pkg` this would be a breaking change once they upgrade to this new client version.

## Changelog

* `Mount.from_local_python_packages` now places mounted packages at `/root` in the Modal runtime by default (used to be `/pkg`). To override this behavior, the function now takes a `remote_dir: Union[str, PurePosixPath]` argument.

### DIFF
--- a/modal/_function_utils.py
+++ b/modal/_function_utils.py
@@ -15,10 +15,8 @@ from modal_proto import api_pb2
 from ._serialization import serialize
 from .config import config, logger
 from .exception import InvalidError
-from .mount import _Mount
+from .mount import ROOT_DIR, _Mount
 from .object import Object
-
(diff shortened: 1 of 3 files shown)


freider:
Maybe this motivates a minor version bump on the client version

aksh-at (approved):
Nice, makes sense!

PR #1226, opened 2024-01-22
Reuse `ClientSession` in `_blob_utils.py`

thecodingwizard:
Resolves MOD-2004


ekzhang:
Going to unassign you, might be better for someone with more Python experience to take a look.

PR #1230, opened 2024-01-24
Mount cache improvements

freider:
## Describe your changes

Moving most of the changeset from the old automount deprecation PR here: #645 

* Reuse `Mount.from_python_packages()` for auto-mount creation, avoiding duplicated logic
* Stub level mount deduplication to avoid creating the same mount multiple times in the same stub (with some caveats)
* Bonus: Python packages mounted using `Mount.from_local_python_packages()` (including auto-mounts) now gets a distinct look in the app creation output:

    ~~├── 🔨 Created mount /Users/elias/code/client/playground/toplevel_pak~~
    ├── 🔨 Created mount PythonPackage:toplevel_pak

## Changelog

* Mounted packages are now deduplicated across functions in the same stub
* Mounting of local Python packages are now marked as such in the mount creation output, e.g. `PythonPackage:my_package`
* Automatic mounting now includes packages outside of the function file's own directory. Mounted packages are mounted in /root/<module path>


freider:
TODO:
* Minor outstanding issue with tests on Windows, need to pass through more env vars.
* TODO Followup PR: exclude `__pycache__` in package mounts
* Add some test assertions for mount cache handling typical use cases where user might "extend" automounts by adding a custom version of the entrypoint and/or automounted packages

freider:
TODO (updated):
* See if we can move the deduplication into the resolver/load phase, in order to do it per-file (not great UX to that during module load/decoration time since it's slow)
* Make sure file watching works for multi-root Mounts

mwaskom (approved):
LGTM!

PR #1234, opened 2024-01-24
Use datachannels for task process stdio

thecodingwizard:
See https://github.com/modal-labs/modal/pull/9600#discussion_r1465288600

### DIFF
--- a/modal/cli/container.py
+++ b/modal/cli/container.py
@@ -191,8 +191,7 @@ async def handle_exec_output(client: _Client, exec_id: str, on_connect: Optional
                 exit_status = batch.exit_code
                 break
 
-            if batch.batch_index:
-                last_batch_index = batch.batch_index
+            last_batch_index = batch.batch_index
 
     while exit_status is None:
         try:
--- a/modal/cli/container.py
+++ b/modal/cli/container.py
@@ -187,7 +187,9 @@ async def handle_exec_output(client: _Client, exec_id: str, on_connect: Optional
                 connected = True
                 on_connect()
 
-            if batch.exit_code is not None:
+            # for some unclear reason, modal-worker -> modal-server is unable to transmit None for exit_code.
+            # as a temporary workaround, I just set it to i32::MIN.
+            if batch.exit_code is not None and batch.exit_code != -2_147_483_648:
                 exit_status = batch.exit_code
                 break
 
--- a/modal/cli/container.py
+++ b/modal/cli/container.py
@@ -9,6 +9,7 @@ import select
 import sys
 from typing import List, Optional, Union
 
+import rich
 import typer
 from grpclib import Status
 from grpclib.exceptions import GRPCError, StreamTerminatedError
@@ -68,13 +69,13 @@ async def exec(task_id: str, command: str):
     except GRPCError as err:
         connecting_status.stop()
         if err.status == Status.NOT_FOUND:
-            console.print(f"Container ID {task_id} not found", style="red")
+            rich.print(f"Container ID {task_id} not found", file=sys.stderr)
             raise typer.Exit(code=1)
         raise
 
     async with handle_exec_input(client, res.exec_id):
         if await handle_exec_output(client, res.exec_id, console, on_connect=connecting_status.stop):
-            console.print("Failed to establish connection to process", style="red")
+            rich.print("Failed to establish connection to process", file=sys.stderr)
             raise typer.Exit(code=1)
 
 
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -1712,15 +1712,14 @@ message RuntimeOutputMessage {
 }
 
 message RuntimeOutputBatch {
-  string entry_id = 1;
+  int32 batch_index = 1; 
   repeated RuntimeOutputMessage items = 2;
   bool eof = 3;
 }
```
luiscape, line 1718: In gRPC, it's not a good idea to overwrite fields that need to be backwards compatible. You should add a new field and deprecate the old one. And then, when backwards compatibility isn't an issue, delete the deprecated field. 

```proto
message RuntimeOutputBatch {
  string entry_id = 1 [deprecated=true];
  repeated RuntimeOutputMessage items = 2;
  bool eof = 3;
  int32 batch_index = 4; 
}
```

modal_proto/api.proto
```diff
@@ -1712,15 +1713,16 @@ message RuntimeOutputMessage {
...
 
 message RuntimeOutputBatch {
-  string entry_id = 1;
+  string entry_id = 1 [deprecated=true];
```
ekzhang, line 1716: We don't need to deprecate these fields since there's no one actively using them! You can just remove them :)

modal/cli/container.py
```diff
@@ -54,12 +55,15 @@ async def exec(task_id: str, command: str):
...
-    if res.exec_id == "":
-        print(f"Failed to execute command. Is the container ID ({task_id}) correct?")
-        return
+    try:
+        res: api_pb2.ContainerExecResponse = await client.stub.ContainerExec(
+            api_pb2.ContainerExecRequest(task_id=task_id, command=command, pty_info=get_pty_info(shell=True))
+        )
+    except GRPCError as err:
+        if err.status == Status.NOT_FOUND:
+            print(f"Container ID {task_id} not found!")
```
ekzhang, line 64: Shouldn't this print to stderr?

Also might be better to remove the exclamatory `!`, and return a nonzero exit code

ekzhang (approved):
✔️ Nice

modal_proto/api.proto
ekzhang: I saw @luiscape's comment above

modal/cli/container.py
thecodingwizard: I used rich.Console, printed it in red, removed the exclamation mark, and returned a nonzero exit code. (I didn't print to stderr though, only stdout -- is it important to print to stderr?)

ekzhang: Yeah, I think stderr is better, since people often use stdout for piping.

You can do `rich.print("...", file=sys.stderr)`

ekzhang (approved):

modal/cli/container.py
```diff
@@ -122,33 +170,52 @@ def try_write():
...
 
-    while not completed:
+    while exit_status is None:
         try:
-            await _get_output()
+            if not connected:
+                try:
+                    output_task = asyncio.create_task(_get_output())
+                    await asyncio.wait_for(asyncio.shield(output_task), timeout=FIRST_OUTPUT_TIMEOUT)
+                except asyncio.TimeoutError or TimeoutError:
```
ekzhang, line 205:
```suggestion
                except (asyncio.TimeoutError, TimeoutError):
```

modal/cli/container.py
```diff
@@ -122,33 +170,52 @@ def try_write():
...
                 await write_data(message.file_descriptor, str.encode(message.message))
 
-            if batch.eof:
-                completed = True
+            if not connected:
+                connected = True
+                on_connect()
+
+            # for some unclear reason, modal-worker -> modal-server is unable to transmit None for exit_code.
+            # as a temporary workaround, I just set it to i32::MIN.
```
ekzhang, line 191: In Python, you can check presence with `batch.HasField("exit_code")`. The value will never be `None`.

thecodingwizard: oops, that's good to know, thanks!

modal/cli/container.py
```diff
@@ -122,33 +170,50 @@ def try_write():
...
+                on_connect()
+
+            if batch.HasField("exit_code"):
+                exit_status = batch.exit_code
                 break
 
-            if batch.entry_id:
-                last_entry_id = batch.entry_id
+            if batch.batch_index:
+                last_batch_index = batch.batch_index
```
thecodingwizard, line 195: todo: I don't think this will ever not be set, so the if statement seems unnecessary.

ekzhang (approved):
Like I said it looks good — but the last change you quickly made seems to have failed type checking, `style="red"` doesn't work here.

PR #1235, opened 2024-01-24
Enhance `modal profile list` to show each profile's workspace

mwaskom:
## Describe your changes

- Resolves MOD-2154

## Changelog

The `modal profile list` command now shows the workspace associated with each profile.

### DIFF
--- a/modal/cli/profile.py
+++ b/modal/cli/profile.py
@@ -3,11 +3,14 @@
 from typing import Optional
 
 import typer
+from rich.console import Console
+from rich.table import Table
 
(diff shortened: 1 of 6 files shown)


mwaskom:
Ah hum it occurs to me that we'll want the new protos and new feature code in separate PRs; consider this a PoC and I'll split up the commit.

mwaskom:
Also not clear to me whether we have a fixture that ensures the `modal.toml` in the test environment has a profile in it which is causing my test to fail

PR #1238, opened 2024-01-24
Enhance `modal profile list` to show each profile's workspace

mwaskom:
## Describe your changes

This PR improves the information shown in `modal profile list` to include the workspace that each profile corresponds to:

image

As previously, it indicates the currently active profile with a green highlight.

Very open to feedback on the formatting for the "active" profile and the error handling when we can't lookup the workspace.

- Resolves MOD-2154

## Changelog

The `modal profile list` command now shows the workspace associated with each profile.


mwaskom:
Oops, a TODO here is that we no longer support `modal profile list --json`. Missed that at first (I guess we don't have tests for it...)

aksh-at (approved):
lgtm

PR #1245, opened 2024-01-25
Send generator output through the fc data channel

ekzhang:
This change sends generator outputs through the function call data channel.

This deprecates `GenericResult.gen_status` and `FunctionPutOutputsItem.gen_index`. Instead, generator outputs are now sent on the data channel for that function call, and a single numeric value is returned on the function call data channel after all items are complete. This value is the total number of generator items.

This works on its own, but there needs to be a corresponding internal PR to handle the changes to `modal-http`.

## Changelog

- Greatly improved streaming performance of generators and WebSocket web endpoints.
- **Breaking change:** You cannot use `.map()` to call a generator function. (In previous versions, this merged the results onto a single stream, but the behavior was undocumented and not widely used.)
- **Incompatibility:** Generator outputs are now on a different internal system. Modal code on client versions before 0.57 cannot trigger [deployed functions](https://modal.com/docs/guide/trigger-deployed-functions) with `.remote_gen()` that are on client version 0.57, and vice versa.


modal/functions.py
```diff
@@ -176,6 +178,45 @@ async def _create_input(args, kwargs, client, idx=None) -> api_pb2.FunctionPutIn
...
+                if chunk.index <= last_index:
+                    continue
+                last_index = chunk.index
+                if chunk.data_blob_id:
+                    message_bytes = await blob_download(chunk.data_blob_id, client.stub)
+                else:
+                    message_bytes = chunk.data
+                message = deserialize_data_format(message_bytes, chunk.data_format, client)
+                yield message
+        except (GRPCError, StreamTerminatedError) as exc:
```
aksh-at, line 208: Might be useful to separate this retry logic out at some point, as all `unary_stream` calls need something like this.

modal/functions.py
```diff
@@ -176,6 +178,45 @@ async def _create_input(args, kwargs, client, idx=None) -> api_pb2.FunctionPutIn
...
+            if retries_remaining > 0:
```
aksh-at, line 209: Should we reset `retries_remaining` if there was no exception? (in `else`?) Otherwise a really long-running generator could naturally exhaust the 10 retries, even if they're spread apart.

modal/_container_entrypoint.py
```diff
@@ -522,32 +515,44 @@ def call_function_sync(
...
+                    generator_queue: asyncio.Queue[Any] = function_io_manager._queue_create(1024)
+                    generator_output_task = function_io_manager.generator_output_task(
+                        function_call_id,
+                        imp_fun.data_format,
+                        generator_queue,
+                        _future=True,  # Synchronicity magic to return a future.
+                    )
+                    try:
+                        while True:
+                            value = next(res)
```
aksh-at, line 539: Why not iterate normally using `for value in res` here?

modal/functions.py
```diff
@@ -271,37 +312,26 @@ async def poll_function(self, timeout: Optional[float] = None):
...
+        items_total: Union[int, None] = None  # populated when generator_finished_task completes
+
+        while True:
+            done, _ = await asyncio.wait([data_task, generator_finished_task], return_when=asyncio.FIRST_COMPLETED)
+            if data_task in done:
+                yield data_task.result()
+                items_received += 1
+                if items_received == items_total:
+                    break
+                data_task = asyncio.create_task(data_stream.__anext__())
```
aksh-at, line 328: This feels a bit messy. Could we simplify this using [aiostream](https://aiostream.readthedocs.io/en/stable/core.html)? Something like

```python
aiostream.stream.merge(data_stream, aiostream.stream.once(generator_finished_task))
```

aksh-at (commented):
Nice! There's a lot of complex loop logic here so might do another pass with fresh eyes tomorrow. 

Does this change the behavior of retries with generators? I believe the current logic is that we retry the entire generator if we don't finish. I think now it's possible for the generator to not push all the data all the way, but return the `DATA_FORMAT_GENERATOR_DONE` value?

ekzhang:
> Does this change the behavior of retries with generators in any way?

No, I don't believe so — retries with generators still work for deterministic generators and are undefined behavior for generators that have nondeterministic output

modal/functions.py
ekzhang: It seems reasonable, but I also think limiting the number of retries is safer for load shedding during periods of high traffic.

If we aim to have (1-eps) reliability for this RPC, then 10 retries gives us (1/eps)^10 reliability

modal/functions.py
ekzhang: Will try that!

ekzhang: Thanks, looks like this also fixed a bug with asynchronous shutdown thanks to the [stream context](https://aiostream.readthedocs.io/en/stable/core.html#stream-context-manager) manager

modal/functions.py
aksh-at: Yeah that's a good point. It's just that we have a natural failure rate per individual RPC. Since longer-lived generators will require this function to do a few orders of magnitudes more RPC calls, longer-lived generators will have a higher overall failure rate, which feels weird.

ekzhang: There are only 1440 minutes in a day!

modal/_container_entrypoint.py
```diff
@@ -587,32 +590,47 @@ async def call_function_async(
...
                 if imp_fun.is_generator:
                     if not inspect.isasyncgen(res):
                         raise InvalidError(f"Async generator function returned value of type {type(res)}")
-                    async for value in res:
-                        await function_io_manager.push_generator_value.aio(
-                            input_id, started_at, output_index.value, value, imp_fun.data_format
+
+                    # Send up to this many outputs at a time.
+                    generator_queue: asyncio.Queue[Any] = await function_io_manager._queue_create.aio(1024)
+                    generator_output_task = asyncio.create_task(
```
aksh-at, line 605: Should we use a `TaskContext` here instead?

aksh-at (approved):
Thanks for the updates! LGTM besides a couple comments

modal/functions.py
aksh-at: Fair enough, let's leave it as is

modal/_container_entrypoint.py
```diff
@@ -169,31 +158,65 @@ def deserialize_data_format(self, data: bytes, data_format: int) -> Any:
...
+    async def generator_output_task(self, function_call_id: str, data_format: int, message_rx: asyncio.Queue) -> None:
+        """Task that feeds generator outputs into a function call's `data_out` stream."""
+        index = 1
+        received_sentinel = False
+        while not received_sentinel:
+            message = await message_rx.get()
+            if message is self._GENERATOR_STOP_SENTINEL:
+                break
+            messages_bytes = [serialize_data_format(message, data_format)]
+            total_size = len(messages_bytes[0]) + 512
```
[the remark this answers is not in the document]
erikbern: +1

modal/_container_entrypoint.py
```diff
@@ -169,31 +158,65 @@ def deserialize_data_format(self, data: bytes, data_format: int) -> Any:
...
             try:
-                async for chunk in unary_stream(client.stub.FunctionCallGetDataIn, req):
-                    if chunk.index <= last_index:
-                        continue
-                    last_index = chunk.index
-                    if chunk.data_blob_id:
-                        message_bytes = await blob_download(chunk.data_blob_id, client.stub)
+                while total_size < 16 * 1024 * 1024:  # 16 MiB
```
erikbern, line 200: what's the 16MB constant? would be nice to put a name on it

modal/_container_entrypoint.py
```diff
@@ -534,20 +526,31 @@ def run_inputs(input_id: str, function_call_id: str, args: Any, kwargs: Any) ->
...
+                    # Send up to this many outputs at a time.
+                    generator_queue: asyncio.Queue[Any] = function_io_manager._queue_create(1024)
+                    generator_output_task = function_io_manager.generator_output_task(
+                        function_call_id,
+                        imp_fun.data_format,
+                        generator_queue,
+                        _future=True,  # Synchronicity magic to return a future.
+                    )
+
+                    item_count = 0
```
erikbern, line 538: I wonder if the control flow would be nicer if you have a context manager on the function io manager that handles cleanup, e.g.

```python
with function_io_manager.generator_result() as q:
    for value in res:
        q.put(value)
```
That way you can move the finalization logic to the context manager instead

This is just a thought, not at all a request! It's also entirely possible I'm misunderstanding the control flow and my suggestion is dumb

erikbern: Also with the current code, what happens if an exception is raised – does the result get sent properly to the server? I don't see a `finally:` handler or similar but maybe it's not needed

erikbern (approved):
This looks mostly good – left a few minor comments.

Do we have tests for exception handling with this? Just making sure any user code exception is handled properly

modal/functions.py
ekzhang: Open to suggestions! I thought about it for a bit and wasn't sure how to refactor this. It is repeated in:

- `_container_exec.py`
- `_output.py`
- `functions.py`
- `sandbox.py`

modal/_container_entrypoint.py
ekzhang: Hmm, I tried this and it doesn't seem resolve the issue of `q` being on the synchronicity event loop and needing the hacky `function_io_manager._queue_put()` implementation. :/

modal/_container_entrypoint.py
ekzhang: Not sure! The sync code path uses a synchronicity call, so it lives on the other event loop. This one uses an async synchronicity call. We might be able to remove the asyncio.create_task() here entirely actually, except due to a limitation of synchronicity where the `asyncio.run_coroutine_threadsafe()` call is placed inside another async function, so it actually doesn't make progress until awaited.

image

ekzhang:
> This looks mostly good – left a few minor comments.
> 
> Do we have tests for exception handling with this? Just making sure any user code exception is handled properly

Just double-checked, we do have tests for exception handling with generators! Thanks for looking at this in depth.

PR #1247, opened 2024-01-25
Include environment_name when sending FunctionBindParams

gongy:
This allows us to place the function into the correct environment for the caller to see.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -923,6 +923,7 @@ class _Function(_Object, type_prefix="fu"):
                 function_id=self._object_id,
                 serialized_params=serialized_params,
                 function_options=options,
+                environment_name=_get_environment_name(None, resolver),
             )
             response = await retry_transient_errors(self._client.stub.FunctionBindParams, req)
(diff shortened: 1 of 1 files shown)


freider:
Lgtm, but I can't quite wrap my head around the different ways this would play out for users.

What are the cases where this environment would differ from the parent function's environment? Only for published functions used across different workspaces?

Are there cases where using the resolver's environment would be counter-intuitive (as opposed to using the parent function's environment)?

freider (approved):

freider:
Might want to add a small unit test to prevent regressions as well, by just intercepting the `FunctionBindParams` call and ensuring the environment_name is what we expect it to be in different cases

gongy:
Great points I will address in a following PR (expediting this merge for functionality) [MOD-2191]
* Test regressions by ensuring `environment_name` is correct in different cases
* Don't populate`environment_name` for non-cross-workspace lookups and test this (currently, it is ignored server-side)

In cross-workspace cases, I believe it's intuitive that a caller environment must be inferred/extracted since it cannot be inherited from the parent function.

PR #1252, opened 2024-01-26
Improve config file isolation in tests

mwaskom:
## Describe your changes

The CLI tests modify the main config file (`~/.modal.toml`) when run locally. Additionally, I want to be able to dump information into a config file so that I can test some of the `modal token` changes I am in the process of making. This PR adds a new fixture that sets up an isolated config file and uses it in (a) cli tests that are currently modifying `~/.modal.toml` and (b) an existing config test that sets up an isolated config in a non-reusable way.

### DIFF
--- a/test/cli_test.py
+++ b/test/cli_test.py
@@ -108,12 +108,14 @@ def test_secret_list(servicer, set_env_client):
     assert "dummy-secret-3" not in res.stdout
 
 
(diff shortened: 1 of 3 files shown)


erikbern (approved):
Nice!

freider (commented):
👍

PR #1255, opened 2024-01-27
Migrate `modal shell` to `container exec`

thecodingwizard:
Depends on https://github.com/modal-labs/modal/pull/9902

Resolves MOD-2175, MOD-1427


modal_proto/api.proto
```diff
@@ -1338,6 +1338,14 @@ message SandboxCreateResponse {
   string sandbox_id = 1;
 }
 
+message SandboxGetTaskIdRequest {
```
aksh-at, line 1341: I don't see the server PR for this up btw. We'll want to merge the proto definitions first, then deploy the server API change, and then the rest of this PR.

modal/runner.py
```diff
@@ -292,11 +302,32 @@ async def _interactive_shell(_function: _Function, cmd: str, environment_name: s
...
+            await asyncio.sleep(0.5)
+            resp = await sb._client.stub.SandboxGetTaskId(api_pb2.SandboxGetTaskIdRequest(sandbox_id=sb._object_id))
+            if resp.task_id != "":
+                task_id = resp.task_id
+                break
+            # else: sandbox hasn't been assigned a task yet
+        else:
+            loading_status.stop()
+            rich.print("Error: timed out waiting for sandbox to start", file=sys.stderr)
+            await sb.terminate()
```
aksh-at, line 325: Don't think this is needed explicitly, since the `with _run_stub` context manager will terminate all apps on exit.

modal/runner.py
```diff
@@ -14,10 +18,10 @@
 from . import _pty
 from ._output import OutputManager, get_app_logs_loop, step_completed, step_progress
 from .app import _LocalApp, is_local
+from .cli.container import _container_exec
```
aksh-at, line 21: We shouldn't import from `.cli` into here. `.cli` is a "peripheral" module that should import from the core modules. So that means we should move the `_container_exec` function out of it somewhere else.

modal/cli/run.py
```diff
@@ -368,24 +369,30 @@ def shell(
...
+            **asdict(function_env),
+            timeout=3600,
+        )
     else:
-        image_obj = Image.from_registry(image, add_python=add_python) if image else None
-        stub = Stub("modal shell", image=image_obj)
-        function = stub.function(
-            serialized=True,
+        image = Image.from_registry(image, add_python=add_python) if image else None
+        interactive_shell(
```
aksh-at, line 388: Minor but would be cleaner to have the if/else just set the function environment, and then just one call to `interactive_shell`.

aksh-at (approved):
nice! Lgtm besides comments

erikbern:
Will this enable us to remove some of this code as well?

* https://github.com/modal-labs/modal-client/blob/main/modal/_container_entrypoint.py#L670
* https://github.com/modal-labs/modal-client/blob/main/modal/_container_entrypoint.py#L827
* https://github.com/modal-labs/modal-client/blob/main/modal/functions.py#L749

aksh-at (approved):
Thanks for making the changes, lgtm. I believe we can get rid of the container_entrypoint PTY code when we remove `interactive=True` right?

thecodingwizard:
Yes, I think so. My interactive branch updates the `_container_entrypoint` code, so perhaps we can keep the code in for now and remove it when we implement interactive functions with `modal container exec`?

If we want to delete it now too that's fine as well, I can just rebase.

I also still have to write unit tests... sorry, I got very distracted fixing gvisor issues :(

test/cli_test.py
```diff
@@ -340,35 +342,44 @@ def mock_get_pty_info(shell: bool) -> api_pb2.PTYInfo:
...
+    ), mock.patch(
+        "modal._container_exec._write_to_fd", write_to_fd
+    ):
+        yield captured_out
 
 
 @pytest.mark.usefixtures("mock_shell_pty")
-def test_shell(servicer, set_env_client, test_dir):
+@skip_windows("modal shell is not supported on Windows.")
+def test_shell(servicer, set_env_client, test_dir, mock_shell_pty):
```
aksh-at, line 365: Can remove the `mark.usefixtures` if we're also explicitly adding it as an argument

test/cli_test.py
```diff
@@ -340,35 +342,44 @@ def mock_get_pty_info(shell: bool) -> api_pb2.PTYInfo:
...
     with mock.patch("rich.console.Console.is_terminal", True), mock.patch(
         "modal._pty.get_pty_info", mock_get_pty_info
-    ), mock.patch("modal._pty.write_stdin_to_pty_stream", asyncnullcontext):
-        yield
+    ), mock.patch("modal._container_exec.get_pty_info", mock_get_pty_info), mock.patch(
+        "modal._pty.write_stdin_to_pty_stream", asyncnullcontext
+    ), mock.patch(
+        "modal._container_exec.handle_exec_input", asyncnullcontext
+    ), mock.patch(
+        "modal._container_exec._write_to_fd", write_to_fd
```
aksh-at, line 358: Could we use pytest's [capsys](https://docs.pytest.org/en/7.1.x/how-to/capture-stdout-stderr.html#accessing-captured-output-from-a-test-function) fixture instead of mocking `write_to_fd`?

thecodingwizard: I think I tried it and it didn't work, but I can test this again -- maybe I made a mistake while testing.

thecodingwizard: Oh, maybe I need `capfd` instead of `capsys`.

thecodingwizard: Never mind, I still can't get it to work, even with `capfd` and `capfdbinary`. Maybe `loop.add_writer(fd, try_write)` doesn't work (ie. `try_write` never gets called)?

test/cli_test.py
```diff
@@ -340,35 +340,43 @@ def mock_get_pty_info(shell: bool) -> api_pb2.PTYInfo:
...
     _run(["shell", webhook_stub_file.as_posix() + "::foo"])
-    assert ran_cmd == "/bin/bash"
+    assert mock_shell_pty == [(1, b"Hello World")]
+    mock_shell_pty.clear()
 
     # Function must be inferred
     _run(["shell", webhook_stub_file.as_posix()])
-    assert ran_cmd == "/bin/bash"
+    assert mock_shell_pty == [(1, b"Hello World")]
+    mock_shell_pty.clear()
```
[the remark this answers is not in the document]
freider: One thing you could do to get truly "clean" fixtures is to use use a `@pytest.mark.paramaterize(...)` to run each of these `_run`s and assertions in "their own test", but I'd say using `clear()` is fine here :)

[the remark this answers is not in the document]
thecodingwizard: thanks!!

PR #1256, opened 2024-01-27
Return Functions in DeployResult

aksh-at:
Allows users to programmatically get `web_url` for a function when they're programmatically deploying apps:

```python
res = deploy_stub(stub)
print(res["Model.my_function"].web_url)
```

Considered returning a separate dataclass, but returning `Function` seems cleaner.

### DIFF
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -186,6 +186,7 @@ class DeployResult:
     """Dataclass representing the result of deploying an app."""
 
     app_id: str
+    functions: dict[str, _Function]
 
 
 async def _deploy_stub(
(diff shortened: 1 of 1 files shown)


erikbern:
we already set the `web_url` property on all the functions when we deploy – is that enough?

no strong feelings about this, just pointing out it may or may not be needed

erikbern (approved):

aksh-at:
> we already set the web_url property on all the functions when we deploy – might be enough in most cases

Actually yeah, if the user has the `stub`, they could just look it up directly. Let me see if that works.

PR #1261, opened 2024-01-29
Use workspace as name for profiles created by `modal token` CLI

mwaskom:
## Describe your changes

A few things are happening in this PR:

First, I refactored the existing code underlying `modal token new` and `modal token set`:
  - I moved the underlying functions into `modal/token_flow.py` so the CLI code stays very thin.
  - `_new_token` now calls into `_set_token` after grabbing credentials from the server, to reduce duplicated code and ensure consistent console logging.
  - I converted the underlying functions to be async as they call into an async function for fetching the workspace name

Then, I changed the behavior of these functions to let us move away from having a "default" profile concept:
- The default profile that is created / updated by these functions will be named after the workspace that the token points to, rather than being named `"default"`
- If no profiles currently exist, the new profile will have its `active` metadata set to `True`
- There's a new `--activate` flag for both CLI commands that will activate the profile that gets created / modified
 
Open questions:
- Should `_new_token` and `_set_token` be public functions in the API?
    - One thing to consider is that the functions as written won't modify the config state in the active process, so I think we'd want to make them reload the config if we do that.
- Maybe `modal/token_flow.py` should be `modal/token.py` now? This is safe to do as it does not currently contain public interfaces.
- Would it make more sense to have `activate=True` by default?

Part of MOD-2168

## Changelog

The `modal token new` and `modal token set` commands now create profiles that are more closely associated with workspaces, and they have more explicit profile activation behavior:

- By default, these commands will create/update a profile named after the workspace that the token points to, rather than a profile named "default"
- Both commands now have an `--activate` flag that will activate the profile associated with the new token
- If no other profiles exist at the time of creation, the new profile will have its `active` metadata set to True

With these changes, we are moving away from the concept of a "default" profile. Implicit usage of the "default" profile will be deprecated in a future update.

### DIFF
--- a/modal/cli/entry_point.py
+++ b/modal/cli/entry_point.py
@@ -6,6 +6,8 @@ import typer
 from rich.console import Console
 from rich.rule import Rule
 
+from modal_utils.async_utils import synchronizer
+
 from . import run
 from .app import app_cli
 from .config import config_cli
@@ -71,11 +73,12 @@ def check_path():
     console.print(Rule(style="white"))
 
 
-def setup(profile: Optional[str] = None):
+@synchronizer.create_blocking
+async def setup(profile: Optional[str] = None):
     check_path()
 
     # Fetch a new token (same as `modal token new` but redirect to /home once finishes)
-    _new_token(profile=profile, next_url="/home")
+    await _new_token(profile=profile, next_url="/home")
 
 
 entrypoint_cli_typer.add_typer(app_cli)
--- a/modal/cli/profile.py
+++ b/modal/cli/profile.py
@@ -30,9 +30,15 @@ def current():
 async def list(json: Optional[bool] = False):
     config = Config()
     profiles = config_profiles()
-    responses = await asyncio.gather(
-        *(_lookup_workspace(config, profile) for profile in profiles), return_exceptions=True
-    )
+    lookup_coros = [
+        _lookup_workspace(
+            config.get("server_url", profile),
+            config.get("token_id", profile),
+            config.get("token_secret", profile),
+        )
+        for profile in profiles
+    ]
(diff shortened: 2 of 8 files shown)


freider:
Great cleanup! UX wise this makes a lot of sense! If we want to expose the functions in token(_flow).py (especially set_token, but maybe new as well?) we might want to refactor them to be "non-interactive" (optional console output, no input) at the same time, but that's probably not high prio at the moment.
Most important is that the CLI is intuitive and leads the user in the right direction, which I think this change does

mwaskom:
> we might want to refactor them to be "non-interactive" 

Good point — I left the `input()` calls for `set_token` in the CLI layer, but we'd probably not want an API call that goes through a manual web browser input step. I do see some potential benefit for programmatic token creation / setting (e.g. there was a question from a user who is running modal in CI and wanted PRs from forks to be able to kick off jobs) but probably best to defer exposing it until we understand that use case better

howard36:
When creating a new token on the website, it recommends using `--profile` with a different name (`modal_labs` instead of `modal-labs`).

image

If we're making this change, we should also update the website to either recommend `--profile modal-labs`, or just use `modal token set` without `--profile` (which will then check the token's workspace)

mwaskom:
Great catch! I think that we can remove the `--profile` suggestion from this suggestion after this change lands.

howard36:
We can't remove it immediately, since not all users will have the latest client version (older clients will apply it to the wrong profile if `--profile` isn't specified).

I think the website should suggest `--profile=modal-labs` for now, and in a few months after everyone has upgraded, we can remove `--profile` while keeping the behavior the same.

mwaskom:
Good point — yeah, agreed, we should not normalize the suggested profile name here.

PR #1264, opened 2024-01-30
Return a specific error code for missing requests

ekzhang:
This should help diagnose the error for users in the future

### DIFF
--- a/modal/_asgi.py
+++ b/modal/_asgi.py
@@ -29,7 +29,22 @@ def asgi_app_wrapper(asgi_app, function_io_manager):
             try:
                 first_message = await asyncio.wait_for(message_gen.__anext__(), 5.0)
             except asyncio.TimeoutError:
-                await messages_to_app.put({"type": "http.disconnect"})
+                if scope["type"] == "http":
+                    await messages_to_app.put({"type": "http.response.start", "status": 502})
+                    await messages_to_app.put(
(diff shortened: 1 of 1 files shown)


ekzhang:
bump @danielnorberg @freider, trying to keep you in the loop

modal/_asgi.py
```diff
@@ -29,7 +29,22 @@ async def fetch_data_in():
             try:
                 first_message = await asyncio.wait_for(message_gen.__anext__(), 5.0)
             except asyncio.TimeoutError:
-                await messages_to_app.put({"type": "http.disconnect"})
+                if scope["type"] == "http":
+                    await messages_to_app.put({"type": "http.response.start", "status": 502})
```
danielnorberg, line 33: Does this mean that we'll respond with `502` also when users take too long to send the request body?

danielnorberg (approved):

modal/_asgi.py
ekzhang: No, this is just for the first protocol message!

```
            # Cancel an ASGI app call if the initial message is not received within a short timeout.
            #
            # This initial message, "http.request" or "websocket.connect", should be sent
            # immediately after starting the ASGI app's function call. If it is not received, that
            # indicates a request cancellation or other abnormal circumstance.
```

PR #1268, opened 2024-01-31
Add error (and deprecation warning) for invalid profile configuration

mwaskom:
## Describe your changes

This adds some checks that enforce the following rule for profile configuration: if more than one profile is defined in `.modal.toml`, exactly one must be explicitly marked as active.

We are deprecating the legacy behavior where we implicitly fall back to a profile named 'default' if one exists. There's a deprecation warning, and we'll start enforcing that in a future release.

I've also added an error when more than one profile is explicitly marked as active. Currently we just silently use the first profile; this feels "wrong" enough to break without a gradual deprecation, but I'm happy to make this a warning if others feel differently.

- Resolves MOD-2211
- Resolves MOD-2228

## Changelog

- Support for an implicit 'default' profile is now deprecated. If you have more than one profile in your Modal config file, one must be explicitly set to `active` (use `modal profile activate` or edit your `.modal.toml` file to resolve).
- An error is now raised when more than one profile is set to `active`.

### DIFF
--- a/modal/config.py
+++ b/modal/config.py
@@ -142,7 +142,8 @@ def _check_config() -> None:
     if num_active > 1:
         raise InvalidError(
             "More than one Modal profile is active. "
-            "Please fix with `modal profile activate` or by editing your Modal config file."
+            "Please fix with `modal profile activate` or by editing your Modal config file "
+            f"({user_config_path})."
         )
     elif num_profiles > 1 and num_active == 0 and _profile == "default":
         # Eventually we plan to have num_profiles > 1 with num_active = 0 be an error
--- a/modal/config.py
+++ b/modal/config.py
@@ -76,6 +76,7 @@ import os
 import typing
 import warnings
 from datetime import date
+from textwrap import dedent
 from typing import Any, Dict, Optional
 
 import toml
@@ -84,7 +85,7 @@ from google.protobuf.empty_pb2 import Empty
 from modal_proto import api_pb2
 from modal_utils.logger import configure_logger
 
-from .exception import deprecation_error
+from .exception import InvalidError, deprecation_error, deprecation_warning
 
 # Locate config file and read it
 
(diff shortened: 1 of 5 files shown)


modal/config.py
```diff
@@ -135,6 +136,32 @@ def config_set_active_profile(env: str) -> None:
...
+            WARNING:
+
+            Support for using an implicit 'default' profile is deprecated.
+            Please use `modal profile activate` to activate one of your profiles.
+            (Use `modal profile list` to see the options.)
+
+            This will become an error in a future update.
+            """
+        )
+        deprecation_warning(date(2024, 2, 6), message)
```
mwaskom, line 162: I am a little sad that this is now a boring old deprecation warning rather than the nice Rich panel in my previous implementation. It feels wrong to be using rich from within the library here (rather than the CLI). But maybe we should have something that boosts the salience of our `deprecation_warning` messages?

modal/config.py
```diff
@@ -135,6 +136,32 @@ def config_set_active_profile(env: str) -> None:
     _write_user_config(_user_config)
 
 
+def _check_config() -> None:
+    num_profiles = len(_user_config)
+    num_active = sum(v.get("active", False) for v in _user_config.values())
+    if num_active > 1:
+        raise InvalidError(
+            "More than one Modal profile is active. "
+            "Please fix with `modal profile activate` or by editing your Modal config file."
```
freider, line 145: nit: could add `{user_config_path}` here too to let the user know where the config file is

modal/config.py
freider: Should hopefully be a relatively rare warning

test/client_test.py
```diff
@@ -151,3 +151,44 @@ def test_client_from_env(servicer):
         assert client_4 == client_3
     finally:
         Client.set_env_client(None)
+
+
+def test_multiple_profile_error(servicer, modal_config):
+    config = """
```
freider, line 157: Love the structure of these tests after your addition of the `modal_config` ctx manager :)

freider (approved):
lgtm!

erikbern (approved):
lgtm!

modal/config.py
mwaskom: 🤞

modal/config.py
mwaskom: Ooo nice idea!

PR #1270, opened 2024-02-01
Misc `container exec` improvements

thecodingwizard:
- Sort `modal container list` by start time
- Allow Ctrl+C while the client is still connecting

### DIFF
--- a/modal/cli/container.py
+++ b/modal/cli/container.py
@@ -149,7 +149,7 @@ async def handle_exec_input(client: _Client, exec_id: str, use_raw_terminal=Fals
     write_task.cancel()
 
 
-async def handle_exec_output(client: _Client, exec_id: str, on_connect: Optional[asyncio.Event] = None):
+async def handle_exec_output(client: _Client, exec_id: str, on_connect: Optional[asyncio.Event] = None) -> int:
     """
     Streams exec output to stdout.
 
--- a/modal/cli/container.py
+++ b/modal/cli/container.py
@@ -35,8 +35,8 @@ async def list():
 
     column_names = ["Container ID", "App ID", "App Name", "Start time"]
     rows: List[List[Union[Text, str]]] = []
-    res.tasks.sort(key=lambda task: task.started_at)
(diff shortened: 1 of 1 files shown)


modal/cli/container.py
```diff
@@ -82,21 +83,22 @@ async def exec(
...
-        except TimeoutError:
-            connecting_status.stop()
-            rich.print("Failed to establish connection to process", file=sys.stderr)
-            exec_failed = True
-
-    if exec_failed:
-        # we don't want to raise this inside the context manager
-        # since otherwise the context manager cleanup doesn't get called
+    on_connect = asyncio.Event()
+    exec_output_task = asyncio.create_task(handle_exec_output(client, res.exec_id, on_connect=on_connect))
```
aksh-at, line 87: use a `TaskContext` here, it handles a bunch of things related to cancellation / waiting for tasks to finish

modal/cli/container.py
```diff
@@ -82,21 +83,22 @@ async def exec(
...
+    try:
+        # time out if we can't connect to the server fast enough
+        await asyncio.wait_for(on_connect.wait(), timeout=15)
+        connecting_status.stop()
```
aksh-at, line 91: Should this be in a `finally:`? It's in both branches

modal/cli/container.py
```diff
@@ -35,7 +35,8 @@ async def list():
 
     column_names = ["Container ID", "App ID", "App Name", "Start time"]
     rows: List[List[Union[Text, str]]] = []
-    for task_stats in res.tasks:
+    res.tasks.sort(key=lambda task: task.started_at)
```
aksh-at, line 38: Can just `reverse=True` here, as an argument to `.sort()`

aksh-at (approved):
Lgtm besides comments

modal/cli/container.py
thecodingwizard: I think it needs to be called before we print the error message, so at least for the `TimeoutError` except case, I don't think it can be moved to `finally` (?)

PR #1271, opened 2024-02-01
Indicate when token is active in env vars in `modal profile list`

mwaskom:
## Describe your changes

image

- Resolves MOD-2194

## Changelog

The `modal profile list` output now indicates when the workspace is determined by a token stored in environment variables.

### DIFF
--- a/modal/cli/profile.py
+++ b/modal/cli/profile.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2022
 
 import asyncio
+import os
 from typing import Optional
 
 import typer
@@ -33,8 +34,8 @@ async def list(json: Optional[bool] = False):
     lookup_coros = [
         _lookup_workspace(
             config.get("server_url", profile),
-            config.get("token_id", profile),
-            config.get("token_secret", profile),
+            config.get("token_id", profile, use_env=False),
+            config.get("token_secret", profile, use_env=False),
         )
         for profile in profiles
     ]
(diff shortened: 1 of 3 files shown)


modal/cli/profile.py
```diff
@@ -53,7 +54,20 @@ async def list(json: Optional[bool] = False):
         content = ["•" if active else "", profile, workspace]
         rows.append((active, content))
 
+    env_based_workspace: Optional[str] = None
+    if "MODAL_TOKEN_ID" in os.environ:
```
erikbern, line 58: shouldn't we just use `config.get` to grab the tokens just like the server url?

mwaskom: That will tell you the token that is being used, but what we want to know here is whether that is different from the token associated with the “active profile”.

freider: Ah, so this is specifically for when a user specifies TOKEN_ID and TOKEN_SECRET as env vars but has an active profile which is otherwise associated with another workspace? I guess we could skip showing the message if the env-var workspace and the profile workspace are the same?

freider (approved):

modal/cli/profile.py
mwaskom: Yeah exactly. It could also make sense to only show this when they disagree. Maybe one reason to show this kind of message even when they agree is that it might help people problem solve faster if they subsequently get themselves into a state when they disagree. Maybe that suggests that `modal profile activate` should also provide some sort of feedback if you activate a profile with a workspace that won't actually get used because the environment variable tokens will override it.

PR #1276, opened 2024-02-02
Run client tests on M1 macOS Sonoma

ekzhang:
These are free.

I think it's a good idea to run client tests on M1 since a lot of our users use M1 macbooks as a development machine. We also use them often though, so this is unlikely to catch any new bugs 😅

https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

### DIFF
--- a/test/gpu_test.py
+++ b/test/gpu_test.py
@@ -142,25 +142,13 @@ def test_a100_20gb_gpu_unsupported():
         stub.function(gpu=modal.gpu.A100(memory=20))(dummy)
 
 
-A100_GPU_COUNT_MAPPING = {1: api_pb2.GPU_TYPE_A100, **{i: api_pb2.GPU_TYPE_A100 for i in range(2, 5)}}
-
-
-@pytest.mark.parametrize("count,gpu_type", A100_GPU_COUNT_MAPPING.items())
-def test_gpu_type_selection_from_count(client, servicer, count, gpu_type):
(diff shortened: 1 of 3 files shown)


test/gpu_test.py
```diff
@@ -142,30 +142,18 @@ def test_a100_20gb_gpu_unsupported():
...
-    # Functions that use A100 20GB can only request one GPU
-    # at a time.
-    with pytest.raises(ValueError):
-        with pytest.warns(DeprecationError):
-            stub.function(gpu=modal.gpu.A100(count=2, memory=20))(dummy)
-        with stub.run(client=client):
-            pass
-
-    # Task type changes whenever user asks more than 1 GPU on
-    # an A100.
```
mwaskom, line 163: These don't run properly on M1?

ekzhang: The tests are kind of broken, `A100()` constructor already raises a ValueError. I think in Pytest v8, the behavior of the warns() context manager changed. Anyway the test is not relevant anymore.

I had to change back to pytest v7 manually in a later commit anyway because some pytest-markdown-docs uses some internal API

ekzhang: (pytest v8 was released on Jan 27, which is why this PR suddenly upgraded versions, but only for non-cached dependencies)

PR #1279, opened 2024-02-02
Migrate interactive functions to `modal container exec`

thecodingwizard:


modal/_container_entrypoint.py
```diff
@@ -36,7 +37,7 @@
 from ._blob_utils import MAX_OBJECT_SIZE_BYTES, blob_download, blob_upload
 from ._function_utils import LocalFunctionError, is_async as get_is_async, is_global_function
 from ._proxy_tunnel import proxy_tunnel
-from ._pty import exec_cmd, run_in_pty
+from ._pty import exec_cmd
```
erikbern, line 40: do we still need `exec_cmd`?

thecodingwizard: hmm good point, I don't think so!

erikbern: it's used on L681:
```python
    if pty_info.pty_type == api_pb2.PTYInfo.PTY_TYPE_SHELL:
        fun = exec_cmd
```

but I think that branch might be dead now? but not sure

modal/functions.py
```diff
@@ -203,7 +207,7 @@ async def create(function_id: str, args, kwargs, client: _Client):
 
         function_call_id = response.function_call_id
 
-        item = await _create_input(args, kwargs, client)
+        item = await _create_input(args, kwargs, client, final_input=True)
```
thecodingwizard, line 210: Note: I'm not sure if this change is correct. See Elias's comment on Slack: https://modal-com.slack.com/archives/C056CGAANRM/p1707123558333609?thread_ts=1706890751.262849&cid=C056CGAANRM

The reason I need this is I need the entire `_container_entrypoint.py` command to exit in order to signal the client that the PTY shell has closed.

If we do not want to rely on `final_input=True`, we will have to find another way to detect when the PTY shell has closed, perhaps by detecting when one input is finished.

I think it would be even better to detect when `breakpoint()` finishes rather than when the PTY shell has closed, but I don't know how to do this.

modal/_output.py
```diff
@@ -420,6 +436,14 @@ async def _get_logs():
                 # TODO (akshat): have a better way of differentiating between
                 # statically and dynamically built images.
                 pass
+            elif log_batch.pty_exec_id:
+                if pty_shell_finish_event:
+                    print("ERROR: concurrent PTY shells are not supported.")
```
thecodingwizard, line 441: with just this PR, this should never happen.

modal/functions.py
aksh-at: We should not set `final_input=True` for all invocations. This would mean that 
calls like this would not reuse a container when they should:

```python
model = MyModel()
model.inference.remote(...)
model.inference.remote(...)
```

ekzhang (approved):

ekzhang:
Looks good to me, may need to edit it to pass tests (lmk if you need help with this!)

PR #1280, opened 2024-02-02
DictGetOrCreate - use new endpoint client side 

erikbern:
This starts using `DictGetOrCreate` on the client side for `persisted`, `from_name`, and `lookup` (but not `new`)

A lot of redundant code right now but once this is done we should be able to clean up quite a lot

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -12,14 +12,14 @@ from ._types import typechecked
 from .client import _Client
 from .config import logger
 from .exception import deprecation_error
-from .object import _get_environment_name, _StatefulObject, live_method
+from .object import _get_environment_name, _Object, live_method
 
 
 def _serialize_dict(data):
     return [api_pb2.DictEntry(key=serialize(k), value=serialize(v)) for k, v in data.items()]
 
 
-class _Dict(_StatefulObject, type_prefix="di"):
+class _Dict(_Object, type_prefix="di"):
     """Distributed dictionary for storage in Modal apps.
 
     Keys and values can be essentially any object, so long as they can be
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -9,9 +9,10 @@ from modal_utils.grpc_utils import retry_transient_errors
 from ._resolver import Resolver
 from ._serialization import deserialize, serialize
 from ._types import typechecked
+from .client import _Client
 from .config import logger
 from .exception import deprecation_error
-from .object import _StatefulObject, live_method
+from .object import _get_environment_name, _StatefulObject, live_method
 
 
 def _serialize_dict(data):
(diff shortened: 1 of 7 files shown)


erikbern:
TODO: tests – will add later

erikbern:
Also passes all integration tests

modal/dict.py
```diff
@@ -93,17 +94,49 @@ def persisted(
...
-        self, label: str, namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE, environment_name: Optional[str] = None
+        async def _load(provider: _Dict, resolver: Resolver, existing_object_id: Optional[str]):
+            req = api_pb2.DictGetOrCreateRequest(
+                deployment_name=label,
+                namespace=namespace,
+                environment_name=_get_environment_name(environment_name),
+                object_creation_type=(api_pb2.OBJECT_CREATION_TYPE_CREATE_IF_MISSING if create_if_missing else None),
+            )
+            response = await resolver.client.stub.DictGetOrCreate(req)
+            logger.debug("Created dict with id %s" % response.dict_id)
```
mwaskom, line 106: Why use %-based interpolation here? I know you can do `logger.debug("id = %s", dict_id)` but why not use an fstring if interpolating directly?

modal/dict.py
```diff
@@ -93,17 +94,49 @@ def persisted(
...
+            provider._hydrate(response.dict_id, resolver.client, None)
+
+        return _Dict._from_loader(_load, "Dict()")
+
+    @staticmethod
+    def persisted(
```
mwaskom, line 112: Do we plan to deprecate this at some point?

modal/dict.py
```diff
@@ -75,13 +76,13 @@ def __init__(self, data={}):
         self._init_from_other(obj)
 
     @staticmethod
-    def persisted(
-        label: str, namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE, environment_name: Optional[str] = None
+    def from_name(
+        label: str,
+        namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
+        environment_name: Optional[str] = None,
+        create_if_missing: bool = False,
```
mwaskom, line 83: Thoughts on the default here? It feels like you'd usually want this to be `True`, and the main reason you would want `False` would be to avoid bugs from typo-ing the name or something similar. Maybe that's a good reason to set it to `False` in code that you want to be defensive about but I'm not sure if it's the friendliest default.

mwaskom (commented):
Should we add a changelog entry about `from_name`? That's new right?

erikbern:
> Should we add a changelog entry about `from_name`? That's new right?

No, it's just dropped from an (intermediate) baseclass `_StatefulObject` – I'm basically reimplementing this method on each subclasss (and removing the intermediate class – implementation inheritance is bad imo)

modal/dict.py
erikbern: Yes! But not quite yet

modal/dict.py
erikbern: This is super old code that I copy pasted! Old habits from my C days but I've come around to appreciate f-strings

modal/dict.py
erikbern: I want to deprecate all `.persisted` methods once I've gone through all object classes

modal/dict.py
erikbern: feels like `False` is the most defensive default since it has no side effects

mwaskom (approved):

PR #1285, opened 2024-02-05
Allow variadic parameters in scheduled functions

freider:
Previously you couldn't do things like:

```
@stub.function(schedule=...)
def my_func(*args):
    ...
```

But this now works (the scheduled function will still be called without arguments, so args will be an empty tuple)

## Changelog

* Variadic parameters (e.g. *args and **kwargs) can now be used in scheduled functions as long as the function doesn't have any other parameters without a default value


mwaskom (approved):
:+1:

PR #1290, opened 2024-02-06
Use QueueGetOrCreate client-side

erikbern:
This starts to use the new endpoint. Still uses `QueueCreate` for ephemeral queues though (might delete this later)

A bit of copy paste with dict but I think we need to do it to get to the other side of the river. We can simplify a lot later (will take some deprecation pain)

### DIFF
--- a/modal/queue.py
+++ b/modal/queue.py
@@ -13,11 +13,12 @@ from modal_utils.grpc_utils import retry_transient_errors
 
 from ._resolver import Resolver
 from ._serialization import deserialize, serialize
+from .client import _Client
 from .exception import deprecation_error
-from .object import _StatefulObject, live_method
+from .object import _get_environment_name, _Object, live_method
 
 
-class _Queue(_StatefulObject, type_prefix="qu"):
+class _Queue(_Object, type_prefix="qu"):
     """Distributed, FIFO queue for data flow in Modal apps.
 
     The queue can contain any object serializable by `cloudpickle`, including Modal objects.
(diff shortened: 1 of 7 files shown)


modal/queue.py
```diff
@@ -80,7 +81,48 @@ def persisted(
...
+                environment_name=_get_environment_name(environment_name),
+                object_creation_type=(api_pb2.OBJECT_CREATION_TYPE_CREATE_IF_MISSING if create_if_missing else None),
+            )
+            response = await resolver.client.stub.QueueGetOrCreate(req)
+            provider._hydrate(response.queue_id, resolver.client, None)
+
+        return _Queue._from_loader(_load, "Queue()")
+
+    @staticmethod
+    def persisted(
```
erikbern, line 98: will be deprecated later

modal/queue.py
```diff
@@ -80,7 +81,48 @@ def persisted(
...
+        label: str, namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE, environment_name: Optional[str] = None
+    ) -> "_Queue":
+        return _Queue.from_name(label, namespace, environment_name, create_if_missing=True)
+
+    @staticmethod
+    async def lookup(
```
erikbern, line 104: will be combined with `from_name` later

erikbern:
passes all integration tests

mwaskom (approved):

erikbern:
Thanks @mwaskom !

PR #1294, opened 2024-02-06
Raise error if generator functions are called with .map(...)

ekzhang:
Resolves MOD-2283

## Changelog

- Improve error message when generator functions are called with `.map(...)`.


ekzhang:
@prbot approve

PR #1299, opened 2024-02-06
Implement reuse_containers flag (and rename it to max_inputs)

howard36:
Continuation of #1295

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -282,7 +282,8 @@ class _FunctionIOManager:
                         yield (item.input_id, item.function_call_id, input_pb)
                         yielded = True
 
-                        if item.input.final_input or self.function_def.reuse_containers is False:
+                        # We only support max_inputs = 1 at the moment
+                        if item.input.final_input or self.function_def.max_inputs == 1:
                             eof_received = True
                             break
             finally:
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -282,7 +282,7 @@ class _FunctionIOManager:
                         yield (item.input_id, item.function_call_id, input_pb)
                         yielded = True
 
-                        if item.input.final_input:
+                        if item.input.final_input or self.function_def.reuse_containers is False:
                             eof_received = True
                             break
             finally:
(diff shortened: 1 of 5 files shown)


modal/_container_entrypoint.py
```diff
@@ -282,7 +282,7 @@ async def _generate_inputs(self) -> AsyncIterator[tuple[str, str, api_pb2.Functi
                         yield (item.input_id, item.function_call_id, input_pb)
                         yielded = True
 
-                        if item.input.final_input:
+                        if item.input.final_input or self.function_def.reuse_containers is False:
```
erikbern, line 285: I think this will be true for all old clients?

howard36: Oh yeah, I just realized a minute ago. Looks like we still need to negate this in function_def

howard36: Actually, since the change is only in the client, old clients won't even have this condition in their source code, so it should still work

gongy: I think that in this case, it could be cleaner to go for a `max_total_inputs_container` on the function definition where 0 is the default no limit.

User-facing field can still be `reuse_containers=False` (translate that to `max_total_inputs_container=1` in the definition). WDYT?

howard36: I'm also fine with `reuse_containers=False` being a special case of `max_inputs=1`. @aksh-at actually [suggested](https://modal-com.slack.com/archives/C02TENSMNMQ/p1706988491710919) this initially, as a user-facing field. The concern at the time was implementation effort, but now we know that's not an issue.

aksh-at: Oh right, if we can implement it client-side, `max_inputs` is better. The worker doesn't optimistically fetch more inputs right? I don't remember how that works, but that would break the "client-side only" implementation.

erikbern: Btw why `x is False` and not just `not x`. Its equivalent for booleans

erikbern: Btw I was stupid and didn't realize this code is in the client. Yeah it will work!

howard36: > Btw why x is False and not just not x. Its equivalent for booleans

I wrote `x is False` back when I thought `x` could be `True`, `False`, or `None`, and only wanted `False`. `not x` would accept both `None` and `False`.

But this is no longer relevant now that we've switched to `max_inputs`, which isn't a boolean

PR #1300, opened 2024-02-07
Fix bad default value for max_inputs in from_args

gimaker:
Afaict, the `True` default value doesn't break anything - the only place the default value will be used is for `run_function`, in which case it's moot anyway.


gongy:
Good find @gimaker merging this and watching

PR #1302, opened 2024-02-07
[MOD-2209] Add force flag for volume uploads

gimaker:
Previously, we silently overwrite any existing files. This is error-prone so we're changing the default behaviour to not overwrite existing files.

This adds a `force` flag which you can use to explicitly allow overwriting existing files.

Even with `force=True` we will not allow overwriting existing _directories_ with uploaded files.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1722,8 +1722,8 @@ message VolumePutFilesRequest {
   // TODO(staffan): This is obviously unfortunately named, but provides what we need - consider renaming.
   repeated MountFile files = 2;
   // If set to true, prevent overwriting existing files. (Note that we don't allow overwriting
-  // existing directories with uploaded files even when clobbering is allowed.)
-  bool no_clobber = 3;
+  // existing directories with uploaded files regardless.)
+  bool disallow_overwrite_existing_files = 3;
 }
 
 message VolumeRemoveFileRequest {
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1721,6 +1721,9 @@ message VolumePutFilesRequest {
   string volume_id = 1;
   // TODO(staffan): This is obviously unfortunately named, but provides what we need - consider renaming.
   repeated MountFile files = 2;
+  // If set to true, prevent overwriting existing files. (Note that we don't allow overwriting
+  // existing directories with uploaded files even when clobbering is allowed.)
+  bool no_clobber = 3;
 }
 
 message VolumeRemoveFileRequest {
--- a/test/cli_test.py
+++ b/test/cli_test.py
@@ -460,16 +460,22 @@ def test_volume_put_force(servicer, set_env_client):
             f.flush()
 
         # File upload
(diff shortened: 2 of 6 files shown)


freider:
Maybe it's just me but I feel "-f/--force" is a more intuitive name even if it's not exactly the same thing in mv/cp, I've never used the `--no-clobber` one. But I don't have a super strong opinion on it.
Maybe someone else should chime in on the name

test/cli_test.py
```diff
@@ -448,6 +448,30 @@ def test_volume_get(servicer, set_env_client):
...
+
+        _run(["volume", "put", vol_name, upload_path, file_path], expected_exit_code=2, expected_stderr=None)
+        _run(["volume", "put", vol_name, upload_path, file_path, "--clobber"])
+
+        # Dir upload
+        _run(["volume", "put", vol_name, tmpdir])
+
+        _run(["volume", "put", vol_name, tmpdir], expected_exit_code=2, expected_stderr=None)
+        _run(["volume", "put", vol_name, tmpdir, "--clobber"])
+
```
freider, line 473: Do we want to add some assertions here, e.g. that the flag on the RPC call is set?

Should be relatively easy with
```python
with servicer.intercept() as ctx:
    ...
req = ctx.pop_request("VolumePutFiles")
assert not req.no_clobber
```

test/volume_test.py
```diff
@@ -185,6 +185,26 @@ async def test_volume_batch_upload(servicer, client, tmp_path):
...
+        with stub.vol.batch_upload() as batch:
+            batch.put_file(local_file_path, "/some_file")
+
+        with pytest.raises(FileExistsError):
+            with stub.vol.batch_upload() as batch:
+                batch.put_file(local_file_path, "/some_file")
+
+        with stub.vol.batch_upload(clobber=True) as batch:
+            batch.put_file(local_file_path, "/some_file")
+
```
freider, line 206: Some assertions might be nice

freider (approved):

modal_proto/api.proto
```diff
@@ -1721,6 +1721,9 @@ message VolumePutFilesRequest {
   string volume_id = 1;
   // TODO(staffan): This is obviously unfortunately named, but provides what we need - consider renaming.
   repeated MountFile files = 2;
+  // If set to true, prevent overwriting existing files. (Note that we don't allow overwriting
+  // existing directories with uploaded files even when clobbering is allowed.)
+  bool no_clobber = 3;
```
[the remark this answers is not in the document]
gimaker: Yeah. Old clients will get the old behaviour of always overwriting existing files.

Wouldn't be the end of the world to make it so that you have to upgrade your client if you want the force/clobber behaviour either :shrug:

test/cli_test.py
gimaker: Added!

test/volume_test.py
gimaker: Added!

PR #1303, opened 2024-02-07
Removes unnecessary branch when checking for `MODAL_CHECKPOINT_ID`

luiscape:


modal/config.py
```diff
@@ -171,6 +171,7 @@ class _Setting(typing.NamedTuple):
     "default_cloud": _Setting(None, transform=lambda x: x if x else None),
     "worker_id": _Setting(),  # For internal debugging use.
     "restore_state_path": _Setting("/opt/modal/restore-state.json"),
+    "container_id": _Setting(),
```
ekzhang, line 174: Maybe remove this setting?

ekzhang (commented):
Thanks for doing this!

modal/config.py
luiscape: Removed and re-added: it's good to have `MODAL_CHECKPOINT_ID` populated for debugging with users.

PR #1312, opened 2024-02-07
Highlight client deprecation warnings in the CLI

mwaskom:
## Describe your changes

This uses rich to make Modal deprecation warnings more salient in the CLI:

image

Note that monkey-patching `warnings.showwarnings` is the [officially blessed](https://docs.python.org/3/library/warnings.html#warnings.showwarning) approach for customizing warning display. That said, I'm not totally sure how to write a unit test for this...

Open to suggestions about the formatting!

I've also enhanced our deprecation warning function to make it optional to include a source line. In some cases (e.g. commandline or config deprecation) the current logic attributes the warning to an asyncio entrypoint or some other unrelated line. Feels useful to supress when we know that will be the case.

## Changelog

- Modal client deprecation warnings are now highlighted in the CLI

### DIFF
--- a/modal/__main__.py
+++ b/modal/__main__.py
@@ -1,11 +1,12 @@
 # Copyright Modal Labs 2022
-from ._traceback import setup_rich_traceback
+from ._traceback import highlight_modal_deprecation_warnings, setup_rich_traceback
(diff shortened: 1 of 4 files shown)


freider (approved):
I like it!

PR #1313, opened 2024-02-07
Add GENERIC_STATUS_INIT_FAILURE; add comments

thecodingwizard:
Adds some comments to `GenericResult` and adds `GENERIC_STATUS_INIT_FAILURE`, used for S3 mount initialization failures.

Thanks @ekzhang !

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1344,7 +1344,6 @@ message S3Mount {
   string bucket_name = 1;
   string mount_path = 2;
   string credentials_secret_id = 3;
-  // if false, --allow-delete and --allow-overwrite are set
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1341,6 +1344,8 @@ message S3Mount {
   string bucket_name = 1;
   string mount_path = 2;
   string credentials_secret_id = 3;
+  // if false, --allow-delete and --allow-overwrite are set
```
luiscape, line 1347: I suggest removing this comment because it's an indirect reference to what we do internally.

luiscape (commented):
nit

PR #1315, opened 2024-02-08
New attempt: Skip synchronicity inspection of arguments/return values

freider:
The previous patch broke since the `modal.functions.Function` methods didn't have no-wrap, so Modal objects passed to `Function.remote()` or similar would become the "internal" represenation of the object, i.e. async without a separate event loop.

This also adds some more tests w/ assertions of data types going in and out of the user-level API

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -19,6 +19,9 @@ jobs:
       - name: Build protobuf
         run: inv protoc
 
+      - name: Install package + deps
+        run: pip install -e .  # Makes sure doc generation doesn't break on client imports etc.
+
       - name: Generate reference docs
         run: python -m modal_docs.gen_reference_docs reference_docs_output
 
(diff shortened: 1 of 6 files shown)


.github/workflows/docs.yml
```diff
@@ -19,6 +19,9 @@ jobs:
       - name: Build protobuf
         run: inv protoc
 
+      - name: Install package + deps
+        run: pip install -e .  # Makes sure doc generation doesn't break on client imports etc.
```
freider, line 23: This is somewhat unrelated, but the build failed because we don't have explicit requirements on client dependencies here. Presumably succeeded before this since dependencies were cached, but now that we needed some new globals/imports from a new synchronicity version things broke

modal/functions.py
```diff
@@ -1347,6 +1355,7 @@ def __call__(self, *args, **kwargs) -> Any:  # TODO: Generics/TypeVars
                 " a Modal container in the cloud",
             )
 
+    @synchronizer.no_input_translation
```
freider, line 1358: Esp. important here to *only* skip input translations and not output, since the return value from `spawn()` *should* be synchronicity-wrapped.

mwaskom (approved):

PR #1319, opened 2024-02-09
Add support for volumes in modal shell

thecodingwizard:
https://github.com/modal-labs/modal/pull/10304

Changes to volumes in `modal shell` persist with `_allow_background_volume_commits`, but I don't know if they commit properly upon exit. I think once I made a new file and very quickly exited the shell and the file did not persist, but if I wait a few seconds before exiting, it seems to commit fine. Maybe my test was just a fluke though.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -13,13 +13,14 @@ import typer
 from rich.console import Console
 from typing_extensions import TypedDict
 
+from modal.volume import _Volume
+
 from ..config import config
 from ..environments import ensure_env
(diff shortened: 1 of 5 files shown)


aksh-at (commented):
I don't think we should expose this to users until we have a handle on committing successfully. We can take it from here and look into it, thanks for doing this!

thundergolfer:
Superseded by https://github.com/modal-labs/modal-client/pull/1376 🙂

PR #1321, opened 2024-02-09
Fix response for missing first ASGI message

gongy:
When the first message is missing, we need to send a `502` response in lieu of `FastAPI`. This fix places the response in the correct outgoing message queue.

Further, this fix adds an ingoing message to the `FastAPI` app to register the a client disconnect. This is not necessary but was useful to expose an issue in our ASGI serialization, where `headers` became optional in a variety of places in the 3.0 specification.


ekzhang:
bug introduced in #1264, sorry!

ekzhang (approved):
Thank you!

PR #1322, opened 2024-02-09
Add automatic breakpoints, connect_stdin() function

thecodingwizard:
- deprecates (and breaks) `interactive=True` in functions
- adds an `--interactive` flag
- if that flag is enabled, functions will be run within a PTY. one function per `modal run` can call `enable_interactivity()` to connect their output to the user's CLI
- adds a custom breakpoint hook so that breakpoint()s automatically call `enable_interactivity()`

Depends on #1279 

It doesn't work too well if two distinct functions both call breakpoint(). The key issue is we need to know when the function our PTY shell is attached to exits. One way is to modify `container_entrypoint` to send a "Disconnect" message when it exits.

## Changelog

`modal run --interactive` or `modal run -i` run the app in "interactive mode". This allows any remote code to connect to the user's local terminal by calling `modal.interact()`. 

```python
@stub.function()
def my_fn(x):
    modal.interact()

    x = input()
    print(f"Your number is {x}")
```

This means that you can dynamically start an IPython shell if desired for debugging:

```python
@stub.function()
def my_fn(x):
    modal.interact()

    from IPython import embed
    embed()
```

For convenience, breakpoints automatically call `interact()`:

```python
@stub.function()
def my_fn(x):
    breakpoint()
```


modal/_output.py
```diff
@@ -448,15 +449,17 @@ async def _get_logs():
...
                     pty_shell_finish_event = asyncio.Event()
+                    pty_shell_task_id = log_batch.task_id
                     asyncio.create_task(stream_pty_shell_input(client, log_batch.pty_exec_id, pty_shell_finish_event))
             else:
                 for log in log_batch.items:
                     await _put_log(log_batch, log)
 
-            if log_batch.eof:
+            if log_batch.eof and log_batch.task_id == pty_shell_task_id:
                 await stop_pty_shell()
```
thecodingwizard, line 463: I'm trying to detect when the current PTY shell's task ends, and stop the pty shell. However, this doesn't seem to work very reliably.

modal/stub.py
```diff
@@ -478,6 +478,10 @@ def function(
         if _warn_parentheses_missing:
             raise InvalidError("Did you forget parentheses? Suggestion: `@stub.function()`.")
 
+        if interactive:
+            # todo(nathan): should we add a migration message here?
+            raise InvalidError("interactive is now deprecated.")
```
erikbern, line 483: yeah it would be great to use `deprecation_warning` here (and in a month or two we can use `deprecation_error` instead)

thecodingwizard: I used the `PTYInfo` field to signal to `modal-worker` that the `INTERACTIVE_FUNCTIONS` config flag is set. I believe this is the same field that `interactive=True` uses. So my understanding is if we want to support both `INTERACTIVE_FUNCTIONS` and `interactive=True` at the same time during the deprecation warning period, we will need to change some code to make `INTERACTIVE_FUNCTIONS` and `interactive=True` distinct from each other from `modal-worker`'s perspective.

erikbern: ok – maybe go straight to `deprecation_error` then. I don't think we need a grace period in this case since it's not something used in production

aksh-at:
@prbot approve

thecodingwizard:
🎉

PR #1269, opened 2024-01-31
Deprecate the singular form `secret` function parameter

mwaskom:
## Describe your changes

- Resolves MOD-2237

## Changelog

The singular form of the `secret` parameter in `Stub.function`, `Stub.cls`, and `Image.run_function` has been deprecated. Please update your code to use the plural form instead:`secrets=[Secret(...)]`.

PR #1324, opened 2024-02-09
fix: incorrect example docs in modal.Image

thundergolfer:
## Describe your changes

Broken in https://github.com/modal-labs/modal-client/pull/1269. 

It still may be dubious to fully reject the `secrets` value. But I've explicitly rejected it to make things less confusing.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1065,11 +1065,13 @@ class _Image(_Object, type_prefix="im"):
         ```python
         modal.Image.from_gcp_artifact_registry(
             "us-east1-docker.pkg.dev/my-project-1234/my-repo/my-image:my-version",
-            secrets=[modal.Secret.from_name("my-gcp-secret")],
+            secret=modal.Secret.from_name("my-gcp-secret"),
             add_python="3.11",
         )
         ```
         """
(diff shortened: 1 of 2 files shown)


mwaskom:
Ah thanks, my bad from find/replace when implementing the singular `secret` deprecation in the rest of the modal API.

modal/image.py
```diff
@@ -1065,11 +1065,13 @@ def from_gcp_artifact_registry(
...
         modal.Image.from_gcp_artifact_registry(
             "us-east1-docker.pkg.dev/my-project-1234/my-repo/my-image:my-version",
-            secrets=[modal.Secret.from_name("my-gcp-secret")],
+            secret=modal.Secret.from_name("my-gcp-secret"),
             add_python="3.11",
         )
         ```
         """
+        if "secrets" in kwargs:
+            raise ValueError("Passing a list of 'secrets' is not supported; use the singular 'secret' argument.")
```
mwaskom, line 1074: Nit: normally a `TypeError` is raised when you pass a parameter that a function does not accept.

mwaskom (approved):

PR #1327, opened 2024-02-09
Install `wheel` in Image.from_registry(add_python=)

ekzhang:
This fixes some issues with `pip_install` failing with PEP517-style installations, due to missing a wheel package, for the `torch` module and some others.

This will invalidate people's cache if they're using `Image.from_registry()`! Which is a bit concerning.

## Changelog

- Install the `wheel` package in `Image.from_registry()` constructors, resolving various installation issues. If you built an image before this version with `Image.from_registry()`, rerunning the app in this version will rebuild your image.


ekzhang:
It's implemented but it will force rebuilds

PR #1328, opened 2024-02-10
Add types to functions.py

savarin:
## Describe your changes

This PR adds mypy types to `functions.py`, `mypy --follow-imports=skip modal/functions.py` will now pass.


modal/functions.py
```diff
@@ -1386,7 +1387,7 @@ async def get_current_stats(self) -> FunctionStats:
 Function = synchronize_api(_Function)
 
 
-class _FunctionCall(_Object, type_prefix="fc"):
+class _FunctionCall(_Object, type_prefix="fc"):  # type: ignore[call-arg]
```
savarin, line 1390: I'm less familiar with `type_prefix`, but on a grep looks like it adds a two char prefix to objects.

savarin:
Hi there! I add types to open source projects for funsies, recently for [instructor](https://github.com/jxnl/instructor/issues/390). If this is helpful, I can create an issue and start from upstream nodes.

modal/functions.py
```diff
@@ -1460,8 +1461,8 @@ async def _gather(*function_calls: _FunctionCall):
 gather = synchronize_api(_gather)
 
 
-_current_input_id = ContextVar("_current_input_id")
-_current_function_call_id = ContextVar("_current_function_call_id")
+_current_input_id: ContextVar = ContextVar("_current_input_id")
+_current_function_call_id: ContextVar = ContextVar("_current_function_call_id")
```
ekzhang, line 1465: This type annotation conflicts with our auto-generated `inv type-stubs` machinery, which generates type stubs for the client library. That's it was omitted. Not sure what the difference between omitting and having it is here?

modal/functions.py
```diff
@@ -924,7 +925,7 @@ def from_parametrized(
         obj,
         from_other_workspace: bool,
         options: Optional[api_pb2.FunctionOptions],
-        args: Iterable[Any],
+        args: Sized,
```
ekzhang, line 928: I haven't heard of this before, what is the difference between `Iterable[Any]` -> `Sized`?

ekzhang (commented):
Hi @savarin — we already have type annotations in our client library, and we run type-checking in CI with `mypy .`

This is passing already so we're curious how you're thinking about going through this or any best practices you were following to edit these types?

savarin:
> Hi @savarin — we already have type annotations in our client library, and we run type-checking in CI with `mypy .`
> 
> This is passing already so we're curious how you're thinking about going through this or any best practices you were following to edit these types?

@ekzhang Got it. Let me try getting the mypy checks in CI to pass locally, using [this](https://github.com/modal-labs/modal-client/actions/runs/7875583349/job/21487764615) successful Github Actions run as example.

In the meantime, I can share that running `mypy --follow-imports=skip` on `modal.functions` gets me these [errors](https://gist.github.com/savarin/36f2fe5001531dd8e9b16d2aeb042fbe) (setting `--python-version=3.8` or `--python-version=3.10` raises about the same). Running `mypy .` gets [this](https://gist.github.com/savarin/b9c11027494b957981481b8552f5f153).

The errors raised appear directionally correct, for example the `secrets` argument being passed to `FunctionEnv` [here](https://github.com/modal-labs/modal-client/pull/1328/commits/16ba69a4fa0a782df9410ebe135531296a44c69c#diff-87e1f4fd05c5e1972a1b8d781a518822704dd8fba567111f9b74ad8a3c9959fdL634) should be a `Sequence` (matching [definition](https://github.com/modal-labs/modal-client/pull/1328/commits/16ba69a4fa0a782df9410ebe135531296a44c69c#diff-87e1f4fd05c5e1972a1b8d781a518822704dd8fba567111f9b74ad8a3c9959fdL533) in `FunctionEnv`) and not a `Collection` (as per [definition](https://github.com/modal-labs/modal-client/pull/1328/commits/16ba69a4fa0a782df9410ebe135531296a44c69c#diff-87e1f4fd05c5e1972a1b8d781a518822704dd8fba567111f9b74ad8a3c9959fdL566) in `Function`). I'll address the `Sized` comment below.

modal/functions.py
savarin: `Iterable` supports any object that can be iterated over (implements `__iter__`) vs `Sized` supports any object that has a length (implements `__len__`). Generator can be iterated over but will fail this line [here](https://github.com/modal-labs/modal-client/pull/1328/commits/16ba69a4fa0a782df9410ebe135531296a44c69c#diff-87e1f4fd05c5e1972a1b8d781a518822704dd8fba567111f9b74ad8a3c9959fdR949).

modal/functions.py
savarin: I'm also less clear why this is raised; I usually see mypy requiring a declaration over compound data structures like lists and dictionaries (say needing `List[bool]` and `Dict[str, int]` vs `List` and `Dict`). Let me try getting `inv mypy` to pass locally and see what else I find.

savarin:
> @ekzhang Got it. Let me try getting the mypy checks in CI to pass locally, using [this](https://github.com/modal-labs/modal-client/actions/runs/7875583349/job/21487764615) successful Github Actions run as example.
> 
> In the meantime, I can share that running locally `mypy --follow-imports=skip` on `modal.functions` gets me these [errors](https://gist.github.com/savarin/36f2fe5001531dd8e9b16d2aeb042fbe) (setting `--python-version=3.8` or `--python-version=3.10` raises about the same). Running `mypy .` gets [this](https://gist.github.com/savarin/b9c11027494b957981481b8552f5f153).
> 
> The errors raised appear directionally correct, for example the `secrets` argument being passed to `FunctionEnv` [here](https://github.com/modal-labs/modal-client/pull/1328/commits/16ba69a4fa0a782df9410ebe135531296a44c69c#diff-87e1f4fd05c5e1972a1b8d781a518822704dd8fba567111f9b74ad8a3c9959fdL634) should be a `Sequence` (matching [definition](https://github.com/modal-labs/modal-client/pull/1328/commits/16ba69a4fa0a782df9410ebe135531296a44c69c#diff-87e1f4fd05c5e1972a1b8d781a518822704dd8fba567111f9b74ad8a3c9959fdL533) in `FunctionEnv`) and not a `Collection` (as per [definition](https://github.com/modal-labs/modal-client/pull/1328/commits/16ba69a4fa0a782df9410ebe135531296a44c69c#diff-87e1f4fd05c5e1972a1b8d781a518822704dd8fba567111f9b74ad8a3c9959fdL566) in `Function`). I'll address the `Sized` comment below.

@ekzhang OK cool got `inv mypy` to pass locally.

I haven't done much type checking against `.pyi` stub files, but my understanding is that this checks that the function signatures in the `.pyi` files are being used correctly. In other words, it's a check against the interfaces rather than the implementation itself.

For example, in the `functions.pyi` file I see the following. This will raise an error if `FunctionEnv` is used downstream with an incorrect type for the argument `mounts`.

```python
class FunctionEnv:
    ...

    def __init__(self, image: typing.Union[modal.image._Image, None], mounts: typing.Sequence[modal.mount._Mount], ...) -> None:
        ...
```

However suppose a `modal.mount._Mount` object is appended to `FunctionEnv.mounts`, this should raise a type error because `Sequence` is read-only, which will be raised under a standard `mypy` check but not for the check against `.pyi` files.

```python
class FunctionEnv:
    ...

    def __init__(self, image: typing.Union[modal.image._Image, None], mounts: typing.Sequence[modal.mount._Mount], ...) -> None:
        mounts.append(modal.mount._Mount(...))
```

Now whether checking `.pyi` is sufficient (enforcing type constraints only with `mypy modal/functions.pyi`) vs checking the implementation (enforcing behavioral constraints like mutability with `mypy modal/functions.py`), that's the team's call.

I just wanted to raise my hand to offer to help out checking that implementation is typed correctly, should this be desired. What I've done in the past is add the types file-by-file, and ensure CI only runs on the files that have been typed.

ekzhang:
First, thanks so much for contributing here!

Next, with this in mind we're looking into the differences in type annotations between the `.pyi` and `.py` files. There definitely seems to be some slightly inconsistent type annotations that we haven't thoroughly audited yet. :)

savarin:
> First, thanks so much for contributing here!
> 
> Next, with this in mind we're looking into the differences in type annotations between the `.pyi` and `.py` files. There definitely seems to be some slightly inconsistent type annotations that we haven't thoroughly audited yet. :)

OK I see this now - there's a bit more 'magic' here that I expected (with stubs codegen from protos etc), very cool though. I was curious what the difference was between `contextvars.ContextVar` and `_contextvars.ContextVar`, ChatGPT had this to say:

> This is because the stub generator recognizes ContextVar as coming from the _contextvars module at a lower implementation level. Python's contextvars module is a high-level interface that may use _contextvars internally, and the stub generator opts to reference this more specific implementation detail in the generated stub file.
> 
> However, this should not affect the usability or the correctness of your stubs. From a type-checking perspective, _contextvars.ContextVar and contextvars.ContextVar are considered equivalent, assuming the type checker properly respects the relationship between these modules. The stub files are primarily for type checkers and tools that analyze your code rather than for direct use in your Python runtime.

Sounds good, will keep an eye out for changes!

modal/functions.py
freider: An explanation as for why this makes a difference:

We use synchronicity's type-stub generation to emit .pyi-files for these files in order to get type info for the runtime-generated blocking/async interfaces of `Function` and not just the internal `_Function` type.

To do that, synchronicity reads annotations, conditionally transforms them, and then re-emits them in the output .pyi-files. In the case of module globals they are only emitted in the .pyi-file if they are explicitly annotated in the original file, so without this annotation here these variables aren't even included in the .pyi-files

modal/functions.py
```diff
@@ -1260,7 +1263,7 @@ async def remote_gen(self, *args, **kwargs) -> AsyncGenerator[Any, None]:
         async for item in self._call_generator(args, kwargs):  # type: ignore
             yield item
 
-    def call(self, *args, **kwargs) -> Awaitable[Any]:
+    def call(self, *args, **kwargs) -> Awaitable[Any]:  # type: ignore[return]
```
freider, line 1266: Why was this added? It also won't get transferred to the generated type stubs in the pyi-file so it shouldn't make a difference

savarin: `mypy` sees this function as returning `None`. This doesn't raise error when we type check against stubs but does so when we type check the implementation (which this PR is suggesting we do).

To avoid confusion I'll close out this PR for now.

freider: Ah, I see now that it's the deprecated/non-function `call` method you're changing. We might as well just remove the return type (or change it to `None`). No need to close the PR, I think it's valuable and we'll merge it as soon as we've ironed out the questionmarks :)

freider:
I was quite surprised that mypy appears to entirely skip the implementation checks since we don't directly refer to our `.py`-files and when it does directory traversal it will prefer to use the `pyi`-files. Realizing there are quite a few places we need to fix in our implemenation to get rid of incorrect types or similar, so thanks for bringing attention to that!

I'd be happy to merge this if we get rid of the # type: ignore additions which I don't think are needed? (I seem to get no mypy warnings for the functions.py when removing them and the incorrect return type on `call`)

modal/functions.py
savarin: Got it! My understanding here is to incorporate these changes, let me know otherwise.

That being said, I do see these in my `.pyi` file (after running `inv protoc` and `inv type-stubs`:

```python
_current_input_id: _contextvars.ContextVar

_current_function_call_id: _contextvars.ContextVar
```

modal/functions.py
savarin: @freider I've removed this. I was getting mypy error previously but might also have been when I was doing mypy checks before running inv etc.

savarin:
> I was quite surprised that mypy appears to entirely skip the implementation checks since we don't directly refer to our `.py`-files and when it does directory traversal it will prefer to use the `pyi`-files. Realizing there are quite a few places we need to fix in our implemenation to get rid of incorrect types or similar, so thanks for bringing attention to that!
> 
> I'd be happy to merge this if we get rid of the # type: ignore additions which I don't think are needed? (I seem to get no mypy warnings for the functions.py when removing them and the incorrect return type on `call`)

@freider Glad to hear this is helpful! I've since removed the type ignores, agreed no longer seeing errors now.

I've seen codebases running `.pyi` checks or running `.py` checks but I don't believe I've seen both - hence my thought of putting this PR on hold pending discussion. Happy to merge this as is, one idea is to add an additional mypy check on an allowlist of files that succeed `.py` type checks (example [here](https://github.com/jxnl/instructor/blob/504c15374ef4d88521a084aa9cd7502ac18fa5bf/.github/workflows/mypy.yml#L12-L29)).

savarin:
@freider I added [this](https://github.com/modal-labs/modal-client/pull/1328/commits/eb790491a7b25080e5321b24f1a006f4d8d2b27d#diff-2e746c9eacad6e2c2fdefdc6a665b5ce4607ea384ebf97a729044b79acc778d6R49) additional `.py` mypy check. What's a tad strange is I now need to add this [line](https://github.com/modal-labs/modal-client/pull/1328/commits/eb790491a7b25080e5321b24f1a006f4d8d2b27d#diff-3b14aa4f3f0def234612480b5e9928a776a92307732d01e8daadf0dc9a6e6e32R53).

modal/functions.py
freider: Those should be fine

freider:
> @freider I added [this](https://github.com/modal-labs/modal-client/pull/1328/commits/eb790491a7b25080e5321b24f1a006f4d8d2b27d#diff-2e746c9eacad6e2c2fdefdc6a665b5ce4607ea384ebf97a729044b79acc778d6R49) additional `.py` mypy check. What's a tad strange is I now need to add this [line](https://github.com/modal-labs/modal-client/pull/1328/commits/eb790491a7b25080e5321b24f1a006f4d8d2b27d#diff-3b14aa4f3f0def234612480b5e9928a776a92307732d01e8daadf0dc9a6e6e32R53).

I think that's unrelated to your change - I'm getting that warning too when I run mypy locally on main. It's due to the mypy version of the repo being pinned to version ~1.4, and github actions caching has probably cached an older version than 1.8 that I have locally which triggers this warning. Should upgrade mypy in the requirements file too, but I can do that separately

freider (approved):

savarin:
@freider I'm afraid I can't merge (I don't have write access), let me know if there's anything else I can do to be helpful. Thanks!

freider:
Thanks for the contribution, it's merged now!

savarin:
> Thanks for the contribution, it's merged now!

@freider Would it be helpful if I did similar changes to other files?

freider:
@savarin Definitely, much appreciated! :)

savarin:
@freider Sounds good, I created the issue https://github.com/modal-labs/modal-client/issues/1372 and will start from the top of the dependency tree.

PR #1338, opened 2024-02-12
Use SecretGetOrCreate for all other use cases

erikbern:
Continuation of #1337 – I want to merge that one first since it's lower risk.

This one, like the other PR, depends on server changes to be deployed.

Will run integration tests with this one.

PR #1340, opened 2024-02-12
Move the secret deployment code to _Secret

erikbern:
This just helps me proactively fix a bunch of integration tests that #1338 otherwise breaks


erikbern:
@prbot approve

PR #1341, opened 2024-02-12
Enable checkpointing in @enter lifecycle methods

mwaskom:
## Describe your changes

This provides a basic implementation for checkpointing `@enter` methods, which run some class setup logic prior to function checkpointing. For example this could be useful for loading model weights into CPU memory

```python
@stub.cls(enable_checkpointing=True)
class Model:
    def __init__(self, name: str):
        self.name

    @enter(checkpoint=True)
    def load_model(self):
        self.model = AutoModel.from_pretrained(self.name)

    @enter()
    def put_model_on_gpu(self):
        self.model.to("cuda")

    @method()
    def inference(self, x):
        return self.model(x)
```

In this example, the function will be checkpointed after the class has been instantiated and the `load_model` method has executed. After resuming from the checkpoint, the `put_model_on_gpu` method will run and then the `inference` method will be available for remote execution.

There are some additional usability aspects of method checkpointing (for example, neither the internet nor GPUs are available while making a checkpoint, although the error messages that you get by default don't make this very clear). Ideally we'll sort those out somewhat; I'm curious about feedback on the basic implementation first, though.

Resolves MOD-2170

This change does not affect client <> server communication

## Changelog

- The `@enter()` lifecycle method can now be used to run additional setup code prior to function checkpointing (when the class is decorated with `stub.cls(enable_checkpointing=True)`. Note that there are currently some limitations on function checkpointing:
  - Checkpointing only works for CPU memory; any GPUs attached to the function will not available
  - Networking is disabled while the checkpoint is being created
- Please note that function checkpointing is still a beta feature.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -46,7 +46,7 @@ from .cls import Cls
 from .config import config, logger
 from .exception import InvalidError
 from .functions import Function, _Function, _set_current_context_ids, _stream_function_call_data
-from .partial_function import _find_callables_for_obj, _find_partial_methods_for_cls, _PartialFunctionFlags
+from .partial_function import _find_callables_for_obj, _PartialFunctionFlags
 
 if TYPE_CHECKING:
     from types import ModuleType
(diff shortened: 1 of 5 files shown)


modal/_container_entrypoint.py
```diff
@@ -861,6 +848,26 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
             dep_object_ids: list[str] = [dep.object_id for dep in container_args.function_def.object_dependencies]
             container_app.hydrate_function_deps(imp_fun.function, dep_object_ids)
 
+        # Identify all "enter" methods and split into groups that should run pre/post checkpointing
```
mwaskom, line 851: This feels messy and I would like to refactor some of the code in `_partial_functions` to push the logic over there but it wasn't trivial so I wanted to make progress on testing basic functionality.

erikbern:
Nice – looks pretty clean and a lot less code than I expected!

modal/_container_entrypoint.py
mwaskom: Refactored by splitting the enum to have `ENTER_PRE_CHECKPOINT` and `ENTER_POST_CHECKPOINT` members. This feels less clean than the previous approach but in lieu of a larger refactor of the code that uses these flags, it ends up being a lot simpler.

modal_test_support/functions.py
```diff
@@ -274,3 +274,21 @@ def build2(self):
     @method()
     def f(self, x):
         return self._k * x
+
+
+@stub.cls(checkpointing_enabled=True)
```
luiscape, line 279: Can you update the class to test for multiple checkpointed enter methods? The test should verify that they are executed in order. Wdyt?

luiscape (approved):
LGTM -- added a small request to update test.

PR #1342, opened 2024-02-13
Adds test that we don't actually expand mount files until an app is run

freider:
This is correct and it's the current behavior, but in order to do some larger refactors of mounts in a safe way we should add this test to ensure we don't regress.

What would be bad is if mounts are loaded at module import time, since they can potentially inspect a large amount of files in the file system which can take a significant amount of time, making imports themselves slow.

## Describe your changes


freider:
@prbot approve

PR #1344, opened 2024-02-13
Add help to `--force` option

ekzhang:


erikbern:
Thanks – not sure why I didn't get it working before. I must have been confused about typer somehow.

PR #1346, opened 2024-02-13
adds env/config-level control of force_build for all images

charlesfrye:
## Describe your changes

Add config-level setting of `force_build` to `true` for monitoring/debugging, e.g. so that we can catch flaky behavior of example builds in synmon.

## Changelog

Adds a new environment variable/config setting, `MODAL_FORCE_BUILD`/`force_build`, that coerces all images to be built from scratch, rather than loaded from cache.


modal/config.py
```diff
@@ -56,6 +56,10 @@
   Defaults to True.
   By default, Modal automatically mounts modules imported in the current scope, that
   are deemed to be "local". This can be turned off by setting this to False.
+* ``force_build`` (in the .toml file) / ``MODAL_FORCE_BUILD`` (as an env var).
+  Defaults to False.
+  By default, Modal automatically caches and re-uses container images. This can be
+  turned off by setting this to True. Meant to be used for debugging and profiling.
```
ekzhang, line 62: I think we should probably not document this one, it's not meant for users to use.

ekzhang (commented):
Looks good! Changelog sounds a bit wordy. See https://modal.com/docs/reference/changelog

PR #1347, opened 2024-02-13
Checks for open connections before checkpointing

luiscape:
Adds a check for remote connections. This informs user that checkpointing isn't possible if connections are open. It also shows which targets currently have connections open.

This is what users will see:
```python
@stub.cls(checkpointing_enabled=True,
          concurrency_limit=1, retries=0, container_idle_timeout=2)
class ConnectionOpen:
    def __init__(self):
        self.new_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    
    @modal.enter(checkpoint=True)
    def open_connection(self):
        remote_ip = socket.gethostbyname("modal.com")
        self.new_socket.connect((remote_ip, 80))
    
    @modal.method()
    def foo(self):
        print("hello from remote function")
        return "done"    
```

```shell
$ modal run .local/hello_checkpoint.py 
✓ Initialized. View run at http://localhost:7601/luiscape/apps/ap-4kH9suzabkIPSdNZtRIeIt
✓ Created objects.
├── 🔨 Created mount /home/ubuntu/modal/.local/hello_checkpoint.py
├── 🔨 Created ConnectionOpen.foo.
└── 🔨 Created f.
Starting function memory checkpointing. Worker fingerprint 37616ec2dbf3210613d0f1916a3236610c64845ebb6dc68c36299b8ddfba0f65
2024-02-13T20:48:54+0000 Found 1 open network connections:
2024-02-13T20:48:54+0000 Remote Address: 54.167.233.8:80, Status: ESTABLISHED
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/pkg/modal/_container_entrypoint.py", line 916, in <module>
    main(container_args, client)
  File "/usr/local/lib/python3.11/contextlib.py", line 222, in __aexit__
    await self.gen.athrow(typ, value, traceback)
  File "/pkg/modal/_container_entrypoint.py", line 136, in heartbeats
    yield
  File "/pkg/modal/_container_entrypoint.py", line 874, in main
    function_io_manager.checkpoint()
  File "/pkg/synchronicity/synchronizer.py", line 526, in proxy_method
    return wrapped_method(instance, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pkg/synchronicity/combined_types.py", line 28, in __call__
    raise uc_exc.exc from None
[...]

PR #1357, opened 2024-02-15
Re-implements network check with vendored `psutil`

luiscape:
Adds the same logic from [this PR](https://github.com/modal-labs/modal-client/pull/1347) but with vendored `psutil` so we don't have to install an additional package.


ekzhang:
Can you include the copyright notice?

https://github.com/giampaolo/psutil?tab=BSD-3-Clause-1-ov-file

vendor/psutil.py
```diff
@@ -0,0 +1,356 @@
+# Copyright Modal Labs 2022
```
mwaskom, line 1: Wrong!

mwaskom:
Also can we prefix the location with an underscore (e.g. `_vendor` so that it doesn't get confused for public API?

mwaskom:
What is our approach to tests going to be with vendored code? Are we just going to assume that our existing unit tests cover all the behaviors that we need?

mwaskom:
It also feels like we want some way of tracking what version the vendored code is sourced from. How will we know when we need to update it, e.g. if there is an important bug in the source repository?

ekzhang:
While these are reasonable comments, I don't think they are relevant to this PR, as the use of `psutil` here is just a particular function that just reads and parses `/proc`. We can revisit it if we need more functionality.

mwaskom:
Now is probably the only time we will know what version this corresponds to though :)

ekzhang:
It shouldn't matter what version it corresponds to — barring any changes to the POSIX API, it should not need updates.

luiscape:
> Are we just going to assume that our existing unit tests cover all the behaviors that we need?

This should be the case, yes. The [tests I added](https://github.com/modal-labs/modal-client/pull/1357/files#diff-91e91e76633aefa71ffbbd8080e1176cc7f61a2d01e3d387cb1f05fc058a4f40R915) test our interface and assumes the vendored code works as it should. That's is acceptable imo.

> It also feels like we want some way of tracking what version the vendored code is sourced from.

I'll add a note to comments if only to point to the right license / code version. I agree with Eric that anything more is overkill.

PR #1359, opened 2024-02-16
Standardize jupyter and vscode names

ajmarcus:
## Describe your changes

Small change: 
- standardize naming for the `launch` jupyter & vscode examples. 

In particular, I was confused when running the vscode example since the stub's function name was `run_jupyter()`:

Screenshot 2024-02-16 at 2 14 00 PM

Totally ok if you all think this isn't the best approach. Is there a way to test these changes? Didn't see any existing tests.

## Changelog

standardize naming for the jupyter & vscode examples


luiscape (approved):
LGTM -- any issues @ekzhang?

ekzhang:
Ah — this doesn't quite work because programs with the name `jupyter.py` can cause issues with Python's import system because they end up conflicting with the actual `jupyter` library.

We also need to refactor this code in the future to not create a temp folder and dump code in so I'll leave it until then!

ekzhang:
We're fixing the name of the app in #1627

PR #1361, opened 2024-02-17
Persist event loop across cls method execution

mwaskom:
## Describe your changes

Fixes a regression in #1341 which caused different event loops to be used when executing `@enter` and `@method` methods. (Unrelated but we need an easier way to talk about the "@method method").

Now the event loop is persisted with a context manager.

- Resolves MOD-2383

## Changelog

- Fixes a regression from 0.57.40 where `@enter` methods used a separate event loop.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -61,7 +61,7 @@ class UserException(Exception):
     pass
 
 
-class EventLoop:
+class SignalHandlingEventLoop:
     """Manage an event loop for executing coroutines while handling SIGINT/SIGTERM.
 
     Prevents stray cancellation errors from propagating up.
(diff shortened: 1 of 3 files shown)


modal/_container_entrypoint.py
```diff
@@ -61,20 +61,24 @@ class UserException(Exception):
...
+        self.loop = asyncio.new_event_loop()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.loop.close()
+
+    def run(self, coro):
+        task = asyncio.ensure_future(coro, loop=self.loop)
+        for s in [signal.SIGINT, signal.SIGTERM]:
+            self.loop.add_signal_handler(s, task.cancel)
```
mwaskom, line 80: Not totally clear to me whether, now that we're going to run multiple coroutines, if we should remove existing signal handlers when this method is called for the second+ time.

aksh-at (approved):
nice, thanks for fixing! Is it easy to add a test that checks if `get_running_loop()` is the same between `@enter` and the `@method` method?

mwaskom:
> nice, thanks for fixing! Is it easy to add a test that checks if `get_running_loop()` is the same between `@enter` and the `@method` method?

Oh yeah good thinking, I'll adapt my reprex (doesn't do exactly what you say but I think it covers the failure mode properly).

aksh-at:
awesome, thanks!

PR #1363, opened 2024-02-18
web endpoint optimization: or, the unreasonable effectiveness of asyncio.gather?

thundergolfer:
### Describe your changes

- In **local dev**, this reduces the high freq web endpoint's execution latency from ~9ms to ~3ms (66% latency reduction).  
- In **dev cluster**, this reduces the high freq web endpoint's execution latency from ~25ms to ~3ms (**88%** latency reduction). 
    - We can expect similar results in prod

There's no behavior change in this PR. The `generator_output_task` will still not exit until it receives `_GENERATOR_STOP_SENTINEL`. But with the concurrent coro creation, the sentinel receive happens more or less instantly.

### Repro

Take the highfreq `webhook.py` and deploy it with this change using `MODAL_SYNC_ENTRYPOINT=1`. Then hit it with `hyperfine` and observe the execution times in the app's logs. 

`hyperfine -N --runs 100 'curl https://thundergolfer--highfreq-webhook-synthetic-monitor-webhook.modal-dev.run'`

## Changelog

* Performance optimization to web endpoints, reducing 5-10ms of overhead.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -596,9 +596,7 @@ async def call_function_async(
             started_at = time.time()
             reset_context = _set_current_context_ids(input_id, function_call_id)
             async with function_io_manager.handle_input_exception.aio(input_id, started_at):
-                print(f"{time.time() - started_at:.5f}s before calling fn")
                 res = imp_fun.fun(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


thundergolfer:
did I `await` that gather?

image

PR #1364, opened 2024-02-18
add concurrency to reduce web endpoint latency

thundergolfer:
## Describe your changes

- In **local dev**, this reduces the high freq web endpoint's execution latency from ~9ms to ~3ms (66% latency reduction).  
- In **dev cluster**, this reduces the high freq web endpoint's execution latency from ~25ms to ~3ms (**88%** latency reduction). 
    - We can expect similar results in prod

### Repro

Take the highfreq `webhook.py` and deploy it with this change using `MODAL_SYNC_ENTRYPOINT=1`. Then hit it with `hyperfine` and observe the execution times in the app's logs. e.g.

`hyperfine -N --runs 100 'curl https://thundergolfer--highfreq-webhook-synthetic-monitor-webhook.modal-dev.run'`

## Changelog

* Performance optimization to web endpoints, reducing 5-10ms of overhead.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -554,9 +554,9 @@ def call_function_sync(
                         item_count += 1
 
                     function_io_manager._queue_put(generator_queue, _FunctionIOManager._GENERATOR_STOP_SENTINEL)
-                    generator_output_task.result()  # Wait to finish sending generator outputs.
                     message = api_pb2.GeneratorDone(items_total=item_count)
                     function_io_manager.push_output(input_id, started_at, message, api_pb2.DATA_FORMAT_GENERATOR_DONE)
+                    generator_output_task.result()  # Wait to finish sending generator outputs.
                 else:
                     if inspect.iscoroutine(res) or inspect.isgenerator(res) or inspect.isasyncgen(res):
                         raise InvalidError(
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -618,7 +618,6 @@ async def call_function_async(
                         await function_io_manager._queue_put.aio(generator_queue, value)
                         item_count += 1
 
-                    message = api_pb2.GeneratorDone(items_total=item_count)
                     await function_io_manager._queue_put.aio(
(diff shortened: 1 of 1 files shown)


modal/_container_entrypoint.py
```diff
@@ -554,9 +554,9 @@ def run_inputs(input_id: str, function_call_id: str, args: Any, kwargs: Any) ->
                         item_count += 1
 
                     function_io_manager._queue_put(generator_queue, _FunctionIOManager._GENERATOR_STOP_SENTINEL)
-                    generator_output_task.result()  # Wait to finish sending generator outputs.
                     message = api_pb2.GeneratorDone(items_total=item_count)
                     function_io_manager.push_output(input_id, started_at, message, api_pb2.DATA_FORMAT_GENERATOR_DONE)
+                    generator_output_task.result()  # Wait to finish sending generator outputs.
```
thundergolfer, line 559: Don't know if this change is of much benefit to sync generators.
The equivalent change in the async path is efficacious because it allows the HTTP response to return more quickly to the client.

modal/_container_entrypoint.py
```diff
@@ -621,10 +621,19 @@ async def run_input(input_id: str, function_call_id: str, args: Any, kwargs: Any
...
+                    # will correctly report the total number of items to receive.
                     message = api_pb2.GeneratorDone(items_total=item_count)
-                    await function_io_manager.push_output.aio(
-                        input_id, started_at, message, api_pb2.DATA_FORMAT_GENERATOR_DONE
+                    await asyncio.gather(
+                        generator_output_task,
+                        function_io_manager.push_output.aio(
+                            input_id, started_at, message, api_pb2.DATA_FORMAT_GENERATOR_DONE
+                        ),
                     )
```
ekzhang, line 637: I would just move `await generator_output_task` one line down

ekzhang (changes requested):
Discussed in-person but it's desirable to make sure that all data chunks are actually pushed before sending the GENERIC_RESULT_SUCCESS / `GeneratorDone` output

thundergolfer:
Thanks for review @ekzhang. Closing as this isn't correct behavior in the failure case.

PR #1369, opened 2024-02-20
Do not merge: Graceful input cancellation - client changes

freider:
Change list:
* Adjust heartbeat loop for longpolling heartbeats
* Add ContainerHeartbeatResponse which may or may not contain a cancellation event
* If there is a cancellation event in the container heartbeat response, deal with the cancellation:
    * If concurrency > 1: Kill the task from within - we don't support graceful termination for those yet
    * Otherwise, check if current input is in the set of input ids to cancel - if so, signal a SIGUSR1 to self, which will cause an exception to be raised in the main thread
    * Catch this exception after it has stopped user code, and continue on with the next input (the expectation is that the cancelled inputs are all marked as cancelled in the db already so no need to send back a result)
    * In case of async functions, the running user task is cancelled with a custom message instead of raising an exception

## Describe your changes

There are a bunch of potential compatibility issues since this functionality requires rollout of three different components:
* Client
* Server
* Worker

## Changelog

* Function calls can now be cancelled without killing the container running the inputs. This allows new inputs by different function calls to the same function to be picked up immediately without having to cold-start new containers after cancelling calls.


modal/_container_entrypoint.py
```diff
@@ -395,11 +449,28 @@ async def handle_user_exception(self) -> AsyncGenerator[None, None]:
...
+            # won't be retried
+            logger.info(f"The current input ({input_id=}) was cancelled by a user request")
+            await self.complete_call(started_at)
+            return
         except BaseException as exc:
+            if isinstance(exc, asyncio.CancelledError) and (
+                str(exc) == INPUT_CANCELLATION_MESSAGE or _ignore_cancellation
+            ):
+                _ignore_cancellation = False  # reset for next cancellations
+                # for async functions, InputCancellation is represented by CancelledError with the INPUT_CANCELLATION_MESSAGE message
```
mwaskom, line 469: Just once it would be nice to do things the same way for sync / async functions :)

modal/exception.py
```diff
@@ -159,3 +159,13 @@ def simulate_preemption(wait_seconds: int, jitter_seconds: int = 0):
     signal.signal(signal.SIGALRM, _simulate_preemption_interrupt)
     jitter = random.randrange(0, jitter_seconds) if jitter_seconds else 0
     signal.alarm(wait_seconds + jitter)
+
+
+class InputCancellation(BaseException):
+    """Raised when the current input is cancelled by the task
+
+    Intentionally a BaseException instead of an Exception, so it won't get
+    caught by unspecified user exception clauses that might be used for retries etc.
```
mwaskom, line 168: Nice

modal_proto/api.proto
```diff
@@ -551,9 +551,18 @@ message ContainerArguments {  // This is used to pass data from the worker to th
...
+}
+
+message ContainerHeartbeatResponse {
+  optional CancelInputEvent cancel_input_event = 1;
+}
+
 message ContainerHeartbeatRequest {
   string current_input_id = 1;
   double current_input_started_at = 2;
+  bool supports_graceful_input_cancellation = 3; // implicitly also supports long poll
```
mwaskom, line 565: j/w but is the idea that we need this so that we can tell whether the request came from a sufficiently new client? would we be able to remove it at some point in the future?

mwaskom (approved):
Nice! Didn't look too closely at the new test code but the changes in the client seem pretty manageable!

modal/_container_entrypoint.py
freider: Yeah :( The problem here is that if we raise an exception in the main thread like we do for sync code, it would risk being raised inside of the event loop rather than the user function, basically killing the container...
Maybe we should implement our own event loop (/extensions) at some point to facilitate this sort of thing

modal_proto/api.proto
freider: Yeah exactly. There is a change in the responder API for this call that changes it to a long-polling mechanism when this flag is set to true, and in that case it can also return a payload with some data that the older clients wouldn't listen to. When older client versions are no longer supported, we can get rid of this. (Reminds me, I should bump the minor version of the package here too to make the deprecation path clear)

modal/_container_entrypoint.py
```diff
@@ -114,28 +128,64 @@ async def initialize_app(self) -> _ContainerApp:
...
-    async def _heartbeat(self):
+    async def _run_heartbeat_loop(self):
+        while 1:
+            t0 = time.monotonic()
+            if await self._heartbeat():
+                # got a cancellation event, fine to start another heartbeat soon since we know
+                # the worker supports long-polling
+                time_until_next_hearbeat = 1.0
+            else:
+                heartbeat_duration = time.monotonic() - t0
```
freider, line 139: If there is no cancellation *and* heartbeat returns immediately (old worker), this will ensure we wait ~15s before the next heartbeat.
If we are on a new worker however, the call will be longpolling so it wouldn't return until 15s has passed (unless there is a cancellation), so it wouldn't sleep before the next heartbeat

PR #1371, opened 2024-02-20
Use uv for CI step

ekzhang:
Also cleaning up the CI a bit in general


ekzhang:
It doesn't work due to some protobuf issue on Windows 🤔

image

Issue #1372, opened 2024-02-20
Increase modal-client type coverage


savarin:
This is an umbrella issue created to add types to `modal-client`.

We expect changes to be made in two stages:

1. Expand type coverage to allow mypy to pass beyond the existing check on `.pyi` stubs.
2. Implement strict type checking, as well as remove `Any` and type ignores (where possible) above.

Existing dependency tree as generated with `pydeps` and the following `.pydeps` file.

```
[pydeps]
pylib = False
exclude =
    aiohttp
    aiostream
    aiostream
    asgiref
    click
    cloudpickle
    dotenv
    fastapi
    grpclib
    ipykernel
    rich
    synchronicity
    tblib
    toml
    typer
    typing_extensions
    watchfiles
```

![modal](https://github.com/modal-labs/modal-client/assets/5699412/44c995cc-3534-4fd7-93d3-906e27926abc)

zsiegel92:
Hey @savarin  FYI I opened [this related PR](https://github.com/modal-labs/modal-client/pull/2145) and filed [this related issue](https://github.com/modal-labs/modal-client/issues/2147).

If you have any availability to chip away at this I'll be so grateful! 🙏 Thanks!

mwaskom:
We're making gradual progress here — doing a little repo cleanup and going to close this as an open ticket. Thanks!

PR #1374, opened 2024-02-20
Add types to modal_version and modal_utils

savarin:
## Describe your changes

This PR adds mypy types to `modal_version` and `modal_utils`, as per https://github.com/modal-labs/modal-client/issues/1372.

```
modal_utils/__init__.py
modal_utils/app_utils.py
modal_utils/async_utils.py
modal_utils/grpc_testing.py
modal_utils/grpc_utils.py
modal_utils/hash_utils.py
modal_utils/http_utils.py
modal_utils/logger.py
modal_utils/package_utils.py
modal_utils/rand_pb_testing.py
modal_version/__init__.py
modal_version/_version_generated.py
```

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -196,7 +196,7 @@ async def _stream_function_call_data(
     while True:
         req = api_pb2.FunctionCallGetDataRequest(function_call_id=function_call_id, last_index=last_index)
         try:
-            async for chunk in unary_stream(stub_fn, req):
+            async for chunk in unary_stream(stub_fn, req):  # type: ignore[var-annotated]
                 if chunk.index <= last_index:
                     continue
                 last_index = chunk.index
--- a/tasks.py
+++ b/tasks.py
@@ -43,6 +43,18 @@ def lint(ctx):
 def mypy(ctx):
     mypy_allowlist = [
         "modal/functions.py",
+        "modal_utils/__init__.py",
+        "modal_utils/app_utils.py",
+        "modal_utils/async_utils.py",
+        "modal_utils/grpc_testing.py",
+        "modal_utils/grpc_utils.py",
+        "modal_utils/hash_utils.py",
+        "modal_utils/http_utils.py",
+        "modal_utils/logger.py",
+        "modal_utils/package_utils.py",
+        "modal_utils/rand_pb_testing.py",
+        "modal_version/__init__.py",
+        "modal_version/_version_generated.py",
     ]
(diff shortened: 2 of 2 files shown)


modal/functions.py
```diff
@@ -196,7 +196,7 @@ async def _stream_function_call_data(
     while True:
         req = api_pb2.FunctionCallGetDataRequest(function_call_id=function_call_id, last_index=last_index)
         try:
-            async for chunk in unary_stream(stub_fn, req):
+            async for chunk in unary_stream(stub_fn, req):  # type: ignore[var-annotated]
```
savarin, line 199: @freider I'm a little puzzled why this now raises a mypy error, since `unary_stream` returns `AsyncIterator[_RecvType]` https://github.com/modal-labs/modal-client/blob/83e90f273bfcb6b10c568f17bf7baca0e366e903/modal_utils/grpc_utils.py#L205.

Not sure why this isn't picked up in `functions.py`.

freider: I think it's due to `--follow-imports=skip`, so the imported type isn't available (removing that flag, this error disappears but you get tons of errors *in the imported files* instead...). I wonder if there is a way to check a single file for type errors but still follow imports in order to get types

freider:
I updated the repo to use `pyright` instead of mypy for checking implemenation files, since `mypy` seems basically broken in this regard. See: https://github.com/modal-labs/modal-client/pull/1403

savarin:
> I updated the repo to use `pyright` instead of mypy for checking implemenation files, since `mypy` seems basically broken in this regard. See: #1403

Sounds good! Sorry for delay, had been away last week and into this week. I'll take a look at #1403.

freider:
> > I updated the repo to use `pyright` instead of mypy for checking implemenation files, since `mypy` seems basically broken in this regard. See: #1403
> 
> @freider Sounds good! Sorry for delay, had been away last week and into this week. I'll take a look at #1403.

Just merged #1403 so it should be straight-forward to take on typing fixes for additional files now if you want to :)

savarin:
> Just merged #1403 so it should be straight-forward to take on typing fixes for additional files now if you want to :)

@freider Thanks for the heavy lifting! I'm afraid I haven't gotten the chance to take a closer look at `pyright` but can move tad slower and go for the low hanging fruit first as I feel my way around - added https://github.com/modal-labs/modal-client/pull/1438

PR #1386, opened 2024-02-21
Update Modal dict reference page notes

yirenlu92:
Add additional information about individual key-value TTL in Modal dicts

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -28,7 +28,7 @@ class _Dict(_Object, type_prefix="di"):
 
     A `Dict` matches the lifetime of the app it is attached to, but individual key-value pairs will expire 30 days after it was last added to its ​​Dict object.
     Because of this, `Dict`s are best not used for
-    long-term storage. All data is deleted when the app is stopped. 
+    long-term storage. All data is deleted when the app is stopped.
 
     **Usage**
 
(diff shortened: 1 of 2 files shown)


modal/dict.py
```diff
@@ -26,9 +26,9 @@ class _Dict(_Object, type_prefix="di"):
 
     **Lifetime of a Dict and its items**
 
-    A `Dict` matches the lifetime of the app it is attached to, but invididual
-    keys expire after 30 days. Because of this, `Dict`s are best not used for
-    long-term storage. All data is deleted when the app is stopped.
+    A `Dict` matches the lifetime of the app it is attached to, but individual key-value pairs will expire 30 days after it was last added to its ​​Dict object.
```
mwaskom, line 29: Nit, tense alignment

```suggestion
but an individual entry will expire 30 days after it was last added to its ​​Dict object.
```

modal/dict.py
```diff
@@ -26,8 +26,8 @@ class _Dict(_Object, type_prefix="di"):
 
     **Lifetime of a Dict and its items**
 
-    A `Dict` matches the lifetime of the app it is attached to, but invididual
-    keys expire after 30 days. Because of this, `Dict`s are best not used for
+    but an individual entry will expire 30 days after it was last added to its ​​Dict object.
```
danielnorberg, line 29: Should this be as below?

> A `Dict` matches the lifetime of the app it is attached to, but an individual entry will expire 30 days after it was last added to its ​​Dict object.

PR #1388, opened 2024-02-21
`modal.experimental.stop_fetching_inputs`

aksh-at:
## Changelog

When called from within a container, `modal.experimental.stop_fetching_inputs()` causes it to gracefully exit after the current input has been processed.

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -74,7 +74,6 @@ __all__ = [
     "forward",
     "is_local",
     "method",
-    "stop_fetching_inputs",
     "web_endpoint",
     "wsgi_app",
     "interact",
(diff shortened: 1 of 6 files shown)


modal/__init__.py
```diff
@@ -74,6 +74,7 @@
     "forward",
     "is_local",
     "method",
+    "stop_fetching_inputs",
```
mwaskom, line 77: If this is exposed in the top-level namespace, won't most users miss that it is `experimental`? Any cost to asking beta testers import it from the full path?

aksh-at: It's not even imported anywhere, so this would not have worked

PR #1398, opened 2024-02-22
[skip ci] Remove logo for now

ekzhang:
Some things about the color, size, and aspect ratio of the logo seem off, and it doesn't work on dark mode. Happy to revisit later! But simple can sometimes be better.


mwaskom:
We could just trim the margins on the logo? It would be nice to have it flush to the left, I don't think that filling the horizontal space is critical.

ekzhang:
Discussed separately outside of GitHub. Happy to have a PR that adds it back later. with proper dark mode rendering and margins and so on. Just think right now it's a bit off, and simple isn't bad.

PR #1399, opened 2024-02-22
Add experimental `SchedulerPlacement` class

irfansharif:
It's hidden in docs, etc. We're going to use it in the new scheduler to very selectively place where tasks can run. Export zone + spot for now too, we can get rid of it or disable it later.


irfansharif:
@prbot approve

PR #1401, opened 2024-02-22
Fix passing exception info to @exit method

mwaskom:
## Describe your changes

We claim to support `__exit__` style exception propagation but because we have already handled any user-code exceptions `sys.exc_info` is null in the `finally` block where we use it. This is a fairly naive fix and there's a lot of complexity here (concurrent inputs, etc.) so I'm not totally confident about it.

I'm also pretty uncertain about how to unit test it. AFAICT we don't really exercise the `exit()` behavior anywhere in the tests, and even in our integration tests, we have a note about not being able to. Open to suggestions.

- Fixes MOD-2417

## Changelog

- Fixed a bug where exception info was not propagated to `@exit` lifecycle methods

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -66,8 +66,7 @@ if TYPE_CHECKING:
     else:
         TracebackType = Any
 
-
-ExceptionInfo = Tuple[Optional[Type[BaseException]], Optional[BaseException], Optional[TracebackType]]
+    ExceptionInfo = Tuple[Optional[Type[BaseException]], Optional[BaseException], Optional[TracebackType]]
 
 MAX_OUTPUT_BATCH_SIZE: int = 49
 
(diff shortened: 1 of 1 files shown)


modal/_container_entrypoint.py
```diff
@@ -149,6 +168,9 @@ async def get_serialized_function(self) -> tuple[Optional[Any], Callable]:
 
         return cls, fun
 
+    def exc_info(self) -> ExceptionInfo:
+        return self._exc_info
```
mwaskom, line 172: I don't really understand why this needs to be a method — I assume it's some synchronicity thing, but when I tried just populating an attribute on the `FunctionIOManager` class it didn't work, the client code couldn't see that the attribute existed.

mwaskom:
Superseded by #1404 — testing this seems pretty challenging and  I think the user value of the feature is either very low or potentially negative.

PR #1402, opened 2024-02-22
Remove references to -musl standalone Python

mwaskom:
## Describe your changes

We realized that images don't build properly on top of the `-musl` standalone Python builds, potentially due to a [bug](https://github.com/python/cpython/issues/81241) in CPython.

This PR deletes the suggestions to use `add_python=f"{version}-musl"` for Alpine Linux images, which I could not get to work.

I've also stopped building the `-musl` base mount. But otherwise preserved the flexibility to have alternate `libc` builds in the future.

- Resolves MOD-2419


ekzhang (approved):
Nice!

PR #1403, opened 2024-02-23
Use pyright for static type checking of implemenation code

freider:
mypy can not and will not get functionality to be able to type check files that have both implemenation and type stubs (both .py and .pyi files). Pyright does seem to support it, so this adds it to the CI pipeline for checking implemenations.

This only adds `functions.py` to the whitelist of files to test, since it's quite an undertaking to fix all of the client code at once. Once we have fixes for all files we can let pyright traverse the entire repo instead of using this whitelist.

If I understand things correctly, pyright is also what's used as the python language server by vscode, so getting rid of type warnings for vscode devs is another benefit.

Initial batch of fixes for: #1372 and MOD-2387

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -34,7 +34,7 @@ from grpclib.exceptions import StreamTerminatedError
 from synchronicity.exceptions import UserCodeException
 
 from modal import _pty
-from modal_proto import api_pb2
+from modal_proto import api_grpc, api_pb2
 from modal_utils.async_utils import (
     queue_batch_iterator,
     synchronize_api,
@@ -228,13 +228,13 @@ class _OutputValue:
 class _Invocation:
     """Internal client representation of a single-input call to a Modal Function or Generator"""
 
-    def __init__(self, stub, function_call_id, client=None):
+    def __init__(self, stub: api_grpc.ModalClientStub, function_call_id: str, client: _Client):
         self.stub = stub
         self.client = client  # Used by the deserializer.
         self.function_call_id = function_call_id  # TODO: remove and use only input_id
 
     @staticmethod
-    async def create(function_id: str, args, kwargs, client: _Client):
+    async def create(function_id: str, args, kwargs, client: _Client) -> "_Invocation":
         assert client.stub
         item = await _create_input(args, kwargs, client)
 
@@ -560,7 +560,7 @@ class _Function(_Object, type_prefix="fu"):
     """
 
     # TODO: more type annotations
-    _info: FunctionInfo
+    _info: Optional[FunctionInfo]
     _all_mounts: Collection[_Mount]
     _stub: "modal.stub._Stub"
     _obj: Any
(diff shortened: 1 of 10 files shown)


modal/functions.py
```diff
@@ -817,44 +832,44 @@ async def _load(provider: _Function, resolver: Resolver, existing_object_id: Opt
...
                 ),
                 volume_mounts=volume_mounts,
                 proxy_id=(proxy.object_id if proxy else None),
                 retry_policy=retry_policy,
-                timeout_secs=timeout_secs,
-                task_idle_timeout_secs=container_idle_timeout,
-                concurrency_limit=concurrency_limit,
+                timeout_secs=timeout_secs or 0,
+                task_idle_timeout_secs=container_idle_timeout or 0,
+                concurrency_limit=concurrency_limit or 0,
```
mwaskom, line 854: IIRC there was some discussion recently that container timeouts _can't_ be set to 0. I don't know if that's really a problem here but seems kinda weird.

mwaskom (commented):
This is a good example of why retroactive type annotation in Python can look so messy; I worry that some of the `assert obj` calls or secret defaults could end up leaking confusing failures to users. But I guess there's not much we can do?

modal/functions.py
freider: These defaults additions were previously implicit due to protobuf's serialization format, since nullity (?) isn't really a thing there unless you box your primitive types in a `message` type or similar. So for example `task_idle_timeout_secs` is currently determined to be unset by the backend when it's `0`, which isn't great since it prevents us from setting an actual 0-value. According to the docs we have a minimum value of `2` seconds, which is arguably even weirder :D

freider:
Yeah it's not very pretty. Perhaps we can refactor the code at some point to make a few of the "connected" types like `client` have a separate type for their connected/initialized version and unconnected versions, and then make sure all "active" wrappers like FunctionCall only contain non-optional references to the connected versions, which would get rid of some of the assertions.

But this should at least not lead to any errors that weren't previously errors further down the call stack :)

setup.cfg
```diff
@@ -19,7 +19,7 @@ packages = find:
 python_requires = >=3.8
 install_requires =
     aiohttp
-    aiostream
```
freider, line 22: This shouldn't trigger image rebuilds right @mwaskom ? Afaicr only the modal/requirements.txt file does that 🤔

mwaskom: That’s my understanding!

savarin:
@freider Interesting, I didn't realize `pyright` reviews both `.py` and `.pyi` files. I still don't have a good feel for what error it raises so I'll play around with it a bit more.

One thought - `modal.functions` is a bit further down the dependency tree, might be easier to start off at the top like in https://github.com/modal-labs/modal-client/pull/1374.

freider:
@prbot approve

PR #1404, opened 2024-02-23
Support unparameterized methods decorated with @exit

mwaskom:
## Describe your changes

Alternative solution to https://github.com/modal-labs/modal-client/pull/1401

The user value of passing exception information to methods decorated with `@exit` is pretty low, and it's fairly difficult to test well. In fact, the user value is potentially negative, in that you need to remember that the method decorated with `@exit` needs parameters that you probably won't use.

So we're going to deprecate the feature. Most of the code for doing that is here, but I've commented out the actual deprecation warning because I need to update internal code and docs first. So in effect, this PR just adds silent support for unparameterized `@exit` methods.

- Part of MOD-2417

### DIFF
--- a/modal_test_support/functions.py
+++ b/modal_test_support/functions.py
@@ -281,7 +281,7 @@ class BuildCls:
         return self._k
 
     @exit()
-    def exit1(self, *args):
+    def exit1(self):
         raise Exception("exit called!")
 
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot lgtm

PR #1417, opened 2024-02-26
Exit from container_exec with modal exceptions

mwaskom:
## Describe your changes

This updates some error handling in the function(s) underlying `modal container exec` and `modal shell` to raise modal exceptions rather than using `typer.Exit` or doing `rich.print` and returning.

The primary goal was to keep the use of `typer` limited to `modal/cli`, but I also think it's more Pythonic to handle errors this way in library functions.

Originally I was hoping to remove `rich` from the libarary functions too; I think that ideally the rich display would be completely owned by the CLI, but that's tricker here; we at least would need to pass in callbacks or some other way of terminating the status spinner when it's time to hand over the terminal to the shell. I tried a few larger refactors but the current situation with these functions being used by both `modal container exec` and `modal shell` left me unable to come up with a satisfactory solution.

- Resolves MOD-2435

### DIFF
--- a/modal/_container_exec.py
+++ b/modal/_container_exec.py
@@ -10,13 +10,13 @@ from typing import List, Optional
 
 import rich
 import rich.status
-import typer
 from grpclib import Status
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 from rich.console import Console
 
 from modal._pty import get_pty_info, raw_terminal, set_nonblocking
 from modal.client import _Client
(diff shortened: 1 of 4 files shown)


ekzhang (commented):
Can we catch the exception and have it print a message, rather than sending the full traceback?

https://clig.dev/#errors

ekzhang (approved):
Discussed that we'll resolve catching exceptions and printing tracebacks separately.

PR #1419, opened 2024-02-27
Stop using SecretCreate in client

erikbern:
This will let us delete the endpoint server-side in like 4-6 months :)


freider (approved):
lgtm!

PR #1422, opened 2024-02-27
Improve error display (with @mwaskom)

ekzhang:
This splits up errors into two classes: user errors and CLI errors.

## CLI errors

Before, CLI errors displayed a long traceback.

image

Now they display just the error message, relevant to the user, but the original behavior can be restored by running with the `MODAL_TRACEBACK=1` config option.

image

## User errors

Before, user code errors in a file loaded by Modal printed out a long traceback.

image

Now, they print out a shorter traceback.

image

### DIFF
--- a/modal/__main__.py
+++ b/modal/__main__.py
@@ -1,13 +1,35 @@
 # Copyright Modal Labs 2022
+import sys
+
 from ._traceback import highlight_modal_deprecation_warnings, setup_rich_traceback
 from .cli.entry_point import entrypoint_cli
+from .cli.import_refs import CliUserExecutionError
+from .config import config
 
 
 def main():
     # Setup rich tracebacks, but only on user's end, when using the Modal CLI.
     setup_rich_traceback()
     highlight_modal_deprecation_warnings()
-    entrypoint_cli()
+
+    if config.get("traceback"):
+        try:
+            entrypoint_cli()
+        except CliUserExecutionError as exc:
+            raise exc.__cause__ from None
+    else:
+        try:
+            entrypoint_cli()
+        except CliUserExecutionError as exc:
+            raise exc.__cause__ from None
+        except Exception as exc:
+            from rich.console import Console
+            from rich.panel import Panel
+
+            console = Console(stderr=True)
+            panel = Panel(str(exc), border_style="red", title="Error", title_align="left")
+            console.print(panel, highlight=False)
+            sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/modal/__main__.py
+++ b/modal/__main__.py
@@ -12,17 +12,14 @@ def main():
     setup_rich_traceback()
     highlight_modal_deprecation_warnings()
 
-    if config.get("traceback"):
-        try:
-            entrypoint_cli()
-        except CliUserExecutionError as exc:
-            raise exc.__cause__ from None
-    else:
-        try:
-            entrypoint_cli()
-        except CliUserExecutionError as exc:
-            raise exc.__cause__ from None
-        except Exception as exc:
+    try:
+        entrypoint_cli()
+    except CliUserExecutionError as exc:
+        raise exc.__cause__ from None
+    except Exception as exc:
+        if config.get("traceback"):
+            raise
+        else:
             from rich.console import Console
             from rich.panel import Panel
 
--- a/modal/cli/import_refs.py
+++ b/modal/cli/import_refs.py
@@ -1,8 +1,10 @@
 # Copyright Modal Labs 2023
-"""Utilities for CLI references to Modal entities
+"""Load or import Python modules from the CLI.
 
 For example, the function reference of `modal run some_file.py::stub.foo_func`
-or the stub lookup of `modal deploy some_file.py`
+or the stub lookup of `modal deploy some_file.py`.
+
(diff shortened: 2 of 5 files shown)


modal/config.py
```diff
@@ -194,6 +193,7 @@ class _Setting(typing.NamedTuple):
     "worker_id": _Setting(),  # For internal debugging use.
     "restore_state_path": _Setting("/__modal/restore-state.json"),
     "force_build": _Setting(False, transform=lambda x: x not in ("", "0")),
+    "traceback": _Setting(False, transform=lambda x: x not in ("", "0")),
```
mwaskom, line 196: Maybe `show_tracebacks`?

modal/__main__.py
```diff
@@ -1,13 +1,35 @@
...
+from .config import config
 
 
 def main():
     # Setup rich tracebacks, but only on user's end, when using the Modal CLI.
     setup_rich_traceback()
     highlight_modal_deprecation_warnings()
-    entrypoint_cli()
+
+    if config.get("traceback"):
```
mwaskom, line 15: Could you move this inside the `except Exception as exc` block so you don't need the duplicate code?

modal/cli/import_refs.py
```diff
@@ -42,6 +44,14 @@ class NoSuchObject(modal.exception.NotFoundError):
     pass
 
 
+class CliUserExecutionError(Exception):
```
mwaskom, line 47: Just bike-shedding, but this name is not super clear to me.

modal/cli/import_refs.py
```diff
@@ -42,6 +44,14 @@ class NoSuchObject(modal.exception.NotFoundError):
...
+    """Private wrapper for exceptions during stub imports in the CLI.
+
+    This intentionally does not inherit from `modal.exception.Error`. Exceptions
```
mwaskom, line 50: > This intentionally does not inherit from `modal.exception.Error`. 

J/w but why is this important?

modal/cli/import_refs.py
```diff
@@ -1,8 +1,10 @@
 # Copyright Modal Labs 2023
-"""Utilities for CLI references to Modal entities
+"""Load or import Python modules from the CLI.
 
 For example, the function reference of `modal run some_file.py::stub.foo_func`
-or the stub lookup of `modal deploy some_file.py`
+or the stub lookup of `modal deploy some_file.py`.
+
+These functions are only called by the Modal CLI, not in tasks.
```
mwaskom, line 7: If I'm following the code properly, we also have `modal.serving.serve_stub` which uses some of these functions. That's pretty convoluted though!

mwaskom (approved):
Do we need those first 5 frames in the "user code" traceback?

ekzhang:
> Do we need those first 5 frames in the "user code" traceback?

I don't think we need them! I tried to get the exception to avoid those first 5 frames but couldn't figure it out yet.

modal/config.py
ekzhang: I'm basing this off the `RUST_BACKTRACE=1` environment variable

modal/__main__.py
ekzhang: I could, but then it would add another stack frame to the error

modal/cli/import_refs.py
ekzhang: Yeah it's an internal error and won't be ever seen by users. It's not meant to be caught.

modal/cli/import_refs.py
ekzhang: Updated the comment

modal/cli/import_refs.py
ekzhang: Nice find. I think `modal.serving` isn't public API and we should mention this!

modal/__main__.py
mwaskom: Hm I think if you have something like
```python
try:
    entrypoint_cli()
except Exception as exc:
    if show_traceback:
        raise exc
    else:
        print(...)
```
then it would render the as if you'd just called `entrypoint_cli()` with no try/except.

modal/cli/import_refs.py
mwaskom: Not public API or not *intended* to be public API? ;)

modal/cli/import_refs.py
mwaskom: Yeah I just mean as a developer trying to understand why we have it. Not a big deal though.

modal/cli/import_refs.py
mwaskom: Thanks (though I still don't follow the implication — why would inheriting from `modal.exception.Error` interfere with the way we use this exception class?)

modal/cli/import_refs.py
ekzhang: Not intended to be public API!

modal/cli/import_refs.py
ekzhang: It wouldn't interfere, it's just a separation of concerns! The `modal.exception.Error` is our public error interface and this is a minimal wrapper error.

modal/__main__.py
ekzhang: Got it!

mwaskom:
So pumped about the example show in the PR description; this is such an improvement!

PR #1425, opened 2024-02-27
Create a git tag for each release

mwaskom:
This should make it easier to track down the specific code / commit for each version that we publish to PyPI.

I've some mild concern about the burden of tags that we'll create doing this. I'm using "unannotated" tags to keep it lighter weight and hopefully it won't be more than O(100) before we reach a stable release. We can always prune tags if it becomes a problem.

Resolves MOD-2444


mwaskom:
I made a dummy repo to test the basic idea here but unfortunately we're just gonna have to do it live.

PR #1426, opened 2024-02-27
Add CI job to test minimal container dependencies

mwaskom:
We're undertaking an effort to reduce the set of client dependencies that are required in the container runtime, and we want a way to automatically check the known set (and to prevent future regressions). This PR adds a new CI job to do that. The steps are:

1) Install the dependencies specified by `container_requirements.txt` (a new file currently used only for CI)
2) Activate a venv, install some additional gRPC utilities; compile the client protos, and then deactivate
3) Try to import `modal._container_entrypoint`

The actual verification is a little bit lightweight but I think that this should catch most cases where unwanted dependencies are included on the codepaths that get imported when we run a container.

Resolves MOD-2449

### DIFF
--- a/container_requirements.txt
+++ b/container_requirements.txt
@@ -1,15 +1,12 @@
 # Requirements file for testing the minimal client dependency set in the container environment
 # Not currently used in production; exists for CI usage until we remove modal/requirements.txt
-aiohttp==3.8.3;python_version<'3.12'
-aiohttp==3.9.1;python_version>='3.12'
+aiohttp==3.9.1
 aiostream==0.4.4
 asgiref==3.5.2
 certifi>=2022.12.07
-cloudpickle==2.0.0;python_version<'3.11'
-cloudpickle==2.2.0;python_version>='3.11'
+cloudpickle==2.2.0
 fastapi==0.88.0
-grpclib==0.4.3;python_version<'3.12'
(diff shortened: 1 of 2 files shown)


container_requirements.txt
```diff
@@ -0,0 +1,17 @@
+# Requirements file for testing the minimal client dependency set in the container environment
```
erikbern, line 1: is this file the same as https://github.com/modal-labs/modal-client/blob/main/modal/requirements.txt

erikbern:
is `container_requirements.txt` similar to `modal/requirements.txt`? https://github.com/modal-labs/modal-client/blob/main/modal/requirements.txt

mwaskom:
> is container_requirements.txt similar to modal/requirements.txt?

Yes, it's based on it. But — as far as I understand — touching `modal/requirements.txt` in any way will force image rebuilds on everyone. So the idea is that we can gradually pare back the contents of `container_requirements.txt` as we obviate or vendor dependencies. Then there can be one big event where we change how client dependencies work in the container. (In practice we're probably not going to have a `modal/requirements.txt` file anymore; we'll handle client dependencies through a combination of vendoring and runtime mounting).

mwaskom:
@prbot approve

erikbern:
👍

PR #1427, opened 2024-02-27
Use stdlib importlib.metadata

mwaskom:
We use a third-party `importlib_metadata` package but `importlib.metadata` was [added to to the stdlib in 3.8](https://docs.python.org/3/library/importlib.metadata.html) (it was originally marked as "provisional", but blessed in 3.10).

Converting our (very simple) usage of this package to the stdlib version will allow us to remove another container dependency.

Resolves MOD-2451

### DIFF
--- a/modal_utils/package_utils.py
+++ b/modal_utils/package_utils.py
@@ -2,9 +2,8 @@
 import importlib
 import importlib.util
 import typing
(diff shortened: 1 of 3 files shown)


freider (approved):
Nice!

PR #1430, opened 2024-02-28
Support args and kwargs for `Image.run_function()`

aksh-at:
## Describe your changes

Depends on https://github.com/modal-labs/modal-client/pull/1429 and https://github.com/modal-labs/modal/pull/10746

## Changelog

- `Image.run_function` now allows you to pass args and kwargs to the function. Usage:

```python
def my_build_function(name, size, *, variant=None):
    print(f"Building {name} {size} {variant}")


image = modal.Image.debian_slim().run_function(
    my_build_function, args=("foo", 10), kwargs={"variant": "bar"}
)
```

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -18,6 +18,7 @@ from modal_proto import api_pb2
 from modal_utils.async_utils import synchronize_api
 from modal_utils.grpc_utils import RETRYABLE_GRPC_STATUS_CODES, retry_transient_errors, unary_stream
 
+from ._blob_utils import MAX_OBJECT_SIZE_BYTES
 from ._function_utils import FunctionInfo
 from ._resolver import Resolver
 from ._serialization import serialize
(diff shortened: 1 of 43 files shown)


modal/image.py
```diff
@@ -1273,6 +1277,7 @@ def run_function(
         timeout: Optional[int] = 86400,  # Maximum execution time of the function in seconds.
         force_build: bool = False,
         secret: Optional[_Secret] = None,  # Deprecated: use `secrets`
+        **kwargs,
```
freider, line 1280: I find it's a bit brittle to have `**kwargs` here with so many "reserved" keyword arguments to the function (that we may also extend as we add more features). I wonder if we could do something slightly different to not overload function arguments and "modal options".

The simplest solution might be to make `kwargs` (and maybe args too in that case) into a singular arguments (dict/tuple) rather than variadic ones. It's not super pretty but it's pretty uncommon and it would prevent future modal library updates from breaking people's code if we add more named args

erikbern: Agree, seems cleaner to take `kwargs: Dict[str, Any]`

erikbern: And maybe take `args: List[Any]` too?

aksh-at: Good point, done.

aksh-at:
@prbot approve

PR #1431, opened 2024-02-28
Fix error when getclosurevars() fails

freider:
`getclosurevars()` can apparently fail with weird ValueErrors ("cell empty") which causes loading of the file to fail.

We'll probably get rid of getclosurevars usage soon anyways, but this should fix it in the meanwhile (at the cost of not being able to use things like `some_global = Dict.new()` in affected functions.

Repro for previous issue:
```python
from sqlalchemy import create_engine
@stub.function(image=modal.Image.debian_slim().pip_install('sqlalchemy')) def foo():
    if False:
        create_engine()
```

## Describe your changes


erikbern:
👍

PR #1432, opened 2024-02-28
Make loader take `self`

erikbern:
The load method used to be a class method that would take the "provider" as the first argument and return a "handle".

But since Aug 2023, the load method just mutates the provider directly. So we might as well rename it to `self` and let it behave like a regular method.

In particular this lets us define the loader function on the class itself, and use it as a normal bound method. This is roughly happening `_Mount._load_mount` already, so we can simplify this a bit and get rid of a usage of `functools.partial`.

(The reason I created this PR was that I needed to add a bunch of extra arguments to the partial, and I realized it's unnecessarily complex)

In the future this lets us move more loader functions to sit on the class instead. This will let us get rid of the nested initialization code.

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -204,12 +204,12 @@ class _Cls(_Object, type_prefix="cs"):
         def _deps() -> List[_Function]:
             return list(functions.values())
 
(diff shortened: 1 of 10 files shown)


erikbern:
Not sure why tests are failing – will debug later

mwaskom (approved):

PR #1435, opened 2024-02-28
Deploy persisted mounts directly

erikbern:
`_Mount` previously inherited from `_StatefulObject`, which is an intermediate object we've needed for deploying single-object apps client-side. But with this change, we deploy mounts directly using `MountGetOrCreate`, and we don't rely on the client code to construct an app etc – this logic will happen server-side instead.

Mounts are only deployed by our own code, in `modal_base_images`, so this hopefully shouldn't affect users.

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -24,7 +24,7 @@ from ._blob_utils import FileUploadSpec, blob_upload_file, get_file_upload_spec_
 from ._resolver import Resolver
 from .client import _Client
 from .config import config, logger
-from .object import _get_environment_name, _StatefulObject
+from .object import _get_environment_name, _Object
 
 ROOT_DIR: PurePosixPath = PurePosixPath("/root")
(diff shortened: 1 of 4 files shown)


modal/mount.py
```diff
@@ -558,6 +567,21 @@ async def lookup(
...
+        environment_name: Optional[str] = None,
+        client: Optional[_Client] = None,
+    ) -> "_Mount":
+        self._deployment_name = deployment_name
+        self._namespace = namespace
+        self._environment_name = environment_name
+        if client is None:
+            client = await _Client.from_env()
+        resolver = Resolver(client=client)
+        await resolver.load(self)
```
mwaskom, line 583: I'm a little confused by this; does deployment happen as a side-effect of calling `Resolver.load` on an object??

mwaskom (approved):

modal/mount.py
erikbern: Yeah in this case – `Resolver.load` is the way you trigger an object's initialization code. It's a bit confusing.

PR #1436, opened 2024-02-28
Remove modal import machinery preamble from user code exceptions

mwaskom:
Follow-on to #1422, this now modifies the traceback printed when we except on user code import to show the user's target script / module as the first frame.

AFAICT, we don't have any test framework that exercises `modal/__main__.py` (there is `cli_test`, but it uses the `cli_entrypoint` that is downstream of where we're doing this). I did add some robustness so that we fall back to the original traceback if we can't find the user's frame.

I still think it's sort of bad CLI UX that there's extra noise from the path taken inside modal for things like invalid argument values:

image

And in some cases we end up with extra synchronicity frames after the line that actually triggered the exception, which is weird:

image

So probably more than we can do here. It's possible that `modal.exception.InvalidError` should be something like just an error panel with the invalid message and then maybe the relevant filename / line number in the user's codebase, but finding that could potentially be hard for more complex applications.

### DIFF
--- a/modal/__main__.py
+++ b/modal/__main__.py
@@ -15,7 +15,18 @@ def main():
     try:
         entrypoint_cli()
     except CliUserExecutionError as exc:
-        raise exc.__cause__ from None
+        # Try to step forward in the traceback until we get to the user code that failed to import
+        tb = orig_tb = exc.__cause__.__traceback__
+        if exc.user_source.endswith(".py"):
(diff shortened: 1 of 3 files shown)


ekzhang:
Wow I'm going to be honest, I don't understand how your code works but this is awesome!

image

ekzhang (approved):

PR #1437, opened 2024-02-28
Restrict toml usage / imports to local environment

mwaskom:
We don't need `toml` in the container runtime: the two places where it's currently used are only relevant locally, so it makes sense to gate the codepaths by `MODAL_IS_REMOTE` and then defer the `toml` import. It's not optimal to have deferred imports, but it provides  a lower maintenance burden than vendoring. We can explore the stdlib `tomllib` once we roll up to 3.11+ (although I think that is read-only).

- Resolves MOD-2456

### DIFF
--- a/container_requirements.txt
+++ b/container_requirements.txt
@@ -10,4 +10,3 @@ grpclib==0.4.7
 protobuf>=3.19.0
 rich==12.3.0
 tblib==1.7.0
-toml==0.10.2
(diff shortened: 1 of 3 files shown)


ekzhang (approved):
Nice find!

mwaskom:
Unfortunately a bit trickier than it appears here because we load the config data at import time and that happens before we initialize the container app (which backs `modal.app.is_local`). So this is going to be more complicated.

mwaskom:
`modal.app.is_local` didn't work here because we need to load the config information prior to initializing the app (or at least what's how it currently works). So I went with adding an explicit `MODAL_IS_REMOTE` env var to the container environment and using that instead.

PR #1438, opened 2024-02-29
Add types to most of modal_utils

savarin:
## Describe your changes

This PR adds mypy types to `modal_version` and `modal_utils`, as per https://github.com/modal-labs/modal-client/issues/1372.

```
modal_utils/__init__.py
modal_utils/app_utils.py
modal_utils/async_utils.py
modal_utils/grpc_testing.py
modal_utils/http_utils.py
modal_utils/logger.py
modal_utils/package_utils.py
```

### DIFF
--- a/modal_utils/async_utils.py
+++ b/modal_utils/async_utils.py
@@ -7,7 +7,7 @@ import sys
 import time
 import typing
 from contextlib import asynccontextmanager
-from typing import Any, AsyncGenerator, Awaitable, Callable, Iterator, List, Optional, Set, TypeVar, Union
+from typing import Any, AsyncGenerator, Awaitable, Callable, Iterator, List, Optional, Set, TypeVar, Union, cast
 
 import synchronicity
 from typing_extensions import ParamSpec
@@ -327,7 +327,7 @@ def asyncify(f: Callable[P, T]) -> Callable[P, Awaitable[T]]:
     return wrapper
 
 
-async def iterate_blocking(iterator: Iterator[T]) -> AsyncGenerator[Union[T, object], None]:
+async def iterate_blocking(iterator: Iterator[T]) -> AsyncGenerator[T, None]:
     """Iterate over a blocking iterator in an async context."""
 
     loop = asyncio.get_running_loop()
@@ -336,7 +336,7 @@ async def iterate_blocking(iterator: Iterator[T]) -> AsyncGenerator[Union[T, obj
         obj = await loop.run_in_executor(None, next, iterator, DONE)
         if obj is DONE:
             break
-        yield obj
+        yield cast(T, obj)
 
 
 class ConcurrencyPool:
--- a/modal_utils/async_utils.py
+++ b/modal_utils/async_utils.py
@@ -7,7 +7,7 @@ import sys
 import time
 import typing
 from contextlib import asynccontextmanager
-from typing import Any, AsyncIterator, Awaitable, Callable, Iterator, List, Optional, Set, TypeVar
+from typing import Any, AsyncGenerator, Awaitable, Callable, Iterator, List, Optional, Set, TypeVar, Union
 
 import synchronicity
 from typing_extensions import ParamSpec
@@ -327,7 +327,7 @@ def asyncify(f: Callable[P, T]) -> Callable[P, Awaitable[T]]:
     return wrapper
 
 
-async def iterate_blocking(iterator: Iterator[T]) -> AsyncIterator[T]:
+async def iterate_blocking(iterator: Iterator[T]) -> AsyncGenerator[Union[T, object], None]:
     """Iterate over a blocking iterator in an async context."""
 
     loop = asyncio.get_running_loop()
--- a/modal_utils/package_utils.py
+++ b/modal_utils/package_utils.py
@@ -23,7 +23,7 @@ def get_file_formats(module):
 BINARY_FORMATS = ["so", "S", "s", "asm"]  # TODO
 
 
-def get_module_mount_info(module_name: str) -> typing.List[typing.Tuple[bool, Path]]:
+def get_module_mount_info(module_name: str) -> typing.Sequence[typing.Tuple[bool, Path]]:
     """Returns a list of tuples [(is_dir, path)] describing how to mount a given module."""
     file_formats = get_file_formats(module_name)
     if set(BINARY_FORMATS) & set(file_formats):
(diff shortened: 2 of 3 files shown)


mwaskom:
I'm not seeing all of the files mentioned in the PR description changed in the diff.

We should be careful about touching `modal_version` by the way as some of those files are auto-generated. It's also pretty simple metadata so I'm not sure it needs explicit annotations.

savarin:
> I'm not seeing all of the files mentioned in the PR description changed in the diff.
> 
> We should be careful about touching `modal_version` by the way as some of those files are auto-generated. It's also pretty simple metadata so I'm not sure it needs explicit annotations.

@mwaskom Not all the files need changing. Running `inv protoc`, `inv type-stubs` and `inv type-check` on the following files:

```python
    pyright_allowlist = [
        "modal/functions.py",
        "modal_utils/__init__.py",
        "modal_utils/app_utils.py",
        "modal_utils/async_utils.py",
        "modal_utils/grpc_testing.py",
        "modal_utils/http_utils.py",
        "modal_utils/logger.py",
        "modal_utils/package_utils.py",
    ]
```

I only see the following type errors (which changes in the PR now fix).

```shell
/Users/savarin/Development/python/modal-client/modal_utils/async_utils.py
  /Users/savarin/Development/python/modal-client/modal_utils/async_utils.py:339:15 - error: Return type of async generator function must be compatible with "AsyncGenerator[T@iterate_blocking | object, Any]"
    "AsyncGenerator[T@iterate_blocking | object, Unknown, Unknown]" is incompatible with "AsyncIterator[T@iterate_blocking]"
      Type parameter "_T_co@AsyncIterator" is covariant, but "T@iterate_blocking | object" is not a subtype of "T@iterate_blocking"
        Type "T@iterate_blocking | object" cannot be assigned to type "T@iterate_blocking" (reportReturnType)
/Users/savarin/Development/python/modal-client/modal_utils/package_utils.py
  /Users/savarin/Development/python/modal-client/modal_utils/package_utils.py:13:51 - error: Object of type "None" cannot be used as iterable value (reportOptionalIterable)
  /Users/savarin/Development/python/modal-client/modal_utils/package_utils.py:44:12 - error: Expression of type "list[tuple[Literal[True], Path]] | list[tuple[Literal[False], Path]]" cannot be assigned to return type "List[Tuple[bool, Path]]"
    Type "list[tuple[Literal[True], Path]] | list[tuple[Literal[False], Path]]" cannot be assigned to type "List[Tuple[bool, Path]]"
      "list[tuple[Literal[True], Path]]" is incompatible with "List[Tuple[bool, Path]]"
        Type parameter "_T@list" is invariant, but "tuple[Literal[True], Path]" is not the same as "Tuple[bool, Path]"
        Consider switching from "list" to "Sequence" which is covariant (reportReturnType)
3 errors, 0 warnings, 0 informations
```

Re: `modal_version`, no changes needed but I removed from the list out of abundance of caution.

savarin:
@freider Let me know what's the best way to ping on these PRs - happy to tag in a comment like this one, annoyingly I can't add reviewers

modal_utils/async_utils.py
```diff
@@ -327,7 +327,7 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:
     return wrapper
 
 
-async def iterate_blocking(iterator: Iterator[T]) -> AsyncIterator[T]:
+async def iterate_blocking(iterator: Iterator[T]) -> AsyncGenerator[Union[T, object], None]:
```
freider, line 330: This return signature should not include `object` in the union - it's only used internally as a marker for when to break the loop. Better then to have a type: ignore on the yield, so at least the users of this function won't have to deal with the generic `object` case

modal_utils/package_utils.py
```diff
@@ -10,7 +10,9 @@
 
 def get_file_formats(module):
     try:
-        endings = [str(p).split(".")[-1] for p in files(module) if "." in str(p)]
+        file_list = files(module)
+        assert file_list is not None
```
freider, line 14: Haha nice, this type checking would have caught and fixed a bug I just fixed in a separate PR! #1454 👍

modal_utils/package_utils.py
```diff
@@ -19,7 +21,7 @@ def get_file_formats(module):
 BINARY_FORMATS = ["so", "S", "s", "asm"]  # TODO
 
 
-def get_module_mount_info(module_name: str) -> typing.List[typing.Tuple[bool, Path]]:
+def get_module_mount_info(module_name: str) -> typing.Sequence[typing.Tuple[bool, Path]]:
```
freider, line 24: This is odd - don't we *know* that the return type here is a list specifically? Not sure why pyright would get an error here.

freider (commented):
Great! added some comments. I try to check incoming PRs once in a while, but it's fine to ping me in the thread

modal_utils/async_utils.py
savarin: @freider OK looks like casting it helps avoid needing to type ignore.

modal_utils/package_utils.py
savarin: TIL thanks to ChatGPT. In the past, I've only used `Sequence` when I need a read-only list but the covariant aspect is new.

```
List is invariant in its type parameter, meaning that if you have a List[X], you cannot use it in place where a List[Y] is expected, even if X is a subtype of Y. This invariance is what's causing the error, as List[tuple[Literal[True], Path]] is not considered a subtype of List[Tuple[bool, Path]].

On the other hand, Sequence is covariant in its type parameter. If X is a subtype of Y, then Sequence[X] is considered a subtype of Sequence[Y]. This is why changing List to Sequence resolves the error. When you use Sequence[Tuple[bool, Path]], the function can return a sequence of tuples where the boolean can be specifically True or False, not just a generic boolean.
```

That being said I asked for a specific example that would satisfy `List[bool]` but not `List[Literal[True, False]]` and vice versa, but none given. Full text [here](https://gist.github.com/savarin/447056e690ffbd8b7e363376b2814de7).

freider: TIL - had no idea abou thte List vs Sequence invariant vs. covariant differences! That's cool :) Super annoying that literal `True` and `False` values are treated as subtypes rather than instances of bool though in this case... I'd consider that a bug in pyright (especially given that `mypy` approves...). Anyways, let's go with Sequence

freider (approved):

PR #1440, opened 2024-02-29
queue: document lifetime

danielnorberg:
## Describe your changes

Linear issue: MOD-2459

Document `modal.Queue` lifetime and expiry.

### DIFF
--- a/modal/queue.py
+++ b/modal/queue.py
@@ -22,6 +22,12 @@ class _Queue(_Object, type_prefix="qu"):
 
     The queue can contain any object serializable by `cloudpickle`, including Modal objects.
 
+    **Lifetime of a queue and its contents**
+
+    A `Queue`'s lifetime matches the lifetime of the app it's attached to, but the contents expire after 30 days.
+    Because of this, `Queues`s are best used for communication between active functions and not relied on for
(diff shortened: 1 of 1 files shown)


modal/queue.py
```diff
@@ -22,6 +22,12 @@ class _Queue(_Object, type_prefix="qu"):
 
     The queue can contain any object serializable by `cloudpickle`, including Modal objects.
 
+    **Lifetime of a queue and its contents**
+
+    A `Queue`'s lifetime matches the lifetime of the app it's attached to, but the contents expire after 30 days.
```
freider, line 27: Except when it's a deployed self-contained queue, i.e. `Queue.persisted()` / `Queue.from_name(, create_if_missing=True)` 🤡

freider: So I'm not sure if we should actually talk about lifetimes here, since it's more of a general concept and each object type can have both ephemeral and persisted variants

danielnorberg: No strong opinion, but this is where we document lifetimes for dicts https://modal.com/docs/reference/modal.Dict#modaldictdict. Where should we discuss lifetime/expiry instead?

danielnorberg: Discussed IRL, will refactor/move docs later.

PR #1441, opened 2024-02-29
Disable version tagging while we debug

mwaskom:
Unborks our deployments


mwaskom:
Superseded by #1442; this was user error, not a problem with the automated tagging code

PR #1443, opened 2024-02-29
Show traceback from user code when local entrypoint function raises errors

mwaskom:
Fixes an oversight in our recent change to traceback handling:

image

Resolves MOD-2460


ekzhang (approved):
Looks good to me!

PR #1448, opened 2024-02-29
Vendor cloudpickle

mwaskom:
Part of reducing client runtime dependencies.

We have no choice on vendoring cloudpickle — it is likely to also be depended on by our users, and it's essential that one has the same cloudpickle version on both sides of a transfer.

Fortunately, it is a single file with no third-party dependencies, so vendoring is mostly straightforward.

The only question is whether we should "maintain" the file such that it can pass our various static checks (currently it fails both mypy and ruff). I've excluded it from both checks for now, but open to alternate arguments on this.

- Resolves MOD-2468

### DIFF
--- a/container_requirements.txt
+++ b/container_requirements.txt
@@ -4,7 +4,6 @@ aiohttp==3.9.1
 aiostream==0.4.4
 asgiref==3.5.2
 certifi>=2022.12.07
-cloudpickle==2.2.0
 fastapi==0.88.0
 grpclib==0.4.7
 protobuf>=3.19.0
--- a/modal/_function_utils.py
+++ b/modal/_function_utils.py
@@ -185,7 +185,7 @@ class FunctionInfo:
         return serialized_bytes
 
     def get_globals(self):
-        from cloudpickle.cloudpickle import _extract_code_globals
+        from ._vendor.cloudpickle import _extract_code_globals
 
         func = self.raw_f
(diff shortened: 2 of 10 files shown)


freider:
Nice! I think it makes sense to exclude vendored files from checks, sounds like a pain to maintain otherwise.

Maybe we should store some metadata about what version we have vendored somewhere, in case we want to upgrade at some point and are unsure?
And if we start vendoring more packages we might want tooling to create the vendored files from the version list, maybe using the vendoring lib you found before @mwaskom ? but at this stage it feels a bit overkill

mwaskom:
> Maybe we should store some metadata about what version we have vendored somewhere, in case we want to upgrade at some point and are unsure?

I put the version that I sourced the current file from in  a docstring comment for now; agreed that we might want to be a little more structured about that, but probably best to defer till we have a better sense of how complicated we need the maintenance tooling to be.

> And if we start vendoring more packages we might want tooling to create the vendored files from the version list, maybe using the vendoring lib you found before @mwaskom ? but at this stage it feels a bit overkill

Yeah, my current thinking is pretty negative on vendoring as a solution because anything beyond very self-contained modules like this fundamentally requires hacks. So I'm hoping that we can avoid the need for something like that.

ekzhang:
There are also a couple of minor server changes to have this work, and I'll make those now.

ekzhang:
https://github.com/modal-labs/modal/pull/10816

PR #1449, opened 2024-02-29
Add inv script for showing current deprecation warnings in the client

mwaskom:
Just a little utility for staying on top of what we're currently in the process of deprecating and whether there's anything that is ready to be enforced:

image


mwaskom:
@prbot lgtm

PR #1453, opened 2024-03-01
Close unawaited coroutines when ConcurrencyPool exits

freider:
Prevents a bunch of warnings, noticeable in tests


mwaskom (commented):
Nice!

Issue #1451, opened 2024-03-01
Error when using modal deploy 


shuohaopolyu:
### Version
modal client: 0.60.8
Python: 3.9.7

### Description
I deploy the echobot.py by using the command: **modal deploy echobot.py**, then I have the following error:

![1709268547918](https://github.com/modal-labs/modal-client/assets/87934285/f0bc7933-07c1-4fa8-a240-30e727659dae)

shuohaopolyu:
Hi Freider, sorry for the late reply!
 
I don't think it is because of the code in echobot.py, I tried the get_started.py given officially, and still got the same error. Please check the `get_started.py` file:

```
import modal

stub = modal.Stub("example-get-started")


@stub.function()
def square(x):
    print("This code is running on a remote worker!")
    return x**2


@stub.local_entrypoint()
def main():
    print("the square is", square.remote(42))
```
and the 'echobot.py' file:
```
"""

Sample bot that echoes back messages.

This is the simplest possible bot and a great place to start if you want to build your own bot.

"""

from __future__ import annotations

from typing import AsyncIterable

import fastapi_poe as fp
from modal import Image, Stub, asgi_app


class EchoBot(fp.PoeBot):
    async def get_response(
        self, request: fp.QueryRequest
    ) -> AsyncIterable[fp.PartialResponse]:
        last_message = request.query[-1].content
        yield fp.PartialResponse(text=last_message)


REQUIREMENTS = ["fastapi-poe==0.0.25"]
image = Image.debian_slim().pip_install(*REQUIREMENTS)
stub = Stub("echobot-poe")


@stub.function(image=image)
@asgi_app()
def fastapi_app():
    bot = EchoBot()
    # Optionally, provide your Poe access key here:
    # 1. You can go to https://poe.com/create_bot?server=1 to generate an access key.
    # 2. We strongly recommend using a key for a production bot to prevent abuse,
    # but the starter examples disable the key check for convenience.
    # 3. You can also store your access key on modal.com and retrieve it in this function
    # by following the instructions at: https://modal.com/docs/guide/secrets
    # POE_ACCESS_KEY = ""
    # app = make_app(bot, access_key=POE_ACCESS_KEY)
    app = fp.make_app(bot, allow_without_key=True)
    return app
```

shuohaopolyu:
the updated version fixes my problem, thanks.

freider:
DIdn't release the fix until now as it had some issues. But v0.61.4 fixes this properly, I think

PR #1454, opened 2024-03-01
Fix for weird edge case when module files is None

freider:
I think this should fix: https://github.com/modal-labs/modal-client/issues/1451


mwaskom:
I don't understand the CI failures at all.

freider:
> I don't understand the CI failures at all.

My guess is PackageNotFound errors are raised immedately now when calling `files()` now, whereas importlib_metadata instead raised it when iterating over the result... Expanding the try block seems to fix

PR #1459, opened 2024-03-01
Vendor (part of) tblib; add tests for traceback utilities

mwaskom:
## Describe your changes

This vendors part of the `tblib` library (BSD 2-clause license, ≈300 LoC) and removes it from the client dependencies.

We didn't have any tests for the places we use `tblib`, so I added some of those too. 

- Resolves MOD-2474


aksh-at (approved):
nice, thanks for adding tests!

PR #1460, opened 2024-03-01
Use VolumeGetOrCreate client-side

erikbern:
This is the final one of all the RPC changes to objects. This one is mostly copy paste from `NetworkFileSystem` (aka `SharedVolume`) and similar to other PRs, adds quite a lot of redundant code in the short and medium term.

Once this is merged, we can start to delete some amount of code. Sadly not a ton initially, because a lot of it will require deprecations. But we can at least delete `_StatefulObject` a few more things

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -105,7 +105,7 @@ class _Volume(_Object, type_prefix="vo"):
         environment_name: Optional[str] = None,
         create_if_missing: bool = False,
     ) -> "_Volume":
-        """Create a reference to a persisted network filesystem
+        """Create a reference to a persisted volume
 
         """
 
@@ -158,10 +158,10 @@ class _Volume(_Object, type_prefix="vo"):
         environment_name: Optional[str] = None,
         create_if_missing: bool = False,
     ) -> "_Volume":
-        """Lookup a network file system with a given name and tag.
+        """Lookup a volume with a given name
 
         ```python
-        n = modal.Volume.lookup("my-nfs")
+        n = modal.Volume.lookup("my-volume")
         print(n.listdir("/"))
         ```
         """
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -9,7 +9,7 @@ from typing import IO, AsyncGenerator, AsyncIterator, BinaryIO, Callable, Genera
 import aiostream
 from grpclib import GRPCError, Status
 
-import modal.exception
+from modal.exception import VolumeUploadTimeoutError
 from modal_proto import api_pb2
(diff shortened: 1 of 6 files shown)


modal/volume.py
```diff
@@ -124,7 +148,50 @@ def f():
...
+
+    @staticmethod
+    async def lookup(
+        label: str,
+        namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
+        client: Optional[_Client] = None,
+        environment_name: Optional[str] = None,
+        create_if_missing: bool = False,
+    ) -> "_Volume":
+        """Lookup a network file system with a given name and tag.
```
mwaskom, line 161: Nit: "network file system"

Also I don't understand "given name *and tag*"; the example only uses a single argument.

modal/volume.py
```diff
@@ -98,11 +98,35 @@ async def _load(self: _Volume, resolver: Resolver, existing_object_id: Optional[
...
         return _Volume._from_loader(_load, "Volume()")
 
+    @staticmethod
+    def from_name(
+        label: str,
+        namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
+        environment_name: Optional[str] = None,
+        create_if_missing: bool = False,
+    ) -> "_Volume":
+        """Create a reference to a persisted network filesystem
```
mwaskom, line 108: Nit: "network filesystem"

modal/volume.py
```diff
@@ -124,7 +148,50 @@ def f():
...
+        label: str,
+        namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
+        client: Optional[_Client] = None,
+        environment_name: Optional[str] = None,
+        create_if_missing: bool = False,
+    ) -> "_Volume":
+        """Lookup a network file system with a given name and tag.
+
+        ```python
+        n = modal.Volume.lookup("my-nfs")
```
mwaskom, line 164: Nit: "my-nfs"

mwaskom (approved):
Looks good, aside from the copypasta on NFS terminology in the documentation

modal/volume.py
erikbern: sorry, this is copy paste two generations – let me fix it and also the other place

modal/volume.py
erikbern: thanks

modal/volume.py
erikbern: d'oh thanks

PR #1462, opened 2024-03-01
Drop from_id down to subclasses, remove AppLookupObject usage

erikbern:
The `from_id` is only ever used by function calls (millions of times during the last week) and sandboxes (handful of times). There are some stray calls for other objects, but they are insanely rare (like, 1 function lookup for the entire last month).

I've dropped the implementation down to respective subclasses. I think we can pull this thread a bit further and get rid of the awkward overloaded handle_metadata stuff later, but it will take a bit more work.

This change gets rid of the remaining usage of `AppLookupObject` by the client so I'm bumping the client version – that way we can drop the old endpoint once 0.61 is the minimum version.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -2,6 +2,7 @@
 import os
 from typing import Dict, List, Optional, Sequence, Union
 
+from google.protobuf.message import Message
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 
 from modal.exception import InvalidError, SandboxTerminatedError, SandboxTimeoutError
@@ -189,6 +190,11 @@ class _Sandbox(_Object, type_prefix="sb"):
 
         return _Sandbox._from_loader(_load, "Sandbox()", deps=_deps)
 
+    def _hydrate_metadata(self, handle_metadata: Optional[Message]):
+        self._stdout = LogsReader(api_pb2.FILE_DESCRIPTOR_STDOUT, self.object_id, self._client)
+        self._stderr = LogsReader(api_pb2.FILE_DESCRIPTOR_STDERR, self.object_id, self._client)
+        self._result = None
+
     @staticmethod
     async def from_id(sandbox_id: str, client: Optional[_Client] = None) -> "_Sandbox":
         """Construct a Sandbox from an id and look up the sandbox result."""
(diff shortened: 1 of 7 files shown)


modal_version/__init__.py
```diff
@@ -7,7 +7,7 @@
 major_number = 0
 
 # Bump this manually on breaking changes, then reset the number in _version_generated.py
```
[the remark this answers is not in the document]
erikbern: oh right, i forgot

test/function_test.py
```diff
@@ -467,12 +467,6 @@ def foo():
     assert function_id
     assert foo.web_url
 
-    rehydrated_function = Function.from_id(function_id, client=client)
```
mwaskom, line 470: Confused about why this test is being removed — don't we still have `Function.from_id`?

mwaskom (approved):
Question about the test

test/function_test.py
erikbern: I dropped it from the base class in order to remove the need for `AppLookupObject`

Looking at internal dashboards and there is exactly zero backend calls in the last week. I see 1 (!) call in the last month. But it seems relatively safe to remove this for now and reintroduce if it's something users really really care about.

mwaskom: Right, but we're not testing the base class here we're testing `Function` right?

erikbern: Yeah but I didn't drop the implementation from the base class to all subclasses – just to `FunctionCall` and `Sandbox`. The other classes lost the method. So `Function` doesn't have `from_id` anymore.

mwaskom: Ah, missed the `Function`/`FunctionCall` distinction, thanks 👍

modal/sandbox.py
```diff
@@ -193,11 +193,21 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
...
-            self._result = None
+        self._result = None
+
+    @staticmethod
+    async def from_id(sandbox_id: str, client: Optional[_Client] = None) -> "_Sandbox":
+        """Construct a Sandbox from an id and look up the sandbox result."""
+        if client is None:
+            client = await _Client.from_env()
+
+        req = api_pb2.SandboxWaitRequest(sandbox_id=sandbox_id, timeout=0)
```
erikbern, line 204: @aksh-at sanity checking that this will work? doing this lets me remove the last use of `AppLookupObject`

These lookups are incredibly rare btw

erikbern:
Merging this so we can start the deprecation clock!

modal/sandbox.py
aksh-at: There's a _slight_ difference now. We'll set the `obj._result` to an empty `api_pb2.GenericResult()`, where it used to be `None`. But looking at the code, that shouldn't break anything.

PR #1465, opened 2024-03-01
Overhaul signal handling in container_entrypoint

ekzhang:
Simplifies the signal handling logic quite a bit, removes some hacks, and fixes various bugs.

Resolves MOD-2476.

### Changes

- Add a few docstrings
- Update a lot of comments
- Remove SIGINT and SIGTERM handlers from `SignalHandlingEventLoop`
  - SIGTERM shouldn't be raised by the Modal runtime, might be outdated
  - SIGINT should just be a standard KeyboardInterrupt for consistency between the sync / async functions
- Refactor run_function_[a]sync to avoid signal handling race conditions that would skip exit lifecycle methods and volume commits
- Fixes SIGINT causing container exit handler to be interrupted
- Fixes SIGTERM raised from #1369 on container heartbeat terminating container without an exit handler
- Fixes MOD-2490, and added a test
- Fixes MOD-2493, and this is checked by an existing test now that the SIGTERM bug is fixed

## Changelog

- Bug fix: Stopping an app while a container's `@exit()` lifecycle methods are being run no longer interrupts the lifecycle methods.
- Bug fix: Worker preemptions no longer interrupt a container's `@exit()` lifecycle method (until 30 seconds later).
- Bug fix: Async `@exit()` lifecycle methods are no longer skipped for sync functions.
- Bug fix: Stopping a sync function with `allow_concurrent_inputs>1` now actually stops the container. Previously, it would not propagate the signal to worker threads, so they would continue running.
- Bug fix: Input-level cancellation no longer skips the `@exit()` lifecycle method.
- Improve stability of container entrypoint against race conditions in task cancellation.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -175,11 +175,23 @@ class _FunctionIOManager:
                     )
                     # This is equivalent to a task cancellation or preemption from worker code,
                     # except we do not send a SIGKILL to forcefully exit after 30 seconds.
+                    #
+                    # SIGINT always interrupts the main thread, but not any auxiliary threads. On a
+                    # sync function without concurrent inputs, this raises a KeyboardInterrupt. When
+                    # there are concurrent inputs, we cannot interrupt the thread pool, but the
+                    # interpreter stops waiting for daemon threads and exits. On async functions,
+                    # this signal lands outside the event loop, stopping `run_until_complete()`.
                     os.kill(os.getpid(), signal.SIGINT)
 
-                if self.current_input_id in input_ids_to_cancel:
+                elif self.current_input_id in input_ids_to_cancel:
                     # This goes to a registered signal handler for sync Modal functions, or to the
                     # `SignalHandlingEventLoop` for async functions.
+                    #
+                    # We only send this signal on functions that do not have concurrent inputs enabled.
+                    # This allows us to do fine-grained input cancellation. On sync functions, the
+                    # SIGUSR1 signal should interrupt the main thread where user code is running,
+                    # raising an InputCancellation() exception. On async functions, the signal should
+                    # reach a handler in SignalHandlingEventLoop, which cancels the task.
                     os.kill(os.getpid(), signal.SIGUSR1)
             return True
         return False
@@ -984,7 +996,8 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
                 function_io_manager.stop_heartbeat()
 
             finally:
-                # Restore the original signal handler, needed for container_test hygiene.
+                # Restore the original signal handler, needed for container_test hygiene since the
+                # test runs `main()` multiple times in the same process.
                 signal.signal(signal.SIGINT, int_handler)
                 signal.signal(signal.SIGUSR1, usr1_handler)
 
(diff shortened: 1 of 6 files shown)


modal/_container_entrypoint.py
```diff
@@ -671,14 +688,16 @@ def run_inputs(input_id: str, function_call_id: str, args: Any, kwargs: Any) ->
...
-            logger.debug("Running cls exit methods (sync)")
-            exit_methods: Dict[str, Callable] = _find_callables_for_obj(imp_fun.obj, _PartialFunctionFlags.EXIT)
-            for exit_method in exit_methods.values():
-                with function_io_manager.handle_user_exception():
-                    # We are deprecating parameterized exit methods but want to gracefully handle old code.
-                    # We can remove this once the deprecation in the actual @exit decorator is enforced.
-                    args = (None, None, None) if method_has_params(exit_method) else ()
-                    exit_method(*args)
+            # Ignore termination signals from the worker while running container @exit() methods.
+            with ignore_sigint():
```
erikbern, line 692: The closing of this context manager re-enables the signal handler, but do we need to do that? After the exit handler is done, the rest is just Modal code to exit the program right?

ekzhang: Probably not — SIGINT shouldn't interrupt volume commits. We can refactor it some more here!

ekzhang: Done! I ended up having to restore the signal handler at the end regardless in order for `container_test.py` to work due to some hygiene, but I added a comment explaining what's going on.

modal/_container_entrypoint.py
```diff
@@ -962,22 +944,49 @@ def breakpoint_wrapper():
...
+                # Finally, commit on exit to catch uncommitted volume changes and surface background
+                # commit errors.
+                function_io_manager.volume_commit(
+                    [v.volume_id for v in container_args.function_def.volume_mounts if v.allow_background_commits]
+                )
+                # Avoid "Canceling remaining unfinished task" warnings.
+                function_io_manager.stop_heartbeat()
+
+            finally:
+                # Restore the original signal handler, needed for container_test hygiene.
```
freider, line 987: This has to do with us running multiple tests of the container in the same process I assume?

modal_test_support/functions.py
```diff
@@ -173,6 +173,42 @@ def generator(self, x):
...
+
+    @enter()
+    def enter_sync(self):
+        self.events.append("enter_sync")
+
+    @enter()
+    async def enter_async(self):
+        self.events.append("enter_async")
+
+    # TODO: We have no way of verifying @exit() handler output in container_test.py right now.
```
freider, line 191: Should be possible when using `_run_container_process` (see the last few tests in container_test.py) since those use subprocesses for the container tests, although it would have to rely on printing stuff to stdout or similar to communicate since no more outputs are returned after exit handlers. 
We can also assert that the VolumeCommit rpc is called

freider (approved):
Nice cleanup! Lgtm

modal_test_support/functions.py
ekzhang: Thanks for the advice, I just added the new tests using `_run_container_process`!

modal/_container_entrypoint.py
ekzhang: Yes, I updated the comment to be clearer here

PR #1466, opened 2024-03-01
Deprecate `.persisted` in favor of `.from_name(..., create_if_missing=True)`

erikbern:
These methods are almost the same at this point, so let's merge.

We'll merge `.lookup` as well later.

Before we merge this, we need to update examples, integration tests, and docs. Otherwise those will fail.

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -10,7 +10,7 @@ from ._serialization import deserialize, serialize
 from ._types import typechecked
 from .client import _Client
 from .config import logger
-from .exception import deprecation_error
+from .exception import deprecation_error, deprecation_warning
 from .object import _get_environment_name, _Object, live_method
 
 
@@ -111,8 +111,8 @@ class _Dict(_Object, type_prefix="di"):
     def persisted(
         label: str, namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE, environment_name: Optional[str] = None
     ) -> "_Dict":
-        """Create a persisted modal.Dict which as a lifetime beyond the app it was created in.
-        The object will persist until it is deleted by the user."""
+        """Deprecated! Use `Dict.from_name(name, create_if_missing=True)`."""
+        deprecation_warning((2024, 3, 1), _Dict.persisted.__doc__)
         return _Dict.from_name(label, namespace, environment_name, create_if_missing=True)
 
     @staticmethod
--- a/modal/network_file_system.py
+++ b/modal/network_file_system.py
@@ -114,8 +114,17 @@ class _NetworkFileSystem(_Object, type_prefix="sv"):
         environment_name: Optional[str] = None,
         create_if_missing: bool = False,
     ) -> "_NetworkFileSystem":
-        """Create a reference to a persisted network filesystem
+        """Create a reference to a persisted network filesystem, optionally creating it lazily.
 
+        **Examples**
+
+        ```python notest
+        volume = NetworkFileSystem.from_name("my-volume", create_if_missing=True)
+
+        @stub.function(network_file_systems={"/vol": volume})
+        def f():
+            pass
+        ```
         """
 
         async def _load(self: _NetworkFileSystem, resolver: Resolver, existing_object_id: Optional[str]):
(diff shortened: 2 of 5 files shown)


freider:
Might be bike-shedding names at this point, but wouldn't a default-creation of missing objects be more convenient?

I'd even go so far as to say `.persisted()`  and `.ephemeral()` would be preferable to `.from_name()` and `new()` to make things a bit more clear, with `.persisted()` having a default `create_if_missing=True` that can be set to `False` if the user prefers it to not exist (like if they are using a local env var to configure it). Wdyt?

Just a slight preference. But this is fine too :)

mwaskom:
+1 on `.from_name(create_if_missing=True)` being a little verbose for the most common (one might say "modal") case. Though I also have a preference for constructors being named `from_*}` unless the name other obviously connotes creation. 🤔

erikbern:
I sort of agree with your point @freider and there's a school of thought that [boolean arguments are "bad"](https://martinfowler.com/bliki/FlagArgument.html) vs having multiple functions.

However:

1. My concern is that `.persisted` and `.from_name` have such different names that I think it's really confusing what's the difference.
2. `.persisted` has a side-effect which isn't obvious at all right now, and pretty confusing
3. I think the mental model of how users think about the cloud is that you create buckets before you run the code, and I'd rather push for that in our docs – i.e. you run `modal volume create` and _then_ you use `.from_name` to refer to it.

If we keep it as two separate methods then I'd rather have names that are somewhat similar and also make it clear that there's a side effect. I can't think of anything obvious though... `from_name` vs `from_name_create_if_missing` makes it very obvious but it seems pretty ugly

mwaskom:
> there's a school of thought that [boolean arguments are "bad"](https://martinfowler.com/bliki/FlagArgument.html) vs having multiple functions.

Interesting! I don't find this super compelling though:

>> My reasoning here is that the separate methods communicate more clearly what my intention is when I make the call. Instead of having to remember the meaning of the flag variable when I see book(martin, false) I can easily read regularBook(martin).

since in Python you can always spell that with the keyword argument (and the API can even require that if it makes sense).

I think a counterargument from a _user_ perspective (rather than a code maintainer) is that an unambiguous "call this to create the object" method that has a few parameters makes it easier to understand the full space of options, otherwise e.g. a user who sees most of our examples using `Volume.persisted` might not even realize that there is the option of having `Volume.ephemeral` until they find themselves browsing the API docs one day.

> I think the mental model of how users think about the cloud is that you create buckets before you run the code, and I'd rather push for that in our docs – i.e. you run modal volume create and then you use .from_name to refer to it.

Not sure I feel too strongly here but just noting that this be thought of as running counter to the infra 2.0 manifesto ;) 

> I don't ever want to provision anything in advance of load.

(Though granted it's not exactly the same as what you're talking about there).

> from_name_create_if_missing

🤮

erikbern:
> running counter to the infra 2.0 manifesto ;)

That's roughly true, although I want to bring back ephemeral objects in a slightly different shape later – something like `Dict.ephemeral()` in a way where it's not associated with apps. That way you can also use it in non-app contexts.

The main challenge is that the garbage collection logic would be somewhat more complex.

freider:
> I sort of agree with your point @freider and there's a school of thought that [boolean arguments are "bad"](https://martinfowler.com/bliki/FlagArgument.html) vs having multiple functions.
> 
> However:
> 
> 1. My concern is that `.persisted` and `.from_name` have such different names that I think it's really confusing what's the difference.
> 2. `.persisted` has a side-effect which isn't obvious at all right now, and pretty confusing
> 3. I think the mental model of how users think about the cloud is that you create buckets before you run the code, and I'd rather push for that in our docs – i.e. you run `modal volume create` and _then_ you use `.from_name` to refer to it.
> 
> If we keep it as two separate methods then I'd rather have names that are somewhat similar and also make it clear that there's a side effect. I can't think of anything obvious though... `from_name` vs `from_name_create_if_missing` makes it very obvious but it seems pretty ugly

Ah if we are pushing users toward prior object creation then this definitely makes sense. I also prefer a single function w/ boolean flags rather than two separate functions in this case.

Perhaps we could even make the `create_if_missing` flag configurable at a CLI level, e.g. `modal deploy --create-missing bla.py` in order to have clean-slate deployments more streamlined without having to modify the code?

erikbern:
> Perhaps we could even make the `create_if_missing` flag overrideable at a CLI level, e.g. `modal deploy --create-missing bla.py` in order to have clean-slate deployments more streamlined without having to modify the code?

Yeah I think this might be good!

freider (approved):
I like it

erikbern:
Thank you @freider!

PR #1470, opened 2024-03-04
Pdm cache fix

freider:
## Describe your changes
Fix issue where all pdm cached packages would be automounted (MOD-2485)

The underlying issue was that pdm with cache symlinks the actual package data into site-packages, and we called `resolve()` on the package paths before comparing them with SYS_PREFIXES.

This patch changes the logic to preserve the un-expanded symlink paths for purposes of automount evaluation, but still resolves paths for serve/watch and deduplication

## Changelog

* Fix issue with pdm where all installed packages would be automounted when using package cache (MOD-2485)

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -97,7 +97,7 @@ class _MountFile(_MountEntry):
         return str(self.local_file)
 
     def get_files_to_upload(self):
-        local_file = self.local_file.expanduser().absolute()
+        local_file = self.local_file.resolve().expanduser()
         if not local_file.exists():
             raise FileNotFoundError(local_file)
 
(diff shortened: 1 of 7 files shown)


freider:
@prbot approve

modal/mount.py
```diff
@@ -120,10 +120,10 @@ class _MountDir(_MountEntry):
     recursive: bool
 
     def description(self):
-        return str(self.local_dir)
+        return str(self.local_dir.resolve().expanduser())
```
gimaker, line 123: this looks problematic?

```
In [22]: pathlib.Path("~/foo").resolve().expanduser()
Out[22]: PosixPath('/home/staffan/src/modal/~/foo')

In [23]: pathlib.Path("~/foo").expanduser().resolve()
Out[23]: PosixPath('/home/staffan/foo')
```

gimaker (commented):
:heavy_check_mark: lgtm overall, but see comment about `expanduser()` and `resolve()` interacting weirdly

modal/mount.py
mwaskom: +1

PR #1471, opened 2024-03-04
Adds failing test for cancelling multiple concurrent async inputs gracefully

freider:
This adds a (currently failling) test that we can cancel multiple async inputs at the same time without terminating the container or unaffected inputs.

We should be able to refactor the container entrypoint to cancel inputs *in case of async functions* by keeping a map of input_id -> asyncio task and then only cancelling the task(s) related to the relevant input ids.

This has several advantages:
* We can gracefully terminate multiple async inputs without affecting other concurrent inputs in the same container 
* It would let the SIGUSR1/SIGINT triggers be reserved only for the non-async case (sigusr1 without concurrency and sigint/task termination for input_concurrency > 1).
* We can probably get rid of the ugly "global marker" cancellation message hack for older Python, since it would then only be the user catching these cancellations anyway.

cc @ekzhang

### DIFF
--- a/test/container_test.py
+++ b/test/container_test.py
@@ -1135,15 +1135,10 @@ def test_cancellation_aborts_current_input_on_match(
 
 @skip_windows("signals not supported on windows and this only runs on containers")
 @pytest.mark.usefixtures("server_url_env")
-@pytest.mark.parametrize(
-    ["function_name"],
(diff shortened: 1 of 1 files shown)


freider:
Similar test incorporated in the patch that fixed concurrent cancellation of async tasks, closing this

PR #1472, opened 2024-03-04
Validate functions with `concurrency_limit < keep_warm`

irfansharif:
## Changelog

- For modal functions/classes with `concurrency_limit` < `keep_warm`, we'll raise an exception now. Previously we (silently) respected the `concurrency_limit` parameter.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -599,7 +599,7 @@ class _Function(_Object, type_prefix="fu"):
         allow_concurrent_inputs: Optional[int] = None,
         container_idle_timeout: Optional[int] = None,
         cpu: Optional[float] = None,
-        keep_warm: Optional[int] = None,
+        keep_warm: Optional[int] = None,  # keep_warm=True is equivalent to keep_warm=1
         cloud: Optional[str] = None,
         _experimental_boost: bool = False,
         _experimental_scheduler: bool = False,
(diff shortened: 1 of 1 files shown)


modal/functions.py
```diff
@@ -695,15 +695,22 @@ def from_args(
                 )
                 image = image.extend(build_function=snapshot_function, force_build=image.force_build)
 
+        if keep_warm is not None:
+            assert isinstance(keep_warm, int)
```
mwaskom, line 699: I know you're just moving this around but it caught my eye: for user-facing errors I think it's preferable to do something like `raise TypeError(f"keep_warm must be an int, not {type(keep_warm).__name__}"`.

mwaskom (commented):
How would you feel about raising instead of warning here? Warnings are easy to miss (they might get quickly buried under lots of image build logs) and it feels like it's going to be hard for users to debug that `keep_warm` is being ignored, since they have to notice slow cold starts.

modal/functions.py
irfansharif: Done.

mwaskom (approved):
LGTM! Might make sense to add a changelog entry since this may "break" some people (even if it's probably for the best).

irfansharif:
Done, thanks for the review.

PR #1474, opened 2024-03-04
Client interface for setting warm pool on functions

gongy:


modal/functions.py
```diff
@@ -1009,6 +1009,26 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
 
         return fun
 
+    async def set_keep_warm(self, warm_pool_size: int) -> None:
```
mwaskom, line 1012: Pure bike-shedding but when the attribute is already a verb, not sure you need `set_` (especially when there's no corresponding `get_` method).

gongy: Yeah, that seems reasonable. @aksh-at what do you think?

I was mainly following this: https://linear.app/modal-labs/issue/MOD-1684/make-it-possible-to-update-keep-warm-dynamically

Server changes need to go in first, so no rush on this PR.

aksh-at: No strong pref either. The `set_*` could be nicer if we have more of these, to be consistent. E.g. if you can dynamically `set_concurrency_limit`, etc.

aksh-at: I guess the "root cause" is that `keep_warm` is a verb that we use as a declarative parameter. Should have used a noun there.

mwaskom: That's a good point.

OTOH, if we anticipate other options that are too risky to set via `with_options` would it make sense to group them under a different batch method? (something like `with_options_i_really_mean_it`, but not literally that)?

gongy: How about `set_min_tasks` and migrate `keep_warm` to `min_tasks` like we've always wanted to?

PR #1477, opened 2024-03-05
Don't use `_run_stub` within `_serve_stub`

aksh-at:
Previously we had a complicated setup where we'd initially use `_run_stub` to initialize the stub the first time, but future updates would recreate the app within a separate process. This would leave the original `stub` in a weird state, where `LocalApp` would get set to `None` after the update. 

This PR moves all the necessary logic from `_run_stub` to `_serve_stub` — most of it was pretty unnecessary for this use case anyway. Note that it changes the interface for `serve_stub` so it now yields an `app_id` and not a `stub` object. That should probably be fine? (We have an internal integration test that relied on it, but we can update it)

### DIFF
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -195,7 +195,7 @@ async def _serve_update(
         )
 
         # Communicate to the parent process
-        is_ready.put(app.app_id)
+        is_ready.put((app.app_id, app.log_url()))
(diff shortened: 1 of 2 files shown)


aksh-at:
Actually will have to move more exception handling over as well, so this might not be worth it.

PR #1479, opened 2024-03-05
Stub composition/blueprints

freider:
## Describe your changes

Adds basic stub composition (aka blueprints) support. This makes multi-file modal apps much more natural to compose, without having to rely on side-effect imports or work around circular imports.

Useful splitting up Modal apps across different self-contained files
```python
stub_a = modal.Stub("a")
@stub.function()
def foo():
    ...

stub_b = modal.Stub("b")  # this would typically be in a different file
@stub.function()
def bar():
    ...

stub_a.include(stub_b)
@stub_a.local_entrypoint()
def main():
    # use function declared on the included stub
    bar.remote()
```

TODO (not necessarily in first revision):
* Add a `prefix` argument or similar to the `include()` method to prefix (or "namespace") any included objects under a name in the target stub, in order to not accidentally overwrite objects with the same names. This would be relatively straight forward for functions and classes, but tricky if we ever want it to propagate to "external" named objects, like a `Dict.from_name()` since it's not tied to the stub. To facilitate writing "libraries" we might have to add some way of at least "lightly" tie such objects to stubs similar to our soon-to-be-deprecated tagged objects.

## Changelog

* Stubs/apps can now be "composed" from several smaller stubs using `stub.include(...)`. This allows more ergonomic setup of multi-file Modal apps.


test/stub_composition_test.py
```diff
@@ -0,0 +1,10 @@
+# Copyright Modal Labs 2024
+from test.helpers import deploy_stub_externally
+
+
+def test_stub_composition_includes_all_functions(servicer, supports_dir, monkeypatch, client):
+    print(deploy_stub_externally(servicer, "main.py", cwd=supports_dir / "multifile_project"))
```
mwaskom, line 6: Intentional?

erikbern (approved):
Nice! Game changer

erikbern:
presumably sub-stubs don't need names, just the main stub?

mwaskom (commented):
Is there an easy way for the tests to assert that `a_func` would be able to call `c_func`? That seems like the critical behavior that we want to test.

mwaskom:
Also should we demonstrate this in https://modal.com/docs/guide/project-structure?

mwaskom:
One thing that isn't totally clear to me is the effect on deployments. With this setup, if you deploy `stub_a`, does that also deploy `stub_b` as a side effect? Or is the composed `stub_a` separate?

freider:
> Also should we demonstrate this in https://modal.com/docs/guide/project-structure?

Yes, I'll do that as a followup once we have tested it out a bit more

freider:
> One thing that isn't totally clear to me is the effect on deployments. With this setup, if you deploy `stub_a`, does that also deploy `stub_b` as a side effect? Or is the composed `stub_a` separate?

It would only deploy the deployed stub, but it would contain all of the elements from the other stubs

test/stub_composition_test.py
freider: Not really, I think it was mostly for debugging

freider:
> presumably sub-stubs don't need names, just the main stub?

Yes, unless they are in the same file, I think? At least that used to be the case, but it might have changed

freider:
> Is there an easy way for the tests to assert that `a_func` would be able to call `c_func`? That seems like the critical behavior that we want to test.


I'll try to add a test for this in a followup! An integration test should be easy to do (in the server repo)

PR #1485, opened 2024-03-05
Updates interface from `S3Mount` to `CloudBucketMount`

luiscape:
The `S3Mount` constructor is now superseded with `CloudBucketMount`:

 ```python
import modal
import subprocess

stub = modal.Stub()
@stub.function(
    volumes={
        "/my-mount": modal.CloudBucketMount("s3-bucket-name", secret=modal.Secret.from_dict({
            "AWS_ACCESS_KEY_ID": "...",
            "AWS_SECRET_ACCESS_KEY": "...",
        }), read_only=True)
    })
def f():
    subprocess.run(["ls" ,"/my-mount"])
```

`CloudBucketMount` only supports S3 buckets at the moment.

### CODE REVIEW

**Comment by luiscape** [comment]
@prbot approve

**Review by erikbern — COMMENTED** [review]
👍 

Since we're mounting both `Volume` and `CloudBucketMount`, should we make it possible to mount `NFS` and `Mount` using the same syntax? And deprecate the old arguments?

**Comment by luiscape** [comment]
Yes, good idea.

**Comment by mwaskom** [comment]
Mounts don't currently have the same type as the others right?
```
    mounts: Sequence[_Mount] = (),
    network_file_systems: Dict[Union[str, PurePosixPath], _NetworkFileSystem] = {},
    volumes: Dict[Union[str, PurePosixPath], _Volume] = {},
```

luiscape:
No, they don't.

PR #1490, opened 2024-03-05
Fix a few small issues in the Stub.function / Stub.cls parameters + descriptions

mwaskom:
I noticed that we had some inaccurate parameter descriptions (e.g. `volumes` was described as experimental). I fixed that, added a few missing descriptions, and tweaked a few others here and there.

I also reorganized the parameter lists a little bit by pushing the deprecated parameters to the bottom (this is fine as these are all keyword-only parameters).

I think there's more that can be done here but I'd rather defer it until we (1) discuss whether these should be separate decorators or (2) implement actual docstrings with a single-source of truth for parameter descriptions.


mwaskom:
@prbot lgtm

gimaker (commented):
:heavy_check_mark: nice, lgtm

PR #1493, opened 2024-03-06
Update image.py

NicholasKao1029:
expose GPU for `image.extend` function

## Describe your changes

- expose the `gpu` parameter for `image.extend` similar to `image.pip_install`

## Changelog

Let's a user enable a `gpu` when extending

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -353,11 +353,13 @@ class _Image(_Object, type_prefix="im"):
                 'RUN echo "hello world" > hello.txt',
             ],
             secrets=[secret1, secret2],
+            gpu="T4",
         )
         ```
         """
-            gpu_config = parse_gpu_config(gpu)
-            return _Image._from_args(base_images={"base": self}, gpu_config=gpu_config, **kwargs)
+        
+        gpu_config = parse_gpu_config(gpu)
+        return _Image._from_args(base_images={"base": self}, gpu_config=gpu_config, **kwargs)
 
     @typechecked
     def copy_mount(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Image":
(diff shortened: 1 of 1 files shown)


mwaskom:
Hey @NicholasKao1029 thanks for the PR!

Layers defined through `Image.extend` already can be run with a GPU, ala the following example:

```python
import modal

image = (
    modal.Image.from_registry("nvcr.io/nvidia/pytorch:22.12-py3")
    .extend(dockerfile_commands=["FROM base", "RUN nvidia-smi"], gpu_config=modal.gpu.parse_gpu_config("T4"))
)
stub = modal.Stub(image=image)

@stub.function()
def f():
    pass
```

Unlike most of the `Image` convenience method, `gpu_config` is narrowly typed as a `modal_proto.api_pb2.GPUConfig` object. But `Image.extend` is a low-level command so that makes sense.

I'm curious if there's a specific usecase that's required you to drop down to using `extend` directly. Is there something we could support better with a convenience method?

NicholasKao1029:
@mwaskom Thanks this makes sense. No we use `extend` because for a portion of our install commands we use docker definitions already. Appreciate the explanation!

mwaskom:
Would [`Image.dockerfile_commands'](https://modal.com/docs/reference/modal.Image#dockerfile_commands) work for you here?

```
def run_commands(
    self,
    *commands: Union[str, List[str]],
    secrets: Sequence[_Secret] = [],
    gpu: GPU_T = None,
    force_build: bool = False,
) -> "_Image":
"""Extend an image with arbitrary Dockerfile-like commands."""
```

We're discussing making `Image.extend` private as it doesn't really have a user-friendly public interface and we can't think of any usecases for running it directly that aren't currently satisfied by the higher-level `Image` methods.

NicholasKao1029:
I think this could work and better!

PR #1495, opened 2024-03-06
Setup external client dependency mounts

mwaskom:
*WIP: we may not go in this direction*

## Describe your changes

This PR adds everything we need to build and use an "external client package" mount. These are core client dependencies needed in the container runtime that will be mounted to a prioritized location on sys.path. However this PR *does not actually enable the mount*. I want to be able to test the system in prod (by manually adding the deployed mount) before that happens.

The PR does change both the creation and the location of the "internal client package" (it will be at `/pgk/internal` rather than `/pkg`). That has an upstream dependency on a worker PR (https://github.com/modal-labs/modal/pull/10946). Analogously, the naming pattern for the deployed client mount has now changed to `modal-client-internal-mount-{client-version}`.

Additionally, I reorganized some of the code in the client, pulling the code relating to *building and publishing* mounts out of `mount.py` itself. I've put it into a new `_packaging.py` module. I don't love the name, but it's private, so maybe not a super high priority to get right.

- Resolves MOD-2495


freider:
I may not fully understand this yet, but why are the internal and external packages in separate directories? Couldn't all be in `/pkg`? (Mounts should be able to "overlap", I think)

Nit: If we do need them to be in separate directories, maybe they shouldn't be subdirectories of the old `pkg`-dir, since that is also on the path and might cause confusion for the ~0 users that use either of:
https://pypi.org/project/external/
https://pypi.org/project/internal/

:D

mwaskom:
> Couldn't all be in /pkg? (Mounts should be able to "overlap", I think)

Ah, I didn't know that! But in this new approach (that handles Python versions) the layout is eventually going to look like 

```
pkg
├── external
│   ├── 3.8
        ├── aiohttp
        ├── aiosignal
        ├── ...
│   ├── 3.9
│   ├── 3.10
│   ├── ...
└── internal
    ├── modal
    ├── modal_proto
    ├── ...
```

And then `sys.path` will have `["/pkg/internal", "/pkg/external/{python_version}", ...]`

> maybe they shouldn't be subdirectories of the old pkg-dir

Hm yeah I wasn't sure what was less disruptive, putting the mount on `/mnt could also conflict with people who have set mounts there? My impression was that we consider `/pkg` to be an "internal" directory but we could probably document and/or enforce that better.

PR #1496, opened 2024-03-07
Fixes asgi_wrapper not showing exceptions from user code

freider:
Previously we never called .result() on our app_task, so even though the underlying fastapi endpoints would emit a 500 error due to the exception, we wouldn't reraise that exception on the Modal side and therefore not technically *fail* the request or show a traceback

Also makes sure we clean up tasks in our asgi wrapper, and raise potential exceptions from those tasks when necessary to surface problems.

Adds a bunch of more "isolated" tests for the asgi wrapper

### DIFF
--- a/modal/_asgi.py
+++ b/modal/_asgi.py
@@ -1,22 +1,27 @@
 # Copyright Modal Labs 2022
 import asyncio
-from typing import Any, Callable, Dict
+from typing import Any, AsyncGenerator, Callable, Dict, List
 
 from asgiref.wsgi import WsgiToAsgi
 
 from modal_utils.async_utils import TaskContext
 
 from ._blob_utils import MAX_OBJECT_SIZE_BYTES
+from .config import logger
 from .functions import current_function_call_id
 
 
-def asgi_app_wrapper(asgi_app, function_io_manager):
+class NoRequestReceived(Exception):
+    pass
+
+
+def asgi_app_wrapper(asgi_app, function_io_manager) -> Callable[..., AsyncGenerator]:
     async def fn(scope):
         function_call_id = current_function_call_id()
         assert function_call_id, "internal error: function_call_id not set in asgi_app() scope"
 
         # TODO: Add support for the ASGI lifecycle spec.
-        messages_from_app: asyncio.Queue[Dict[str, Any]] = asyncio.Queue(1)
+        messages_from_app: asyncio.Queue[List[Dict[str, Any]]] = asyncio.Queue(1)
         messages_to_app: asyncio.Queue[Dict[str, Any]] = asyncio.Queue(1)
 
         async def fetch_data_in():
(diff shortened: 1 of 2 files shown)


modal/_asgi.py
```diff
@@ -28,26 +33,32 @@ async def fetch_data_in():
...
+                        [
+                            {
+                                "type": "websocket.close",
+                                "code": 1011,
+                                "reason": "Missing request, possibly due to cancellation or crash",
+                            },
+                            {"type": "websocket.disconnect"},
+                        ]
                     )
-                    await messages_to_app.put({"type": "websocket.disconnect"})
```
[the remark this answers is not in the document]
freider: Oh wow, nice catch! Can't believe I didn't notice the disconnect was on a different channel, I kind of just mechanically moved all the "messages" to the same batch without looking to closely. Will add a test since it's quite bad this wasn't caught (even if it's an edge case/error state)

modal/_asgi.py
```diff
@@ -28,26 +33,32 @@ async def fetch_data_in():
...
-                return
+                else:
+                    logger.error(f"scope is neither http nor websocket {scope['type']})")
+                raise NoRequestReceived()
```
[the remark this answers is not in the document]
freider: This would be caught and reraised by the task.result() call in the main loop below:
```
logger.error(f"Data fetching task stopped unexpectedly: {fetch_data_in_task.exception()}")
fetch_data_in_task.result()
```
And that reraised exception would bubble all the way up to the modal function running, so it would basically error spit out a traceback and fail the function call for the user, which is probably (?) good since it's unexpected and the response will be a 502 to the client anyways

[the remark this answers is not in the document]
freider: The exception isn't actually handled in our code until after the function finishes, but this is mostly by design, since the asgi handler itself *should* raise a ClientDisconnect error before that (since we send an http.disconnect message to the asgi app).

I think this failure state isn't greatly handled by our asgi wrapper today either, since it typically leads to both us and whatever asgi framework the user uses to sending error asgi messages back to modal-http...

modal/_asgi.py
```diff
@@ -26,26 +31,36 @@ async def fetch_data_in():
...
+                            }
+                        ]
+                    )
+                    await messages_to_app.put(
+                        {"type": "websocket.disconnect"},
                     )
-                    await messages_to_app.put({"type": "websocket.disconnect"})
-                return
+                else:
+                    logger.error(f"scope is neither http nor websocket {scope['type']})")
```
ekzhang, line 62: This is not an error, there can be `lifespan` protocol scopes

modal/_asgi.py
```diff
@@ -71,10 +86,10 @@ async def send(msg):
                     indices = list(range(0, size, chunk_size))
                     for i in indices[:-1]:
                         chunk = msg["body"][i : i + chunk_size]
-                        await messages_from_app.put({"type": "http.response.body", "body": chunk, "more_body": True})
+                        await messages_from_app.put([{"type": "http.response.body", "body": chunk, "more_body": True}])
```
ekzhang, line 89: Why did we turn this into lists?

modal/_asgi.py
```diff
@@ -1,20 +1,25 @@
...
 from modal_utils.async_utils import TaskContext
 
 from ._blob_utils import MAX_OBJECT_SIZE_BYTES
+from .config import logger
 from .functions import current_function_call_id
 
 
-def asgi_app_wrapper(asgi_app, function_io_manager):
+class NoRequestReceived(Exception):
+    pass
```
ekzhang, line 13: Could we be consistent about placing exceptions in `modal.exception`?

modal/_asgi.py
freider: To ensure atomicity when adding multiple "related" messages at once (which we do in when the input generation task gets an error). At some point I managed to get into a weird error state where our own internal error got intermixed with responses from the app.

modal/_asgi.py
freider: I see! Will fix

modal/_asgi.py
freider: You might be right, and I think some of my other changes here might make this unnecessary

modal/_asgi.py
ekzhang: Hmm, we're still `yield`-ing individual items though, so there's no guarantee that they'll be put on the `DataChannel` at the same time. Also I don't think our internal error is sent on this asyncio Queue, since the exception goes directly through the output channel

modal/_asgi.py
freider: Managed to get rid of it in the followup PR

modal/_asgi.py
freider: No, I mean the internal error we produce in the timeout of first-input, which still uses the `messages_from_app` Queue and added multiple messages there sequentially before this patch:

```
await messages_from_app.put({"type": "http.response.start", "status": 502})
await messages_from_app.put(
    {
    "type": "http.response.body",
     "body": b"Missing request, possibly due to cancellation or crash",
}
```

We should probably also ensure in the http case that these messages only get sent/queued if the asgi app hasn't already started sending responses and vice versa (but maybe modal-http would error if multiple `http.response.start` came in?)

PR #1498, opened 2024-03-07
Deprecate Image.extend

mwaskom:
## Describe your changes

`Image.extend` does not have a user-friendly public interface and provides no additional value over the other user-facing image methods. This PR deprecates it and updates our internal code to use `Image._from_args` directly. This adds a small amount of additional boilerplate. I'd be open to keeping `_Image._extend` as an internal convenience method, but as a counterargument I kind of like how this makes it more clear that the image methods do not mutate `self`.

Upstream dependency to remove usage from integration tests: https://github.com/modal-labs/modal/pull/10972

- Closes MOD-2513
- Closes #1493 

## Changelog

- The `Image.extend` method has been deprecated. This is a low-level interface and can be replaced by other `Image` methods that offer more flexibility, such as `Image.from_dockerfile`, `Image.dockerfile_commands`, or `Image.run_commands`.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -700,7 +700,11 @@ class _Function(_Object, type_prefix="fu"):
                     is_auto_snapshot=True,
                     _experimental_scheduler_placement=_experimental_scheduler_placement,
                 )
-                image = image.extend(build_function=snapshot_function, force_build=image.force_build)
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot approve

PR #1499, opened 2024-03-07
Revert 26aa0574 with conflict resolution

mwaskom:
Reverts #1479, which breaks integration tests.

Unfortunately this did not revert cleanly due to #1497 so re-applying it is probably going to be pretty annoying.


mwaskom:
Superseded by #1500

PR #1501, opened 2024-03-08
Fail on Python 3.7 and below

erikbern:
User ran into this!


mwaskom:
Surprising that `requires_python` didn’t take care of it, I wonder how they installed it?

erikbern:
yeah it's a bit strange

PR #1500, opened 2024-03-07
Revert recent changes to mounts to unbreak integration tests

mwaskom:
Reverts the changes from #1479 and #1497.

We think that #1479 broke integration tests but I was unable to cleanly revert it due to a subtle conflict with #1497.

I've also removed the note about stub composition from the changelog for now.

PR #1502, opened 2024-03-08
Revert "Revert recent changes to mounts to unbreak integration tests"

freider:
Reverts modal-labs/modal-client#1500

Work in progress to fix the integration tests breaking

## Changelog

* Stubs/apps can now be "composed" from several smaller stubs using `stub.include(...)`. This allows more ergonomic setup of multi-file Modal apps.

### DIFF
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -92,13 +92,16 @@ class Resolver:
             await obj._preload(obj, self, existing_object_id)
 
     async def load(self, obj: "_Object", existing_object_id: Optional[str] = None):
+        breakpoint()
+        print("LOADING", obj, existing_object_id, "DEDUP", self._deduplication_cache)
         deduplication_key = await obj._deduplication_key()
-        cached_future = None
-        if deduplication_key is not None:
-            cached_future = self._deduplication_cache.get(deduplication_key)
+        cached_future = self._local_uuid_to_future.get(obj.local_uuid)
 
-        if not cached_future:
-            cached_future = self._local_uuid_to_future.get(obj.local_uuid)
+        if not cached_future and deduplication_key is not None:
+            # deduplication cache makes sure duplicate mounts are resolved only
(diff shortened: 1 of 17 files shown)


modal/_resolver.py
```diff
@@ -122,20 +131,33 @@ async def loader():
 
             cached_future = asyncio.create_task(loader())
             self._local_uuid_to_future[obj.local_uuid] = cached_future
+            if deduplication_key is not None:
+                self._deduplication_cache[deduplication_key] = cached_future
+
+        def hydrate_original(fut):
+            # in case an object is omitted due to content duplication, make sure the original reference
```
freider, line 138: The previous bug making integration tests fail was here - there was a check `if obj.is_hydrated: return` which made stale anonymous mounts from previous deploys (typically auto mounted packages) not get new ids on a second deploy...

test/function_test.py
```diff
@@ -682,3 +682,21 @@ def test_mount_deps_have_ids(client, servicer, monkeypatch, test_dir):
     function_create = ctx.pop_request("FunctionCreate")
     for dep in function_create.function.object_dependencies:
         assert dep.object_id
+
+
+def test_no_state_reuse(client, servicer, supports_dir):
```
freider, line 687: Added an additional test here that reproduced the subtle state reuse issue we previously only detected only in integration tests

modal/object.py
```diff
@@ -204,9 +204,12 @@ async def resolve(self):
                 " wasn't defined in global scope."
             )
         else:
-            resolver = Resolver()
+            resolver = Resolver()  # TODO: this resolver has no attached Client!
```
freider, line 207: Added this TODO, as I noticed this is the only place we initialize a Resolver without a working client, which caused issues for lazily hydrated objects until I patched my way around it. Would still cause issues in the future if a lazy object used the resolver to load additional entities that relies on the client

freider:
@prbot approve

PR #1503, opened 2024-03-10
Bump interactive shell timeout to 60s

aksh-at:


mwaskom:
:+1:, could also make this a flag in the CLI.

ekzhang:
It's also a bit weird that the timeout is this high though, we don't know the cause of the delay right?

mwaskom:
Closing as stale — I think we suspect the issue is more complicated in any case. (Feel free to reopen and land if you still think it's a good idea though!)

PR #1506, opened 2024-03-11
Remove asgiref client dependency and fix request streaming in WSGI

ekzhang:
Resolves MOD-2517. This vendors the `a2wsgi` library, which is a pure-Python dependency. I tried to include the license properly.

## Changelog

- The `@wsgi_app()` decorator now uses a different backend based on `a2wsgi` that streams requests in chunks, rather than buffering the entire request body.

### DIFF
--- a/test/container_test.py
+++ b/test/container_test.py
@@ -385,7 +385,7 @@ def _get_web_inputs(path="/"):
         "type": "http",
         "path": path,
         "headers": {},
-        "query_string": "arg=space",
+        "query_string": b"arg=space",
         "http_version": "2",
     }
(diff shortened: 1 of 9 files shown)


mwaskom (approved):
LGTM!

test/container_test.py
```diff
@@ -500,6 +499,33 @@ def test_asgi(unix_servicer, event_loop):
     assert json.loads(second_message["body"]) == {"hello": "space"}
 
 
+@skip_windows_unix_socket
+def test_wsgi(unix_servicer, event_loop):
```
erikbern, line 503: great with test!

PR #1507, opened 2024-03-11
Consolidate all util modules into `modal/_utils`

mwaskom:
Resolves MOD-2491

We use some of these server-side too so I'm going to do this in a couple steps:

This PR:
- [x] Run CI against the repo with `modal_utils` moved to `modal._utils`
- [x] Copy `modal._utils` back to `modal_utils`
- [ ] Merge and publish

Subsequent PRs:
- [ ] Update the server code
- [ ] Remove `modal_utils` from this repository

Additionally, we had a few `_*_utils.py` files in `modal/`; I've moved them to `modal/_utils` so that everything is together.


ekzhang:
Nice! Just need to add a period in the right place :)

ekzhang (approved):
I have a slight preference to just move the files directly instead of doing the duplicated-code trick because the current method will lose all of our git history, and some of these files are quite tricky :/

It wouldn't be too bad to just merge something really quickly afterward on the monorepo? 😬

mwaskom:
> I have a slight preference to just move the files directly instead of doing the duplicated-code trick because the current method will lose all of our git history, and some of these files are quite tricky :/

 valid concern though it looked when I did the commits that it registered a move and then new files when I added it back. Maybe we will lose that if we squash?

> It wouldn't be too bad to just merge something really quickly afterward on the monorepo? 😬

modal_utils is used pretty extensively in the server so I was hoping to be able to rely on tests there … I guess I could run ci against this branch or something but still feels risky

ekzhang:
> Maybe we will lose that if we squash?

Yeah I think so, since Git doesn't actually store diffs, it only stores the snapshots for each commit

> I was hoping to be able to rely on tests there … I guess I could run ci against this branch

might save some time regardless :)

mwaskom:
Right now the blame is showing source information for lines in `modal/_utils/`: https://github.com/modal-labs/modal-client/blame/michael/incorporate-modal-utils/modal/_utils/grpc_testing.py

Should be fine if we don't squash on merge?

mwaskom:
Confirming that the git history has been maintained successfully after merging to main!

ekzhang:
Nice! Git master :D

PR #1508, opened 2024-03-12
Sandbox Stdin

brianshih1:
[MOD-2483](https://linear.app/modal-labs/issue/MOD-2483/support-writing-to-stdin-for-sandboxes)

[PR](https://github.com/modal-labs/modal/pull/11047) for server

This PR supports writing to a sandbox process's `STDIN`. Here is how to use the `STDIN` API:
```
@stub.local_entrypoint()
def main():
    sandbox = stub.spawn_sandbox(
        "bash",
        "-c",
        "while read line; do echo $line; done",
    )

    sandbox.stdin.write("foo\n")
    sandbox.stdin.write("bar\n")

    sandbox.stdin.write_eof()

    sandbox.wait()
    print(sandbox.stdout.read())

```
Output:
```
foo
bar
```

Async version:
```
@stub.local_entrypoint()
async def main():
    sandbox = stub.spawn_sandbox(
        "bash",
        "-c",
        "while read line; do echo $line; done;",
    )
    # not wrapping
    sandbox.stdin.write(b"curry\n")
    sandbox.stdin.write(b"iverson\n")
    sandbox.stdin.write_eof()
    await sandbox.stdin.drain.aio()
    sandbox.wait()

```

This API closely mirrors `asyncio.subprocess`'s [StreamWriter](https://docs.python.org/3/library/asyncio-stream.html#streamwriter) API:
```
import asyncio

async def run_executable():
    process = await asyncio.create_subprocess_exec(
        '/bin/bash', '-c', 'while read line; do echo $line; done',
        stdin=asyncio.subprocess.PIPE,
        stdout=asyncio.subprocess.PIPE,
        stderr=asyncio.subprocess.PIPE
    )

    # Write data to the subprocess stdin
    process.stdin.write(b'foo\n')
    process.stdin.write(b'bar\n')
    await process.stdin.drain()
    process.stdin.write_eof()

    # Wait for the subprocess to finish
    await process.wait()

   # Read output from the subprocess stdout
    output = await process.stdout.read()
    print(output.decode())

```

Also, my current implementation assumes that `Sandbox` is not `thread-safe`. (i.e. there won't be two Sandbox instances with the same `sandbox_id` making `stdin.write` calls)


modal_proto/api.proto
```diff
@@ -623,6 +623,7 @@ message DataChunk {
     string data_blob_id = 3;
   }
   uint64 index = 4; // Index of this data chunk in the stream.
+  optional bool is_eof = 5; // Used for sandbox stdin
```
brianshih1, line 626: Added this so that I could reuse the same `DataChannel` used for storing user's `stdin` (`ContainerInput`) in interactive mode for storing Sandbox's `stdin.write` and `stdin.write_eof` data.

modal/sandbox.py
```diff
@@ -91,7 +91,36 @@ async def _get_logs():
...
+    """Provides an interface to buffer and write logs to a sandbox stream (`stdin`)."""
+
+    def __init__(self, sandbox_id: str, client: _Client):
+        self._index = 1
+        self._sandbox_id = sandbox_id
+        self._client = client
+
+    def get_next_index(self):
+        index = self._index
+        self._index += 1
```
brianshih1, line 104: This implementation means that the `Sandbox` is not `thread-safe`. If there are two instances of Sandbox performing writes, they will overwrite the `_index` here. Going to assume `Sandbox` isn't thread-safe in general now.

modal/sandbox.py
```diff
@@ -91,7 +91,36 @@ async def _get_logs():
         return data
 
 
+class _StreamWriter:
```
brianshih1, line 94: I call it the `_StreamWriter` to mirror `asyncio.subprocess`'s `StreamWriter`. Happy to rename to `LogWriter` or something else.

modal_proto/api.proto
```diff
@@ -1492,6 +1493,25 @@ message SandboxGetLogsRequest {
   string last_entry_id = 4;
 }
 
+message SandboxStdinWriteRequest {
```
brianshih1, line 1496: Created a separate [PR](https://github.com/modal-labs/modal-client/pull/1510) for proto changes

modal_proto/api.proto
```diff
@@ -1492,6 +1493,25 @@ message SandboxGetLogsRequest {
...
+  string sandbox_id = 1;
+  bytes input = 2;
+  uint32 index = 3;
+}
+
+message SandboxStdinWriteResponse {
```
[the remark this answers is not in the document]
brianshih1: keeping this here in case we want to return things in the future?

modal/sandbox.py
```diff
@@ -91,7 +91,38 @@ async def _get_logs():
...
+        self._index = 1
+        self._sandbox_id = sandbox_id
+        self._client = client
+
+    def get_next_index(self):
+        index = self._index
+        self._index += 1
+        return index
+
+    async def write(self, input: str):
```
thundergolfer, line 107: Should this be `bytes` not `str`? Don't know why we'd force the client to write strings.

modal/sandbox.py
thundergolfer: Called out IRL that the write methods should actually be sync if we're following the asyncio interface

brianshih1: good call!

modal/sandbox.py
```diff
@@ -91,7 +91,45 @@ async def _get_logs():
...
+        self._client = client
+        self._is_closed = False
+        self._buffer = bytearray()
+
+    def get_next_index(self):
+        index = self._index
+        self._index += 1
+        return index
+
+    def write(self, data):
```
brianshih1, line 109: I made `data` type-less but asserted that `isinstance(data, (bytes, bytearray, memoryview))`. This mirrors the API and semantics of `StreamWriter::write`

mwaskom: Why not have a union type annotation?

test/sandbox_test.py
```diff
@@ -105,3 +105,49 @@ def test_sandbox_terminate(client, servicer):
         sb.terminate()
 
         assert sb.returncode != 0
+
+
+@skip_non_linux
+@pytest.mark.asyncio
+async def test_sandbox_stdin_async(client, servicer):
```
brianshih1, line 112: I have an `async` test and a `sync` test that pretty much does the same thing. Should I remove one of them?

test/sandbox_test.py
```diff
@@ -105,3 +105,49 @@ def test_sandbox_terminate(client, servicer):
...
+    with stub.run(client=client):
```
thundergolfer, line 113: I think this should be `stub.aio.run(`

modal/sandbox.py
```diff
@@ -91,7 +91,44 @@ async def _get_logs():
...
 
+class _StreamWriter:
+    """Provides an interface to buffer and write logs to a sandbox stream (`stdin`)."""
+
+    def __init__(self, sandbox_id: str, client: _Client):
+        self._index = 1
+        self._sandbox_id = sandbox_id
+        self._client = client
+        self._is_closed = False
+        self._buffer = bytearray()
```
thundergolfer, line 102: Probably want a max size check somewhere to prevent clients from filling their process memory indefinitely with unflushed buffer

modal/sandbox.py
brianshih1: that's true I guess we're not tied to `asyncio.subprocess`'s API

modal/sandbox.py
```diff
@@ -91,7 +91,44 @@ async def _get_logs():
...
+
+    def get_next_index(self):
+        index = self._index
+        self._index += 1
+        return index
+
+    def write(self, data):
+        if self._is_closed:
+            return
```
thundergolfer, line 111: Shouldn't this be an error? Does writing to a asyncio `StreamWriter` silently drop the write?

thundergolfer (commented):
Almost there!

modal/sandbox.py
brianshih1: Yeah I confirmed and looked at their code, they fail silently. But after a certain number of failed writes they log a warning. Maybe we can log a warning for our user?

Screenshot 2024-03-13 at 6 14 18 PM

modal/sandbox.py
brianshih1: I was originally going to leave the flushing mechanism for the upcoming PR. But let me put together a basic implementation.

modal/sandbox.py
mwaskom: Yeah — I'd argue anyway that we have the same API, we're just enforcing it only at runtime. Adding annotations will allow us to enforce is at check time too.

modal/sandbox.py
thundergolfer: Wow that is surprising. There must be some use-case where hard failing 'late' writes doesn't work. I'd opt to be strict until we discover a problem with the strictness.

modal/sandbox.py
```diff
@@ -91,7 +91,49 @@ async def _get_logs():
         return data
 
 
+MAX_BUFFER_SIZE = 4096
```
thundergolfer, line 94: I think I was wrong the first time, this limit is too low. Let's do 128 KiB

thundergolfer (approved):
LGTM after addressing outstanding comments. 

(Can't really merge this though until workers have rolled out the modal-side change)

modal/sandbox.py
brianshih1: ah makes sense, we can't use the same constant as `asyncio`'s `StreamWriter` because they immediately write to the underlying file and only buffers if there is a `BlockingIOError`.

thundergolfer:
🥳

PR #1509, opened 2024-03-12
Asgi wrapper: Only raise error due to input loop failure if the input is actually used

freider:
## Slightly improved asgi wrapper error handling

If the asgi input queue is empty and the underlying asgi app requests additional data, it should raise an exception in case the input loop shuts down unexpectedly while waiting for the queue to get data.

Additionally, we don't have to raise an error and fail the Modal function if `fetch_data_in_task` fails unless the user's asgi app actually needs the data (i.e. calling `receive()`). (We do however log a warning since it's unexpected behavior)


ekzhang (commented):
Why mix `asyncio.create_task` with `tc.create_task`?

Also, `fetch_data_in_task` should not normally have an exception. Is this something we need to check for explicitly? I'm worried that the additional 30 lines of complex asyncio event loop code will surface other bugs.

Also see comments on #1496

freider:
> Why mix `asyncio.create_task` with `tc.create_task`?
Unintentional, fixed

> Also, `fetch_data_in_task` should not normally have an exception. Is this something we need to check for explicitly? I'm worried that the additional 30 lines of complex asyncio event loop code will surface other bugs.
Not normally, but I'm a bit worried about unexpected errors leading to hard-to-debug issues. Refactored the input loop itself to log errors and send a disconnect to the app instead, to get rid of the extra async complexity in `receive()` now. 
 
> Also see comments on #1496
Thanks for the thorough review. I got rid of the new exception through some refactoring and added a bunch of tests for coverage of most of the `_asgi` module.

freider:
@prbot approve

mwaskom (commented):
LGTM (cleaner now — thanks!) though also cc @ekzhang who has a much deeper background than me here

PR #1510, opened 2024-03-12
Proto changes for Sandbox STDIN

brianshih1:
## Describe your changes

Proto changes for this [PR](https://github.com/modal-labs/modal-client/pull/1508)

[PR](https://github.com/modal-labs/modal/pull/11067) for server changes.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1500,7 +1500,6 @@ message SandboxStdinWriteRequest {
 }
 
 message SandboxStdinWriteResponse {
-  GenericResult result = 1;
 }
 
 message SandboxHandleMetadata {
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2073,6 +2073,7 @@ service ModalClient {
   rpc SandboxWait(SandboxWaitRequest) returns (SandboxWaitResponse);
   rpc SandboxList(SandboxListRequest) returns (SandboxListResponse);
   rpc SandboxTerminate(SandboxTerminateRequest) returns (SandboxTerminateResponse);
+  rpc SandboxStdinWrite(SandboxStdinWriteRequest) returns (SandboxStdinWriteResponse);
 
   // Secrets
   rpc SecretCreate(SecretCreateRequest) returns (SecretCreateResponse);  // Not used by client anymore
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -623,6 +623,7 @@ message DataChunk {
     string data_blob_id = 3;
   }
   uint64 index = 4; // Index of this data chunk in the stream.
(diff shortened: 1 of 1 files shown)


ekzhang (dismissed):
DataChunk is a generic message used for a lot of data channel messages and shouldn't have sandbox-specific fields. If it needs fields like that, it should go on the `data_oneof` part

modal_proto/api.proto
```diff
@@ -623,6 +623,7 @@ message DataChunk {
     string data_blob_id = 3;
   }
   uint64 index = 4; // Index of this data chunk in the stream.
+  optional bool is_eof = 5; // Used for sandbox stdin
```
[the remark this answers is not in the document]
brianshih1: yeah I chatted with @thundergolfer about this one. We decided to put up a PR to see how you think about it. Putting it in `data_oneof` makes more sense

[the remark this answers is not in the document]
brianshih1: we also considered using a unique `UUID` to denote that it's an `is_eof` (only in the context of sandbox). Do you prefer the `UUID` approach or adding it to `data_oneof`?

modal_proto/api.proto
```diff
@@ -1948,6 +1968,7 @@ message RuntimeOutputBatch {
 message RuntimeInputMessage {
   bytes message = 1;
   uint64 message_index = 2;
+  bool is_eof = 3;
```
brianshih1, line 1971: I considered moving `message` and `is_eof` into a `oneof`, because `message` would be ignored if `is_eof` is true. However, that complicates the Rust code quite a bit so keeping it as is. Any strong opinions here?

modal_proto/api.proto
```diff
@@ -621,6 +621,7 @@ message DataChunk {
   oneof data_oneof {
     bytes data = 2;
     string data_blob_id = 3;
+    bool is_eof = 5; // Used for sandbox stdin
```
brianshih1, line 624: one approach we considered is to use a hardcoded `UUID` to represent an `is_eof` in the context of Sandbox. @ekzhang do you have a preference?

modal_proto/api.proto
```diff
@@ -1492,6 +1493,25 @@ message SandboxGetLogsRequest {
...
+message SandboxStdinWriteResponse {
+  GenericResult result = 1;
+}
+
+message SandboxStdinEofRequest {
+  string sandbox_id = 1;
+  uint32 index = 2;
+}
+
+message SandboxStdinEofResponse {
```
brianshih1, line 1511: remove these RPCs

modal_proto/api.proto
brianshih1: `is_eof` -> `eof`

thundergolfer (approved):

PR #1518, opened 2024-03-13
Consolidate modal_test_support into test/supports

mwaskom:
We had test support files distributed across `modal_test_support` and `test/supports`. This moves everything to the latter location. It needed a few updates to the testing code, but I didn't see any reason why we need these to be separate. Let me know if I'm missing something!


mwaskom:
@prbot approve based on @freider giving me a thumbs up on slack :)

PR #1522, opened 2024-03-14
Cleanup test spam

freider:
Minor changes and cleanups of tests to not have as much stdout/stderr spam. Also fixes a couple of instances of harmless unfinalized asyncio tasks causing warnings when their TaskContexts shut down


mwaskom (approved):
Thanks for this!!

ekzhang:
Do we still need the warning when we cancel tasks on TaskContext shutdown? Maybe we can just get rid of the warning? iirc @irfansharif turned it off for Python <3.11 already due to issues!

PR #1523, opened 2024-03-14
Remove on_shutdown/set_pre_stop

freider:
It causes occational exit spam, especially in tests, when an error occurs in the exit handler due to not being able to connect etc.

I *think* we don't need it anymore since `_run_stub`'s finally-block appears to trigger disconnect properly on KeyboardInterrupt now. I haven't been able to repro the old race (CC @thundergolfer who fought with this bug back in the day)


mwaskom:
Ah I think I had tried this and ran into the same issue — so lovely to be able to scroll right to the test results though!!!

mwaskom:
Guessing this is pretty stale / conflicted at this point. Probably better to start over if we want to make more progress on this.

PR #1524, opened 2024-03-14
Don't show output from non-failing tests when running in CI

mwaskom:
Revert to default pytest behavior of only showing stdout/stderror when a test fails. This will prevent the modal CLI output (primarily) from cluttering up the GitHub actions logs.


mwaskom:
Can always flip this back if we miss it but IMO this makes diagnosing CI failures much easier!

PR #1525, opened 2024-03-14
Fix mounted directories with symlinked files

mwaskom:
We introduced a bug in https://github.com/modal-labs/modal-client/pull/1470 while trying to address an edge case in automounting behavior that arises with `pdm`. The consequence of the bug is that symlinks get dropped when we mount a local directory. This PR attempts to fix that problem but in a fairly naive way — I just replaced calls to `Path.resolve` (which resolves symlinks) with `Path.absolute` (which does not).

This fixes my test case for the bug and passes our test suite, but I am a little concerned that it is going to reintroduce the issue we tried to fix in #1470  — I am not sure I have enough context there to say.

## Changelog

- Fixed an bug where` Mount` was failing to include symbolic links.

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -98,7 +98,7 @@ class _MountFile(_MountEntry):
         return str(self.local_file)
 
     def get_files_to_upload(self):
-        local_file = self.local_file.resolve().expanduser()
+        local_file = self.local_file.expanduser().absolute()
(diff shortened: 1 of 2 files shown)


freider (approved):
Nice! As long as the pdm test case passes, which it does, I think this wouldn't cause any issues

PR #1527, opened 2024-03-14
Minor simplification of asgi_app_wrapper / TODOs

ekzhang:
Thanks @freider for fixing up the error handling. I had some minor changes on this function as well / TODO handling.

### DIFF
--- a/modal/_asgi.py
+++ b/modal/_asgi.py
@@ -1,7 +1,6 @@
 # Copyright Modal Labs 2022
 import asyncio
-import time
-from typing import Any, AsyncGenerator, Callable, Dict, List
+from typing import Any, AsyncGenerator, Callable, Dict
 
 from ._utils.async_utils import TaskContext
 from ._utils.blob_utils import MAX_OBJECT_SIZE_BYTES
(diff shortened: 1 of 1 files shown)


modal/_asgi.py
```diff
@@ -17,45 +16,31 @@ async def fn(scope):
         assert function_call_id, "internal error: function_call_id not set in asgi_app() scope"
 
         # TODO: Add support for the ASGI lifecycle spec.
-        messages_from_app: asyncio.Queue[List[Dict[str, Any]]] = asyncio.Queue(1)
+        messages_from_app: asyncio.Queue[Dict[str, Any]] = asyncio.Queue(1)
```
ekzhang, line 19: This doesn't need to be a `List[]` — and I think that's a bit confusing because the messages aren't sent atomically regardless due to being outputted from a generator.

modal/_asgi.py
```diff
@@ -17,45 +16,31 @@ async def fn(scope):
...
-                await messages_to_app.put(
-                    {"type": "websocket.disconnect"},
-                )
+                await messages_to_app.put({"type": "websocket.disconnect"})
 
         async def handle_first_input_timeout():
             if scope["type"] == "http":
-                # TODO: prevent this from being sent if a response was already sent by the asgi app
-                #       also, prevent the (http) asgi app from sending additional data at this point
-                #       since this queues a full response
```
ekzhang, line 37: We have handling in `modal-http` to make sure that responses are handled properly. The `ResponseDriver` state machine in `modal-http/src/service.rs` handles this.

modal/_asgi.py
```diff
@@ -17,45 +16,31 @@ async def fn(scope):
...
+                await messages_from_app.put({"type": "http.response.start", "status": 502})
                 await messages_from_app.put(
-                    [
-                        {
-                            "type": "http.response.start",
-                            "status": 502,
-                        },  # TODO: should this have headers w/ content length/type?
```
ekzhang, line 43: Doesn't need content-length, since the `modal-http` server all sends responses with chunked encoding.

modal/_asgi.py
```diff
@@ -67,20 +52,15 @@ async def fetch_data_in():
...
-            t0 = time.monotonic()
             try:
                 first_message = await asyncio.wait_for(message_gen.__anext__(), FIRST_MESSAGE_TIMEOUT_SECONDS)
-            except asyncio.TimeoutError:
-                await handle_first_input_timeout()
-                return
-            except StopAsyncIteration:
-                # the generator shouldn't typically exit, but we handle it like a timeout in that case
-                remaining_grace_period = max(0.0, FIRST_MESSAGE_TIMEOUT_SECONDS - time.monotonic() - t0)
-                await asyncio.sleep(remaining_grace_period)
```
ekzhang, line 79: Good point about `StopAsyncIteration` — however, I think this code has a typo:

```
FIRST_MESSAGE_TIMEOUT_SECONDS - time.monotonic() - t0
```

Should be


```
FIRST_MESSAGE_TIMEOUT_SECONDS - time.monotonic() + t0
```

I also think it's probably not necessary to sleep additionally here?

ekzhang:
@prbot approve

Going to merge because I don't want to deal with two layers of merge conflicts in #1513 — and the behavior is similar to an older version of the client.

Happy to receive changes afterward though, will try to get #1513 out soon.

modal/_asgi.py
freider: Yeah that was a typo. I think the sleep can still be useful to prevent shutting down an asgi call prematurely if it would return data within the first five seconds without actually needing any input data to do so, which would be consistent with how we handle the timeout here.

modal/_asgi.py
freider: The output is sent individually from a generator, and this isn't for preventing partial responses but rather for preventing intermixed responses from the asgi app itself and our error messages, for example:

```python
async def handle_first_input_timeout():
            if scope["type"] == "http":
                await messages_from_app.put({"type": "http.response.start", "status": 502})
                # <---- At this point the asgi app could be queuing another message on our queue, effectively breaking up the two "related" queue entries we are adding
                await messages_from_app.put(
                    {
                        "type": "http.response.body",
                        "body": b"Missing request, possibly due to cancellation or crash",
                    }
                )
```

Another option for preventing this would be to add some kind of simple locking mechanism (a simple started_sending=True bool should suffice?), preventing either the user app or our own error responses from being sent if the other has already started queuing responses.

ekzhang: Do you have a reproduction of this? It doesn't really make sense to me that a user app could send a response before actually receiving a request.

modal/_asgi.py
ekzhang: Hmm, so the five second timeout is a very abnormal circumstance and indicates some kind of server failure or very rare cancellation race condition. I don't think it's intended to be explicit behavior!

GET requests should have a http.request message as well.

ekzhang: In the past this behavior was also only triggered on server failures where web endpoints got backed up for 15 minutes.

PR #1529, opened 2024-03-15
Fix presumed mixup of is_another_app tagging in Dict

freider:
I assume this was unintentional (or I'm misunderstanding what the flag is used for)


erikbern:
Yes this looks like a mistake. Not sure what happened but one theory is I did a rebase at some point and git broke up the functions in the wrong way. Or maybe I was just drunk 🍻

erikbern (approved):

PR #1530, opened 2024-03-15
Fix sandbox ephemeral object reuse

freider:
Fix for `spawn_sandbox()` with ephemeral object references created using `.new()` recreating the object for each call (MOD-2578)

However, this fix also breaks multiple subsequent `stub.run()` since they will also reuse stale ephemeral objects from previous runs using this logic. To make this work we'd need `stub._uncreated_all_objects` to properly unhydrate *all* object references linked to the app, not just the named references.


freider:
Non-trivial to solve in a way that works both across stub runs and across sandboxes within a stub - closing this since we are deprecating .new() anyways and this is a non-issue for named or appless objects.

PR #1531, opened 2024-03-15
Fix for potential dying container heartbeat loop (!)

freider:
Realized there might have been a regression in error resilience in our heartbeat loop as part of the long polling refactor of container heartbeats a couple of weeks ago!

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -141,9 +141,10 @@ class _FunctionIOManager:
             # pre Python3.8, CancelledErrors were a subclass of exception
             except asyncio.CancelledError:
                 raise
-            except Exception as exc:
+            except Exception:
                 # don't stop heartbeat loop if there are transient exceptions!
-                logger.warning("Heartbeat failure", exc_info=exc)
+                time_elapsed = time.monotonic() - t0
+                logger.exception(f"Heartbeat attempt failed (time_elapsed={time_elapsed})")
 
             heartbeat_duration = time.monotonic() - t0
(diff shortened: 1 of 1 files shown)


advay-modal:
i think this is related to an issue i've been debugging, will confirm

what do we generally write tests for? might be good to add a test for this since it was a regression?

freider:
Yeah will definitely write a test, just felt this is urgent enough that I'll push it out and see if it resolves issues for some users first

mwaskom (commented):
Great catch!

modal/_container_entrypoint.py
```diff
@@ -129,16 +129,25 @@ async def initialize_app(self) -> _ContainerApp:
...
+                if await self._heartbeat():
+                    # got a cancellation event, fine to start another heartbeat immediately
+                    # since the cancellation queue should be empty on the worker server
+                    # however, we wait at least 1s to prevent short-circuiting the heartbeat loop
+                    # in case there is ever a bug. This means it will take at least 1s between
+                    # two subsequent cancellations on the same task at the moment
+                    await asyncio.sleep(1.0)
+                    continue
+            # pre Python3.8, CancelledErrors were a subclass of exception
+            except asyncio.CancelledError:
```
erikbern, line 142: this handler isn't actually needed since it inherits from `BaseException`: https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError

aksh-at: yeah I guess we dropped support for Python 3.7 so we don't need it now, but we did in the past.

aksh-at: image

PR #1533, opened 2024-03-15
Do not warn when TaskContext cancels unfinished tasks

ekzhang:
See https://github.com/modal-labs/modal-client/pull/1522#issuecomment-1999782348

The history of this is that @irfansharif silenced the warning for Python versions up to 3.10 and below, two months ago. But it's still been happening for version 3.11, and it's a bit of an unhelpful warning in our case.

Note: `asyncio.TaskGroup` from the Python standard library waits for all tasks to finish when it exits normally. It only cancels tasks when it sees that one of its subtasks throws an exception other than `CancelledError`. However, our `async_utils.TaskContext` object is quite different in behavior, since it actually cancels _all_ tasks when it exits after a short grace period.

Code that is written for `TaskContext` will not work with `asyncio.TaskGroup`, and vice versa.

### DIFF
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -143,11 +143,7 @@ class TaskContext:
                 if task.done() or task in self._loops:
                     continue
 
-                if sys.version_info >= (3, 11):
-                    already_cancelling = task.cancelling() > 0
-                    if not already_cancelling:
-                        logger.warning(f"Canceling remaining unfinished task: {task}")
-
+                # Cancel any remaining unfinished tasks.
                 task.cancel()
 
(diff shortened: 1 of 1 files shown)


mwaskom (approved):
+1 from me, doesn't feel like an actionable warning for most users

freider:
Definitely makes sense from an end user perspective. It might make it a bit harder for us to spot unintentional cancellations internally though, so there's something to be mindful of.

freider (approved):

mwaskom:
Could be useful to have a way to track client behaviors that are relevant to us but not to users that don't show up in visible logs / output.

ekzhang:
Oof this is old

PR #1535, opened 2024-03-15
Refactor traceback handling into the library and add a test

mwaskom:
I felt bad that this logic had gotten fairly complicated without being subjected to at least a simple unit test.


mwaskom:
What was that I was saying about dropping 3.8 😠

ekzhang (approved):
Nice!

PR #1537, opened 2024-03-18
Use sandbox to implement shell

brianshih1:
This PR adds the ability to stream data out of a `Sandbox`'s `LogReader`. It also reimplements `modal shell` with `Sandbox API` (`stdin` and `async for`).

For example, you can now do:
```
import modal

stub = modal.Stub("examples")

@stub.local_entrypoint()
async def main():
    sandbox = stub.spawn_sandbox(
        "bash",
        "-c",
        "while true; do echo foo; sleep 1; done",
    )

    async for message in sandbox.stdout:
        print(f"Message: {message}")
```

The `async for` API is the same as `asyncio::StreamReader`:
```
process = await asyncio.create_subprocess_exec(
    'bash', '-c', 'echo foo', 
    stdout=asyncio.subprocess.PIPE
)
async for line in process.stdout:
    print(f'Output: {line.decode()}')
```

This PR does change the semantics of `read` for our Sandbox's `LogReader` to be more aligned with `asyncio`'s [StreamReader::Read](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.read) API. Before this PR, calling `sb.stdout.read()` would return the full stream content both times. Now, the second `read` would return an empty string.

Similarly, if you do an `async for` after a `read`, no messages would be streamed, as the reader has received an `EOF` already.

### DIFF
--- a/modal/_sandbox_shell.py
+++ b/modal/_sandbox_shell.py
@@ -36,7 +36,7 @@ async def _stream_logs_to_stdout(sandbox: _Sandbox, on_connect: asyncio.Event) -
     # slave. The PTY shell will then relay data from PTY master to stdout.
     # Therefore, we only need to stream from/to stdout here.
     async for message in sandbox.stdout:
-        await write_to_fd(1, str.encode(message))
+        await write_to_fd(1, message.encode("utf-8"))
 
         if not connected:
             connected = True
--- a/modal/_sandbox_shell.py
+++ b/modal/_sandbox_shell.py
@@ -1,0 +1,49 @@
+# Copyright Modal Labs 2024
+import asyncio
+
+from ._utils.shell_utils import connect_to_terminal, write_to_fd
+from .sandbox import _Sandbox
+
+
+async def connect_to_sandbox(sandbox: _Sandbox):
+    """
+    Connects the current terminal to the Sandbox process.
+    """
+
+    async def _handle_input(data: bytes, _):
+        sandbox.stdin.write(data)
+        await sandbox.stdin.drain.aio()  # type: ignore
+
+    async def _stream_to_stdout(on_connect: asyncio.Event) -> int:
+        return await _stream_logs_to_stdout(sandbox, on_connect)
+
+    await connect_to_terminal(_handle_input, _stream_to_stdout, pty=True)
+
+
+async def _stream_logs_to_stdout(sandbox: _Sandbox, on_connect: asyncio.Event) -> int:
+    """
+    Streams sandbox output logs to the current terminal's stdout.
+
+    The on_connect event will be set when the client connects to the running process,
+    and the event loop will be released.
+    """
+
+    # we are connected if we received at least one message from the server
+    # (the server will send an empty message when the process spawns)
+    connected = False
+
+    # Since the sandbox process will run in a PTY, stderr will go to the PTY
+    # slave. The PTY shell will then relay data from PTY master to stdout.
+    # Therefore, we only need to stream from/to stdout here.
+    async for message in sandbox.stdout:
+        await write_to_fd(1, str.encode(message))
+
+        if not connected:
+            connected = True
+            on_connect.set()
+            # give up the event loop
+            await asyncio.sleep(0)
+
+    # Right now we don't propagate the exit_status to the TaskLogs, so setting
+    # exit status to 0.
+    return 0
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -10,8 +10,9 @@ from rich.console import Console
 
 from modal_proto import api_pb2
 
-from ._container_exec import container_exec
 from ._output import OutputManager, get_app_logs_loop, step_completed, step_progress
+from ._pty import get_pty_info
+from ._sandbox_shell import connect_to_sandbox
 from ._utils.app_utils import is_valid_app_name
 from ._utils.async_utils import TaskContext, synchronize_api
 from ._utils.grpc_utils import retry_transient_errors
@@ -313,13 +314,12 @@ async def _interactive_shell(_stub: _Stub, cmd: List[str], environment_name: str
         loading_status = console.status("Starting container...")
         loading_status.start()
 
-        sb = await _stub.spawn_sandbox("sleep", "360000", **kwargs)
+        sb = await _stub.spawn_sandbox("bash", pty_info=get_pty_info(shell=True), **kwargs)
 
         for _ in range(40):
             await asyncio.sleep(0.5)
             resp = await sb._client.stub.SandboxGetTaskId(api_pb2.SandboxGetTaskIdRequest(sandbox_id=sb._object_id))
             if resp.task_id != "":
-                task_id = resp.task_id
                 break
             # else: sandbox hasn't been assigned a task yet
         else:
@@ -327,7 +327,7 @@ async def _interactive_shell(_stub: _Stub, cmd: List[str], environment_name: str
             raise InteractiveTimeoutError("Timed out while waiting for sandbox to start")
 
         loading_status.stop()
-        await container_exec(task_id, cmd, pty=True, client=client, terminate_container_on_exit=True)
+        await connect_to_sandbox(sb)
 
 
 run_stub = synchronize_api(_run_stub)
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -314,8 +314,8 @@ async def _interactive_shell(_stub: _Stub, cmd: List[str], environment_name: str
         loading_status = console.status("Starting container...")
         loading_status.start()
 
-        sb = await _stub.spawn_sandbox("bash", pty_info=get_pty_info(shell=True), **kwargs)
-
+        sandbox_cmds = cmd if len(cmd) > 0 else ["/bin/bash"]
+        sb = await _stub.spawn_sandbox(*sandbox_cmds, pty_info=get_pty_info(shell=True), **kwargs)
         for _ in range(40):
             await asyncio.sleep(0.5)
             resp = await sb._client.stub.SandboxGetTaskId(api_pb2.SandboxGetTaskIdRequest(sandbox_id=sb._object_id))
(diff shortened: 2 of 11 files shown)


modal/_sandbox_shell.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if not connected:
+            connected = True
+            on_connect.set()
+            # give up the event loop
+            await asyncio.sleep(0)
+
+    # Right now we don't propagate the exit_status to the TaskLogs, so setting
+    # exit status to 0.
+    return 0
```
brianshih1, line 49: In `ContainerExec`, the `exit_status` of the process is added to the output messages [here](https://github.com/modal-labs/modal/blob/main/crates/modal-runtime/src/task/container.rs#L679).

However, in the case of `run_task`, the exit_status isn't added to the `TaskLogs`. TODO:

modal/sandbox.py
```diff
@@ -26,17 +27,33 @@
...
         """mdmd:hidden"""
 
         self._file_descriptor = file_descriptor
         self._sandbox_id = sandbox_id
         self._client = client
+        self._stream = None
+        self._last_log_batch_entry_id = ""
+        # Whether the reader received an EOF. Once eof is True, it returns
+        # an empty string for any subsequent reads (including async for)
+        self.eof = False
```
brianshih1, line 52: The semantic for `LogsReader` in this PR is based on `asyncio.subprocess`'s semantic.

If we perform `reader.read()` after an `async for` (or vice versa), the latter will not read anything. This is because the reader has already encountered an `eof`. 

This behavior is shown in the example below:

```

import asyncio

async def read_output(process):
    async for line in process.stdout:
        print(f'Output: {line.decode().strip()}')

async def main():
    process = await asyncio.create_subprocess_exec(
        'bash', '-c', 'echo foo', 
        stdout=asyncio.subprocess.PIPE
        )
    async for line in process.stdout:
        print(f'Output: {line.decode()}')
    await read_output(process)

    first = await process.stdout.read()
    print(f"First: {first}")

    second = await process.stdout.read()
    print(f"Second: {second}")

asyncio.run(main())
```

brianshih1: Before this PR, calling `read` twice would return the contents of the entire stream. Now, the second read would return an empty string.

modal/runner.py
```diff
@@ -313,21 +314,20 @@ async def _interactive_shell(_stub: _Stub, cmd: List[str], environment_name: str
...
             if resp.task_id != "":
-                task_id = resp.task_id
                 break
             # else: sandbox hasn't been assigned a task yet
         else:
             loading_status.stop()
             raise InteractiveTimeoutError("Timed out while waiting for sandbox to start")
 
         loading_status.stop()
-        await container_exec(task_id, cmd, pty=True, client=client, terminate_container_on_exit=True)
```
thundergolfer, line 330: ⚠️ not using `cmd` from the user

modal/sandbox.py
```diff
@@ -48,47 +65,72 @@ async def read(self) -> str:
...
+                        yield message
+                    if log_batch.eof:
+                        self.eof = True
+                        completed = True
+                        yield None
+                        break
             except (GRPCError, StreamTerminatedError) as exc:
-                if isinstance(exc, GRPCError):
-                    if exc.status in RETRYABLE_GRPC_STATUS_CODES:
+                if retries_remaining > 0:
```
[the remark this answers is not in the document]
brianshih1: I realized that methods like `_stream_function_call_data` have additional error handling mechanism. Just added it here to make the stream more fault tolerant. Not for anything I encountered during development.

test/conftest.py
```diff
@@ -796,54 +794,56 @@ async def QueueLen(self, stream):
...
+        if self.is_shell:
+            # sends an empty message to simulate PTY
+            await stream.send_message(
+                api_pb2.TaskLogsBatch(items=[api_pb2.TaskLogs(data="", file_descriptor=request.file_descriptor)])
+            )
+
         if request.file_descriptor == api_pb2.FILE_DESCRIPTOR_STDOUT:
             # Blocking read until EOF is returned.
-            data = self.sandbox.stdout.read()
+            io = self.sandbox.stdout
```
thundergolfer, line 830: nit: will shadow the `io` stdlib library if it's imported. Maybe just `file`, or `f`?

mwaskom: just had a boomer moment where i was about to "well, actually" you on `file` also being a builtin name, but it turns out they changed that in Python 3 🙈

thundergolfer: haha nice. I've written <1000 lines of Python 2 and never really had to mentally transition. TIL https://portingguide.readthedocs.io/en/latest/builtins.html#removed-file

test/cli_test.py
```diff
@@ -348,38 +350,70 @@ def mock_get_pty_info(shell: bool) -> api_pb2.PTYInfo:
...
+    captured_out.clear()
+
+
+@skip_windows("modal shell is not supported on Windows.")
+def test_shell_cmd(servicer, set_env_client, test_dir, mock_shell_pty):
+    stub_file = test_dir / "supports" / "app_run_tests" / "default_stub.py"
+    _, captured_out = mock_shell_pty
+    _run(["shell", "--cmd", "pwd", stub_file.as_posix() + "::foo"])
+    p = subprocess.Popen("pwd", stdout=subprocess.PIPE)
+    expected_output = p.stdout.read()
```
thundergolfer, line 415:
```suggestion
    expected_output = subprocess.run(["pwd"], capture_output=True, check=True).stdout
```

micro-nit: bit of cold golf to save a line, take it or leave it 👍

test/conftest.py
```diff
@@ -795,34 +795,55 @@ async def QueueLen(self, stream):
...
         self.sandbox_defs.append(request.definition)
+
         await stream.send_message(api_pb2.SandboxCreateResponse(sandbox_id="sb-123"))
 
+    def is_shell_cmds(self, cmds):
+        shell_list = ["bash", "sh", "/bin/bash"]
+        if len(cmds) == 1 and cmds[0] in shell_list:
+            return True
+        else:
+            return False
```
thundergolfer, line 816:
```suggestion
        try:
            return cmds[0] in ["bash", "sh", "/bin/bash"]
        except KeyError:
            return False
```

test/conftest.py
```diff
@@ -152,7 +151,8 @@ def __init__(self, blob_host, blobs):
         self.volume_reloads: Dict[str, int] = defaultdict(lambda: 0)
 
         self.sandbox_defs = []
-        self.sandbox: subprocess.Popen = None
+        self.sandbox: asyncio.subprocess.Process = None
+        self.is_shell = False
```
thundergolfer, line 155: `self` is here the Servicer, so this naming suggests that the variable is concerned with whether _the Servicer_ is a shell or not. That obviously makes no sense, so it should be something like `self.sandbox_is_shell` 🙂

test/cli_test.py
```diff
@@ -348,38 +350,70 @@ def mock_get_pty_info(shell: bool) -> api_pb2.PTYInfo:
...
+
+    # first captured message is the empty message the mock server sends
+    assert captured_out[1:] == [(1, b"Hello World\n")]
+    captured_out.clear()
 
     # Function is explicitly specified
     _run(["shell", webhook_stub_file.as_posix() + "::foo"])
-    assert mock_shell_pty == [(1, b"Hello World")]
-    mock_shell_pty.clear()
+    assert captured_out[1:] == [(1, b"Hello World\n")]
```
thundergolfer, line 400: Hmm could we make this clearer. I think this has something to do with the empty initial message for shells, but it's not obvious without context.

modal/sandbox.py
```diff
@@ -48,47 +65,72 @@ async def read(self) -> str:
...
+    def __aiter__(self):
+        """mdmd:hidden"""
+        self._stream = self._get_logs()
+        return self
+
+    async def __anext__(self):
+        """mdmd:hidden"""
+        value = await self._stream.__anext__()
+
+        # The stream yields None if it receives an eof batch.
```
thundergolfer, line 129:
```suggestion
        # The stream yields None if it receives an EOF batch.
```

nit: consistency throughout module

modal/sandbox.py
```diff
@@ -26,17 +27,33 @@
...
 
     def __init__(self, file_descriptor: int, sandbox_id: str, client: _Client) -> None:
         """mdmd:hidden"""
 
         self._file_descriptor = file_descriptor
         self._sandbox_id = sandbox_id
         self._client = client
+        self._stream = None
+        self._last_log_batch_entry_id = ""
+        # Whether the reader received an EOF. Once eof is True, it returns
```
thundergolfer, line 50:
```suggestion
        # Whether the reader received an EOF. Once EOF is True, it returns
```

nit: consistency within module (and indeed within same comment block)

modal/_sandbox_shell.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # we are connected if we received at least one message from the server
+    # (the server will send an empty message when the process spawns)
+    connected = False
+
+    # Since the sandbox process will run in a PTY, stderr will go to the PTY
+    # slave. The PTY shell will then relay data from PTY master to stdout.
+    # Therefore, we only need to stream from/to stdout here.
+    async for message in sandbox.stdout:
+        await write_to_fd(1, str.encode(message))
```
thundergolfer, line 39:
```suggestion
        await write_to_fd(1, message.encode("utf-8"))
```

nit: the above is the more typical encoding expression

thundergolfer: annoying that the logs are `str` and have to be encoded to bytes. 🤷

thundergolfer (approved):
I think it's good to go 🙂 (assuming server is compatible)

Glad we got that testing mess sorted!

test/conftest.py
brianshih1: modified to `return len(cmds) == 1 and cmds[0] in ["bash", "sh", "/bin/bash"]` because we don't want the shell prompt in the case of `.spawn_sandbox("bash", "-c", "..."`

PR #1538, opened 2024-03-18
[MOD-2582] fix multipart upload concurrency

thundergolfer:
## Describe your changes

- MOD-2582

### Problem

At some point after I last looked at `volume put` reliability, something changed to surface this concurrency bug in our multi-part upload code. When I was testing against prod with a 30 GiB file, I got three hash mismatch errors in a row, each caused by corrupted data being uploaded to S3.

There's plenty of details in the referenced Linear ticket, but the crux of this issue is that this piece of code was making an unsynchronized mutation to the shared file position pointer, and it was doing so whenever a part upload finished, not just on error: 

```
    @contextmanager
    def reset_on_error(self):
        try:
            yield
        finally:
            self.reset_state()
```

This caused upload parts to read from the start of the file instead of the intended read position, corrupting the upload. 

### Fix

The chosen fix is to do away with the lock altogether and just have each upload part use it's own file position pointer. For files a part gets its own file descriptor; for in-memory `BinaryIO` objects we create N copies which reference the original using memory view and don't copy the underlying buffer.

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -161,7 +161,7 @@ async def _upload_to_s3_url(
 
 
 async def perform_multipart_upload(
-    data_file: Union[io.BytesIO, io.FileIO],
+    data_file: Union[BinaryIO, io.BytesIO, io.FileIO],
     *,
     content_length: int,
     max_part_size: int,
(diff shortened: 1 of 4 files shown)


freider:
Outch, nice find and fix!

freider (approved):

thundergolfer:
Finally back in business 🙂 
image

PR #1539, opened 2024-03-18
Queue.ephemeral

erikbern:
This lets you create app-less ephemeral queues:

```python
with Queue.ephemeral() as q:
    q.put(123)
```

### DIFF
--- a/modal/queue.py
+++ b/modal/queue.py
@@ -2,10 +2,10 @@
 import queue  # The system library
 import time
 import warnings
-from contextlib import asynccontextmanager
 from typing import Any, AsyncIterator, List, Optional, Type
 
 from grpclib import GRPCError, Status
+from synchronicity.async_wrap import asynccontextmanager
 
 from modal_proto import api_pb2
 
--- a/modal/queue.py
+++ b/modal/queue.py
@@ -2,7 +2,8 @@
 import queue  # The system library
 import time
 import warnings
-from typing import Any, List, Optional
+from contextlib import asynccontextmanager
+from typing import Any, AsyncIterator, List, Optional, Type
 
 from grpclib import GRPCError, Status
 
@@ -10,7 +11,7 @@ from modal_proto import api_pb2
 
 from ._resolver import Resolver
 from ._serialization import deserialize, serialize
-from ._utils.async_utils import synchronize_api
+from ._utils.async_utils import TaskContext, synchronize_api
 from ._utils.grpc_utils import retry_transient_errors
 from .client import _Client
 from .exception import deprecation_error, deprecation_warning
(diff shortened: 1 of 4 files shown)


modal/queue.py
```diff
@@ -67,6 +69,21 @@ def __init__(self):
         obj = _Queue.new()
         self._init_from_other(obj)
 
+    @classmethod
+    @asynccontextmanager
+    async def ephemeral(cls: Type["_Queue"], client: Optional[_Client] = None, environment_name: Optional[str] = None):
```
freider, line 74: I *think* we can get correct type annotations/api autocomplete on the context variable in our type stubs if you use `@synchronicity.async_wrap.asynccontextmanager` instead (and annotate the return value as an `AsyncIterator["Queue"]`)

modal/queue.py
```diff
@@ -67,6 +69,21 @@ def __init__(self):
...
+        if client is None:
+            client = await _Client.from_env()
+        request = api_pb2.QueueGetOrCreateRequest(
+            object_creation_type=api_pb2.OBJECT_CREATION_TYPE_EPHEMERAL,
+            environment_name=_get_environment_name(environment_name),
+        )
+        response = await client.stub.QueueGetOrCreate(request)
+        async with TaskContext() as tc:
+            request = api_pb2.QueueHeartbeatRequest(queue_id=response.queue_id)
+            tc.infinite_loop(lambda: client.stub.QueueHeartbeat(request), sleep=300)
```
freider, line 84: Perhaps put the 300s in a module-level constant + add some simple test?

freider (approved):

erikbern:
Btw was thinking maybe in notebooks and other places it's not convenient to use a context manager. So we could also support something like this

```python
q = Queue.ephemeral()
q.put(123)
```

This would require `Queue.ephemeral` to return a queue and start the heartbeat task.

freider:
It could return a queue with a mixin `__aexit__` so it can be used in both ways :crazyeyes:

Potentially larger issue - we might need a way to prevent the 🦶 🔫  of trying to use this in global scope, right? Otherwise people might declare an ephemeral queue/dict in the global scope and then would end up creating a new one in every container 😬

erikbern:
Yeah I have the same thought about the global scope foot gun. Maybe that's a good reason to enforce a context manager for now. That way it's harder to make the mistake. We can maybe revisit it later since it would be backwards compatible.

mwaskom:
> Btw was thinking maybe in notebooks and other places it's not convenient to use a context manager.

I think I'm still a little confused about where it _is_ convenient to use this as a context manager. Do you define the whole app in the context manager block?

erikbern:
> I think I'm still a little confused about where it _is_ convenient to use this as a context manager. Do you define the whole app in the context manager block?

The point of this is to create objects outside of the app's lifecycle. I honestly think the main use case right now is for integration tests, since you can test object behavior in isolation this way. Another real use case is with sandboxes (we currently suggest [something pretty janky](https://modal.com/docs/guide/sandbox#returning-or-persisting-data) imo)

freider:
> > It could return a queue with a mixin `__aexit__` so it can be used in both ways :crazyeyes:
> 
> There's actually no exit handler rn anyway so it wouldn't be needed. We only delete because of lack of heartbeat. We could maybe add an explicit delete endpoint later though.

I mean for shutting down the heartbeat loop - it's kind of nice to stop it from running somehow

freider:
> 2. Sandboxes – we currently suggest [something pretty janky](https://modal.com/docs/guide/sandbox#returning-or-persisting-data) imo
Outch, I didn't know we had this as an example in our docs:
```
@stub.local_entrypoint()
def main():
    stub.nfs = modal.NetworkFileSystem.new()
    sb = stub.spawn_sandbox(
        "bash",
        "-c",
        "echo foo > /cache/a.txt",
        network_file_systems={"/cache": stub.nfs},
    )
    sb.wait()
    for data in stub.nfs.read_file("/a.txt"):
        print(data)
```

This usage is actually pretty broken at the moment since the AppSetObjects "update" notifier (that's triggered when the app starts up) will often garbage collect the NetworkFileSystem before the sandbox container starts up! I'm quickfixing the docs now...

erikbern:
> I mean for shutting down the heartbeat loop - it's kind of nice to stop it from running somehow

Yeah that's true – the exit handler will cause this! I wasn't thinking

PR #1545, opened 2024-03-18
Add Dict.ephemeral

erikbern:
Almost identical to #1539

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -1,17 +1,19 @@
 # Copyright Modal Labs 2022
-from typing import Any, Optional
+from typing import Any, AsyncIterator, Optional, Type
+
+from synchronicity.async_wrap import asynccontextmanager
 
 from modal_proto import api_pb2
 
 from ._resolver import Resolver
 from ._serialization import deserialize, serialize
 from ._types import typechecked
-from ._utils.async_utils import synchronize_api
+from ._utils.async_utils import TaskContext, synchronize_api
 from ._utils.grpc_utils import retry_transient_errors
 from .client import _Client
 from .config import logger
 from .exception import deprecation_error, deprecation_warning
-from .object import _get_environment_name, _Object, live_method
+from .object import EPHEMERAL_OBJECT_HEARTBEAT_SLEEP, _get_environment_name, _Object, live_method
 
 
 def _serialize_dict(data):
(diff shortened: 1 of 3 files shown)


modal/dict.py
```diff
@@ -71,8 +73,26 @@ async def _load(self: _Dict, resolver: Resolver, existing_object_id: Optional[st
     def __init__(self, data={}):
         """mdmd:hidden"""
         deprecation_error((2023, 6, 27), "`Dict({...})` is deprecated. Please use `Dict.new({...})` instead.")
-        obj = _Dict.new(data)
-        self._init_from_other(obj)
+
+    @classmethod
+    @asynccontextmanager
+    async def ephemeral(
```
mwaskom, line 79: I know we just got _rid_ of the intermediate `StatefulObject` class but the amount of copy-paste we'll need to do to support all of these is still a little unfortunate. Not sure if it's necessarily better but maybe we could have like an `EphemeralMixin` to handle that?

test/dict_test.py
```diff
@@ -26,3 +27,13 @@ def test_dict_app(servicer, client):
...
+
+
+def test_dict_ephemeral(servicer, client):
+    assert servicer.n_dict_heartbeats == 0
+    with Dict.ephemeral(client=client, _heartbeat_sleep=1) as d:
+        d["foo"] = 42
+        assert d.len() == 1
+        assert d["foo"] == 42
+        time.sleep(1.5)  # Make time for 2 heartbeats
+    assert servicer.n_dict_heartbeats == 2
```
mwaskom, line 39: Is there a good way to assert that the dict actually dies as expected after the context manager exit?

mwaskom (approved):
Looks good aside from comment about code duplication.

test/dict_test.py
erikbern: We don't actually delete the dict on exit, we just rely on a heartbeat timeout after 3600 server-side for that.

It seems nice to keep the objects around in the system for some brief period in case people want to inspect them.

I thought about adding maybe a `DictDelete` method and make it do nothing for now. No strong feelings about it.

modal/dict.py
erikbern: I prefer code duplication in this case since in theory these constructor might have different constructor arguments. They also use different protobuf classes and grpc endpoints etc, so it's not trivial to use a mixin.

There's separately a bit too much code duplication elsewhere on the objects though, but imo the `ephemeral` method seems fine to me.

PR #1546, opened 2024-03-18
Fix support for modal types in constructor args

erikbern:
I didn't quite get it working in #1543 – this properly fixes it.

I don't love this, but let's revisit later.


erikbern:
@prbot approve

PR #1549, opened 2024-03-19
Grpc testing responders

freider:
Needed for some tests I'm working on

Makes overriding servicer functions a bit more flexible:
* override_default (now set_responder) now allows setting an async responder function with the same signature as the original method, instead of a blocking one
* errors in the mock servicer methods are now caught and logged properly, regardless how/when they are added


freider:
@prbot approve

gimaker (commented):
:+1: :heavy_check_mark:

PR #1550, opened 2024-03-19
Add tests for heartbeat loop

freider:
* Adds test that heartbeat loop doesn't die from grpc errors
* Make heartbeat loop failures less scary by not printing full tracebacks
* Adds test that heartbeat loop doesn't die from other random exceptions (not sure if good, but probably doesn't hurt?)


freider:
@prbot approve

gimaker (commented):
nice :heavy_check_mark:

PR #1552, opened 2024-03-19
Support for ephemeral NFS

erikbern:
```python
import modal
with modal.NetworkFileSystem.ephemeral() as nfs:
    print(nfs.listdir("/"))
```

Added a test for `nfs.listdir` too – we didn't have anything


erikbern:
The high amount of redundancy between this and the other objects is a bit unfortunate. Soon we'll be able to delete a bunch of old code though!

erikbern:
@prbot approve

PR #1555, opened 2024-03-19
Set Env to main for Testing

brianshih1:
## Describe your changes

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -48,6 +48,7 @@ class VolumeFile:
     mode: int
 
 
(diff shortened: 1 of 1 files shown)


mwaskom:
Feels like we shouldn't need to do this?

thundergolfer (approved):

mwaskom:
Long in-person discussion: we _shouldn't_ need to do this, but proper test isolation from the host config (modal.toml and environment variables) is a hard problem to solve, so this is a short-term (🤞) workaround.

PR #1556, opened 2024-03-19
Add docs for Sandbox Stdin StreamWriter

brianshih1:
Add docs for `Sandbox StreamWriter`

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -154,7 +154,7 @@ class _StreamWriter:
 
     def write(self, data: Union[bytes, bytearray, memoryview]):
         """
-        This methods writes the data to an internal writer buffer waiting to be drained.
+        Write data to stream's internal buffer, but does not drain/flush the write.
 
         This method needs to be used along with the `drain()` method which flushes the buffer.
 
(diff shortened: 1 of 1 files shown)


modal/sandbox.py
```diff
@@ -147,11 +147,33 @@ def __init__(self, sandbox_id: str, client: _Client):
...
 
     def get_next_index(self):
+        """mdmd:hidden"""
         index = self._index
         self._index += 1
         return index
 
     def write(self, data: Union[bytes, bytearray, memoryview]):
+        """
+        This methods writes the data to an internal writer buffer waiting to be drained.
```
thundergolfer, line 157:
```suggestion
        Write data to stream's internal buffer, but does not drain/flush the write.
```

minor rewording, but you also don't need to start a docstring with "This methods [sic]"

modal/sandbox.py
```diff
@@ -147,11 +147,33 @@ def __init__(self, sandbox_id: str, client: _Client):
...
+
+        ```python
+        sandbox = stub.spawn_sandbox(
+            "bash",
+            "-c",
+            "while read line; do echo $line; done",
+        )
+        sandbox.stdin.write("foo\n")
+        sandbox.stdin.write("bar\n")
+        sb.stdin.write_eof()
```
thundergolfer, line 171: `sb` should be `sandbox` no? The var name seems to have changed mid-snippet.

thundergolfer (approved):
Nice, thanks. Just needed a bit of proof reading 🙂

PR #1559, opened 2024-03-20
Remove Volume.new from volume_test.py unit tests

erikbern:
Need to do this as prep work for deprecation. There's two remaining places in container_test.py that uses them – it wasn't trivial to remove them, so I'll save that for a separate PR.


mwaskom (approved):

erikbern:
Thanks @mwaskom

PR #1561, opened 2024-03-20
Fix container sigint handling for enter/function/exit

freider:
1. Fixes issue where async functions would spit out long asyncio tracebacks when sigint:ed.
2. Adds tests for sigint handling during enter/function/exit handlers to ensure they are all handled as expected:
    * During enter/function, the task should shut down immediately and gracefully, letting the server/runner handle retries etc. No additional outputs or task results should be emitted (?)
    * During exit, sigint should be ignored, allowing exit handlers to run.

This is achieved by installing a `sigint` handler on the user code event loop that does basically what `asyncio.run()`'s sigint handler does: the first sigint triggers a task cancellation and then raises KeyboardInterrupt when the task has successfully exited. However, we make sure to only install this signal handler on the event loop if the global signal handler hasn't been set to SIGIGN.

Fixes MOD-2588


freider:
@prbot approve

modal/_container_entrypoint.py
```diff
@@ -683,6 +712,8 @@ def worker_thread():
                 try:
                     run_input(*args)
                 except BaseException:
+                    # This should basically never happen, since only KeyboardInterrupt is the only error that can
+                    # bubble out of from handle_input_exception and those wouldn't be raised outside the main thread
                     pass
```
ekzhang, line 717: Thanks!

modal/_container_entrypoint.py
```diff
@@ -60,14 +60,18 @@ class UserException(Exception):
...
 
+    This will run all *user supplied* async code, i.e. async functions, as well as async enter/exit managers
+
     The following signals are handled while a coroutine is running on the event loop until
     completion (and then handlers are deregistered):
 
     - `SIGUSR1`: converted to an async task cancellation. Note that this only affects the event
       loop, and the signal handler defined here doesn't run for sync functions.
+    - `SIGINT`: Unless the global signal handler has been set to SIGIGN, the loop's signal handler
+        is set to cancel the current task and raise KeyboardInterrupt to the caller.
```
ekzhang, line 74: I think SIG_IGN not SIGIGN right?

ekzhang (commented):
It looks good! Sorry for not getting back to it.

modal/_container_entrypoint.py
```diff
@@ -701,7 +732,10 @@ def worker_thread():
         for input_id, function_call_id, args, kwargs in function_io_manager.run_inputs_outputs(
             imp_fun.input_concurrency
         ):
-            run_input(input_id, function_call_id, args, kwargs)
+            try:
+                run_input(input_id, function_call_id, args, kwargs)
+            except:
+                raise
```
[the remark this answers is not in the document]
freider: Haha, it does basically nothing, I must have accidentally left it in here from some previous debugging session 🤔  Will remove

PR #1562, opened 2024-03-20
POC for looking up the image_builder_version in the client

mwaskom:
Sharing for discussion


mwaskom:
Superseded for an approach based on #1584

PR #1563, opened 2024-03-20
Remove inert typechecked decorator

mwaskom:
A lot of our functions are decorated with `@typechecked` but, as of https://github.com/modal-labs/modal-client/pull/447, that's misleading.

AFAICT we're not likely to take on this TODO in the near future. Automatic type error messages would be cool, but we're just at Python status quo by not having them, and lots of people develop in IDEs with nice integrated static type checking these days. And the story with PEP563 / PEP649 still seems a little bit unclear; maybe there will be better tooling introduced in the future?

If a good solution emerges, we can always put these back; but for now they seem like cruft.


freider:
Makes sense to me!

freider (approved):

PR #1565, opened 2024-03-20
Remove almost all NetworkFileSystem.new

erikbern:
There's an annoying thing triggered by a sandbox test that's holding up the deprecation. Removing all other uses in the meantime.


erikbern:
@prbot approve

PR #1567, opened 2024-03-20
Add changelog for Sandbox stdin/async for

brianshih1:
Add changelog for Sandbox `stdin/async for`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,8 +45,8 @@ def my_fn():
     sandbox.stdin.write("foo\n")
     sandbox.stdin.write("bar\n")
     sandbox.stdin.write_eof()
-    sb.stdin.drain()
-    sb.wait()
+    sandbox.stdin.drain()
+    sandbox.wait()
 ```
 
 ### 0.61.53 (2024-03-15)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,11 +10,44 @@ We appreciate your patience while we speedily work towards a stable release of t
 
 <!-- NEW CONTENT GENERATED BELOW. PLEASE PRESERVE THIS COMMENT. -->
 
+### 0.61.75 (2024-03-19)
```
mwaskom, line 13: I think this should be v0.61.76; looks like .75 was the test fix

brianshih1: oof sorry, good catch

thundergolfer (approved):

CHANGELOG.md
```diff
@@ -10,11 +10,44 @@ We appreciate your patience while we speedily work towards a stable release of t
...
 
+```python
+@stub.function()
+def my_fn():
+    sandbox = stub.spawn_sandbox(
+            "bash",
+            "-c",
+            "while read line; do echo $line; done",
+        )
+    sandbox.stdin.write("foo\\n")
```
thundergolfer, line 45: These need to be bytes I think

PR #1570, opened 2024-03-20
Update usage of ephemeral queues in modal launch apps

mwaskom:
These were using the deprecated `Queue.new` method. Good opportunity to figure out the right way to use `Queue.ephemeral`!


erikbern (approved):
Nice catch, thanks for fixing!

PR #1578, opened 2024-03-21
Dynamically identify modules for typestub gen, address new issues

mwaskom:
We were failing to generate a type stubs for a few modules that use synchronicity because we had a hardcoded list. I addressed a TODO to make generate the list of modules dynamically. This caused a few new typing issues to pop up; I think we might encounter a few more in CI too.


mwaskom:
@prbot approve

freider (commented):
lgtm!

PR #1581, opened 2024-03-22
Fix for unindexed inclusion of already-hydrated objects in apps

freider:
Prevously they would not get included in unindexed_object_ids and cause the container entrypoint to crash when trying to look up their metadata from the app object metadata dict


freider:
Function deps that were already hydrated wouldn't previously be included in unindexed_object_ids, so unless they were in indexed_object_ids by being assigned to the stub, they wouldn't get included at app creation.

PR #1584, opened 2024-03-22
Defer dockerfile command resolution until Image is loaded

mwaskom:
This pushes the logic of building dockerfile syntax and defining local context files to the `_load` function used to define the `Image` object. The practical upsides are

1. will make it cleaner to implement versioned builds
2. we can remove some short-circuiting guards that we had in place to prevent image methods from trying to access local data when they run remotely

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -5,6 +5,7 @@ import shlex
 import sys
 import typing
 import warnings
+from dataclasses import dataclass, field
 from inspect import isfunction
 from pathlib import Path, PurePosixPath
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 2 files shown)


modal/image.py
```diff
@@ -501,26 +533,28 @@ def pip_install_private_repos(
             )
 
         secret_names = ",".join([s.app_name if hasattr(s, "app_name") else str(s) for s in secrets])  # type: ignore
-        dockerfile_commands = ["FROM base"]
-        if any(r.startswith("github") for r in repositories):
-            dockerfile_commands.append(
-                f"RUN bash -c \"[[ -v GITHUB_TOKEN ]] || (echo 'GITHUB_TOKEN env var not set by provided modal.Secret(s): {secret_names}' && exit 1)\"",
```
erikbern, line 507: this is existing code so not related to this PR, but just noticed it and... it's pretty ugly?

erikbern (approved):
Big PR but I trust the tests on this one.

Would be great to run this against integration tests as well if you didn't already

PR #1586, opened 2024-03-23
Deprecate stub getitem/setitem

erikbern:
Ahead of deprecating `stub.x`, let's deprecate `stub["x"]`

Seems pretty rare

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -236,11 +236,25 @@ class _Stub:
         self._indexed_objects[tag] = obj
 
     def __getitem__(self, tag: str):
-        deprecation_warning((2024, 3, 23), "stub[...] is deprecated!")
+        """Deprecated!
+
+        The only use case for `stub[...]` assignments is in conjunction with `.new()`, which is
+        in itself deprecated. If you are constructing objects with `.from_name(...)`, there is no
+        need to assign those objects to the stub. Example:
+
+        ```python
+        d = modal.Dict.from_name("my-dict", create_if_missing=True)
+
+        @stub.function()
+        def f(x, y):
+            d[x] = y  # Refer to d in global scope
+        ```
+        """
+        deprecation_warning((2024, 3, 25), _Stub.__getitem__.__doc__)
         return self._indexed_objects[tag]
 
     def __setitem__(self, tag: str, obj: _Object):
-        deprecation_warning((2024, 3, 23), "stub[...] is deprecated!")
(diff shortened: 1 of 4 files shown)


test/stub_test.py
```diff
@@ -25,11 +25,11 @@ async def test_kwargs(servicer, client):
...
-        # TODO: interface to get type safe objects from live apps
-        await stub["d"].put.aio("foo", "bar")  # type: ignore
-        await stub["q"].put.aio("baz")  # type: ignore
-        assert await stub["d"].get.aio("foo") == "bar"  # type: ignore
-        assert await stub["q"].get.aio() == "baz"  # type: ignore
+        with pytest.warns(DeprecationError):
+            await stub["d"].put.aio("foo", "bar")  # type: ignore
+            await stub["q"].put.aio("baz")  # type: ignore
+            assert await stub["d"].get.aio("foo") == "bar"  # type: ignore
+            assert await stub["q"].get.aio() == "baz"  # type: ignore
```
mwaskom, line 32: Nit: this test will pass if only one of these warns

modal/stub.py
```diff
@@ -236,10 +236,11 @@ def _add_object(self, tag, obj):
         self._indexed_objects[tag] = obj
 
     def __getitem__(self, tag: str):
-        # Deprecated? Note: this is currently the only way to refer to lifecycled methods on the stub, since they have . in the tag
+        deprecation_warning((2024, 3, 23), "stub[...] is deprecated!")
```
mwaskom, line 239: Nit: I had trouble parsing this for a second, maybe elaborating a bit would help, e.g.

```
"Stub subscription (`stub[...]`) is deprecated"
```

(though "subscription" is a fairly arcane term, the slightly less accurate "indexing" may be more familiar).

Also do we want to push people towards setattr/getattr or no because we also plan on deprecating those? I don't think deprecation churn is great UX but neither is a deprecation message that doesn't suggest a remediation.

mwaskom (approved):
LGTM just a question about what the message should say

modal/stub.py
erikbern: There really isn't much of a reason to use either setattr or setitem at this point – people should be able to remove it. The only exception would be `.new()` methods which are themselves deprecated.

I can make that clear in the error message!

mwaskom: Yeah — saying something like "It is no longer necessary as objects can be defined in global scope" seems helpful!

PR #1593, opened 2024-03-25
Add unit test to catch changes to image definition logic

mwaskom:
In prep for some refactoring of the image building logic and then expansion to support versioned builds, I am adding a unit test to assert that image definitions (i.e. the two attributes that we currently consider to be "versioned" — dockerfile commands and context file path / contents) remain unchanged.

To avoid storing a lot of data, I am doing this in a harsh way: hashing the the repr of this data and storing the hash in the unit test itself. Maybe this is a bad idea?


mwaskom:
@prbot approve; discussed on slack that this makes sense

erikbern (approved):
👍

mwaskom:
Hah well this was a useful exercise if for no other reason than surfacing a bug in `Image.micromamba`

PR #1598, opened 2024-03-25
Don't import support CLI imports of the form `stub.func`

erikbern:
As a step towards deprecating stub assignments, I ran into a bit of a weird case. When deprecating `stub.foo` getattrs, we're indirectly breaking CLI invocations of that form. E.g. `modal run my_file.py::stub.foo`

I think that's fine given that `modal run my_file.py::foo` still works?

We could in theory still support it, but I would rather maintain consistency between the CLI and Python. And in all normal cases, the function is always defined in the global scope so you should be able to import it using the function name instead (not using the stub as a qualifier).

The only case that I can think of where this breaks is if you have two different stub in the same file, each adding a function _with the same name_. This seems like a weird case?

### DIFF
--- a/test/cli_imports_test.py
+++ b/test/cli_imports_test.py
@@ -8,7 +8,6 @@ from modal.cli.import_refs import (
     import_file_or_module,
     parse_import_ref,
 )
-from modal.functions import _Function
 from modal.stub import _LocalEntrypoint, _Stub
 
 # Some helper vars for import_stub tests:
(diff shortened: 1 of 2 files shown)


freider:
I see, I didn't realize we were removing `getattr` for stub functions as well. I like it, but I have a vague feeling of a few cases where it's been useful. Hopefully we can solve it in some other way if the need arises again. For the CLI I think it's definitely acceptable to only allow references to module-level objects.

One other case where it could be a bit annoying is if you are programmatically adding functions in a loop or something so you don't have a "nice" way of adding the resulting Functions to the module scope, e.g.
```
for i in range(n):
    stub.function(name=f"func_{i}")(my_func)
```

But I suppose in this edge case you can still work around it by assigning to `globals()`, e.g. `globals()[name] = ...` 🤮

freider (approved):

freider:
One other advantage of removing this is that we can be more certain that any `.` delimiters in the pathspec of the cli commands are actually referring to methods of classes, which might clean up some code?

erikbern:
Great that you're ok with this change @freider!! 🙏

PR #1600, opened 2024-03-25
Remove stub assignments: image_test.py

erikbern:
Slightly larger change, broke up a test into three smaller tests since they were basically independent


erikbern:
@prbot approve

pure test change

mwaskom (commented):
:+1:

PR #1602, opened 2024-03-26
Deprecate stub assignments

erikbern:
This finally deprecates `getattr` and `setattr` on stubs.

~This PR includes #1598 since it's needed for tests to pass~ EDIT: that one is now merged, and this one is rebased on main

### DIFF
--- a/modal/stub.py
+++ b/modal/stub.py
@@ -236,9 +236,9 @@ class _Stub:
         self._indexed_objects[tag] = obj
 
     def __getitem__(self, tag: str):
-        """Deprecated!
+        """Stub assignments of the form `stub.x` or `stub["x"]` are deprecated!
 
-        The only use case for `stub[...]` assignments is in conjunction with `.new()`, which is
+        The only use cases for these assignments is in conjunction with `.new()`, which is now
         in itself deprecated. If you are constructing objects with `.from_name(...)`, there is no
         need to assign those objects to the stub. Example:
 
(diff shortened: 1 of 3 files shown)


modal/stub.py
```diff
@@ -265,15 +265,20 @@ def __getattr__(self, tag: str) -> _Object:
             # Hacky way to avoid certain issues, e.g. pickle will try to look this up
             raise AttributeError(f"Stub has no member {tag}")
         # Return a reference to an object that will be created in the future
-        return self._indexed_objects[tag]
+        obj: _Object = self._indexed_objects[tag]
+        deprecation_warning((2024, 3, 25), _Stub.__getitem__.__doc__)
```
mwaskom, line 269: Do we want to gate the deprecation warning on `obj` being an `_Object`?

The main reason I ask is that I've tripped our deprecation warnings in various annoying ways, especially in tests or support code, e.g. calling `hasattr()` on an object in some duck typing scenario.

But also in Python it's not generally illegal to set / get random attributes on objects. I think it makes perfect sense to deprecate `__getitem__` / `__setitem__`, since those are not generally available on objects. And it makes sense to deprecate any side effects that `__setattr__` has that we don't want uses to rely on. But I'm not sure we need to be noisy about people accessing random attributes from a stub.

erikbern: Doesn't `__getattr__` get called only if the item isn't found in `__dict__`? So I think you can still set/get random attributes if you want to.

I'm hoping to get rid of the entire `__getattr__` method eventually, so once it's gone we'll end up with the behavior your mentioning – it will be possible to set/get arbitrary attributes. But it's going to take probably 3-6 months of deprecations to get there.

mwaskom: > Doesn't __getattr__ get called only if the item isn't found in __dict__? So I think you can still set/get random attributes if you want to.

True, although the specific situation I am thinking of is a case where I was testing for the presence of an attribute (a method) the objects might not have, so it did end up taking this route.

mwaskom (approved):
Aside from question about whether we can avoid false positive warnings, LGTM

PR #1603, opened 2024-03-26
Fixes automount incompatibility with symlinked python installations

freider:
Would previously automount all stdlibs if the used python install root was referenced via a symlink when setting up a virtualenv or similar. For example, this could happen when using [mise](https://mise.jdx.dev/) to set up virtualenvs pointing to a minor python version which was itself a symlink to a micro version (e.g. 3.11 -> 3.11.4)

I have a hunch we used to support both resolved and unresolved SYS_PREFIXES, so I'm adding a test to ensure we don't regress in this way again

MOD-2645

### DIFF
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -266,7 +266,8 @@ class FunctionInfo:
                 continue
 
             for local_path, remote_path in mount_paths:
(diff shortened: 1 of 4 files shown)


mwaskom (approved):
Brilliant!

PR #1605, opened 2024-03-26
Bump client to 0.62

erikbern:
Since we deprecated `.new` and stub assignments, let's bump the client version. That way 0.62+ is a good proxy for how fast we can turn the deprecation warnings into errors (and eventually remove the old code)

## Changelog

- The `.new()` method has now been deprecated on all Modal objects. It should typically be replaced with `.from_name(...)` in Modal app code, or `.ephemeral()` in scripts that use Modal
- Assignment of Modal objects to a `Stub` via subscription (`stub["object"]`) or attribute (`stub.object`) syntax is now deprecated. This syntax was only necessary when using `.new()`.

### DIFF
--- a/modal_version/__init__.py
+++ b/modal_version/__init__.py
@@ -7,7 +7,7 @@ from ._version_generated import build_number
 major_number = 0
 
 # Bump this manually on breaking changes, then reset the number in _version_generated.py
-minor_number = 61
+minor_number = 62
 
 # Right now, automatically increment the patch number in CI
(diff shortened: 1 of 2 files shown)


erikbern:
> Nice — can we get a short changelog entry?

How do I add a changelog entry?

mwaskom:
I added a draft but feel free to edit.

mwaskom:
We auto populate the changelog from the PR description.

mwaskom (approved):

erikbern:
Thanks, reworded the draft slightly!

PR #1610, opened 2024-03-27
Update Volume docs to match actual behaviour

gimaker:


mwaskom:
Thanks for catching this!

PR #1611, opened 2024-03-27
Simplify stub matching

erikbern:
Was looking at the app/stub code and noticed a quick thing to clean up. The flow used to be very confusing and we were doing the stub matching in basically two places:

1. In the `Stub` constructor in global scope (by triggering `_container_app._associate_stub_container`)
2. In `import_function` in `_container_entrypoint.py`

The second one is a bit more safe since you avoid any partially initialized state. This PR basically removes the first one. This also lets us avoid a warning in the cases where the stub can be uniquely inferred from the function (see the test I updated).

The reason the code used to be more complex is we had to support hydrating images in global for `is_inside()`. Since this is deprecated, we can simplify the code.

This PR would have been net red if it wasn't for a bunch of comments I added

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -827,8 +827,28 @@ def import_function(
     function_io_manager,
     client: Client,
 ) -> ImportedFunction:
-    # This is not in function_io_manager, so that any global scope code that runs during import
-    # runs on the main thread.
+    """Imports a function dynamically, and locates the stub.
+
+    This is somewhat complex because we're dealing with 3 quite different type of functions:
+    1. Functions defined in global scope and decorated in global scope (Function objects)
+    2. Functions defined in global scope but decorated elsewhere (these will be raw callables)
+    3. Serialized functions
+
+    In addition, we also need to handle
+    * Normal functions
+    * Methods on classes (in which case we need to instantiate the object)
+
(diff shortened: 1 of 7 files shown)


freider:
Ugh yeah this is a bit complex, it's great to cut down a bit on the complexity. I think one case where this would not work is for serialized functions, since those won't have an active_stub on the `ImportedFunction` instance. But maybe that isn't a problem in practice 🤔
Stub-assigned objects might not have worked well for serialized functions even before, and without stub assignments I think most objects should work anyways (as soon as we add lazy hydration)

freider (approved):

erikbern:
> Ugh yeah this is a bit complex, it's great to cut down a bit on the complexity. I think one case where this would not work is for serialized functions, since those won't have an active_stub on the `ImportedFunction` instance. But maybe that isn't a problem in practice 🤔

It will still work, I'm pretty sure! With this new flow we basically try two things:
1. Import the function, grab the stub from it
2. Get the stub from its name (if 1 doesn't work)

I can maybe document this a bit better in the code, let me add a few more comments

freider (approved):
Great! This refreshed my memory of this a lot more efficiently than reading the code 👍

erikbern:
Thanks @freider

PR #1613, opened 2024-03-27
Avoid errors when Image.run_commands commands are empty strings

mwaskom:
Helps avoids a confusing error that a user ran into (empty or whitespace strings produce Dockerfile syntax that looks like `RUN  `).

Arguably we could raise here, but given that we already "gracefully" handle an empty list, dropping empty commands feels more consistent.

- Resolves MOD-2626

## Changelog

- `Image.run_commands` will now skip empty strings or strings containing only whitespace characters, which otherwise produce malformed Dockerfile syntax.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -773,11 +773,11 @@ class _Image(_Object, type_prefix="im"):
     ) -> "_Image":
         """Extend an image with a list of shell commands to run."""
         cmds = _flatten_str_args("run_commands", "commands", commands)
-        if not cmds:
+        if not cmds or not any(cmds):
             return self
 
         def build_dockerfile() -> DockerfileSpec:
-            return DockerfileSpec(commands=["FROM base"] + [f"RUN {cmd}" for cmd in cmds], context_files={})
+            return DockerfileSpec(commands=["FROM base"] + [f"RUN {cmd}" for cmd in cmds if cmd], context_files={})
 
         return _Image._from_args(
             base_images={"base": self},
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -233,6 +233,35 @@ def test_dockerfile_image(servicer, client):
         assert any("RUN pip install numpy" in cmd for cmd in layers[1].dockerfile_commands)
 
 
+def test_run_commands(servicer, client):
+    base = Image.debian_slim()
+
+    command = "echo 'Hello Modal'"
+    stub = Stub(image=base.run_commands(command))
+    with stub.run(client=client):
+        layers = get_image_layers(stub.image.object_id, servicer)
+        assert layers[0].dockerfile_commands[1] == f"RUN {command}"
+
+    commands = ["echo 'Hello world'", "touch agi.yaml"]
+    stub = Stub(image=base.run_commands(commands))
+    with stub.run(client=client):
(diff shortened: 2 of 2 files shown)


ekzhang:
Hmm, an empty list to me makes sense because it is the natural progression of

```python
image.run_commands("echo 1", "echo 2", "echo 3") # 1, 2, 3
image.run_commands("echo 1", "echo 2") # 1, 2
image.run_commands("echo 1") # 1
image.run_commands() #
```

But not as sure about empty string

```python
>>> subprocess.run(["echo", "0"])
0
CompletedProcess(args=['echo', '0'], returncode=0)
>>> subprocess.run([])
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.11/subprocess.py", line 1819, in _execute_child
    executable = args[0]
                 ~~~~^^^
IndexError: list index out of range
```

ekzhang:
> or strings containing only whitespace characters

Is this a minor typo in changelog?

ekzhang:
`Image.pip_install("")` also fails with an empty string:

```
=> Step 1: RUN python -m pip install '' 
ERROR: Invalid requirement: ''
Terminating task due to error: failed to run builder command "python -m pip install '' "

Caused by:
    container exit status: 1
Runner failed with exception: task exited with failure, status = exit status: 1
Stopping app - uncaught exception raised locally: RemoteError('Image build for im-q5JWJjKi7EuNMgQHfHigKC failed with the exception:\ntask exited with failure, status = exit status: 1').
```

mwaskom:
>  Is this a minor typo in changelog?

author-blindness, but i am not seeing the typo?

mwaskom:
> Image.pip_install("") also fails with an empty string:

True — good point — although this feels more obvious to debug then the one the user reported.

mwaskom:
> But not as sure about empty string

Is this to advocate for raising instead of silently dropping empty strings?

ekzhang:
> Is this to advocate for raising instead of silently dropping empty strings?

Yes I think it might be more expected to raise?

mwaskom:
I’m not really following the argument for why empty lists should be considered valid but empty strings not.

ekzhang:
I also don't really understand why empty strings are an expected API

mwaskom:
The argument is pretty simple: the parameter is typed as Union[list, str], so if we allow empty list, allowing empty string makes sense to. As noted in the PR description, raising an error could make sense too. But this is a very minor change, and either approach would be a marginal improvement.

ekzhang:
> The argument is pretty simple: the parameter is typed as Union[list, str], so if we allow empty list, allowing empty string makes sense to.

I must be crazy because this doesn't make any sense to me, can you please explain more as I have 🤔

ekzhang:
I also wrote the API to take a union only as backwards compatibility with previous list APIs. But I don't get why an empty list and an empty string inside a list are the same regardless??

PR #1614, opened 2024-03-27
Add image_builder_version to ClientHelloResponse

mwaskom:
Adds `image_builder_version` as a field in the `ClientHelloResponse` message. ~Also removes the (forever unused) `ImageBuilderVersion*` message and rpc protos.~

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -547,6 +547,7 @@ message ClientCreateResponse {
 
 message ClientHelloResponse {
   string warning = 1;
+  string image_builder_version = 2;
 }
 
 message ClientHeartbeatRequest {


modal_proto/api.proto
```diff
@@ -1227,10 +1228,6 @@ message Image {
   BuildFunction build_function = 21;
 }
 
-message ImageBuilderVersionLookupResponse {
```
aksh-at, line 1230: If you're removing this, you'll have to merge it in after the RPC handler is removed from the server

aksh-at (approved):
🚢

mwaskom:
Thanks for catching, rebased to just add the `ClientHelloResponse` member for now

PR #1618, opened 2024-03-28
Misc cleanup of `ContainerApp` and `LocalApp`

erikbern:
I'm slowly trying to get rid of those two classes and move the functionality into `Stub`. Getting rid of some stuff that definitely isn't used in this PR. #1611 created a lot of the opportunity for cleanup.

Eventually, this will let us rename `Stub` to `App`


erikbern:
@prbot approve

PR #1623, opened 2024-03-28
Require ClientHello to have run before creating objects

mwaskom:
Moderate refactoring / renaming of methods on `Client` with the main functional change that we now require the `ClientHello` RPC to have been run before creating any objects:

- I renamed `Client._verify` to `Client._init` to make it more clear that it's an initialization step (we plan to use this method to retrieve the image builder version — and perhaps other information — that will be needed later).
- I added a `Client.authenticated` property, which is set when `Client._init` runs. This is checked by `_LocalApp._create_all_objects`. We plan to remove `_LocalApp`; but I assume we'll just port `_create_all_objects` somewhere else? This may affect the behavior of `Client.from_credentials` (presumably by improving the error message returned for bad credentials?)
- I renamed `Client.unauthenticated_client` to `Client.anonymous`, since the client can be in an unauthenticated or authenticated state depending on whether  `._init` has been called, and the purpose of this method is to create a client without a token (i.e., not associated with any specific workspace)
- I made a few marginal improvements to the code quality e.g. filling in some type annotations and docstrings, making the `classmethod` actually use `cls` (no practical effect now since we don't have `Client` subtypes), etc.

- Resolves MOD-2674

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -2,11 +2,13 @@
 import asyncio
 import platform
 import warnings
-from typing import Awaitable, Callable, Dict, Optional, Tuple
+from typing import AsyncIterator, Awaitable, Callable, Dict, Optional, Tuple
 
+import grpclib.client
 from aiohttp import ClientConnectorError, ClientResponseError
 from google.protobuf import empty_pb2
 from grpclib import GRPCError, Status
+from synchronicity.async_wrap import asynccontextmanager
 
 from modal_proto import api_grpc, api_pb2
 from modal_version import __version__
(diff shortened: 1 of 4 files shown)


mwaskom:
BTW I guess `Client.unauthenticated_client` is technically a public method but we discourage users from touching `Client` directly; not sure if it needs a deprecation?

modal/client.py
```diff
@@ -75,46 +77,44 @@ async def _http_check(url: str, timeout: float) -> str:
...
 
     @property
     def stub(self) -> Optional[api_grpc.ModalClientStub]:
         """mdmd:hidden"""
         return self._stub
 
+    @property
+    def authenticated(self) -> bool:
+        """mdmd:hidden"""
+        return getattr(self, "_authenticated", False)
```
aksh-at, line 112: I think this should work here

```suggestion
        return self._authenticated
```

modal/client.py
```diff
@@ -245,16 +249,30 @@ async def from_env(cls, _override_config=None) -> "_Client":
...
+        try:
+            await client._init()
+        except BaseException:
+            await client._close()
+            raise
         async_utils.on_shutdown(client._close())
         return client
 
+    @classmethod
+    async def verify(cls, server_url: str, credentials: Tuple[str, str]) -> None:
```
aksh-at, line 269: Is this used anywhere?

aksh-at (approved):
nice! Lgtm besides comments

modal/client.py
mwaskom: Yeah, we call it in `modal token set`

PR #1625, opened 2024-03-28
Partitions for modal.Queue

gongy:
## Changelog

- Queue methods `put`, `put_many`, `get`, `get_many` and `len` now support an optional `partition` argument (must be specified as a `kwarg`). When specified, users read and write from new partitions of the queue independently. `partition=None` corresponds to the default partition of the queue.

### DIFF
--- a/modal/queue.py
+++ b/modal/queue.py
@@ -71,8 +71,10 @@ class _Queue(_Object, type_prefix="qu"):
     @staticmethod
     def validate_partition(partition: str) -> None:
         partition_key = partition.encode("utf-8")
+
         if len(partition_key) > 64:
             raise InvalidError("Partition names must be 64 characters or fewer.")
+
         return partition_key
 
     @classmethod
--- a/modal/queue.py
+++ b/modal/queue.py
@@ -14,7 +14,7 @@ from ._serialization import deserialize, serialize
 from ._utils.async_utils import TaskContext, synchronize_api
 from ._utils.grpc_utils import retry_transient_errors
 from .client import _Client
-from .exception import deprecation_warning
+from .exception import InvalidError, deprecation_warning
 from .object import EPHEMERAL_OBJECT_HEARTBEAT_SLEEP, _get_environment_name, _Object, live_method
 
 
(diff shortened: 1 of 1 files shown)


modal/queue.py
```diff
@@ -217,19 +232,21 @@ async def get(self, block: bool = True, timeout: Optional[float] = None) -> Opti
...
 
         if values:
             return values[0]
         else:
             return None
 
     @live_method
-    async def get_many(self, n_values: int, block: bool = True, timeout: Optional[float] = None) -> List[Any]:
+    async def get_many(
+        self, n_values: int, block: bool = True, timeout: Optional[float] = None, partition: Optional[str] = None
```
mwaskom, line 248: Here too I guess

modal/queue.py
```diff
@@ -309,9 +332,12 @@ async def _put_many_nonblocking(self, vs: List[Any]):
             raise queue.Full(exc.message) if exc.status == Status.RESOURCE_EXHAUSTED else exc
 
     @live_method
-    async def len(self) -> int:
-        """Return the number of objects in the queue."""
-        request = api_pb2.QueueLenRequest(queue_id=self.object_id)
+    async def len(self, partition: Optional[str] = None) -> int:
+        """Return the number of objects in the queue partition."""
```
mwaskom, line 336: Haven't looked server-side, but what happens if you have a partitioned queue but don't specify `partition` here? Does it return the full queue size? Or is there some default "anonymous" partition? Or does it raise?

mwaskom (approved):
Implementation looks good, just a couple comments:

- Small API thing, but requiring a keyword arg might make sense
- It's not totally obvious what some operations should do when you don't specify a partition, but have previously put objects into a partition. Is there a "default" anonymous partition? I mention it to (a) make sure you're thought it through and (b) suggest documenting where it makes sense.

Also, would be great if we could exercise this in tests and have a changelog entry, since it's a nice user-facing feature.

modal/queue.py
gongy: Yeah good point, the default partition is the empty key. I disallow specifying an empty partition client side (it's either `None` or a non-zero string).

gongy: Going to add some documentation in a follow up!

PR #1626, opened 2024-03-28
Reduce scope of `_LocalApp` (step 1/n of `Stub` -> `App`)

erikbern:
My plan is roughly:
1. Turn `_LocalApp` and `_ContainerApp` into more data-class like (public properties etc)
2. Move all methods that operate on those objects to other places (runner.py in particular)
3. Turn those objects into actual dataclasses
4. Unify it into something like a `RunningApp` dataclass
5. Put that object on the stub instead
6. Rename "stub" to "app"

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -130,10 +130,6 @@ class _LocalApp:
         req_disconnect = api_pb2.AppStopRequest(app_id=self.app_id, source=api_pb2.APP_STOP_SOURCE_PYTHON_CLIENT)
         await retry_transient_errors(self.client.stub.AppStop, req_disconnect)
 
-    def log_url(self):
-        """URL link to a running app's logs page in the Modal dashboard."""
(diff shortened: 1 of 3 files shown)


modal/app.py
```diff
@@ -71,15 +56,15 @@ async def _create_all_objects(
     ):  # api_pb2.AppState.V
         """Create objects that have been defined but not created on the server."""
         resolver = Resolver(
-            self._client,
```
mwaskom, line 74: I think you'll want to make sure you rebase and update this line: https://github.com/modal-labs/modal-client/blob/main/modal/app.py#L73

mwaskom (approved):
Looks good, though see note about a potential silent merge conflict that could break everything :)

modal/app.py
erikbern: Thanks!

PR #1627, opened 2024-03-28
Make the `modal launch` commands less hacky

ekzhang:
They previously spun up an entire subprocess and did find/replace writing to a tempfile within the code, which caused breakage that @gimaker graciously fixed in #1617.

It also makes `modal launch` commands show up in the dashboard like this:

image

We can avoid spinning up an entire subprocess by creating an environment variable instead. This is not completely foolproof, since environment variables don't appear on the remote runner (which is questionable!), but at least it avoids the issue that we ran into previously.

### DIFF
--- a/modal/cli/launch.py
+++ b/modal/cli/launch.py
@@ -1,10 +1,18 @@
 # Copyright Modal Labs 2023
-import subprocess
-import tempfile
+import asyncio
+import inspect
+import json
+import os
 from pathlib import Path
-from typing import Optional
+from typing import Any, Dict, Optional
 
 from typer import Typer
+
+from modal.exception import _CliUserExecutionError
+
+from ..runner import run_stub
+from ..stub import Stub
+from .import_refs import import_function
 
 launch_cli = Typer(
     name="launch",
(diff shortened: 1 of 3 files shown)


modal/cli/launch.py
```diff
@@ -17,16 +24,25 @@
 )
 
 
-def _launch_program(name: str, args) -> None:
-    contents = (Path(__file__).parent / "programs" / name).read_text()
-    contents = contents.replace("args: Dict[str, Any] = {}", f"args: Any = {repr(args)}")
+def _launch_program(name: str, filename: str, args: Dict[str, Any]) -> None:
+    os.environ["MODAL_LAUNCH_LOCAL_ARGS"] = json.dumps(args)
```
erikbern, line 28: maybe i'm missing something but why can't you just pass this to `func` directly? i.e. invoke `func(args)` instead

erikbern:
Would it be cleaner to build up the stub programmatically instead? I.e. construct an image object and a function object on the fly and trigger a function on it. If what I'm saying makes no sense, no worries.

ekzhang:
> Would it be cleaner to build up the stub programmatically instead? I.e. construct an image object and a function object on the fly and trigger a function on it. If what I'm saying makes no sense, no worries.

Makes sense, I can try it with `serialized=True`!

erikbern:
> > Would it be cleaner to build up the stub programmatically instead? I.e. construct an image object and a function object on the fly and trigger a function on it. If what I'm saying makes no sense, no worries.
> 
> 
> 
> Makes sense, I can try it with `serialized=True`!

You don't have to use a serialize=True btw, it should work without it, I think. I can try later if it's not clear what I mean

ekzhang:
I'm not 100% sure — feel free to edit the PR, or send me an outline for how to implement it and I'll follow that.

ekzhang:
Actually for now, I will just merge this branch since it's an incremental improvement over our current state in two ways:

- Not getting lots of deprecation warnings printed in the console.
- No large `/tmp/...` paths in the apps on the console.

And there's no public API changes regardless.

ekzhang:
@prbot approve

erikbern:
Yep agree that this PR is a clear improvement!

PR #1628, opened 2024-03-29
Try testing the client with protobuf 5

mwaskom:
Protobuf 5.26 was released [earlier this month](https://pypi.org/project/protobuf/#history). We pin <5.0 for the local client, I assume just to be conservative, but I am testing whether there are any issues in CI.


mwaskom:
Integration tests also passed using this version locally, by the way.

PR #1637, opened 2024-03-31
Document queue partitions and allow setting TTL on put

gongy:
Same as previous PR - can address comments in follow-ups.

### DIFF
--- a/modal/queue.py
+++ b/modal/queue.py
@@ -2,7 +2,7 @@
 import queue  # The system library
 import time
 import warnings
-from typing import Any, AsyncIterator, List, Optional, Type
+from typing import Any, AsyncGenerator, AsyncIterator, List, Optional, Type
 
 from grpclib import GRPCError, Status
 from synchronicity.async_wrap import asynccontextmanager
@@ -11,11 +11,11 @@ from modal_proto import api_pb2
 
 from ._resolver import Resolver
 from ._serialization import deserialize, serialize
-from ._utils.async_utils import TaskContext, synchronize_api
+from ._utils.async_utils import TaskContext, synchronize_api, warn_if_generator_is_not_consumed
 from ._utils.grpc_utils import retry_transient_errors
 from .client import _Client
 from .exception import InvalidError, deprecation_warning
-from .object import EPHEMERAL_OBJECT_HEARTBEAT_SLEEP, _get_environment_name, _Object, live_method
+from .object import EPHEMERAL_OBJECT_HEARTBEAT_SLEEP, _get_environment_name, _Object, live_method, live_method_gen
 
 
 class _Queue(_Object, type_prefix="qu"):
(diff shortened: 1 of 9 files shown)


gongy:
@prbot approve

modal/queue.py
```diff
@@ -23,11 +23,31 @@ class _Queue(_Object, type_prefix="qu"):
...
+    is also isolated from all other partitions. Please see the Usage section below for an example using partitions.
+
+    **Lifetime of a queue and its partitions**
+
+    By default, each partition is cleared 24 hours after the last `put` operation. A lower TTL can be specified by the `partition_ttl`
+    argument in the `put` or `put_many` methods. Each partition's expiry is handled independently.
+
+    As such, `Queue`s are best used for communication between active functions and not relied on for persistent storage.
+
+    On app completion or after stopping an app any associated `Queue` objects are cleaned up. All its partitions will be cleared.
```
aksh-at, line 44: suggestion: "When an app completes or is stopped, any ..."

modal/queue.py
```diff
@@ -23,11 +23,31 @@ class _Queue(_Object, type_prefix="qu"):
...
+
+    **Limits**
+
+    A single `Queue` can contain up to 100,000 partitions, each with up to 5,000 items. Each item can be up to 256 KiB.
```
aksh-at, line 48: Should we also document the total memory limit?

aksh-at (commented):
Nice! Can we also add it to the guide page on "Dicts and queues"? (maybe we should separate those two into separate pages now?)

modal/queue.py
```diff
@@ -23,11 +23,31 @@ class _Queue(_Object, type_prefix="qu"):
...
+    Specifying partition keys gives access to other independent FIFO partitions within the same `Queue` object.
+    Across any two partitions, puts and gets are completely independent. For example, a put in one partition does not affect
+    a get in any other partition.
+
+    When no partition key is specified (by default), puts and gets will operate on a default partition. This default partition
+    is also isolated from all other partitions. Please see the Usage section below for an example using partitions.
+
+    **Lifetime of a queue and its partitions**
+
+    By default, each partition is cleared 24 hours after the last `put` operation. A lower TTL can be specified by the `partition_ttl`
```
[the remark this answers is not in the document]
gongy: Old clients get the old 30 day window server side. I reached out to openart to discuss their TTL requirements.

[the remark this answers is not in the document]
mwaskom: +1 same question

modal/queue.py
mwaskom: Is this true for "persisted" queues as well, or just ephemeral queues?

modal/queue.py
```diff
@@ -44,6 +64,21 @@ def main():
 
         assert my_queue.get() == "some value"
         assert my_queue.get() == 123
+
+        my_queue.put(0)
+        my_queue.put(1, partition_key="foo")
```
mwaskom, line 69: The `.put` method has `partition` as the parameter name, but you have `partition_key` here (and below).

modal/queue.py
```diff
@@ -44,6 +64,21 @@ def main():
...
+        my_queue.put(2, partition_key="bar")
+
+        # Default and "foo" partition are ignored by the get operation.
+        assert my_queue.get(partition_key="bar") == 2
+
+        # Set custom 10s expiration time on "foo" partition.
```
mwaskom, line 75: I think I still have the question I asked (poorly) on Slack. What happens if you do this:

```
my_queue.put(1, partition="a", partition_ttl=20)
time.sleep(10)
my_queue.put(2, partition="a", partition_ttl=5)
```

I could imagine at least 2 possible states after the third line executes:
1. The `"a"` partition has 5 seconds to live; we've completely reset its TTL
2. The `"a"` partition has 10 seconds to live;  it lives until at least the longest TTL implied by puts during its lifetime.

IMO if (1) is the answer, I sort of think that's an API footgun, since queues are explicitly meant to used across multiple functions, so it can be hard / impossible from the local context where queue is used to tell whether the requested TTL will actually be honored.

I think another unspecified behavior here is what happens if you set a `partition_ttl` when using the default / anonymous partition.

modal/queue.py
```diff
@@ -23,11 +23,31 @@ class _Queue(_Object, type_prefix="qu"):
...
-    **Lifetime of a queue and its contents**
+    By default, the `Queue` object acts as a single FIFO queue which supports puts and gets (blocking and non-blocking).
 
-    A `Queue`'s lifetime matches the lifetime of the app it's attached to, but the contents expire after 30 days.
-    Because of this, `Queues`s are best used for communication between active functions and not relied on for
-    persistent storage. On app completion or after stopping an app any associated `Queue` objects are cleaned up.
+    **Queue partitions (beta)**
+
+    Specifying partition keys gives access to other independent FIFO partitions within the same `Queue` object.
+    Across any two partitions, puts and gets are completely independent. For example, a put in one partition does not affect
```
mwaskom, line 31: Super super nit but some of these lines are very long. I guess there's no way to get ruff to auto-wrap docstrings?

modal/queue.py
mwaskom: Could also mention what happens if you exceed these limits (e.g. if there's a specific error class that could be caught).

PR #1642, opened 2024-04-01
Fix lazy hydration for Dict, Queue, Volume, NetworkFileSystem

freider:
This should fix most, if not all, hydration issues that users have had lately.

Looking into other object types too, and they should be pretty straight forward imo

As a shortcut I just inject the `Client.from_env()` client where a client was currently missing and not readily accessible, which feels a bit like a crutch. Preferably, we should find a nice way to inject a client directly in `from_name()` (mostly to make tests cleaner)

### DIFF
--- a/modal/_utils/grpc_testing.py
+++ b/modal/_utils/grpc_testing.py
@@ -175,7 +175,10 @@ class InterceptionContext:
                 self.calls = self.calls[i + 1 :]
                 return msg
 
-        raise Exception(f"No message of that type in call list: {self.calls}")
+        raise KeyError(f"No message of that type in call list: {self.calls}")
+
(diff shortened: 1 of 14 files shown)


erikbern:
Great, ship it!!

I briefly looked at the same thing a couple of weeks ago and ran into the same issue with the client dependency. I think we should add an optional client to `.from_name`. This would decouple *hydration* from *setting the client* which maybe changes a few things a bit? However I think it would make tests a bit cleaner.

erikbern:
Btw it will be a bit more annoying to solve this for `Cls` than other object types because of how the constructor needs to be synchronous. We'll have to figure this out.

erikbern (approved):

PR #1643, opened 2024-04-01
Update client dependencies in all base Image constructors

mwaskom:
This is the first step in adding versioned build _logic_ to image definitions. It focuses on how we install the client dependencies. I have added a new requirements file and updated all of the base image constructors to use it. I've additionally implemented a few related enhancements, like removing the requirements file after using it to produce a cleaner container 🧽.

<details>
<summary> A few open questions that I resolved in subsequent commits </summary>

A couple open questions / thoughts:

- I'm a little undecided on whether we want to put the requirements in a `requirements.txt` file that gets copied into the image, as we currently do, or whether it's better to specify them directly in a `pip` command (as I do here). The upside to doing everything in the command string is that it means less data to keep organized in the client repository / source distribution. It also makes the image definition logic feel slightly simpler. The downside is that (I think) it precludes us from installing different dependency versions based on the Python version. I don’t believe we need to do this currently. But maybe we will at some point in the future?
- Originally, I had imagined we would have a special `PREVIEW` version tag that we could use to gradually roll out and test updates. That makes it more obviously "unstable" and allows us to easily suppress it from the invalid version message. But it seems like it might be kind of a pain to keep this updated — when we get to "release" the next version, we'll have to go through and update all the if statements.

</details>

--- 
Part of MOD-2659


modal/image.py
```diff
@@ -72,19 +72,59 @@ def _dockerhub_python_version(python_version=None):
...
     # Locate Modal client requirements.txt
     import modal
 
     modal_path = modal.__path__[0]
     if python_version is None:
         major, minor, *_ = sys.version_info
     else:
-        major, minor = python_version.split("-")[0].split(".")[:2]
-    suffix = {(3, 12): ".312"}.get((int(major), int(minor)), "")
+        major, minor = map(int, python_version.split("-")[0].split(".")[:2])
```
mwaskom, line 83: Unrelated, but silences a pyright error

mwaskom:
@prbot approve

Merging as we can continue iterating here before "releasing" `2024.04`

PR #1647, opened 2024-04-03
simplify `_ContainerApp`

erikbern:
Making the initializer a global function instead (and cleaned up some tangentially related stuff)

Once this is merged, I can remove the synchronization of `_ContainerApp` and `_LocalApp` and turn them into simple dataclasses

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -85,6 +85,11 @@ container_app = synchronize_api(_container_app)
 assert isinstance(container_app, ContainerApp)
 
 
+def _set_is_container_app():
+    global _is_container_app
+    _is_container_app = True
+
+
 async def _init_container_app(
     client: _Client,
     app_id: str,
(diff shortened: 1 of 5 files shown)


modal/app.py
```diff
@@ -112,6 +85,36 @@ def stop_fetching_inputs(self):
...
 
 
+async def _init_container_app(
+    client: _Client,
+    app_id: str,
+    environment_name: str = "",
+    function_def: Optional[api_pb2.Function] = None,
+):
+    """Used by the container to bootstrap the app and all its objects. Not intended to be called by Modal users."""
+    global _container_app, _is_container_app
```
mwaskom, line 95: Can we not pass the container app in when we call this? I'm a little confused about the order of operations but not a big deal since it's basically consistent with the status quo.

mwaskom (approved):

modal/app.py
erikbern: we don't have to, but then we'd have to have a separate function to set `_is_container_app = True`

this is fairly temporary anyway so I think let's leave it for now and optimize further later

erikbern:
thanks @mwaskom

freider:
👍

PR #1648, opened 2024-04-03
Support "App" instead of "Stub"

erikbern:
This introduces a dummy subclass of Stub called App.

The point is forward compatibility with future versions (e.g. people look at future docs where we call it "App" and try to run it with older versions of the client). It's not an official launch in any way (it will take another few weeks probably)

Example:
```python
import modal

app = modal.App()

@app.function()
def xyz():
    print("hello")
```

### DIFF
--- a/test/supports/app_run_tests/stub_is_now_app.py
+++ b/test/supports/app_run_tests/stub_is_now_app.py
@@ -1,0 +1,9 @@
+# Copyright Modal Labs 2024
+import modal
+
+app = modal.App()
+
+
+@app.function()
+def foo():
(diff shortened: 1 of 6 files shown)


test/supports/app_run_tests/stub_is_now_app.py
```diff
@@ -0,0 +1,9 @@
+# Copyright Modal Labs 2024
+import modal
+
+app = modal.App()
```
mwaskom, line 4: image

mwaskom (approved):
Great!!

PR #1649, opened 2024-04-03
Update Python versions used in Image constructors

mwaskom:
More image builder updates, focusing on the way that we identify the version of Python to use in our images (specifically with version `2024.04`):

- For `debian_slim`, I updated the full `X.Y.Z` versions that we resolve `X.Y` arguments to, using the currently latest versions on Dockerhub
- For `conda` and `micromamba`, I made a few changes:
  - The default Python version will now match the local version — previously it was hardcoded to "3.9" — to be more consistent with other constructors
  - We now allow `X.Y.Z` specification (I am not sure why we rejected this previously)

Additionally, I improved the way that we validate and give feedback about the Python version specification and added some additional tests for this.

Note that we cannot update the micro versions that we create mounts for (i.e., to support `add_python=...`) because the context mount is not added to the image definition in a a way that can depend on the result of an RPC.

---
Part of MOD-2659

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2022
 import os
 import pytest
+import re
 import sys
 import threading
 from hashlib import sha256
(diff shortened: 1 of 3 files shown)


test/image_test.py
```diff
@@ -664,17 +716,16 @@ def test_get_modal_requirements_path(builder_version, python_version):
...
-                        assert servicer.image_builder_versions
-                        for version in servicer.image_builder_versions.values():
-                            assert version == "2000.01"
+    test_requirements = str(test_dir / "supports" / "test-requirements.txt")
+    with mock.patch("modal.image._get_modal_requirements_path", lambda *_, **__: test_requirements):
+        with mock.patch("modal.image._dockerhub_python_version", lambda *_, **__: "3.11.0"):
+            with mock.patch("test.conftest.ImageBuilderVersion", Literal["2000.01"]):
+                with mock.patch("modal.image.ImageBuilderVersion", Literal["2000.01"]):
+                    with Client(servicer.remote_addr, api_pb2.CLIENT_TYPE_CONTAINER, ("ak-123", "as-xyz")) as client:
+                        with stub.run(client=client):
```
mwaskom, line 725: image

mwaskom:
@prbot approve

PR #1650, opened 2024-04-03
Update Debian version in base images for 2024.04

mwaskom:
Simple image builder update: bump `bullseye` -> `bookworm` everywhere we use a base Debian image.

I also bumped the version tag on the `micromamba` image (it's not actually completely clear what this version corresponds to).

Part of MOD-2659

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -16,6 +16,7 @@ from modal.exception import DeprecationError, InvalidError, VersionError
 from modal.image import (
     SUPPORTED_PYTHON_SERIES,
     ImageBuilderVersion,
+    _dockerhub_debian_codename,
     _dockerhub_python_version,
     _get_modal_requirements_path,
     _validate_python_version,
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot approve

test/image_test.py
```diff
@@ -719,13 +721,16 @@ def test_image_builder_version(servicer, test_dir):
...
-            with mock.patch("test.conftest.ImageBuilderVersion", Literal["2000.01"]):
-                with mock.patch("modal.image.ImageBuilderVersion", Literal["2000.01"]):
-                    with Client(servicer.remote_addr, api_pb2.CLIENT_TYPE_CONTAINER, ("ak-123", "as-xyz")) as client:
-                        with stub.run(client=client):
-                            assert servicer.image_builder_versions
-                            for version in servicer.image_builder_versions.values():
-                                assert version == "2000.01"
+            with mock.patch("modal.image._dockerhub_debian_codename", lambda *_, **__: "bullseye"):
+                with mock.patch("test.conftest.ImageBuilderVersion", Literal["2000.01"]):
+                    with mock.patch("modal.image.ImageBuilderVersion", Literal["2000.01"]):
```
erikbern, line 726: not important but is there a way to do some sort of dependency injection here instead of relying on patching?

and separately maybe it's a bit cleaner syntactically if you use one `with` statement for all of the context managers (so you don't end up with this mega-nesting

mwaskom: > not important but is there a way to do some sort of dependency injection here instead of relying on patching?

Hm not sure — since we're using this type annotation for both static analysis and runtime validation I think it should be (or maybe needs to be, due to synchronicity) in the global scope of that file?

> and separately maybe it's a bit cleaner syntactically if you use one with statement for all of the context managers (so you don't end up with this mega-nesting

Yeah no kidding — but that isn't supported on 3.8! I have a TODO to clean it up when we can :D

PR #1655, opened 2024-04-05
Move maps to own module

freider:
`functions.py` is something like ~1500+ loc, and the `.map()` family of methods account for a sizeable part of its complexity, especially after the recent rather verbose changes in #1588 .

This PR moves code around but makes no significant changes to bodys of functions. A small bonus is that a few of the previously `@synchronicity.nowrap` functions no longer need to be synchronicity-wrapped to begin with since they aren't helper methods on the Function class anymore.


freider:
@prbot approve

modal/__init__.py
```diff
@@ -9,16 +9,16 @@
 from modal_version import __version__
 
 try:
-
     from ._container_io_manager import interact, is_local
     from ._tunnel import Tunnel, forward
+    from ._utils.function_utils import current_function_call_id, current_input_id
```
mwaskom, line 14: Feels a little weird to be importing functions from a private subpackage into the top-level namespace IMO.

mwaskom (commented):
Breaking up the huge module makes sense (I guess I personally don't feel that large modules are per se a problem if that's the best grouping for the functionality — but moving the map-related code to its own module is sensible). I have some aversion to putting functions that are part of public API in the private `_utils` subpackage though as it won't necessarily be obvious that we expose those. Can we find a better home for those?

PR #1657, opened 2024-04-05
rename 'Function env' to 'Function spec'

thundergolfer:
## Describe your changes

image

* Calling this an "environment" clashes with our Environment domain term. 
* We also had this dataclass in the docs 

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -16,7 +16,7 @@ from typing_extensions import TypedDict
 from ..config import config
 from ..environments import ensure_env
 from ..exception import ExecutionError, InvalidError, _CliUserExecutionError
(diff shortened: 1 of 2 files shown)


mwaskom:
Nice, makes sense.

Why do we have public docs for this btw? Is this something users would ever interact with directly?

mwaskom:
Oh I see you're also making it private, so we won't going forward?

thundergolfer:
Yep, it'll be hidden

mwaskom (approved):

mwaskom:
Makes sense. Sort of confusing how most of our objects look private on declaration but then become public when they get synchronicity wrapped. Different issue though :)

PR #1661, opened 2024-04-06
Temporary revert of map refactor

freider:
I think it might have caused a performance and/or freeze regression when mapping over large inputs that require blob uploads. Reverting the relevant patch(es) until I can properly debug this.

Side note: it turns out the client doesn't have any unit tests for e2e function calls with large inputs, and adding a trivial one freezes the test runner for me regardless of this patch... Need more time to investigate that and add tests


freider:
@prbot approve

mwaskom:
👍 thanks for catching it!

PR #1665, opened 2024-04-08
intro memory_mib_max resource field to OOM containers

thundergolfer:
## Describe your changes

Adds new field that is required to handle OOMing containers based on user requested limit. 

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1456,9 +1456,10 @@ message RateLimit {
 }
 
 message Resources {
-  uint32 memory_mb = 2;
-  uint32 milli_cpu = 3;
+  uint32 memory_mb = 2; // MiB
+  uint32 milli_cpu = 3; // milli CPU cores
   GPUConfig gpu_config = 4;
+  uint32 memory_mib_max = 5; // MiB
 }
 
 message S3Mount {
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1456,9 +1456,10 @@ message RateLimit {
 }
 
 message Resources {
-  uint32 memory_mb = 2;
-  uint32 milli_cpu = 3;
+  uint32 memory_mb = 2; // MiB
+  uint32 milli_cpu = 3; // milli CPU cores
   GPUConfig gpu_config = 4;
+  uint32 memory_mib_max = 5; // MiB
```
mwaskom, line 1462: YMMV but it feels arguably more error-prone to have inconsistent naming schemes than to just be consistently imprecise in the language.

mwaskom: Alternatively just bite the bullet on deprecating the _mb suffix so we can be internally and externally consistent at some point in the future.

thundergolfer: I think I can _maybe_ rename the field without much trouble because proto cares about field number not the name.

mwaskom: Oh yeah that’s smart. I’m not experienced enough with protobuf to say that is definitely a good idea but it does correspond to my understanding of how they work.

mwaskom: I guess the tricky part is coordinating the change between between the client / server codebases; you'll still need to start serverside and make it check both attribute names I expect.

thundergolfer: I don't have the time to do this rename carefully so I'll just go with your suggestion of the consistently imprecise naming :ok

mwaskom (approved):

PR #1666, opened 2024-04-09
Fix `spawn_sandbox` inside containers

erikbern:
The issue was introduced in #1659 – the stub would end up with an unsynchronized `Client` (as opposed to `_Client`)

Adding tests for this as well

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -21,6 +21,7 @@ from typing import TYPE_CHECKING, Any, AsyncGenerator, AsyncIterator, Callable, 
 
 from google.protobuf.message import Message
 from grpclib import Status
+from synchronicity import Interface
 
 from modal_proto import api_pb2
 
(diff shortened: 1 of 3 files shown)


mwaskom (approved):
Seems fine as an fix to the immediate problem but needing to access private synchronicity logic like this feels like a deeper issue.

Would an alternate fix to the specific issue be exposing the `image_builder_version` metadata as a property? I think that should resolve this issue, but not sure if we expect other similar issues to pop up after the refactoring.

erikbern:
We need to make sure the objects have the right type inside our code, so `_container_entrypoint.py` already has a bunch of code to translate between synchronized and unsynchronized objects. I'm honestly not sure how to get around this other than maybe rewriting synchronicity completely (which we've talked about).

erikbern:
> Would an alternate fix to the specific issue be exposing the `image_builder_version` metadata as a property? I think that should resolve this issue, but not sure if we expect other similar issues to pop up after the refactoring.

This would have solved it, but would potentially cause much worse bugs. The problem here is we have a synchronized object leak into unsynchronized code, so we need to plug this leak. Otherwise we can end up with deadlocks.

PR #1669, opened 2024-04-10
volumes: Improve error message when reload fails due to open files

gimaker:
If reloading fails because of open files, annotate the error with information with the path to the first open file or problematic cwd we detect.

The paths reported are volume-relative, if the volume is mounted on `/vol` and a file `/vol/foo.txt` is open that will be reported as `foo.txt`. This is not ideal, but unfortunately we don't have an easy way to determine where the volume is mounted.

This may or may not be entirely reliable, but ought to be better than the status quo and catch more common issues.
* The containers view of open files might not be the same as that outside of the container (the latter is what really matters).
* It's racy - the file that prevented reloading might be closed by the time we look for it.

```
import modal
import os

stub = modal.Stub()
volume = modal.Volume.from_name("test123", create_if_missing=True)

@stub.function(volumes={"/volume": volume})
def f():
    with open("/volume/test.txt", "w") as f:
        try:
            volume.reload()
        except RuntimeError as ex:
            print(ex)
    
    os.chdir("/volume")
    try:
        volume.reload()
    except RuntimeError as ex:
        print(ex
```
-->
```
there are open files preventing the operation: path test.txt is open
there are open files preventing the operation: cwd is inside volume
```

## Changelog

- When Volume reloads fail due to open files, we now try to identify and report the relevant paths. Note that there may be some circumstances in which we are unable to identify the specific files blocking a reload and will report a generic error message in that case.

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2023
 import asyncio
 import concurrent.futures
+import os
 import time
 from contextlib import nullcontext
 from pathlib import Path, PurePosixPath
@@ -276,6 +277,28 @@ class _Volume(_Object, type_prefix="vo"):
         try:
             await self._do_reload()
         except GRPCError as exc:
+            # TODO(staffan): This is very brittle and janky. We should return the open files from the server.
+            if exc.message == "there are open files preventing the operation":
+                # Attempt to identify what open files are problematic and include the path in the error message.
+                # This is best-effort and not necessarily bulletproof, as the view of open files inside the container
+                # might differ from that outside - but it will at least catch common errors.
+                vol_path = f"/__modal/volumes/{self.object_id}"
+
+                cwd = PurePosixPath(os.readlink("/proc/self/cwd"))
+                if cwd.is_relative_to(vol_path):
+                    raise RuntimeError(f"{exc.message}: the current working directory is inside the volume")
+
+                open_paths = []
+                for fd in os.listdir("/proc/self/fd"):
+                    try:
+                        path = PurePosixPath(os.readlink(f"/proc/self/fd/{fd}"))
+                        if path.is_relative_to(vol_path):
+                            rel_path = path.relative_to(vol_path)
+                            open_paths.append(rel_path.as_posix())
+                    except FileNotFoundError:
+                        pass
+                if open_paths:
+                    raise RuntimeError(f"{exc.message}: {open_paths} (relative to where the volume is mounted)")
             raise RuntimeError(exc.message) if exc.status in (Status.FAILED_PRECONDITION, Status.NOT_FOUND) else exc
 
     @live_method_gen
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -618,6 +618,14 @@ def _open_files_error_annotation(mount_path: str) -> Optional[str]:
 
     self_pid = os.readlink("/proc/self")
 
+    # TODO(staffan): Replace with path.is_relative_to(other) when support for Python 3.8 is dropped
+    def is_path_relative_to(path: PurePosixPath, other: str) -> bool:
+        try:
+            path.relative_to(other)
+            return True
+        except ValueError:
+            return False
+
     def find_open_file_for_pid(pid: str) -> Optional[str]:
         # /proc/{pid}/cmdline is null separated
         with open(f"/proc/{pid}/cmdline", "rb") as f:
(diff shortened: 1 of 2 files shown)


modal/volume.py
```diff
@@ -276,6 +277,28 @@ async def reload(self):
...
+                # This is best-effort and not necessarily bulletproof, as the view of open files inside the container
+                # might differ from that outside - but it will at least catch common errors.
+                vol_path = f"/__modal/volumes/{self.object_id}"
+
+                cwd = PurePosixPath(os.readlink("/proc/self/cwd"))
+                if cwd.is_relative_to(vol_path):
+                    raise RuntimeError(f"{exc.message}: the current working directory is inside the volume")
+
+                open_paths = []
+                for fd in os.listdir("/proc/self/fd"):
```
gimaker, line 292: Doesn't work if subprocesses has open files, since this naively only looks at `/proc/self/fd`. Still better than nothing!

Should be simple enough to just traverse everything in `/proc/<pid>/fd` though, to catch those as well

modal/volume.py
```diff
@@ -276,6 +277,28 @@ async def reload(self):
...
+                    try:
+                        path = PurePosixPath(os.readlink(f"/proc/self/fd/{fd}"))
+                        if path.is_relative_to(vol_path):
+                            rel_path = path.relative_to(vol_path)
+                            open_paths.append(rel_path.as_posix())
+                    except FileNotFoundError:
+                        pass
+                if open_paths:
+                    raise RuntimeError(f"{exc.message}: {open_paths} (relative to where the volume is mounted)")
```
mwaskom, line 301: Will we just silently swallow this error if we can't work out the specific list of files? It's not totally obvious to me what's "brittle" or what the failure modes look like.

gimaker: no errors being swallowed - if we can't work out a list of files we will just raise `RuntimeError("there are open files preventing the operation")` error without any more info.

It's brittle / janky because there are things that can change server-side that will break this:
 * We rely on the specific error mesage (`"there are open files preventing the operation"`), which is :vomiting_face: :see_no_evil:
 * The specific path where volumes are mounted: `"/__modal/volumes/{self.object_id}"`

Where "break" just means reverting to the an error message will less info - so not the end of the world.

Ideally we'd figure out what files are causing issues server-side and just return them in the error message. Doing it client-side is a lot easier though.

Another reason to do that is that the container view of what files are open doesn't necessarily match what the host sees outside of the container, which is what really matters. The container view of open files will be whatever gVisor says it is - which may or may not be true :)

mwaskom: Thanks that's helpful. Actually my question was fairly dumb anyway, I missed that the generic runtime error isn't in an else clause here, so we'll fall down to it if we don't exit on the highlighted lines.

I think the change here is a definite improvement, despite not being perfect. Thanks for adding it!

mwaskom (approved):

mwaskom:
BTW I added a short changelog because this is a nice enhancement and also it may be useful for users to know about the potential for less-than-full information; feel free to edit!

gimaker:
Ok, made this cover more cases by checking all processes:

```
import modal
import os
from multiprocessing import Process
import time

stub = modal.Stub()
volume = modal.Volume.from_name("test123", create_if_missing=True)

def subp():
    with open("/volume/test2.txt", "w") as f:
        time.sleep(5)

def subp2():
    os.chdir("/volume")
    time.sleep(5)

@stub.function(volumes={"/volume": volume})
def f():
    with open("/volume/test.txt", "w") as f:
        try:
            volume.reload()
        except RuntimeError as ex:
            print(ex)

    p = Process(target=subp)
    p.start()
    try:
        volume.reload()
    except RuntimeError as ex:
        print(ex)
    p.join()

    p = Process(target=subp2)
    p.start()
    try:
        volume.reload()
    except RuntimeError as ex:
        print(ex)
    p.join()
    
    os.chdir("/volume")
    try:
        volume.reload()
    except RuntimeError as ex:
        print(ex)
```

->

```
there are open files preventing the operation: path test.txt is open

there are open files preventing the operation: path test2.txt is open from 'python-u-<edited-for-brevity>'

there are open files preventing the operation: cwd of 'python-u-<edited-from-brevity>' is inside the volume

there are open files preventing the operation: cwd is inside the volume
```

To keep the error message from becoming insanely long I'm now only including the first offending path/cwd found.

gimaker:
Thanks for the review and changelog entry @mwaskom !

mwaskom (approved):
Nice! Learned some unix things.

Is it right that, contra your previous version where a long list of files could be printed, here we'll report at most one open file that is blocking the reload?

gimaker:
> Is it right that, contra your previous version where a long list of files could be printed, here we'll report at most one open file that is blocking the reload?

@mwaskom Correct. Reasoning being that the error messages can be **very** verbose if we include every path and process that is keeping open files. If we had a way to present the user with a nicely bulleted list in the error message that would be more palatable but I didn't find any precedent for that (nor I am sure that's a good idea...).

mwaskom:
Makes sense!

PR #1670, opened 2024-04-10
Improve warning and documentation related to passing objects to Stub()

mwaskom:
I noticed that we were documenting a deprecated usage pattern in the `Stub` constructor docstring. (I am a little confused about how the doc tests didn't flag this).

While fixing that, I tried to improve some of the feedback (warnings and errors) in the same neighborhood. I'm not sure we're being optimally helpful with the warning about what to do instead of passing the object to the stub constructor — the situation is fairly nuanced wrt persistent / ephemeral objects. Maybe can be further improved.

### DIFF
--- a/modal/exception.py
+++ b/modal/exception.py
@@ -121,7 +121,7 @@ def deprecation_error(deprecated_on: Tuple[int, int, int], msg: str):
 
 
 def deprecation_warning(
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot lgtm

erikbern (approved):
Great!

PR #1675, opened 2024-04-11
Update volume.copy_files reference docs

yirenlu92:
- clarify that `volume.copy_files` takes a list as a first parameter
- clarify that you don't have to prepend the mount location

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -476,7 +476,7 @@ class _Volume(_Object, type_prefix="vo"):
         The `src_paths` parameter is a list. If you want to copy a single file, you should pass a list with a
         single element.
 
-        Both the `src_paths` and `dst_path` parameters are the location of the file *inside* the volume. You do not need to prepend
+        `src_paths` and `dst_path` should refer to the desired location *inside* the volume. You do not need to prepend
         the volume mount path.
 
         **Usage**
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -488,9 +488,10 @@ class _Volume(_Object, type_prefix="vo"):
         vol.copy_files(["bar/example.txt"], "bar/example2.txt")  # Rename a file by copying
         ```
 
-        Note that if the volume is already mounted on the Modal function, you should use normal filesystem operations 
-        like `os.rename()` and then `commit()` the volume. The `copy_files()` method is useful when you don't have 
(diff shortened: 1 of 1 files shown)


modal/volume.py
```diff
@@ -419,6 +419,27 @@ async def copy_files(self, src_paths: Sequence[Union[str, bytes]], dst_path: Uni
...
+        @stub.function(volumes={"/root/foo": volume})
+        def g():
+            volume.reload()  # Fetch latest changes
+            with open("/root/foo/bar/example.txt", "r") as f:
+                print(f.read())
+
+            volume.copy_files(["bar/example.txt"], "bar2")  # Copy files to another directory
+            volume.copy_files(
+                ["bar/example.txt"], "bar/example2.txt"
+            )  # Rename a file by copying
```
mwaskom, line 441: I think it's an anti-pattern to use the Volume API methods when the volume is mounted to the function; we should probably demonstrate usage in a non-modal-app context (and maybe say that explicitly).

yirenlu92: yup, we had a separate discussion about this, updated :-)

modal/volume.py
```diff
@@ -472,6 +472,24 @@ async def copy_files(self, src_paths: Sequence[Union[str, bytes]], dst_path: Uni
...
+        ```python notest
+        vol = modal.Volume.lookup("my-modal-volume")
+
+        vol.copy_files(["bar/example.txt"], "bar2")  # Copy files to another directory
+        vol.copy_files(["bar/example.txt"], "bar/example2.txt")  # Rename a file by copying
+        ```
+
+        Note that if the volume is already mounted on the Modal function, you should use normal filesystem operations like `os.rename()` and then `commit()` the volume.
+        The `copy_files()` method is useful when you don't have the volume mounted as a filesystem, e.g. when running a script on your local computer.
         """
```
mwaskom, line 493:
```suggestion
        Note that if the volume is already mounted on the Modal function, you should use normal filesystem operations 
        like `os.rename()` and then `commit()` the volume. The `copy_files()` method is useful when you don't have 
        the volume mounted as a filesystem, e.g. when running a script on your local computer.
```

Just trying to keep everything within a reasonable linewidth

mwaskom (approved):
Thanks for improving this

PR #1678, opened 2024-04-12
Move FunctionIOManager into a separate module

erikbern:
Big PR but all it does is to

1. Move `FunctionIOManager` into a new module `_container_io_manager.py`
2. Rename it to `ContainerIOManager`

Really just a bunch of search & replace.

The direct reason I'm doing that is that we have a `_container_app` singleton that isn't really used, outside of `interact()` and `modal.experimental.stop_fetching_inputs()`. In both those cases it makes a lot more sense to operate on the `ContainerIOManager` which is already a singleton, rather than setting stuff on the fairly useless `_container_app` singleton.

But the indirect reason is that `_container_entrypoint.py` has been growing and growing for a while. At 1,165 lines it's time to split it up imo.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -7,9 +7,8 @@ import signal
 import sys
 import threading
 import time
-from collections.abc import Iterable
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Type
+from typing import TYPE_CHECKING, Any, Callable, List, Optional, Sequence, Type
 
 from google.protobuf.message import Message
 from synchronicity import Interface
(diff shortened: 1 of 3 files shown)


modal/_container_entrypoint.py
```diff
@@ -124,540 +106,14 @@ def _sigint_handler():
...
-            else:
-                logger.debug(f"modal.Volume background commit success for {volume_id}.")
-
-
-FunctionIOManager = synchronize_api(_FunctionIOManager)
-
-
 def call_function_sync(
-    function_io_manager,  #: FunctionIOManager,  TODO: this type is generated at runtime
+    container_io_manager,  #: ContainerIOManager,  TODO: this type is generated at runtime
```
freider, line 110: Not relevant for this PR, but I just realized we have a bunch of these comments around the codebase, but now since we are actually generating type stubs for wrapped types we can annotate these properly and get some better static type analysis (at least using pyright/vscode... mypy has some issues)

freider (approved):
Great!

freider:
Separately, we should unify how we communicate the "container" concept to users. I think at this point we use both "container", "runner" and "task" in different places (also unrelated to this PR)

modal/_container_entrypoint.py
erikbern: yeah i actually ran into a bunch of issues with types that i had to fix, so i think that's good

PR #1679, opened 2024-04-12
Add back `list_apps`

erikbern:
See #1677 

Using an import hack to expose this under `modal.app` even though I'm removing that module shortly.


mwaskom:
You might need to use a module-level `__getattr__` here to avoid the synchronicity complications — though not completely sure if that will work.

erikbern:
@mwaskom wdym? I fixed the mypy complaint by suppressing it for now.

Unfortunately `list_apps` won't be listed in the documentation until we've done the stub.py -> app.py switch. But that's fairly soon.

thundergolfer (approved):

PR #1680, opened 2024-04-12
Unify FileEntry protobufs and add a public dataclass

ekzhang:
This unifies the SharedVolumeListFilesEntry and VolumeListFilesEntry protobufs, which now just both become FIleEntry. I'm also adding a public dataclass for FileEntry, which is used the API for listdir() on both NFS and Volume objects.

Note that the NFS implementation doesn't have support for symlinks yet, or for the mtime / size fields. I quickly added support for those in server code.

## Changelog

- The return type of `Volume.listdir()`, `Volume.iterdir()`, `NetworkFileSystem.listdir()`, and `NetworkFileSystem.iterdir()` is now a `FileEntry` dataclass from the `modal.volume` module. The fields of this data class are the same as the old protobuf object returned by these methods, so it should be mostly backwards-compatible.

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -42,6 +42,7 @@ from ._utils.blob_utils import (
 from ._utils.grpc_utils import retry_transient_errors, unary_stream
 from .client import _Client
 from .config import logger
+from .exception import deprecation_error
 from .object import EPHEMERAL_OBJECT_HEARTBEAT_SLEEP, _get_environment_name, _Object, live_method, live_method_gen
 
 # Max duration for uploading to volumes files
@@ -74,6 +75,12 @@ class FileEntry:
             type=FileEntryType(proto.type),
             mtime=proto.mtime,
             size=proto.size,
+        )
+
+    def __getattr__(self, name: str):
+        deprecation_error(
+            (2024, 4, 15),
+            f"The FileEntry dataclass was introduced to replace a private Protobuf message. This dataclass does not have the {name} attribute.",
         )
 
 
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -1,11 +1,13 @@
 # Copyright Modal Labs 2023
 import asyncio
 import concurrent.futures
+import enum
 import os
 import platform
 import re
 import time
 from contextlib import nullcontext
+from dataclasses import dataclass
 from pathlib import Path, PurePosixPath
 from typing import (
     IO,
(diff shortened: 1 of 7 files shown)


ekzhang:
Hmm I can't get IntEnum to typecheck, the `type-stubs` generates a bad type here:

image

ekzhang:
Adding Enum support here: https://github.com/modal-labs/synchronicity/pull/142

mwaskom:
Do you think it's worth adding a backwards compatibility layer for some of the protobuf methods that would previously have been "public" on the return type here? i.e. what that user was trying to do. It would be more work, but also more friendly — not pushing either way.

mwaskom:
(to be clear just suggesting we add those to support a graceful deprecation)

ekzhang:
> Do you think it's worth adding a backwards compatibility layer for some of the protobuf methods that would previously have been "public" on the return type here? i.e. what that user was trying to do. It would be more work, but also more friendly — not pushing either way.

Good point about the protobuf object having some methods like `CopyFrom()`, `SerializeToString()`, `FromString()`, and so on. Hmm, I don't think we can add compat to all of the protobuf message class's methods though.

Is there a specific protobuf method you had in mind? (No pressure here, just trying to get an sense for what people were doing.)

I kept the properties the same:

- `name: str`
- `type: FileEntryType`
- `mtime: int`
- `size: int`

mwaskom:
yeah agreed that scope is unclear, the only thing i've seen in the wild was the user question that motived this change, where they were doing
```
for field in modal_entry.DESCRIPTOR.fields:
        entry_dict[field.name] = getattr(modal_entry, field.name)
 ```

maybe we could do something even dumber like grabbing all the public methods of the protobuf class and adding a method with that name to the dataclass that prints a "sorry, i am a dataclass, you can't do this any more" deprecation message.

**Comment by ekzhang** [comment]
@prbot approve

File: modal/volume.py [hunk 1]
```diff
@@ -40,13 +42,48 @@
...
+            type=FileEntryType(proto.type),
+            mtime=proto.mtime,
+            size=proto.size,
+        )
+
+    def __getattr__(self, name: str):
+        deprecation_error(
+            (2024, 4, 15),
+            f"The FileEntry dataclass was introduced to replace a private Protobuf message. This dataclass does not have the {name} attribute.",
+        )
```
mwaskom, line 84: Clever!

mwaskom (commented):
Nice change — I gather you're already on top of the server-side implications :+1:

PR #1681, opened 2024-04-12
[MOD-2078] Add memory limit specification

thundergolfer:
## Describe your changes

- MOD-2078

The following program will set a memory limit of 2 GiB on the container, which will cause OOM failure when run. 

```python
import modal
import random

stub = modal.Stub()

@stub.function(memory=(1024, 2048))
def main():
    ten_gib = 4 * 1024 * 1024 * 1024  # 4 GiB in bytes
    memory_hog = bytearray(ten_gib)
    for i in range(10_000_000):
        memory_hog[i] = random.randrange(128) # mutate
    print("done")
```

<kbd>
image
</kbd>

    - old servers will ignore the `memory_mb_max` value.

## Changelog

* Specifying a hard memory limit for a `modal.Function` is now supported. Pass a tuple of `memory=(request, limit)`. Above the `limit`, which is specified in MiB, a Function's container will be OOM killed.

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -1,7 +1,7 @@
 # Copyright Modal Labs 2022
 import os
 import typing
-from typing import Any, Callable, Collection, Dict, List, Optional, Type, TypeVar, Union
+from typing import Any, Callable, Collection, Dict, List, Optional, Tuple, Type, TypeVar, Union
 
 from google.protobuf.message import Message
 from grpclib import GRPCError, Status
@@ -258,7 +258,7 @@ class _Cls(_Object, type_prefix="cs"):
     def with_options(
         self: "_Cls",
         cpu: Optional[float] = None,
-        memory: Optional[int] = None,
+        memory: Optional[Union[int, Tuple[int, int]]] = None,
         gpu: GPU_T = None,
         secrets: Collection[_Secret] = (),
         volumes: Dict[Union[str, os.PathLike], _Volume] = {},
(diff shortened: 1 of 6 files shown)


mwaskom:
I think you want to update the type / comment in `Stub.cls` too — super annoying that we need to manually keep these synced. Not completely certain there's nowhere else...

thundergolfer:
Oh yeh thanks!

thundergolfer:
@prbot approve

modal/functions.py
```diff
@@ -802,6 +802,16 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
...
+            if memory and isinstance(memory, int):
+                memory_mb = memory
+                memory_mb_max = 0  # no limit
+            elif memory and isinstance(memory, tuple):
+                memory_mb, memory_mb_max = memory
+                if memory_mb_max < memory_mb:
+                    raise ValueError(f"Cannot specify a memory limit lower than request: {memory_mb_max} < {memory_mb}")
+            else:
+                memory_mb = 0
+                memory_mb_max = 0
```
mwaskom, line 814: Kind of annoying that we need to have this logic duplicated :/

modal/cls.py
```diff
@@ -292,7 +292,19 @@ def with_options(
...
             milli_cpu = int(1000 * cpu) if cpu is not None else None
             gpu_config = parse_gpu_config(gpu)
-            resources = api_pb2.Resources(milli_cpu=milli_cpu, gpu_config=gpu_config, memory_mb=memory)
+            if memory and isinstance(memory, int):
+                memory_mb = memory
+                memory_mb_max = 0  # no limit
+            elif memory and isinstance(memory, tuple):
+                memory_mb, memory_mb_max = memory
+                if memory_mb_max < memory_mb:
+                    raise ValueError(f"Cannot specify a memory limit lower than request: {memory_mb_max} < {memory_mb}")
```
mwaskom, line 301: It's a little unclear to me whether we have a policy of always raising `modal.exception` classes from the client or whether that is even a good idea, but I think that usually we would raise an `InvalidError` here? Although `ValueError` is also correct.

mwaskom (approved):
Good feature; not wild about the amount of code duplication, curious if we need to do some refactoring at some point.

modal/functions.py
thundergolfer: Yeh I could have factored it out into a helper 😅 . Rule of threes.

modal/cls.py
thundergolfer: Yeh also thought the same. I think `ValueError` is correct, but I'd also want to things to be consistently correct across the codebase.

mwaskom: I'm sure this isn't the only instance; maybe we should just make a ticket and try to standardize prior to v1.0

PR #1682, opened 2024-04-13
Type annotations for client classvars

erikbern:
minor typing thing, want to do something similar with `ContainerIOManager`

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -2,7 +2,7 @@
 import asyncio
 import platform
 import warnings
-from typing import AsyncIterator, Awaitable, Callable, Dict, Optional, Tuple
+from typing import AsyncIterator, Awaitable, Callable, ClassVar, Dict, Optional, Tuple
 
 import grpclib.client
 from aiohttp import ClientConnectorError, ClientResponseError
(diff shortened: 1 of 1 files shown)


modal/client.py
```diff
@@ -81,8 +81,8 @@ async def _grpc_exc_string(exc: GRPCError, method_name: str, server_url: str, ti
 
 
 class _Client:
-    _client_from_env = None
-    _client_from_env_lock = None
+    _client_from_env: ClassVar[Optional["_Client"]] = None
+    _client_from_env_lock: ClassVar[asyncio.Lock] = asyncio.Lock()
```
freider, line 85: This is a bit of a foot gun in Python <3.10, where I think asyncio synchronization primitives are bound to the event loop where they are *constructed* (in more recent Python versions they are lazily bound to the event loop of the "first use").

This means this lock will be bound to the event loop of the main thread rather than the synchronicity thread where it's used, which I assume could cause errors when trying to acquire this if it's locked.

Somewhat worse - if `modal` is imported by a non-main thread, this line will actually raise an error since there isn't an event loop to bind to (just had a user report about this).

I'll add a test and revert the move of this lock to a classvar if that's ok? Will keep the nice annotations though :)

PR #1686, opened 2024-04-13
move `stream_function_call_data` info `function_utils.py`

erikbern:
This is in order to avoid a circular dependency between modules


erikbern:
actually realiezd i can do this in another way

erikbern:
actually, i need to do this

PR #1687, opened 2024-04-13
Break out `stream_function_call_data` to `function_utils.py`

erikbern:
This is so we can avoid a circular dependency between `functions.py` and `_container_io_manager.py`

Complete noop, just moving code.


erikbern:
@prbot approve

very low risk, is a noop

PR #1660, opened 2024-04-05
Use dataclass for the App classes

erikbern:
Almost done putting these objects on the stub

PR #1688, opened 2024-04-13
Move `is_local` and `interact` to `_container_io_manager`

erikbern:
This moves a bunch of complexity out of `modal.app`, partly because getting rid of the module will make it easier to rename `modal.stub` to `modal.app`, partly because most of that code is fairly old and janky.

In particular, this moves `is_local` and `interact` to the `_ContainerIOManager`. Also deletes the `_container_app` global, since the `_ContainerIOManager` singleton replace it

PR #1689, opened 2024-04-14
Use dataclass for the App classes

erikbern:
This change supersedes #1660 which fell too far behind to be worth rebasing. It further simplifies `modal.app` to two simple dataclasses. All in all `app.py` is down to 24 lines now from I think 500 lines a few weeks ago, so we're super close to removing it at this point! The goal of this is to rename `stub.py` to `app.py`.


erikbern:
Holding off to merge this for a few hours just so we can revert #1688 more cleanly in case something breaks

PR #1690, opened 2024-04-15
[MOD-2536] support R2 in CloudBucketMount

thundergolfer:
## Describe your changes

- MOD-2536

## Changelog

* Cloudflare R2 bucket support added to `modal.CloudBucketMount`

### DIFF
--- a/modal/cloud_bucket_mount.py
+++ b/modal/cloud_bucket_mount.py
@@ -85,10 +85,11 @@ def cloud_bucket_mounts_to_proto(mounts: List[Tuple[str, _CloudBucketMount]]) ->
     cloud_bucket_mounts: List[api_pb2.CloudBucketMount] = []
 
     for path, mount in mounts:
+        # TODO: in future this relationship between endpoint URL and type will not hold true.
         if mount.bucket_endpoint_url:
-            bucket_type = api_pb2.CloudBucketMount.BucketType.S3
-        else:
             bucket_type = api_pb2.CloudBucketMount.BucketType.R2
+        else:
+            bucket_type = api_pb2.CloudBucketMount.BucketType.S3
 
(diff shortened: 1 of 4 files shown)


modal/cloud_bucket_mount.py
```diff
@@ -1,85 +1,105 @@
...
-    # The given secret can contain AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.
-    # AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY can be omitted if the bucket is publicly accessible.
+    # Credentials used to access a cloud bucket.
+    # If the bucket is private, the secret **must** contain AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.
+    # If the bucket is publicly accessible, the secret is unnecessary and can be omitted.
     secret: Optional[_Secret] = None
 
     read_only: bool = False
     requester_pays: bool = False
-    bucket_type: Union[
```
thundergolfer, line 59: I think this is currently not needed, so will follow YAGNI and remove it. It's unnecessary to make the user pass this type explicitly when we can just infer it from `bucket_endpoint_url`.

modal/stub.py
```diff
@@ -479,7 +479,9 @@ def function(
         network_file_systems: Dict[
             Union[str, PurePosixPath], _NetworkFileSystem
         ] = {},  # Mountpoints for Modal NetworkFileSystems
-        volumes: Dict[Union[str, PurePosixPath], _Volume] = {},  # Mountpoints for Modal Volumes
+        volumes: Dict[
```
thundergolfer, line 482: I get why we've reused this, but also worried that doing it this was is more error prone.

PR #1691, opened 2024-04-15
Re-apply `.map()` on main thread-refactor

freider:
Fixed the incompatibility with `async for foo.map()` (without `.aio()`) incompatibility of the previous version, which broke integration tests (and would have broken any such user code).

**Related synchronicity note:**
I think it might be a good idea to always require an explicit `.aio` in the future when running async iterators (synchronicity-wrapped or not). This would have some not-obvious annoying implications in for types that are iterable themselves though (adding a .aio property, and making it somehow consistent with how .aio() on methods work). Something to think about for synchronicity 2.0...

## Changelog
* Iterators passed to `Function.map()` and similar parallel execution primitives are now executed on the main thread, preventing blocking iterators from possibly locking up background Modal API calls, and risking task shutdowns.


freider:
@prbot approve

PR #1692, opened 2024-04-15
Combine `_LocalApp` and `_ContainerApp` and move them to `modal.running_app`

erikbern:
1. Combine the two (very simple dataclasses) into one.
2. Move them to a new module.

`modal.app` is basically available now, which will let us rename `modal.stub` to `modal.app` 🎉

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -28,13 +28,13 @@ from ._proxy_tunnel import proxy_tunnel
 from ._serialization import deserialize
 from ._utils.async_utils import TaskContext, synchronizer
 from ._utils.function_utils import LocalFunctionError, is_async as get_is_async, is_global_function, method_has_params
-from .app import RunningApp
 from .client import Client, _Client
 from .cls import Cls
 from .config import logger
 from .exception import ExecutionError, InputCancellation, InvalidError
 from .functions import Function, _Function, _set_current_context_ids
 from .partial_function import _find_callables_for_obj, _PartialFunctionFlags
+from .running_app import RunningApp
 from .stub import Stub, _Stub
 
 if TYPE_CHECKING:
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -28,7 +28,7 @@ from ._proxy_tunnel import proxy_tunnel
 from ._serialization import deserialize
 from ._utils.async_utils import TaskContext, synchronizer
 from ._utils.function_utils import LocalFunctionError, is_async as get_is_async, is_global_function, method_has_params
-from .app import _ContainerApp
+from .app import RunningApp
 from .client import Client, _Client
 from .cls import Cls
 from .config import logger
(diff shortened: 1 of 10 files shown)


erikbern:
Btw is there a way to remove `app.py` in the same PR as I rename `stub.py` to `app.py`? I'm a bit nervous the git history will be lost of we squash the commits (since git might thing `stub.py` was removed and the same code added in `app.py` unless it's a "clean" file rename)

mwaskom:
> Btw is there a way to remove app.py in the same PR as I rename stub.py to app.py? I'm a bit nervous the git history will be lost of we squash the commits (since git might think stub.py was removed and the same code added in app.py unless it's a "clean" file rename). Looks like we don't support other merging strategies than squashing.

We can temporarily disable the squash requirement, I've done that in the past to support other housekeeping where we wanted to preserve history.

modal/running_app.py
```diff
@@ -0,0 +1,15 @@
+# Copyright Modal Labs 2024
+from dataclasses import dataclass, field
+from typing import Dict, Optional
+
+from google.protobuf.message import Message
+
+
+@dataclass
+class RunningApp:
```
freider, line 9: Full circle! Didn't we use to have something called `RunningApp`, that was the context var of the stub.run context manager (which was then `app.run()` iirc) ? :D

freider (approved):
Clean 💯

modal/_container_entrypoint.py
```diff
@@ -490,10 +490,12 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
                 client,
             )
 
+        # Get ids and metadata for objects (primarily functions and classes) on the app
+        container_app: RunningApp = container_io_manager.get_app_objects()
```
mwaskom, line 494: Any functional consequences of moving this out of the if block below?

mwaskom (approved):
`modal.app` is dead; long live `modal.app`!

`RunningApp` is kind of a clumsy name IMO given that there's no application logic here anymore. Maybe `AppInfo` or something? That's lame too, though, so not a final suggestion :)

modal/_container_entrypoint.py
erikbern: I noticed we're actually referring to it in the higher-level scope, in the next block. We basically had code like this

```python
if a:
    x = X()

if b:
    x.f()
```

I _think_ that the 2nd block is never entered unless the 1st block is also entered, so I don't think there's a practical difference. I'm surprised mypy or ruff doesn't catch issues like this actually.

erikbern:
> `RunningApp` is kind of a clumsy name IMO given that there's no application logic here anymore. Maybe `AppInfo` or something? That's lame too, though, so not a final suggestion :)

Yeah I don't love it. It does specifically refer to a running app though, whereas I don't think `AppInfo` makes that clear.

modal/running_app.py
erikbern: I think that was basically the `App` class, which I split up at some point, then eventually merged back (after deleting like 99% of the code however)

PR #1704, opened 2024-04-17
Flip around Stub and App

erikbern:
This flips it around so that `Stub` is a dummy subclass of `App` rather than the other way around.

Will still have like 1000 references to "stub" in the code but we can go through and change it fairly incrementally.


mwaskom (approved):

erikbern:
Thanks @mwaskom

PR #1705, opened 2024-04-17
Type annotations in _container_io_manager

erikbern:
Minor thing, just for extra paranoia


erikbern:
@prbot approve

mwaskom:
👍 

> just for extra paranoia

😳 😵‍💫 💀

mwaskom:
Unfortunately I think we still won't actually type check the implementation here :)

PR #1709, opened 2024-04-17
Reapply "run_app and deploy_app" (#1708)

mwaskom:
Docs should now be updated to not break on this reference: https://github.com/modal-labs/modal/pull/11963


ekzhang:
Thank you 🙏

PR #1710, opened 2024-04-17
Tests: stub -> app

erikbern:
Easiest place to start since it doesn't affect docs or the public SDK

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -6,16 +6,16 @@ import sys
 from typing import Optional
 
 
-def deploy_stub_externally(
+def deploy_app_externally(
     servicer,
     file_or_module: str,
-    stub_variable: Optional[str] = None,
+    app_variable: Optional[str] = None,
     deployment_name="Deployment",
     cwd=None,
     env={},
     capture_output=True,
 ) -> Optional[str]:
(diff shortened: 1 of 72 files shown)


erikbern:
@prbot approve

test/helpers.py
```diff
@@ -6,16 +6,16 @@
...
     file_or_module: str,
-    stub_variable: Optional[str] = None,
+    app_variable: Optional[str] = None,
     deployment_name="Deployment",
     cwd=None,
     env={},
     capture_output=True,
 ) -> Optional[str]:
-    # deploys a stub from another interpreter to prevent leaking state from client into a container process (apart from what goes through the servicer)
+    # deploys a app from another interpreter to prevent leaking state from client into a container process (apart from what goes through the servicer)
```
mwaskom, line 18: "a app" you say? ;)

mwaskom: (More seriously we should probably watch out for this when we update the user-facing docs)

mwaskom (commented):
:+1: lot of changes but safe to rely on CI here

test/helpers.py
erikbern: oops this was a mistake, but it's only in tests

mwaskom: yep not an issue in a test comment but good to think about (if you haven't been) for user-facing text

PR #1711, opened 2024-04-17
Add keys/values/items method to modal.Dict

mwaskom:
Resolves MOD-2754

## Changelog

- The `modal.Dict` object now implements a `keys`/`values`/`items` API. Note that there are a few differences when compared to standard Python dicts:
    - The return value is a simple iterator, whereas Python uses a dictionary view object with more features.
    - The results are unordered.
 - Additionally, there was no key data stored for items added to a `modal.Dict` prior to this release, so empty strings will be returned for these entries.


mwaskom:
@prbot approve

PR #1723, opened 2024-04-19
Simplify Queue and Dict docs

erikbern:
Was going to replace `Stub` with `App` everywhere but for this one there's no point using a stub/app.

We test all docstrings and for these snippets to work we need to run against the cloud env. So I added Modal credentials to the docstring test.


erikbern:
@prbot approve

PR #1724, opened 2024-04-19
Remove grpc_utils.ChannelPool

ekzhang:
## Describe your changes

This removes grpc_utils.ChannelPool, a channel connection manager that appears to slow down certain RPCs by introducing extraneous handshakes, and use extra TCP connections.

This PR only affects the client from outside containers. It doesn't affect existing deployed functions or endpoints.

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -10,9 +10,7 @@ from typing import (
     Any,
     AsyncIterator,
     Dict,
-    List,
     Optional,
-    Type,
     TypeVar,
 )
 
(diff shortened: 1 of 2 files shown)


ekzhang:
@prbot approve

luiscape:
@ekzhang Isn't it a good thing to have clients close connections every few seconds? This would help with NGINX scale down events (for example, a new NGINX target is added, a client connects to it and has a long lived connection, NGINX is removed from the pool closing the connection). What would happen in that case now that we removed the pool?

ekzhang:
That's fair! The TGPS for Nginx controller pods is 90 seconds already though, so clients will just get a disconnection after that period.

We already do this for container proxy clients and internal services, so it's probably OK.

PR #1725, opened 2024-04-19
stub -> app in docstrings

erikbern:
This is the first place where we start using "app" instead of "stub" to the public.

We should fast-follow with equivalent changes to the examples and the online docs.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -155,7 +155,7 @@ def call_function_sync(
                 if inspect.iscoroutine(res) or inspect.isgenerator(res) or inspect.isasyncgen(res):
                     raise InvalidError(
                         f"Sync (non-generator) function return value of type {type(res)}."
-                        " You might need to use @stub.function(..., is_generator=True)."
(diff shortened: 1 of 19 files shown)


erikbern:
@prbot approve

modal/_container_entrypoint.py
```diff
@@ -155,7 +155,7 @@ def run_input(input_id: str, function_call_id: str, args: Any, kwargs: Any) -> N
                 if inspect.iscoroutine(res) or inspect.isgenerator(res) or inspect.isasyncgen(res):
                     raise InvalidError(
                         f"Sync (non-generator) function return value of type {type(res)}."
-                        " You might need to use @stub.function(..., is_generator=True)."
+                        " You might need to use @app.function(..., is_generator=True)."
```
mwaskom, line 158: Feels like we might want to reword this one to be more robust, e.g. "you might need to use `is_generator=True` in your function decorator"

mwaskom (commented):
:+1:

PR #1727, opened 2024-04-19
Re-enable ChannelPool optionally with an environment variable

ekzhang:
This may be useful for debugging network issues in the future, until we're sure we can remove the code.


erikbern:
Thanks! Hopefully we can remove the `ChannelPool` class unceremoniously in a few weeks/month if users don't report any issues.

ekzhang:
No reported issues!

PR #1728, opened 2024-04-20
Flip around app & stub in some tests

erikbern:
Instead of having a bunch of tests that use "stub" and test specifically that "app works", flip it around so that almost all tests use "app" and there's specific tests for the old "stub" name.

Just prepping for _deprecating_ "stub"


erikbern:
@prbot approve

PR #1733, opened 2024-04-20
`Stub` deprecation warning

erikbern:
This enables deprecation warnings for the usage of `Stub` classes, as well as deprecation warnings if the user has a `stub` symbol defined at the module level while running `modal run`.

My suggestion is that we hold off merging this for at least 1-2 weeks, just in case there's some broken code with the app stuff. Also to give people a bit more time to adjust.

Edit: tentatively targeting next Friday, i.e. 2024-04-26. Also made the warnings _pending_ just to be slightly more gentle. But let's raise to non-pending warnings pretty soon after.

# Changelog

* Add a pending deprecation warning for usage of `Stub` in favor of `App`
* Add a pending deprecation warning for CLI module imports using the `stub` variable name, in favor of `app`


ekzhang:
> we hold off merging this for at least 1-2 weeks

Should we wait until all of our examples, docs, and official external repos are updated?

PR #1734, opened 2024-04-20
handle `app` -> `stub` fallback during CLI imports

erikbern:
If a user had both `app` and `stub` present at the module level, and ran `modal run myscript.py`, and `stub` was the main app class, then #1728 would cause `modal run` to break. E.g. this code:

```
import modal

app = 234
stub = modal.Stub()

@stub.function()
def f():
    ...
```

It would cause this issue:

```
erikbern@Eriks-Air modal-client % modal run stub_app_thing.py 
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 234 is not a Modal entity (should be a App or Function)                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Now, after this code, it still works:

```
erikbern@Eriks-Air modal-client % modal run stub_app_thing.py                   
╭─ Modal Deprecation Warning (2024-04-20) ─────────────────────────────────────────────────────────────────╮
│ The symbol `app` is present at the module level but it's not a Modal app. We will use `stub` instead,    │
│ but this will not work in future Modal versions. Suggestion: change the name of `app` to something else. │
│                                                                                                          │
│ Source: /Users/erikbern/python3.10-env/lib/python3.10/site-packages/click/core.py:1729                   │
│   cmd = self.get_command(ctx, cmd_name)                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯
✓ Initialized. View run at https://modal.com/modal-labs/apps/ap-stbr8cgdypPapuhzKIuEP7
✓ Created objects.
├── 🔨 Created mount /Users/erikbern/modal-client/stub_app_thing.py
└── 🔨 Created f.
Stopping app - local entrypoint completed.
✓ App completed. View run at https://modal.com/modal-labs/apps/ap-stbr8cgdypPapuhzKIuEP7
```

### DIFF
--- a/modal/cli/import_refs.py
+++ b/modal/cli/import_refs.py
@@ -20,7 +20,7 @@ from rich.markdown import Markdown
 
 import modal
 from modal.app import App, LocalEntrypoint
-from modal.exception import _CliUserExecutionError
+from modal.exception import _CliUserExecutionError, deprecation_warning
 from modal.functions import Function
 
 
@@ -42,7 +42,6 @@ def parse_import_ref(object_ref: str) -> ImportRef:
 
 
 DEFAULT_APP_NAME = "app"
-POSSIBLE_APP_NAMES = ["app", "stub"]
 
 
 def import_file_or_module(file_or_module: str):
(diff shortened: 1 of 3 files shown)


erikbern:
@prbot approve

mwaskom:
I’d suggest using `show_source=False` in the deprecation warning helper here since it’s just getting attributed to something in the CLI framework.

PR #1737, opened 2024-04-22
Fix typo causing warning/error in modal.serve

freider:
Fixes issue causing:
```
    for entry in mount._entries:
TypeError: 'NoneType' object is not iterable
```
When using `modal serve` (typo in which `is_local()` method was used...)

Adds a test that ensures that only local mounts are used as watch entries


freider:
@prbot automerge

freider:
@prbot approve

freider:
@prbot automerge

mwaskom:
:+1: good catch though confusing to have these two function / method names that mean different things

PR #1739, opened 2024-04-22
Rename test/supports/stub.py

erikbern:
Update: turns out this file _was_ used, but let's rename it at least.


MINHHAI10289 (commented):
tks alot

PR #1740, opened 2024-04-22
Add CLI for interacting with modal.Dict object

mwaskom:
Resolves MOD-2418

## Changelog

- Added a command-line interface for interacting with `modal.Dict` objects. Run `modal dict --help` in your terminal to see what is available.

### DIFF
--- a/modal/cli/dict.py
+++ b/modal/cli/dict.py
@@ -6,6 +6,7 @@ from rich.console import Console
 from rich.json import JSON
 from typer import Argument, Typer
 
+from modal._resolver import Resolver
 from modal._utils.async_utils import synchronizer
 from modal._utils.grpc_utils import retry_transient_errors
 from modal.cli.utils import ENV_OPTION, display_table
@@ -17,8 +18,18 @@ from modal_proto import api_pb2
 dict_cli = Typer(
     name="dict",
     no_args_is_help=True,
-    help="Inspect the contents of modal.Dict objects",
+    help="Manage `modal.Dict` objects and inspect their contents.",
 )
+
+
+@dict_cli.command(name="create")
+@synchronizer.create_blocking
+async def create(name: str, *, env: Optional[str] = ENV_OPTION):
+    """Create a named Dict object."""
+    d = _Dict.from_name(name, environment_name=env, create_if_missing=True)
+    client = await _Client.from_env()
+    resolver = Resolver(client=client)
+    await resolver.load(d)
 
 
 @dict_cli.command(name="list")
@@ -31,53 +42,10 @@ async def list(*, json: bool = False, env: Optional[str] = ENV_OPTION):
     response = await retry_transient_errors(client.stub.DictList, request)
 
     def format_timestamp(t: float) -> str:
-        return datetime.strftime(datetime.fromtimestamp(t), "%Y-%m-%d %H:%M")
+        return datetime.strftime(datetime.fromtimestamp(t), "%Y-%m-%d %H:%M") + " UTC"
 
     rows = [(d.name, format_timestamp(d.created_at)) for d in response.dicts]
     display_table(["Name", "Created at"], rows, json)
-
-
-@dict_cli.command(name="show")
-@synchronizer.create_blocking
-async def show(
-    name: str,
-    n: int = Argument(default=None, help="Retrieve and show no more than this many entries"),
-    *,
-    json: bool = False,
-    env: Optional[str] = ENV_OPTION,
-):
-    """Print the contents of a Dict."""
-    # TODO alternate names: inspect, dump, items
-    # TODO add an n: int option? or alternately a `modal dict peek` command or similar
-    d = await _Dict.lookup(name, environment_name=env)
-    i, items = 0, []
-    async for item in d.items():
-        i += 1
-        items.append(item)
-        if n is not None and i >= n:
-            break
-    if json:
-        # Note, we don't use the json= option of display_table becuase we want to display
-        # the dict itself as a JSON, rather than have a JSON representation of the table.
-        console = Console()
-        console.print(JSON.from_data(dict(items)))
-    else:
-        display_table(["Key", "Value"], [[str(k), str(v)] for k, v in items])
-
-
-@dict_cli.command(name="get")
-@synchronizer.create_blocking
-async def get(name: str, key: str, *, env: Optional[str] = ENV_OPTION):
-    """Print the value for a specific key.
-
-    Note: When using the CLI, keys are always interpreted as having a string type.
-    """
-    # TODO would it be nice to be able to get multiple values? Should we do that here?
-    d = await _Dict.lookup(name, environment_name=env)
-    console = Console()
-    val = await d.get(key)
-    # TODO val will be `None` when key is not found
-    console.print(val)
 
 
 @dict_cli.command("clear")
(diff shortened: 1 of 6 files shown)


modal/cli/dict.py
```diff
@@ -0,0 +1,123 @@
...
+    name: str,
+    n: int = Argument(default=None, help="Retrieve and show no more than this many entries"),
+    *,
+    json: bool = False,
+    env: Optional[str] = ENV_OPTION,
+):
+    """Print the contents of a Dict.
+
+    Note: By default, this command will retrieve the complete Dict contents.
+    Using the N argument is recommended for quickly inspecting a small number of items.
```
mwaskom, line 100: I think it is not great that the default behavior could potentially grab the contents of a huge Dict. But also having a default non-null value for `N` feels like it could be confusing too. I have considered also adding a `modal dict peek` command that has such a default, which makes sense to me. Then people's muscle memory could point them towards the cheap operation. Open to adding if it sounds like a good idea to others.

modal/cli/dict.py
```diff
@@ -0,0 +1,123 @@
...
+    no_args_is_help=True,
+    help="Manage `modal.Dict` objects and inspect their contents.",
+)
+
+
+@dict_cli.command(name="create")
+@synchronizer.create_blocking
+async def create(name: str, *, env: Optional[str] = ENV_OPTION):
+    """Create a named Dict object."""
+    d = _Dict.from_name(name, environment_name=env, create_if_missing=True)
```
erikbern, line 30: super minor but this won't fail if the dict already exists which is probably fine

modal/cli/dict.py
erikbern: seems fine to have a default number for n if we print something like `... 584 more values not shown`

modal/cli/dict.py
```diff
@@ -0,0 +1,123 @@
...
+    d = await _Dict.lookup(name, environment_name=env)
+
+    try:
+        i, items = 0, []
+        async for item in d.items():
+            i += 1
+            items.append(item)
+            if n is not None and i >= n:
+                break
+    except ModuleNotFoundError as exc:
```
erikbern, line 111: maybe this except block should be in the `items()` method?

modal/cli/dict.py
```diff
@@ -0,0 +1,123 @@
...
+        # I think that on 3.10+ we could rewrite this to use anext and attribute errors to specific
+        # items (perhaps represent them in the output as "<library>_object" or something.)
+        msg = f"Dict contains objects from the `{exc.name}` library and cannot be deserialized locally."
+        raise ExecutionError(msg)
+
+    if json:
+        # Note, we don't use the json= option of display_table becuase we want to display
```
erikbern, line 118: *because

modal/cli/dict.py
mwaskom: Yeah curious if you think it would be better to error out when we try to `create` something that exists? Was trying to stay close to the behavior of the API (i.e. we don't do `create_if_missing_else_error`) but I could see the argument for it.

modal/cli/dict.py
mwaskom: We can print something that indicates truncation although currently our estimate of dict size is very wrong so I don't think we can say exactly how many :)

mwaskom: Also as written there is no way to invoke `modal dict show` in a way that *doesn't* truncate ... how would you spell that, `modal dict show -1`? That seems kind of weird.

modal/cli/dict.py
```diff
@@ -0,0 +1,123 @@
...
+    resolver = Resolver(client=client)
+    await resolver.load(d)
+
+
+@dict_cli.command(name="list")
+@synchronizer.create_blocking
+async def list(*, json: bool = False, env: Optional[str] = ENV_OPTION):
+    """List all named Dict objects."""
+    env = ensure_env(env)
+    client = await _Client.from_env()
```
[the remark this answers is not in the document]
mwaskom: Yeah that's a nice idea!

test/conftest.py
```diff
@@ -313,6 +313,13 @@ async def AppList(self, stream):
             apps.append(api_pb2.AppStats(name=app_name, description=app_name, app_id=app_id))
         await stream.send_message(api_pb2.AppListResponse(apps=apps))
 
+    async def AppStop(self, stream):
```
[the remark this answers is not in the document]
mwaskom: Currently `modal dict delete` is implemented by hitting the `AppStop` method — is there a better way to do that?

modal/cli/dict.py
erikbern: `modal dict show --unlimited` or something? idk

modal/cli/dict.py
erikbern: we have that in some other places but imo it's not worth the implementation cost in this case

modal/cli/dict.py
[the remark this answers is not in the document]
mwaskom: Think I might punt this idea from this PR so I can think a little bit more about the name and then roll out the same API to all objects at the same time.

test/conftest.py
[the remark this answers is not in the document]
erikbern: I see, that's not great because it ties the client to the "dummy app data model" we've been trying to get away from (it's a long journey because we need to wait for old clients to be fully deprecated before we can actually change the data model server-side).

Take a look at the `Volume` class which has a `VolumeDelete` endpoint

modal/cli/dict.py
mwaskom: I'll add a comment to the docstring to clarify at least.

modal/cli/dict.py
mwaskom: Went with `modal dict show --all` and a default size of `20`

modal/cli/dict.py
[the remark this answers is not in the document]
erikbern: Seems totally fine! Sorry should have been clear that I don't consider this a blocker at all

modal/cli/dict.py
erikbern: sounds good! We actually have some code here that handles a similar issue: https://github.com/modal-labs/modal-client/blob/main/modal/_utils/function_utils.py#L450

modal/cli/dict.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Note: When using the CLI, keys are always interpreted as having a string type.
+    """
+    d = await _Dict.lookup(name, environment_name=env)
+    console = Console()
+    val = await d.get(key)
+    console.print(val)
+
+
+@dict_cli.command(name="show")
```
freider, line 91: I saw some discussion about this somewhere else. I have a slight preference for "items" vs "show" fwiw

mwaskom: 👍

freider: Agree it's slightly weird, but the json representation will always be a bit weird, since it's basically only compatible with a subset of dicts anyway (the keys can only be strings, and the values need to be json encodable). Also wouldn't be opposed to making the json representation a list of tuples for consistency, although it's a bit messier to work with

mwaskom: Hm that is a good point about key types…

mwaskom:
@prbot approve

merging, we can iterate in some followups if there's need

PR #1742, opened 2024-04-22
Client refactor of volumes listdir() APIs and CLI

ekzhang:
## Describe your changes

Requires us to merge this server change first: https://github.com/modal-labs/modal/pull/12096

Note that although the changelog says that we no longer take `**` arguments, it still works while the backwards compatibility shim exists in the server. I added shims on the server-side, and we can remove them after 0.62 is no longer a supported client version. It's not a particularly widely-used feature.

## Changelog

- Passing a glob `**` argument to the `modal volume get` CLI has been deprecated — instead, simply download the desired directory path, or `/` for the entire volume.
- `Volume.listdir()` no longer takes trailing glob arguments. Use `recursive=True` instead.
- `modal volume get` and `modal nfs get` performance is improved when downloading a single file. They also now work with multiple files when outputting to stdout.
- Fixed a visual bug where `modal volume get` on a single file will incorrectly display the destination path.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1934,7 +1934,7 @@ message VolumeListFilesRequest {
   string volume_id = 1;
   string path = 2;
   bool recursive = 4;
-  optional int32 max_entries = 3;
+  optional uint32 max_entries = 3;
 }
 
 message VolumeListFilesResponse {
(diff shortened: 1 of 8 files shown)


modal_proto/api.proto
```diff
@@ -1920,7 +1920,7 @@ message VolumeListFilesRequest {
   string volume_id = 1;
   string path = 2;
   bool recursive = 4;
-  optional int32 max_entries = 3;
+  optional uint32 max_entries = 3;
```
thundergolfer, line 1923: Nice that proto handles compatibility on this.

thundergolfer (approved):

PR #1747, opened 2024-04-24
Clean up FunctionInfo

freider:
The public interface of FunctionInfo was super hard to understand since it sets a ton of members conditionally on construction. This cleans that up so it's easier to refactor (needed for common-pooling methods)

Also realized `get_auto_mounts()` isn't actually using `self` at all, so moved it out to modal.mount as a function instead.


freider:
@prbot approve

PR #1748, opened 2024-04-24
Adds test that modal can be imported from non-main thread

freider:
Reproduces regression from https://github.com/modal-labs/modal-client/pull/1682/files that a user reported


mwaskom (commented):
Nice catch!

PR #1751, opened 2024-04-24
Improve feedback for deserialization errors on environment mismatch

mwaskom:
Side-quest from the `modal dict` CLI work, adding a modal `DeserializationError` class and using it to provide more context for the two most common errors that arise from an environment mismatch.

## Changelog

- Improved feedback for deserialization failures when objects are being transferred between local / remote environments.


mwaskom:
@prbot approve

PR #1756, opened 2024-04-25
Show human-readable timezone codes in CLI tables

mwaskom:
Small change to truncate the creation timestamp at seconds and show a timezone code instead of UTC offset when we render a timestamp in a CLI table:

image

JSON output continues to use the full ISO format to remain machine-readable:

image

Additionally, I refactored a couple of places where we were not using the common utility function for converting timestamps.


mwaskom:
@prbot approve

PR #1761, opened 2024-04-27
fix: add args check to avoid `tuple index out of range` issue

chenrui333:
## Describe your changes

This is to add args check as it is empty for `modal volume delete --yes test`

The error I have seen:

```
$ modal volume delete --yes test2
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────╮
│ tuple index out of range                                                                            │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

## Changelog

- add args check when deleting the volume

---

I have tested with the local build against my modal volume


mwaskom:
Thanks for flagging! I fixed this in #1764; this surfaced the need from some additional tests

PR #1764, opened 2024-04-29
Fix Volume.delete with kwarg label

mwaskom:
Fixes a bug in the backwards-compatibility layer of `Volume.delete`, raised in #1761

Also adds a test for the `volume create` and `volume delete` CLI


mwaskom:
@prbot approve, new code is mostly tests

PR #1765, opened 2024-04-29
Simplify _asgi code with context

ekzhang:
Follow-up to #1533.

This wasn't possible to do before due to the warning messages. See the new docstring of TaskContext added in #1533 for current semantics.


ekzhang:
@prbot approve

PR #1766, opened 2024-04-29
Add TaskContext.gather, remove ConcurrencyPool / wait

ekzhang:
This adds the `TaskContext.gather()` method, a replacement for `asyncio.gather()` that only takes coroutines and cancels all tasks on an exception being raised on any task.

```python
    @staticmethod
    async def gather(*awaitables: Awaitable) -> Any:
        """Wait for a sequence of awaitables to finish, concurrently.

        This is similar to `asyncio.gather()`, but it uses TaskContext to cancel all remaining tasks
        if one fails with an exception other than `asyncio.CancelledError`. The native `asyncio`
        function does not cancel remaining tasks in this case, which can lead to surprises.

        For example, if you use `asyncio.gather(t1, t2, t3)` and t2 raises an exception, then t1 and
        t3 would continue running. With `TaskGroup.gather(t1, t2, t3)`, they are cancelled.

        (It's still acceptable to use `asyncio.gather()` if you don't need cancellation — for
        example, if you're just gathering quick coroutines with no side-effects. Or if you're
        gathering the tasks with `return_exceptions=True`.)

        Usage:

        ```python notest
        # Example 1: Await three coroutines
        created_object, other_work, new_plumbing = await TaskContext.gather(
            create_my_object(),
            do_some_other_work(),
            fix_plumbing(),
        )

        # Example 2: Gather a list of coroutines
        coros = [a.load() for a in objects]
        results = await TaskContext.gather(*coros)
        ```
        """
        async with TaskContext() as tc:
            results = await asyncio.gather(*(tc.create_task(a) for a in awaitables))
        return results
```

I also removed `TaskContext.wait`, which is unused and could just be replaced with asyncio.wait() on the tasks themselves, as well as `ConcurrencyPool`, which was only being used in one place after a while — made it consistent with other code using `aiostream` for now.

### DIFF
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -201,8 +201,8 @@ class TaskContext:
         return t
 
     @staticmethod
-    async def gather(*awaitables: Awaitable) -> Any:
-        """Wait for a sequence of awaitables to finish, concurrently.
+    async def gather(*coros: Awaitable) -> Any:
+        """Wait for a sequence of coroutines to finish, concurrently.
 
         This is similar to `asyncio.gather()`, but it uses TaskContext to cancel all remaining tasks
         if one fails with an exception other than `asyncio.CancelledError`. The native `asyncio`
@@ -231,7 +231,7 @@ class TaskContext:
         ```
         """
         async with TaskContext() as tc:
-            results = await asyncio.gather(*(tc.create_task(a) for a in awaitables))
+            results = await asyncio.gather(*(tc.create_task(coro) for coro in coros))
         return results
 
 
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -6,7 +6,7 @@ import inspect
 import time
 import typing
 from contextlib import asynccontextmanager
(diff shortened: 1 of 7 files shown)


modal/_utils/async_utils.py
```diff
@@ -200,29 +196,39 @@ async def loop_coro() -> None:
...
+    @staticmethod
+    async def gather(*coros: Awaitable) -> Any:
+        """Wait for a sequence of coroutines to finish, concurrently.
+
+        This is similar to `asyncio.gather()`, but it uses TaskContext to cancel all remaining tasks
+        if one fails with an exception other than `asyncio.CancelledError`. The native `asyncio`
+        function does not cancel remaining tasks in this case, which can lead to surprises.
+
+        For example, if you use `asyncio.gather(t1, t2, t3)` and t2 raises an exception, then t1 and
+        t3 would continue running. With `TaskGroup.gather(t1, t2, t3)`, they are cancelled.
```
freider, line 208: Small typo: TaskGroup.gather -> TaskContext.gather

freider (approved):
nice cleanup!

PR #1768, opened 2024-04-29
Image builder 2024.04 changes: drop to protobuf 4 and install uv

mwaskom:
Two updates to the 2024.04 image builder version before we announce it in "preview" status:

- Install `uv` (the latest version at time of image build) in base image layers so it's easier to use it for image creation — first class API still TK
- Drop the pinned protofbuf version to 4.x as we've had sporadic reports of issues with deprecated functionality in 5.x on some systems

Closes MOD-2820 and partly addresses (but doesn't fix) MOD-2812

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -916,7 +916,7 @@ class _Image(_Object, type_prefix="im"):
                 # .bashrc is explicitly sourced because RUN is a non-login shell and doesn't run bash.
                 "RUN . /root/.bashrc && conda activate base \\ ",
                 # Ensure that packaging tools are up to date and install client dependenices
-                f"&& python -m pip install --upgrade {'pip' if version == '2023.12' else 'pip wheel'} \\ ",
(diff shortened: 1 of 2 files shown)


ekzhang:
Great! Hopefully `uv` doesn't make any breaking changes to their main CLI 🙏

mwaskom:
Agreed that the early-status of `uv` is a good reason to be cautious about how much first-class integration we offer, but this feels like a reasonable compromise in terms of supporting it better without committing too much of our own internals to their potentially-still-moving CLI.

We actually do have a slightly weird situation where you'll tend to get a newer version of pip — and now wheel and uv too — when using some image constructors (`from_registry`, `from_dockerfile`) but not others where we publish global artifacts (`debian_slim`, `micromamba`). I guess that could bite someone at some point. But I have to expect that even if they go in a new direction with the interface (e.g. getting rid of the `pip` in `uv pip install`, it would be a gradual change).

PR #1772, opened 2024-04-29
Add modal queue CLI

mwaskom:
Adds most of what you'd want in a `modal queue` CLI. The only thing that's perhaps missing is a command to list the queue partitions and give you information about them (i.e., length, TTL, etc.) I think that would be nice but can't decide what a good name would be or what the API would look like, so I'll punt that until someone asks for it 🙃.

- Closes MOD-2810

## Changelog

- Added a command-line interface for interacting with `modal.Queue` objects. Run `modal queue --help` in your terminal to see what is available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ We appreciate your patience while we speedily work towards a stable release of t
 
 ### 0.62.116 (2024-04-26)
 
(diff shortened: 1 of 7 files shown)


mwaskom:
@prbot approve, discussed on slack

PR #1773, opened 2024-04-30
Run @enter methods when doing .local() calls on class instances

freider:
Previously only the deprecated `__enter__()` method was being run when a modal class instance was used for a `.local()` call, e.g.

`MyModalCls().my_method.local()`

Fixes MOD-2759


mwaskom (approved):
Nice, thanks for figuring this out!

PR #1775, opened 2024-04-30
Raise RequestSizeError when server returns 413 for Dict/Queue operations

mwaskom:
Should improve the user experience when requests (e.g. `Queue.put`, `Dict.update`) are rejected because the payloads are too large.

I think this is the main place we need to do this because other interfaces (Volumes, Function inputs) route through s3 on large payloads? Note that for Queue, payloads that are >256KiB will get rejected by server _logic_. There's no size limit on `Dict` values, but maybe there should be.

The error handling is a little bit janky because we just get a `Status.UNKNOWN` and need to introspect the error message. Maybe we can improve that somehow on the server side?

I considered putting the size limit (and perhaps measuring the request size in the except block) but that felt like an implementation detail that we didn't need here. Open to it, though!

It's untested because our mock server doesn't have a limit but I could add that if it feels important.

Closes MOD-2761

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2022
 from typing import Any, AsyncIterator, Optional, Tuple, Type
 
+from grpclib import GRPCError
 from synchronicity.async_wrap import asynccontextmanager
 
 from modal_proto import api_pb2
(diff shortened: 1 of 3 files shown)


mwaskom:
@prbot approve

PR #1776, opened 2024-04-30
Add doc comment to force_build: bool parameter

thundergolfer:
## Describe your changes

> Hey team, can we please have a feature to re-build modal containers with a `--no-cache` option. Something like `model [sic] deploy --no-cache script.py` ?

([source for the above in our Slack](https://modal-com.slack.com/archives/C069MLUT4SJ/p1714449602964049))

Don't love repeating this much, but want to at least have a way for users to find this in our docs. 

N/A Comment change only


mwaskom (approved):
:+1: makes sense

PR #1777, opened 2024-04-30
Deprecate invalid object names and standardize error message

mwaskom:
Fixes MOD-2831

## Changelog

* Added a deprecation warning when object names are invalid. This applies to `Dict`, `NetworkFileSystem`, `Secret`, `Queue`, and `Volume` objects. Names must be shorter than 64 characters and may contain only alphanumeric characters, dashes, periods, and underscores. These rules were previously enforced, but the check had inadvertently been dropped in a recent refactor.  Please update the names of your objects and transfer any data to retain access, as invalid names will become an error in a future release.


ekzhang (approved):
This looks good to me!

mwaskom:
Actually even though you can’t create secrets through the client it probably still makes sense to issue the deprecation warning for them to flag for people looking up the secret.

PR #1778, opened 2024-04-30
Improve the error message for unhydrated objects

mwaskom:
This is a persistent source of confusion. I am pretty sure that the main cause of this error (these days at least) is when users try to do `func.remote()` or similar in a "bare" Python script, or when they have multiple apps and are running a different one. I am fairly certain that we now catch "not in global scope" at function wrapping time, and storage objects can now be lazily hydrated.

I also don't think that mentioning "lazily hydrated" is helpful since that's an implementation detail and not something a user can understand or act on.

It might be helpful to add some Function-specific logic to include the function name in the error message, but I wasn't sure what the best way to approach that was, so I went for a smaller improvement.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -680,15 +680,18 @@ class _Function(_Object, type_prefix="fu"):
         """mdmd:hidden"""
 
         async def _load(self: _Function, resolver: Resolver, existing_object_id: Optional[str]):
-            function_name = self._parent.info.function_name
+            try:
+                identity = f"base {self._parent.info.function_name} function"
+            except Exception:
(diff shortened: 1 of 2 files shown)


erikbern:
def seems like an upgrade so 👍 

i think we could probably infer from the environment what's going on even more dynamically. e.g. if it's a function on a non-running app, we could say that.

erikbern (approved):

mwaskom:
OK, a little sketchy, but I added some logic to try and get the function name and whether the app is running in some places. Not totally obvious how to do that so could maybe be improved further.

PR #1784, opened 2024-05-01
Add command groups to modal CLI

mwaskom:
Makes the CLI help a little easier to understand.

The entrypoint now organizes the things you can do better:

image

And for some commands with different kinds of subcommands, there's less opportunity for confusion:

image

Happy to iterate on any of the specific groupings or names

### DIFF
--- a/modal/cli/dict.py
+++ b/modal/cli/dict.py
@@ -21,7 +21,7 @@ dict_cli = Typer(
 )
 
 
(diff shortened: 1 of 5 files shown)


mwaskom:
@prbot approve, but happy to keep iterating on the names and groupings

ekzhang:
👍

PR #1785, opened 2024-05-02
[mod 2845] task secret is expired and failing requests in restored container

thundergolfer:
## Describe your changes

**Reproduction**

```python
import asyncio
import time
import modal

from fastapi import FastAPI
from fastapi.responses import StreamingResponse
from modal import Stub, asgi_app

stub = Stub("repro")
web_app = FastAPI()

async def fake_video_streamer():
    for i in range(10):
        yield f"frame {i}: hello world!".encode()
        await asyncio.sleep(1.0)


@stub.cls(
    enable_memory_snapshot=True,
    container_idle_timeout=3,
)
class ScoreChunk:
    @modal.enter(snap=True)
    async def load_engine(self):
        print("entering (snap=True)")

    @modal.enter(snap=False)
    async def load_models(self):
        print("entering (snap=False)")

    @modal.method()
    async def score_chunk(self):
        print("Done")

@web_app.get("/")
async def main():
    _ = await ScoreChunk().score_chunk.remote.aio()
    return StreamingResponse(
        fake_video_streamer(), media_type="text/event-stream"
    )

@stub.function(enable_memory_snapshot=True, container_idle_timeout=4, max_inputs=1, timeout=10000)
@asgi_app()
def fastapi_app():
    return web_app
```

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -580,7 +580,7 @@ class _ContainerIOManager:
         )
 
         self._waiting_for_checkpoint = True
-        await self._client._close()
+        await self._client._close(forget_credentials=True)
 
         logger.debug("Checkpointing request sent. Connection closed.")
         await self.restore()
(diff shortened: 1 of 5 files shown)


luiscape (approved):
LGTM -- thanks for explaining it to me via call.

PR #1787, opened 2024-05-02
refactor: 'checkpoint' to 'memory snapshot' renaming

thundergolfer:
## Describe your changes

One more step on the road to renaming _checkpoint_ to _memory snapshot_.

    - Should be pure refactor. None of these name/symbol changes cross the boundary between the client and the worker/server/host.
    - Should be pure refactor. None of these name/symbol changes cross the boundary between the client and the worker/server/host.


mwaskom (approved):
:+1:

PR #1790, opened 2024-05-03
Add proto field to support a global image failsafe

mwaskom:
Part of MOD-2866

~- [ ] Client forward compatibility: this change ensures client can accept data intended for later versions of itself~

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1285,6 +1285,8 @@ message ImageGetOrCreateRequest {
   bool force_build = 7;
   DeploymentNamespace namespace = 8;
   string builder_version = 9;
+  // Only admins can publish global images, but this provides an extra failsafe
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1285,6 +1285,8 @@ message ImageGetOrCreateRequest {
   bool force_build = 7;
   DeploymentNamespace namespace = 8;
   string builder_version = 9;
+  // Only admins can publish global images, but this provides an extra failsafe
+  bool global_deployment_allowed = 10;
```
aksh-at, line 1289: Wdyt of `allow_global_deployment`?

mwaskom: No strong feelings either way! Can go with that

mwaskom:
@prbot automerge

PR #1795, opened 2024-05-06
Remove SharedVolume

erikbern:
Has been deprecated for 10 months now! Seems fine to remove.


mwaskom:
Can we also drop `shared_volumes` from `Image.run_function`? Or are you worried about breaking positional argument lists?

freider (approved):

erikbern:
> Can we also drop `shared_volumes` from `Image.run_function`? Or are you worried about breaking positional argument lists?

looks like we're silently ignoring this one actually, so it seems better to remove it

PR #1796, opened 2024-05-06
More docs for .new deprecation warning

erikbern:
Slightly more docs to make it clear that `.new` -> `.ephemeral` is not a straightforward substitution

### DIFF
--- a/modal/queue.py
+++ b/modal/queue.py
@@ -87,7 +87,9 @@ class _Queue(_Object, type_prefix="qu"):
     def new():
         """`Queue.new` is deprecated.
 
-        Please use `Queue.from_name` (for persisted) or `Queue.ephemeral` (for ephemeral) queues.
+        You can switch this out for `Queue.from_name("some-queue-identifier")` for persisted dicts.
+        You can also take a look at the documentation for `Queue.ephemeral` for another option,
+        although it might require more code changes if you're coming from `Queue.new`.
         """
         deprecation_warning((2024, 3, 19), Queue.new.__doc__)
 
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -139,7 +139,9 @@ class _Volume(_Object, type_prefix="vo"):
     def new() -> "_Volume":
         """`Volume.new` is deprecated.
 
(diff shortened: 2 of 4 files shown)


modal/queue.py
```diff
@@ -87,7 +87,9 @@ class _Queue(_Object, type_prefix="qu"):
     def new():
         """`Queue.new` is deprecated.
 
-        Please use `Queue.from_name` (for persisted) or `Queue.ephemeral` (for ephemeral) queues.
+        You can switch this out for `Queue.from_name("some-queue-identifier")` for persisted dicts.
```
mwaskom, line 90:
```suggestion
        You can switch this out for `Queue.from_name("some-queue-identifier")` for persisted queues.
```

modal/volume.py
```diff
@@ -139,7 +139,9 @@ def _initialize_from_empty(self):
     def new() -> "_Volume":
         """`Volume.new` is deprecated.
 
-        Please use `Volume.from_name` (for persisted) or `Volume.ephemeral` (for ephemeral) volumes.
+        You can switch this out for `Volume.from_name("some-volume-identifier")` for persisted dicts.
```
mwaskom, line 142:
```suggestion
        You can switch this out for `Volume.from_name("some-volume-identifier")` for persisted volumes.
```

mwaskom (commented):
See two copy-pasta nits.

IMO this may still be a little confusing since `new` -> `from_name` is a nontrivial behavioral change and i'm not sure that "persisted" is a magic-enough word to communicate that.

modal/volume.py
```diff
@@ -139,7 +139,9 @@ def _initialize_from_empty(self):
...
+        You can also take a look at the documentation for `Volume.ephemeral` for another option,
+        although it might require more code changes if you're coming from `Volume.new`.
```
aksh-at, line 144: I had a hard time finding which docs this refers to. Is it https://modal.com/docs/reference/modal.Volume#ephemeral? Ideally we should add a guide page on "ephemeral objects" and link to that here directly

mwaskom:
Should we close this now that it's an error?

PR #1797, opened 2024-05-06
Reject lambdas and non-callable objects from Image.run_function

mwaskom:
Closes MOD-2874

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1450,6 +1450,12 @@ class _Image(_Object, type_prefix="im"):
         """
         from .functions import _Function
 
+        if not callable(raw_f):
+            raise InvalidError(f"Argument to Image.run_function must be a function, not {type(raw_f).__name__}.")
+        elif raw_f.__name__ == "<lambda>":
+            # It may be possible to support lambdas eventually, but for now we don't handle them well, so reject quickly
+            raise InvalidError("Image.run_function does not support lambda functions.")
+
         info = FunctionInfo(raw_f)
 
         if shared_volumes or network_file_systems:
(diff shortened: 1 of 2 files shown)


mwaskom:
cc @charlesfrye

modal/image.py
```diff
@@ -1450,6 +1450,12 @@ def my_build_function():
         """
         from .functions import _Function
 
+        if not callable(raw_f):
+            raise InvalidError(f"Argument to Image.run_function must be a function, not {type(raw_f).__name__}.")
+        elif raw_f.__name__ == "<lambda>":
```
erikbern, line 1455: i think it's actually more broader than this – any function that isn't defined in the global scope will fail. You can check whether the name contains `<locals>`

feels like there should be some utility for this but I haven't seen one

There's https://github.com/modal-labs/modal-client/blob/main/modal/_utils/function_utils.py#L52 but that one otoh doesn't look for lambdas

mwaskom: We catch functions defined in non-global scope here: https://github.com/modal-labs/modal-client/blob/b6ba0ef2621a906efb03f5c8b447997dcac93da7/modal/_utils/function_utils.py#L99-L106

mwaskom: Prior to this change we actually package the lambda perfectly fine, but it fails container-side in `import_function`.

erikbern: ok maybe the lambda check belongs in the same place then?

mwaskom: Hm yeah maybe, I guess it seemed like you could only really be using a lambda in `Image.run_function` since the decorator pattern otherwise wouldn't work, but maybe someone would try to do

```
f = lambda x: x
app.function(f)
```

?

PR #1800, opened 2024-05-07
Validate passed in App name is a string

devennavani:
## Describe your changes

Validate that the passed in value for `name` during App (formerly Stub) instantiation is a `str`.

- MOD-2639

## Changelog

Added validation that App `name`, if provided, is a string.


mwaskom (approved):
Nice; same comment regarding changelog as other PR (also we don't always add changelog entries for small things like this, but there's no hard or fast rule or problem with doing it).

PR #1801, opened 2024-05-07
Catch when users decorate class with @app.function()

devennavani:
## Describe your changes

Throw an error when users decorate a class with @stub.function()

- MOD-2261

## Changelog

The @app.function decorator now raises an error when it is used to decorate a class.

### DIFF
--- a/test/app_test.py
+++ b/test/app_test.py
@@ -279,7 +279,7 @@ def test_function_decorator_on_class():
     app = App()
     with pytest.raises(TypeError):
 
-        @app.function
+        @app.function()
         class Foo:
             pass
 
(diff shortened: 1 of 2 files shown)


test/app_test.py
```diff
@@ -275,6 +275,17 @@ def f():
...
 
+def test_function_decorator_on_class():
+    app = App()
+    with pytest.raises(TypeError) as excinfo:
+
+        @app.function()
+        class Foo:
+            pass
+
+    assert "cannot be used on a class" in str(excinfo.value)
```
mwaskom, line 286: Note you can also use the `match=` parameter of `pytest.raises` to check the message.

mwaskom (approved):
Nice, thanks for adding a test too!

One nit re: changelog formatting: we typically use bullets and usually phrase things either in past tense ("Added ...") or in describing the new behavior ("The `App.function` decorated now raises...") where past tense is awkward to describe the development change. You can see some examples here: https://modal.com/docs/reference/changelog

PR #1802, opened 2024-05-07
Improve modal app list output

mwaskom:
Closes MOD-2881

## Changelog

- The `modal app list` output has been improved in several ways:
    - Persistent storage objects like Volumes or Dicts are no longer included (these objects receive an app ID internally, but this is an implementation detail and subject to future change). You can use the dedicated CLI for each object (e.g. `modal volume list`) instead.
    - For Apps in a *stopped* state, the output is now limited to those stopped within the past 2 hours.
    - The number of tasks running for each App is now shown.

### DIFF
--- a/modal/cli/app.py
+++ b/modal/cli/app.py
@@ -1,5 +1,6 @@
 # Copyright Modal Labs 2022
 import asyncio
+import time
 from typing import List, Optional, Union
 
 import typer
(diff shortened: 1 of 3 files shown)


modal/cli/app.py
```diff
@@ -31,20 +32,36 @@
...
     for app_stats in apps:
-        state = APP_STATE_TO_MESSAGE.get(app_stats.state, Text("unknown", style="gray"))
+        if (
+            # Previously, all deployed objects (Dicts, Volumes, etc.) created an entry in the App table.
+            # We are waiting to roll off support for old clients before we can clean up the database.
+            # Until then, we filter deployed "single-object apps" from this output based on the object entity.
+            (app_stats.object_entity and app_stats.object_entity != "ap")
+            # AppList always returns up to the 250 most-recently stopped apps, which is a lot for the CLI
+            # (it is also used in the web interface, where apps are organized by tabs and paginated).
+            # So we semi-arbitrarily limit the stopped apps to those stopped within the past 2 hours.
```
ekzhang, line 50: Feel free to add the returned number of recently-stopped apps as a parameter to AppListRequest if you'd like!

PR #1804, opened 2024-05-08
Refactor container entrypoint to use same handler for both sync and sync functions

freider:
This merges the old `call_function_sync` and `call_function_async` input loop functions into a single one that sets up cancellation handling *per input* instead of doing it for the duration of the task.

This will allow us to mix async and sync function executions within the same task, which we need to allow pooling executions of all methods in Modal classes.

I'm sure we can clean this up additionally in the future, but I'm reasonably sure this should work well given our decent test coverage...

~~The most notable thing that might break things for users here is that sync code now technically *also* runs within the main thread event loop, so users might not be able to start new event loops (in the same thread), but I hope that such a use case would be very uncommon given that Modal supports running async functions directly~~

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -62,6 +62,9 @@ class ImportedFunction:
 
 
 class DaemonizedThreadPool:
+    # Used instead of ThreadPoolExecutor, since the latter won't allow
+    # the interpreter to shut down before the currently running tasks
+    # have finished
     def __init__(self, max_threads):
         self.max_threads = max_threads
 
@@ -258,13 +261,12 @@ def call_function(
         reset_context()
 
     if imp_fun.input_concurrency > 1:
-        # all run_input coroutines will have completed by the time we leave the execution context
-        # but the wrapping *tasks* may not yet have been resolved, so we add a 0.01s
-        # for them to resolve gracefully:
-
         with DaemonizedThreadPool(max_threads=imp_fun.input_concurrency) as thread_pool:
 
             async def run_concurrent_inputs():
+                # all run_input coroutines will have completed by the time we leave the execution context
+                # but the wrapping *tasks* may not yet have been resolved, so we add a 0.01s
+                # for them to resolve gracefully:
                 async with TaskContext(0.01) as task_context:
                     async for input_id, function_call_id, args, kwargs in container_io_manager.run_inputs_outputs.aio(
                         imp_fun.input_concurrency
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -1,14 +1,13 @@
 # Copyright Modal Labs 2022
 import asyncio
 import base64
-import datetime
 import importlib
 import inspect
+import queue
 import signal
 import sys
 import threading
 import time
-from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Callable, List, Optional, Sequence, Type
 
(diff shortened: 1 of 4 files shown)


modal/_container_entrypoint.py
```diff
@@ -60,6 +61,50 @@ class ImportedFunction:
     function: _Function
 
 
+class DaemonizedThreadPool:
```
freider, line 64: This isn't exactly new code, I just moved it out of the old `run_function_sync` implementation into it's own context manager class

ekzhang: Thanks for splitting it out. But on a reading I think it might be a bit confusing this way because the exception-handling interface between DaemonizedThreadPool and its caller is tricky.

Usually splitting code up into classes helps with modularity. But in this case, the KeyboardInterrupt / BaseException-handling comments don't quite make sense anymore when reading the class itself, you need to go back into the code where it is used. Compare to the previous implementation where the error handling is perhaps easier to follow / it's also a pretty simple few lines of code?

```python
if imp_fun.input_concurrency > 1:
    # We can't use `concurrent.futures.ThreadPoolExecutor` here because in Python 3.11+, this
    # class has no workaround that allows us to exit the Python interpreter process without
    # waiting for the worker threads to finish. We need this behavior on SIGINT.

    import queue
    import threading

    spawned_workers = 0
    inputs: queue.Queue[Any] = queue.Queue()
    finished = threading.Event()

    def worker_thread():
        while not finished.is_set():
            try:
                args = inputs.get(timeout=1)
            except queue.Empty:
                continue
            try:
                run_input(*args)
            except BaseException:
                # This should basically never happen, since only KeyboardInterrupt is the only error that can
                # bubble out of from handle_input_exception and those wouldn't be raised outside the main thread
                pass
            inputs.task_done()

    for input_id, function_call_id, args, kwargs in container_io_manager.run_inputs_outputs(
        imp_fun.input_concurrency
    ):
        if spawned_workers < imp_fun.input_concurrency:
            threading.Thread(target=worker_thread, daemon=True).start()
            spawned_workers += 1
        inputs.put((input_id, function_call_id, args, kwargs))

    finished.set()
    inputs.join()
```

modal/_container_entrypoint.py
```diff
@@ -122,102 +171,12 @@ def _sigint_handler():
...
-
-
-async def call_function_async(
+def call_function(
+    user_code_event_loop: UserCodeEventLoop,
     container_io_manager,  #: ContainerIOManager,  TODO: this type is generated at runtime
     imp_fun: ImportedFunction,
 ):
-    async def run_input(input_id: str, function_call_id: str, args: Any, kwargs: Any) -> None:
+    async def run_input_async(input_id: str, function_call_id: str, args: Any, kwargs: Any) -> None:
```
ekzhang, line 179: Do you think we could move out `run_input_async` and `run_input_sync` into their own top-level functions now that they've gotten bigger? They could take `container_io_manager` and `imp_fun` as parameters.

ekzhang (approved):
Sounds good to me about supporting both sync and async functions! I have a couple minor asks about structure.

modal/_container_entrypoint.py
```diff
@@ -261,24 +220,87 @@ async def run_input(input_id: str, function_call_id: str, args: Any, kwargs: Any
...
+                        # released all its slots so that they can be acquired by the run_inputs_outputs finalizer
+                        # This prevents leaving the task_context before outputs have been created
+                        # TODO: refactor to make this a bit more easy to follow?
+                        if imp_fun.is_async:
+                            task_context.create_task(run_input_async(input_id, function_call_id, args, kwargs))
+                        else:
+                            # run sync input in thread
+                            thread_pool.submit(run_input_sync, input_id, function_call_id, args, kwargs)
+
+            user_code_event_loop.run(run_concurrent_inputs())
```
ekzhang, line 284: Also, I saw that you refactored it to pass around user_code_event_loop, nice!

modal/_container_entrypoint.py
freider: Yeah that makes sense

modal/_container_entrypoint.py
freider: Good point about the usage-coupled exception handling.

I thought it was a bit too messy to keep the implementation inlined in the input loop now that there is additional branching in there (and the threadpool is only used for a subset of iterations), and I like that the threadpool object becomes a sort of sync equivalent of `task_context.create_task()` in the loop.

I think I'll just refactor the DaemonizedThreadPool to be more akin to a "SyncTaskCancellationContext", and get rid of the context specific errors in favor of more general ones. I also think we can just skip the `inputs.join()` call for all exceptions (similar to the old implementation).

PR #1805, opened 2024-05-08
Add AWS_REGION in code sample for S3 bucket

yirenlu92:
- Add AWS_REGION in code sample for S3 bucket


mwaskom:
What effect does this have?

thundergolfer:
Agree with @mwaskom's question. What's the region of (the bucket) and how do you find the region for the bucket. The other two key-values are more self-explanatory, but arguably could also get documented.

luiscape:
Mountpoint supports `AWS_REGION` [as a parameter](https://github.com/awslabs/mountpoint-s3/blob/bd9b1af33128f235b26bbeaee5c586d22de748e2/doc/CONFIGURATION.md#region-detection). The main use-case seems to be resolving issues that may happen with "cross-region mount with a directory bucket", among other things.

Seems fair to add it as an optional env var.

mwaskom:
This was just a genuine question on my end; it’s not clear from the example what this is supposed to accomplish. (Are you setting the region you are connecting from? The region you want to mount in? Etc.)

mwaskom:
Thanks for adding the comments, makes a lot more sense.

So that they remain readable, can we move the comments to a line above the parameter and wrap them at a reasonable character width?

mwaskom (approved):
Thanks for bearing with me :)

PR #1806, opened 2024-05-09
Lexically split entrypoint command for modal shell

irfansharif:
Something like this didn't previously work:
```
  $ modal shell --cmd 'bash -c bash'
  Failed to run command: No such file or directory
```
Because we were translating it to `CMD ["bash -c bash"]` underneath, which tries to execute a binary literally named "bash -c bash". The right thing to do is to translate it to `CMD ["bash", "-c", "bash"]`, which we now do.

Basically now you can pass in arbitrarily sophisticated commands through --cmd, like:
```
  $ modal shell ollama-modal.py \
    --cmd 'bash -c "systemctl start ollama; ollama run llama3:instruct"'
```
(Maybe there's appetite for everything being passed through 'bash -c' by default?)

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -3,6 +3,7 @@ import asyncio
 import functools
 import inspect
 import re
+import shlex
 import sys
 import time
 from functools import partial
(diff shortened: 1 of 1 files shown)


irfansharif:
@prbot automerge

PR #1810, opened 2024-05-09
Allow `modal app logs` to look up apps by name

mwaskom:
Currently, we support only ID-based lookup in `modal app logs`, which means that users have to look up the ID for the currently deployed version of a given app whenever they want to check its logs.

This PR adds support for name-based lookup, which is more convenient (names are stable across re-deploys and much easier to remember).

Note that name lookup is supported only for apps in a "deployed" state, as the name would otherwise potentially map to multiple App IDs.

Closes MOD-2905

## Changelog

- Added support for looking up a deployed App by its deployment name in `modal app logs`

### DIFF
--- a/modal/cli/app.py
+++ b/modal/cli/app.py
@@ -8,6 +8,7 @@ from click import UsageError
 from grpclib import GRPCError, Status
 from rich.table import Column
 from rich.text import Text
+from typer import Argument, Option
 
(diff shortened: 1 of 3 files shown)


mwaskom:
@prbot lgtm

thundergolfer (commented):
LGTM

PR #1811, opened 2024-05-09
Michael/app lookup proto

mwaskom:
Supports getting the ID for a deployed app with a given name.


mwaskom:
Oops accidentally branched from the tip of a larger feature branch

PR #1813, opened 2024-05-09
[MOD-2766] Wrote function to check/validate environment names

mattnappo:
Addresses MOD-2766 by writing a function to check/verify environment names.

### DIFF
--- a/test/utils_test.py
+++ b/test/utils_test.py
@@ -36,9 +36,12 @@ def test_object_name():
 
 def test_environment_name():
     assert is_valid_environment_name("AB13/root:env1")
+    assert is_valid_object_name("a" * 64)
+    assert not is_valid_object_name("a" * 65)
     assert not is_valid_environment_name("--help")
     assert not is_valid_environment_name(":env")
     assert not is_valid_environment_name("/env")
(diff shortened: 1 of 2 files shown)


mwaskom:
Nice! Would be great to have a couple simple tests; the other code here is tested (somewhat non-obviously) in `test/utils_test.py`

test/utils_test.py
```diff
@@ -29,6 +34,13 @@ def test_object_name():
         check_object_name("foo/bar", "Volume", warn=True)
 
 
+def test_environment_name():
+    assert is_valid_environment_name("AB13/root:env1")
+    assert not is_valid_environment_name("--help")
+    assert not is_valid_environment_name(":env")
+    assert not is_valid_environment_name("/env")
```
thundergolfer, line 41:
```suggestion
    assert not is_valid_environment_name("/env")
    assert not is_valid_environment_name("")
```

thundergolfer (approved):
Nice 👌

test/utils_test.py
mwaskom: Check also exercise the "not too long" rule with something like `"a" * 65`

PR #1814, opened 2024-05-10
A few opportunistic improvements to type annotations

mwaskom:
- Adding a couple of type annotations to give better IDE information
- Fixing a couple of minor annotation issues
- Updating the `inv type-stubs` task to remove all existing `.pyi` files before running

### DIFF
--- a/modal/object.py
+++ b/modal/object.py
@@ -1,7 +1,7 @@
 # Copyright Modal Labs 2022
 import uuid
 from functools import wraps
-from typing import Awaitable, Callable, ClassVar, Dict, Hashable, List, Optional, Type, TypeVar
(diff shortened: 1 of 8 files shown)


modal/object.py
```diff
@@ -27,7 +28,7 @@ def _get_environment_name(environment_name: Optional[str], resolver: Optional[Re
         return config.get("environment")
 
 
-class _Object:
+class _Object(Generic[O]):
```
mwaskom, line 31: I am not totally sure that inheriting from `Generic` here makes sense, although it did silence the typing error we were otherwise getting here. I say that because the `O` TypeVar is used to represent specific subclasses, rather than a different type of object that this class can act on. What do you think @freider?

PR #1815, opened 2024-05-10
Add `--stream-logs` flag to `modal deploy`

devennavani:
## Describe your changes

Added a `--stream-logs` flag to `modal deploy` that if True, begins streaming the app logs once deployment is complete.

- MOD-2906

## Testing

`pytest test/cli_test.py::test_logs`

## Changelog

Added a `--stream-logs` flag to `modal deploy` that if True, begins streaming the app logs once deployment is complete.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -271,7 +271,7 @@ def deploy(
         False, help="[beta] Publicize the deployment so other workspaces can lookup the function."
     ),
     skip_confirm: bool = typer.Option(False, help="Skip public app confirmation dialog."),
-    stream_logs: bool = typer.Option(False, help="Stream logs from the deployment."),
+    stream_logs: bool = typer.Option(False, help="Stream logs from the app upon deployment."),
 ):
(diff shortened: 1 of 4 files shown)


modal/cli/run.py
```diff
@@ -290,6 +291,11 @@ def deploy(
 
     deploy_app(app, name=name, environment_name=env, public=public)
 
+    if stream_logs:
+        from modal.cli.app import logs
+
+        logs("", name=name, env=env)
+
```
mwaskom, line 298: Because `deploy_app` returns an object with the `app_id`, I think we can avoid coupling the CLI endpoints (and the name-to-ID RPC in `logs`) by using the same underlying helper function as `logs()` directly:

```
       try:
            with output_mgr.show_status_spinner():
                await get_app_logs_loop(app_id, client, output_mgr)
        except asyncio.CancelledError:
            pass
```

mwaskom (approved):
Great!

PR #1817, opened 2024-05-12
Support `--mount`, `--volume` in `modal launch {vscode,jupyter}`

irfansharif:
Finding this very handy when playing around with random notebooks just downloaded off of the internet/colab, and/or stashing state away into modal volumes for later use.

### DIFF
--- a/modal/cli/launch.py
+++ b/modal/cli/launch.py
@@ -25,7 +25,7 @@ launch_cli = Typer(
 
 
 def _launch_program(name: str, filename: str, args: Dict[str, Any]) -> None:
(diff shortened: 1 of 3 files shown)


irfansharif:
@prbot approve

mwaskom:
I think that it would be better to allow multiple volume / mounts. The hardcoding of the remote path sort of precludes that. Would it make more sense to mount them by default at `/root/lab/{volume_name}` or `/root/lab/volumes/{volume_name}`?

mwaskom:
I also wonder whether people might want control over the mountpoint, maybe through a spelling like `--volume my-vol-name:/root/mount/point`? Not sure what is most intuitive there.

PR #1820, opened 2024-05-13
remove `shared_volumes` argument from `run_function`

erikbern:
@mwaskom pointed this out in https://github.com/modal-labs/modal-client/pull/1795#issuecomment-2096896943

we don't actually use this argument so it seems worse to keep it than to just remove it (ideally we would have used deprecation warnings/errors in the past though)


mwaskom:
Since we have `kwargs` in the signature here, you could pull it out of there and warn or raise.

mwaskom:
Wait nevermind sorry, that's a dict parameter, not `**kwargs`

PR #1821, opened 2024-05-13
Add check on names for objects created through CLI

mwaskom:
Follow-on to #1777, which missed an alternate way of creating objects

Fixes MOD-2930


mwaskom:
@prbot automerge

PR #1822, opened 2024-05-13
fix: regression in nfs get

thundergolfer:
## Describe your changes

https://github.com/modal-labs/modal-client/pull/1742 caused a regression in `nfs get`. It can be repro'd with: 

```bash
MODAL_PROFILE=modal_labs MODAL_LOGLEVEL=debug modal nfs get txt-to-pokemon-cache-vol pokemon_names/rnn.txt
```

User reported this here https://modal-com.slack.com/archives/C069RAH7X4M/p1715620466703749. 

This wasn't caught by tests because our fake implementation was wrong. Fixed that. 

### DIFF
--- a/modal/cli/network_file_system.py
+++ b/modal/cli/network_file_system.py
@@ -1,7 +1,10 @@
 # Copyright Modal Labs 2022
 import os
+import shutil
 import sys
+from contextlib import contextmanager
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 from typing import Optional
 
 import typer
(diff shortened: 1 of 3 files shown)


modal/cli/network_file_system.py
```diff
@@ -188,7 +193,41 @@ async def get(
...
+
+    @contextmanager
+    def _destination_stream():
+        if destination == PIPE_PATH:
+            yield sys.stdout.buffer
+        else:
+            with NamedTemporaryFile(delete=False) as fp:
+                yield fp
+            shutil.move(fp.name, destination)
+
```
ekzhang, line 218: Can we fix `_volume_download` instead of introducing deviant behavior based on the path containing an asterisk?

ekzhang (changes requested):

PR #1824, opened 2024-05-14
Fix sigint concurrent inputs issues

freider:
Two pretty hard-to-find bugs were present:
* We had a finally clause in the multi-input handling code that tries to acquire all slots of a semaphore, assuming they have been released by each input, but that won't be the case since we can't let a sync worker thread know that the input has been cancelled. This lead to a stalled coroutine, and in a previous version of the client (prior to the unification of sync/async inputs) it even prevented the container_entrypoint from shutting down altogether since this blocked the synchronicity loop from shutting down.
* Our BaseException catch-all for handle_input_exception caught GeneratorExit and emitted that as an input error. This would trigger when the event loop is shut down prematurely like on sigint, since that will shut down all async generators. This *mostly* led to confusing exception spam to users

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -392,7 +392,6 @@ class _ContainerIOManager:
         # - or, when the output for the fetched input is sent, release the semaphore.
         self._input_concurrency = input_concurrency
         self._semaphore = asyncio.Semaphore(input_concurrency)
-
         try:
             async for input_id, function_call_id, input_pb in self._generate_inputs():
                 args, kwargs = self.deserialize(input_pb.args) if input_pb.args else ((), {})
@@ -486,7 +485,10 @@ class _ContainerIOManager:
         """Handle an exception while processing a function input."""
         try:
             yield
-        except KeyboardInterrupt:
+        except KeyboardInterrupt:  # , GeneratorExit):
+            # We need to explicitly reraise these BaseExceptions to not handle them in the catch-all:
+            # 1. KeyboardInterrupt can end up here even though this runs on non-main thread, since the code block yielded to could be sending back a main thread exception
+            # 2. GeneratorExit - raised if this (async) generator is garbage collected while waiting for the yield. Typically on event loop shutdown
             raise
         except (InputCancellation, asyncio.CancelledError):
             # just skip creating any output for this input and keep going with the next instead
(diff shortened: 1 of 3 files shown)


modal/_container_io_manager.py
```diff
@@ -393,16 +393,15 @@ async def run_inputs_outputs(self, input_concurrency: int = 1) -> AsyncIterator[
...
-            for _ in range(input_concurrency):
-                await self._semaphore.acquire()
+        async for input_id, function_call_id, input_pb in self._generate_inputs():
+            args, kwargs = self.deserialize(input_pb.args) if input_pb.args else ((), {})
+            self.current_input_id, self.current_input_started_at = (input_id, time.time())
+            yield input_id, function_call_id, args, kwargs
+            self.current_input_id, self.current_input_started_at = (None, None)
+
+        # collect all active input slots, meaning all inputs have wrapped up.
+        for _ in range(input_concurrency):
```
freider, line 403: Fix 1: Just remove the finally clause and let exceptions from `_generate_inputs` just bubble up. I don't think this should have any crazy effects since we probably want to exit here anyway and it shouldn't matter/should be intentional if some ongoing inputs are dropped?

modal/_container_io_manager.py
```diff
@@ -486,7 +485,10 @@ async def handle_input_exception(self, input_id, started_at: float) -> AsyncGene
         """Handle an exception while processing a function input."""
         try:
             yield
-        except KeyboardInterrupt:
+        except (KeyboardInterrupt, GeneratorExit):
```
freider, line 488: Fix 2: Reraise `GeneratorExit` instead of catching it further down and printing it/emitting a failed input

test/conftest.py
```diff
@@ -543,7 +543,7 @@ def output_lockstep(self) -> Iterator[threading.Barrier]:
         self.put_outputs_barrier = threading.Barrier(1)
 
     async def FunctionGetInputs(self, stream):
-        self.get_inputs_barrier.wait()
+        await asyncio.get_running_loop().run_in_executor(None, self.get_inputs_barrier.wait)
```
freider, line 546: This was mostly a code sanity thing - I realized this could block the test server event loop otherwise if you have a malformed test using input_lockstep

test/container_test.py
```diff
@@ -1309,6 +1314,39 @@ async def custom_heartbeater(self):
     assert "Traceback" not in caplog.text  # should not print a full traceback - don't scare users!
 
 
+@skip_github_non_linux
+@pytest.mark.usefixtures("server_url_env")
+def test_sigint_termination_input_concurrent(servicer):
```
freider, line 1319: Kind of tests the same thing as `test_cancellation_stops_task_with_concurrent_inputs` but for completeness we shouldn't assume that cancellations use sigint

mwaskom (approved):
Nice catch!

PR #1827, opened 2024-05-14
dx: dont silently drop invalid volume= mounts

thundergolfer:
## Describe your changes

Was for a little bit baffled as to what was wrong with my `CloudBucketMount` code before realizing I had a dangling `,`. 

```python
volume = modal.CloudBucketMount(
    secret=cloudflare_creds,
    bucket_name="modal-testing",
    bucket_endpoint_url="https://hello.r2.cloudflarestorage.com",
),
app = modal.App("demobug", volumes={"/vol/": volume})
```

### DIFF
--- a/modal/_utils/mount_utils.py
+++ b/modal/_utils/mount_utils.py
@@ -39,7 +39,7 @@ def validate_volumes(
     volumes: Mapping[Union[str, PurePosixPath], Union["_Volume", "_CloudBucketMount"]],
 ) -> Sequence[Tuple[str, Union["_Volume", "_NetworkFileSystem", "_CloudBucketMount"]]]:
     if not isinstance(volumes, dict):
-        raise InvalidError("volumes must be a dict[str, Volume] where the keys are mount paths")
+        raise InvalidError("volumes must be a dict where the keys are mount paths")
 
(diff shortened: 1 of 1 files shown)


modal/_utils/mount_utils.py
```diff
@@ -42,13 +39,16 @@ def validate_volumes(
     volumes: Mapping[Union[str, PurePosixPath], Union["_Volume", "_CloudBucketMount"]],
 ) -> Sequence[Tuple[str, Union["_Volume", "_NetworkFileSystem", "_CloudBucketMount"]]]:
     if not isinstance(volumes, dict):
-        raise InvalidError("volumes must be a dict[str, Volume] where the keys are paths")
+        raise InvalidError("volumes must be a dict[str, Volume] where the keys are mount paths")
```
mwaskom, line 42: Not something you introduced, but is the value type here no longer correct?

thundergolfer: Yeh it's no longer entirely accurate. Could just be `dict`.

thundergolfer, line 42:
```suggestion
        raise InvalidError("volumes must be a dict where the keys are mount paths")
```

PR #1829, opened 2024-05-16
Run asgi app constructor after lifecycle enter and hydration

freider:
## Describe your changes
Fixes issue were other objects weren't hydrated in an @app.asgi_app constructor function, and lifecycle enter methods hadn't yet run at that point

MOD-2782

## Changelog

* Fix bug where other functions weren't callable from within an `asgi_app` or `wsgi_app` constructor function, and side effects of `@enter` methods weren't available in that scope.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -510,7 +510,6 @@ def finalize_function(
         startup_timeout = imp_fun.webhook_config.web_server_startup_timeout
         wait_for_web_server(host, port, timeout=startup_timeout)
         callable = asgi_app_wrapper(web_server_proxy(host, port), container_io_manager)
-
     else:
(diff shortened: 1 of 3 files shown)


modal/_container_entrypoint.py
```diff
@@ -569,9 +597,11 @@ def breakpoint_wrapper():
             post_snapshot_methods = _find_callables_for_obj(imp_fun.obj, _PartialFunctionFlags.ENTER_POST_SNAPSHOT)
             call_lifecycle_functions(event_loop, container_io_manager, list(post_snapshot_methods.values()))
 
+        finalized_function = finalize_function(imp_fun, container_io_manager)
```
erikbern, line 600: if there's an exception here, will this be handled properly? iirc there's a context manager to handle user exceptions that we might want to user here

although the exception flow is i guess an existing problem, so it's not something we have to solve in this PR

PR #1830, opened 2024-05-16
Make background volume commits default experience

thundergolfer:
## Describe your changes

Previously if a user didn't specify the flag they wouldn't get background commits. When this merges they will. 

https://github.com/modal-labs/modal-client/pull/1823 merged only a few days ago. I'll merge this around mid-week next week. 

cc @gimaker 

## Changelog

* Background committing on `modal.Volume` mounts is now default behavior.


gimaker (approved):
:+1:

PR #1831, opened 2024-05-16
Add new Resource type in proto

thundergolfer:
## Describe your changes

New proto field intended to support a requesting ephemeral disk space for a function. Currently Modal Functions get a fixed disk quota. 

**Prospective API**

```python
@app.function(ephemeral_disk=1_000)
def f():
   print("I have 1 GiB of scratch disk storage!")
```

    - just an addition of a new proto field. that's always compatible
    - just an addition of a new proto field. that's always compatible


mwaskom:
Just bikeshedding but I don’t find the parameter name super clear as to what this does. What makes the disk “ephemeral”?

thundergolfer:
The data on the disk lasts only as long as the Function container. The Slack thread in #client has background.

PR #1832, opened 2024-05-17
Show deprecation warning for A100(memory=40)

thundergolfer:
## Describe your changes

Should have done this much earlier. 

### DIFF
--- a/modal/gpu.py
+++ b/modal/gpu.py
@@ -67,11 +67,18 @@ class A100(_GPUConfig):
         self,
         *,
         count: int = 1,  # Number of GPUs per container. Defaults to 1. Useful if you have very large models that don't fit on a single GPU.
-        memory: int = 0,  # Deprecated. Use `size` instead.
+        memory: Optional[int] = None,  # Deprecated. Use `size` instead.
         size: Union[str, None] = None,  # Select GiB configuration of GPU device: "40GB" or "80GB". Defaults to "40GB".
     ):
         allowed_memory_values = {40, 80}
         allowed_size_values = {"40GB", "80GB"}
+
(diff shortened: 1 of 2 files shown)


modal/gpu.py
```diff
@@ -67,11 +67,18 @@ def __init__(
...
-        memory: int = 0,  # Deprecated. Use `size` instead.
+        memory: Optional[int] = None,  # Deprecated. Use `size` instead.
         size: Union[str, None] = None,  # Select GiB configuration of GPU device: "40GB" or "80GB". Defaults to "40GB".
     ):
         allowed_memory_values = {40, 80}
         allowed_size_values = {"40GB", "80GB"}
+
+        if memory is not None:
+            deprecation_warning(
+                (2024, 5, 16),
```
mwaskom, line 78: Might be good to clarify that it’s not a simple rename, but you should be able to suggest the full value that they should replace with.

test/gpu_test.py
```diff
@@ -138,8 +133,8 @@ def test_a100_20gb_gpu_unsupported():
 
     app = App()
 
-    with pytest.raises(ValueError, match="A100 20GB is unsupported, consider"):
-        app.function(gpu=modal.gpu.A100(memory=20))(dummy)
+    with pytest.raises(ValueError, match="size='20GB' is invalid"):
+        app.function(gpu=modal.gpu.A100(size="20GB"))(dummy)
```
erikbern, line 137: might want to add something like this too
```python
with pytest.warns(match="size='40GB'"):
    app.function(gpu=modal.gpu.A100(memory=40))(dummy)
```

erikbern (approved):

PR #1836, opened 2024-05-20
Add `modal container stop` CLI

mwaskom:
## Changelog

- Added a `modal container stop` CLI command that will kill an active container and reassign its current inputs.


gongy (commented):
Thanks!

PR #1841, opened 2024-05-21
[MOD-2471] Add `key_prefix` to `CloudBucketMount`

mattnappo:
Required client changes to enable https://github.com/modal-labs/modal/pull/12788.

* Add `key_prefix` to the `CloudBucketMount` proto
* Update `CloudBucketMount` API to support `key_prefix`

### DIFF
--- a/modal/cloud_bucket_mount.py
+++ b/modal/cloud_bucket_mount.py
@@ -135,8 +135,12 @@ def cloud_bucket_mounts_to_proto(mounts: List[Tuple[str, _CloudBucketMount]]) ->
         if mount.requester_pays and not mount.secret:
             raise ValueError("Credentials required in order to use Requester Pays.")
 
-        if not mount.key_prefix.endswith('/'):
+        if mount.key_prefix == "":
+            raise ValueError("key_prefix must not be empty. To mount the entire bucket, do not specify any prefix.")
+        elif mount.key_prefix and not mount.key_prefix.endswith('/'):
             raise ValueError("key_prefix will be prefixed to all object paths, so it must end in a '/'")
+        else:
+            key_prefix = mount.key_prefix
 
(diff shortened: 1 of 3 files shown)


modal/cloud_bucket_mount.py
```diff
@@ -133,6 +135,9 @@ def cloud_bucket_mounts_to_proto(mounts: List[Tuple[str, _CloudBucketMount]]) ->
         if mount.requester_pays and not mount.secret:
             raise ValueError("Credentials required in order to use Requester Pays.")
 
+        if not mount.key_prefix.endswith('/'):
+            raise ValueError("key_prefix will be prefixed to all object paths, so it must end in a '/'")
+
```
luiscape, line 140: Don't you have to check for `None` here?

luiscape (approved):

modal/cloud_bucket_mount.py
mattnappo: Are you implying we should have a separate error message for when key_prefix is empty? (since empty string does not end with a `/`)

modal/cloud_bucket_mount.py
```diff
@@ -135,8 +135,12 @@ def cloud_bucket_mounts_to_proto(mounts: List[Tuple[str, _CloudBucketMount]]) ->
         if mount.requester_pays and not mount.secret:
             raise ValueError("Credentials required in order to use Requester Pays.")
 
-        if not mount.key_prefix.endswith('/'):
+        if mount.key_prefix == "":
+            raise ValueError("key_prefix must not be empty. To mount the entire bucket, do not specify any prefix.")
```
mwaskom, line 139: I might be barging in after a useful discussion but is there any specific reason to disallow this?

PR #1843, opened 2024-05-24
Add ephemeral_disk request to client

thundergolfer:
## Describe your changes

```python
import subprocess

import modal

app = modal.App()


@app.function(
    ephemeral_disk=1000 * 1000,
)
def f() -> None:
    subprocess.run("df /", shell=True, check=True)
```

* I chose MiB as the parameter unit, which is a bit annoying because GiB is more natural for disk, but if we're not going to have units in the parameter name I think it's better to be consistent across all space-based parameters. 
* See _"I'm wanting to introduce a way to request disk space as part"_ in internal #client channel for previous discussion. 

    - Old servers are fine accepting the ephemeral disk request value. They'll ignore it. 
    - Not relevant to this changeset

## Changelog

* `modal.Function` now supports requesting ephemeral disk (SSD) via the new `ephemeral_disk` parameter. Intended for use in doing large dataset ingestion and transform.

### DIFF
--- a/modal/_resources.py
+++ b/modal/_resources.py
@@ -12,6 +12,7 @@ def convert_fn_config_to_resources_config(
     cpu: Optional[float],
     memory: Optional[Union[int, Tuple[int, int]]],
     gpu: GPU_T,
+    ephemeral_disk: Optional[int],
 ) -> api_pb2.Resources:
     if cpu is not None and cpu < 0.1:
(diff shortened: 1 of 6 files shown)


mwaskom:
Is this actually per *function*, not per container? I feel confused about the behavior under scaling.

thundergolfer:
Is defined on the Function and determines the amount of disk per-container. Is there an alternative wording that's better? I did avoiding using "container".

thundergolfer:
@prbot approve

PR #1845, opened 2024-05-25
Remove `keep_warm` from Function proto

aksh-at:
Long deprecated in favor of `warm_pool_size`.


modal_proto/api.proto
```diff
@@ -877,7 +877,7 @@ message Function {
 
   uint32 concurrency_limit = 19;
 
-  bool keep_warm = 20;
+  reserved 20; // old fields
```
gongy, line 880: `[deprecated=true]`

PR #1848, opened 2024-05-28
ruff format our .pyi type stubs

thundergolfer:
## Describe your changes

N/A formatting change only

https://modal-com.slack.com/archives/C069MLUT4SJ/p1716840985672899

N/A formatting change only

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ filterwarnings = [
 ]
 
 [tool.ruff]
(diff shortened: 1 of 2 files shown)


mwaskom:
NVM I am dumb — I'd originally read this as being part of the pre-commit logic but doing it in the `type-stubs` task will indeed work for package deployment.

I still think it would be kind of nice to move it upstream to synchronicity, but not critical.

PR #1850, opened 2024-05-28
Allow `modal app stop` CLI to specify app by name not just ID

mwaskom:
## Changelog

- The `modal app stop` CLI command now accepts a `--name` (or `-n`) option to stop an App by name rather than by ID.


mwaskom:
@prbot approve

mwaskom:
Hmm need to add @devennavani to the prbot pool

PR #1852, opened 2024-05-29
Method pooling: Container entrypoint changes

freider:
## Describe your changes

Internal changes that are activated once the new "loading" logic + necessary backend changes goes live.
The new behavior is only activated once the `is_class` boolean attribute on the Function definition proto is set to `true`.

TODO: add/cherrypick tests for the new behavior to container_entrypoint.py from the mega branch

* This should be backwards compatible with the current type of functions/methods that can run.
* It's forward compatible with the upcoming repackaging of classes as unified "services" that execute all methods in the same container pools

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -9,9 +9,10 @@ import signal
 import sys
 import threading
 import time
+import typing
 from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence, Type
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence
 
 from google.protobuf.message import Message
 from synchronicity import Interface
@@ -38,7 +39,7 @@ from ._utils.function_utils import (
 )
 from .app import App, _App
 from .client import Client, _Client
-from .cls import Cls
+from .cls import Cls, Obj
 from .config import logger
 from .exception import ExecutionError, InputCancellation, InvalidError, deprecation_warning
 from .execution_context import _set_current_context_ids, interact
@@ -52,8 +53,6 @@ from .partial_function import (
 from .running_app import RunningApp
 
 if TYPE_CHECKING:
-    from types import ModuleType
-
     import modal._container_io_manager
     import modal.object
 
@@ -111,7 +110,7 @@ class Service(metaclass=ABCMeta):
     of "Services" after the point of import.
     """
 
-    obj: Any
+    user_cls_instance: Any
     app: Optional[_App]
     code_deps: Optional[List["modal.object._Object"]]
 
@@ -124,16 +123,17 @@ class Service(metaclass=ABCMeta):
 
 @dataclass
 class ImportedFunction(Service):
-    obj: Any
-    user_defined_callable: Callable[..., Any]
+    user_cls_instance: Any
     app: Optional[_App]
     code_deps: Optional[List["modal.object._Object"]]
+
+    _user_defined_callable: Callable[..., Any]
 
     def get_finalized_functions(
         self, fun_def: api_pb2.Function, container_io_manager: "modal._container_io_manager.ContainerIOManager"
     ) -> Dict[str, "FinalizedFunction"]:
         # Check this property before we turn it into a method (overriden by webhooks)
-        is_async = get_is_async(self.user_defined_callable)
+        is_async = get_is_async(self._user_defined_callable)
         # Use the function definition for whether this is a generator (overriden by webhooks)
         is_generator = fun_def.function_type == api_pb2.Function.FUNCTION_TYPE_GENERATOR
 
@@ -142,7 +142,7 @@ class ImportedFunction(Service):
             # for non-webhooks, the runnable is straight forward:
             return {
                 "": FinalizedFunction(
-                    callable=self.user_defined_callable,
+                    callable=self._user_defined_callable,
                     is_async=is_async,
                     is_generator=is_generator,
                     data_format=api_pb2.DATA_FORMAT_PICKLE,
@@ -150,7 +150,7 @@ class ImportedFunction(Service):
             }
 
         web_callable = construct_webhook_callable(
-            self.user_defined_callable, fun_def.webhook_config, container_io_manager
+            self._user_defined_callable, fun_def.webhook_config, container_io_manager
         )
 
         return {
@@ -165,16 +165,17 @@ class ImportedFunction(Service):
 
 @dataclass
 class ImportedClass(Service):
-    obj: Any
-    partial_functions: Dict[str, _PartialFunction]
+    user_cls_instance: Any
     app: Optional[_App]
     code_deps: Optional[List["modal.object._Object"]]
+
+    _partial_functions: Dict[str, _PartialFunction]
 
     def get_finalized_functions(
         self, fun_def: api_pb2.Function, container_io_manager: "modal._container_io_manager.ContainerIOManager"
     ) -> Dict[str, "FinalizedFunction"]:
         finalized_functions = {}
-        for method_name, partial in self.partial_functions.items():
+        for method_name, partial in self._partial_functions.items():
             partial = synchronizer._translate_in(partial)  # ugly
             user_func = partial.raw_f
             # Check this property before we turn it into a method (overriden by webhooks)
@@ -183,7 +184,7 @@ class ImportedClass(Service):
             is_generator = partial.is_generator
             webhook_config = partial.webhook_config
 
-            bound_func = user_func.__get__(self.obj)
+            bound_func = user_func.__get__(self.user_cls_instance)
 
             if not webhook_config or webhook_config.type == api_pb2.WEBHOOK_TYPE_UNSPECIFIED:
                 # for non-webhooks, the runnable is straight forward:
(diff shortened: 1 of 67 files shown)


modal/_container_entrypoint.py
```diff
@@ -73,6 +103,108 @@ class FinalizedFunction:
...
+    obj: Any
+    partial_functions: Dict[str, _PartialFunction]
+    app: Optional[_App]
+    code_deps: Optional[List["modal.object._Object"]]
+
+    def get_finalized_functions(
+        self, fun_def: api_pb2.Function, container_io_manager: "modal._container_io_manager.ContainerIOManager"
+    ) -> Dict[str, "FinalizedFunction"]:
+        finalized_functions = {}
+        for method_name, partial in self.partial_functions.items():
```
mwaskom, line 177: I regret that we've not come up with a better term than "partial function" :(

modal/_container_entrypoint.py
```diff
@@ -73,6 +103,108 @@ class FinalizedFunction:
...
+            user_func = partial.raw_f
+            # Check this property before we turn it into a method (overriden by webhooks)
+            is_async = get_is_async(user_func)
+            # Use the function definition for whether this is a generator (overriden by webhooks)
+            is_generator = partial.is_generator
+            webhook_config = partial.webhook_config
+
+            bound_func = user_func.__get__(self.obj)
+
+            if not webhook_config or webhook_config.type == api_pb2.WEBHOOK_TYPE_UNSPECIFIED:
```
mwaskom, line 188: Why does the webhook config check look different between the function / class version?

modal/_container_entrypoint.py
```diff
@@ -73,6 +103,108 @@ class FinalizedFunction:
...
+        # Check this property before we turn it into a method (overriden by webhooks)
+        is_async = get_is_async(self.user_defined_callable)
+        # Use the function definition for whether this is a generator (overriden by webhooks)
+        is_generator = fun_def.function_type == api_pb2.Function.FUNCTION_TYPE_GENERATOR
+
+        webhook_config = fun_def.webhook_config
+        if not webhook_config.type:
+            # for non-webhooks, the runnable is straight forward:
+            return {
+                "": FinalizedFunction(
```
mwaskom, line 144: Is it right that we use an empty string as a key here because we're going to get this from a proto message and so it will be the default for non-class-functions?

modal/_container_entrypoint.py
```diff
@@ -309,18 +457,18 @@ def _cancel_input_signal_handler(signum, stackframe):
...
                 # this lets us cancel it using a signal handler that raises an exception
                 try:
-                    run_input_sync(input_id, function_call_id, args, kwargs)
+                    run_input_sync(finalized_function, input_id, function_call_id, args, kwargs)
                 finally:
                     signal.signal(signal.SIGUSR1, usr1_handler)  # reset signal handler
 
 
-def import_function(
+def import_single_function_service(
```
mwaskom, line 465: What's the "single function" terminology mean? Is that the same thing as a "function that isn't a class"? I guess not though as I would expect the return type to be an `ImportedFunction` in that case.

modal/_container_entrypoint.py
```diff
@@ -441,20 +553,110 @@ def import_function(
...
+        qual_name: str = function_def.function_name
+
+        if not is_global_object(qual_name):
+            raise LocalFunctionError("Attempted to load a class defined in a function scope")
+
+        parts = qual_name.split(".")
+        if not (
+            len(parts) == 2 and parts[1] == "*"
+        ):  # the "function name" of a class service "function placeholder" is expected to be "ClassName.*"
+            raise InvalidError(f"Invalid 'class function' identifier {qual_name}")
```
mwaskom, line 601: Should we try to reserve `InvalidError` for when the *user* does something invalid (that's how we document it)? AFAICT this is more of an internal error check?

modal/_container_entrypoint.py
```diff
@@ -441,20 +553,110 @@ def import_function(
...
+            active_app = _cls._app
+            # TODO: set code deps using the decorator options of the cls - maybe using _cls.deps?
+        else:
+            # This is a raw class - find all methods
+            method_partials = _find_partial_methods_for_cls(cls, _PartialFunctionFlags.all())
+
+    # Instantiate the class if it's defined
+    assert cls  # must be a class
+    if ser_params:
+        _client: _Client = synchronizer._translate_in(client)
```
mwaskom, line 619: Yuck

modal/_container_entrypoint.py
```diff
@@ -441,20 +553,110 @@ def import_function(
...
+        args, kwargs = deserialize(ser_params, _client)
+    else:
+        args, kwargs = (), {}
+    obj = cls(*args, **kwargs)
+    if isinstance(cls, Cls):
+        obj = obj.get_obj()
```
mwaskom, line 625: What's going on here?

modal/_container_entrypoint.py
```diff
@@ -441,20 +553,110 @@ def import_function(
...
+        method_partials,
+        active_app,
+        code_deps,
     )
 
 
+def get_active_app_fallback(function_def: api_pb2.Function) -> Optional[_App]:
+    # This branch is reached in the special case that the imported function/class is:
+    # 1) not serialized, and
+    # 2) isn't a FunctionHandle - i.e, not decorated at definition time
```
mwaskom, line 638: Do we still have `FunctionHandle` as a concept?

mwaskom (approved):
:+1: Felt like I mostly had a couple nitpicky comments / questions; the overall approach feels *slightly* bolted on but quite possible the best we can do right now! Excited to see the rest of the project come together.

modal/_container_entrypoint.py
freider: This is an old comment that was moved here I think. We don't have separate Function/FunctionHandle concepts, but this refers to it being a proper `modal.Function` rather than just a Python vanilla function (i.e. decorated).
This handles the rare case, mostly used in tests, that the imported raw function isn't actually decorated in global scope and is a vanilla Python function that doesn't know which Modal app it belongs to.

modal/_container_entrypoint.py
freider: Yeah this is quite hard to grok, and we should try to clarify the code a bit IMO (this is mostly as is from the pre-patch code).

If you have a `modal.Cls` instance (`cls` in this case, so `Cls` can basically be considered a metaclass...) and `__call__` it (which to a user looks like an instantiation of the class), it returns an `Obj` instance, which is not an actual instance of the user's defined class (the one decorated) but a lazily evaluated placeholder for the args and kwargs that should be used to create the actual instance of the user's class.

`obj.get_obj()` in this case takes the `Obj` and returns an instance of the user's actual class. The reason that it's if-gated is due to the imported `cls` not always necessarily being a `modal.Cls` instance. In case the class isn't decorated in global scope it could **be** the actual user class, in which case we wouldn't have to do the whole `.get_obj()` thing to get to the underlying instance.

Might actually take a stab at putting better names on these entities/methods now that I have it fresh in my head, as it's definitely confusing...

modal/_container_entrypoint.py
freider: 👍

modal/_container_entrypoint.py
freider: Yeah it's a funciton that isn't a class, i.e. all old function definitions. The runtime returned object type is `ImportedFunction` but it's type annotated as a `Service` (its abstract baseclass) to indicate that anything that's not part of the Service interface shouldn't be used by the calling code.

modal/_container_entrypoint.py
freider: There is a subtle difference in that `webhook_config` can be `None` in the case of the class (since it's a normal Python attribute on a `PartialFunction` which is just set to `None` when it's not a web endpoint), but in the case of a function it's actually a sub-message of the larger `Function` proto, which afaik is never actually `None`, but rather a WebhookConfig object with all attributes set to defaults.

modal/_container_io_manager.py
```diff
@@ -398,7 +398,7 @@ async def run_inputs_outputs(self, input_concurrency: int = 1) -> AsyncIterator[
         async for input_id, function_call_id, input_pb in self._generate_inputs():
             args, kwargs = self.deserialize(input_pb.args) if input_pb.args else ((), {})
             self.current_input_id, self.current_input_started_at = (input_id, time.time())
-            yield input_id, function_call_id, args, kwargs
+            yield input_id, function_call_id, args, kwargs, input_pb.method_name
```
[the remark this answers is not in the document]
freider: 1 - fixed!
2 - I can see your point, but it would feel more natural to send it along if the deserialization happened in the receiver (since we send args and kwargs), but then we are basically duplicating this logic in both the async and sync versions so I'm not sure... Can take a look at refactoring after this patch

modal/_container_entrypoint.py
```diff
@@ -73,6 +103,108 @@ class FinalizedFunction:
...
+    ) -> Dict[str, "FinalizedFunction"]:
+        ...
+
+
+@dataclass
+class ImportedFunction(Service):
+    obj: Any
+    user_defined_callable: Callable[..., Any]
+    app: Optional[_App]
+    code_deps: Optional[List["modal.object._Object"]]
```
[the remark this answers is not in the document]
freider: No idea actually, probably a result of how I got here through multiple refactors. Fixing, and also marking the subclass specific ones as "private" by prefixing with underscore to indicate they shouldn't be used outside of the finalization logic inside the class

PR #1854, opened 2024-05-29
Turn `.new` from `deprecation_warning` to `deprecation_error`

erikbern:
Looking at an internal dashboard and I see almost no volume for the old RPC endpoints.

* `SharedVolumeCreate`: 32 calls in the last week
* `VolumeCreate`: 5 calls in the last week
* `DictCreate`: 478 calls in the last week
* `QueueCreate`: 330 calls in the last week

This PR removes the support client-side. We will still keep it server-side for a bit. In maybe a month let's delete the implementation of those RPC endpoints and raise a grpc error that the method is no longer supported.


freider (approved):
sweet!

PR #1855, opened 2024-05-29
Fix modal volume ls --json

mwaskom:
Fixes MOD-3033

Apparently I added `--json` to `modal volume ls` but just forgot to actually wire it up 🤦‍♂️.

I think that's because I got distracted trying to figure out whether there is a reason that we didn't use our utility function for displaying a table here (which also does JSON output). AFAICT there isn't, although we do have a special case where we dump a much simpler representation if the output is not a terminal.

I rewrote the method to use the utility function while preserving the previous behavior for non-terminal output stream (but *not* in the case of JSON, as I think that if you are asking for JSON and piping output you really want JSON output).

Also this command didn't appear to be tested at all, so I added some testing too.

### DIFF
--- a/modal/cli/volume.py
+++ b/modal/cli/volume.py
@@ -10,7 +10,6 @@ from grpclib import GRPCError, Status
 from rich.console import Console
 from rich.live import Live
 from rich.syntax import Syntax
-from rich.table import Table
 from typer import Argument, Option, Typer
 
 import modal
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot lgtm

modal/cli/volume.py
```diff
@@ -143,30 +142,30 @@ async def ls(
...
 
-    if sys.stdout.isatty():
-        console = Console()
-        console.print(f"Directory listing of '{path}' in '{volume_name}'")
-        table = Table()
-        for name in ["filename", "type", "created/modified", "size"]:
-            table.add_column(name)
-
+    if not json and not sys.stdout.isatty():
+        # Legacy behavior -- I am not sure why exactly we did this originally but I don't want to break it
```
gimaker, line 146: Also have no context on this. From a scripting perspective it's nice that we omit the table formatting, but this seems oddly terse!

gimaker (commented):
:heavy_check_mark: LGTM!

PR #1856, opened 2024-05-29
Don't use `indexed_objects` for `app.image`

erikbern:
Fairly minor change. There's no need to write the image object to the blueprint since a long time ago, so let's just store it on the app object instead. This came up because I'm trying to clean up `indexed_objects` to really only be functions and classes.

Unfortunately a lot of image tests depended on this indirectly. So for images to get built inside tests, I had to add a dummy function in all those tests that use the image. So that's why the delta is much bigger.

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -28,6 +28,10 @@ from modal_proto import api_pb2
 from .supports.skip import skip_windows
 
 
+def dummy():
+    ...
+
+
 def test_supported_python_series():
     assert SUPPORTED_PYTHON_SERIES == PYTHON_STANDALONE_VERSIONS.keys()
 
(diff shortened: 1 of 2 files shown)


test/image_test.py
```diff
@@ -449,6 +472,7 @@ def test_image_run_function(builder_version, servicer, client):
 def test_image_run_function_interactivity(builder_version, servicer, client):
     app = App()
     app.image = Image.debian_slim().pip_install("pandas").run_function(run_f)
```
mwaskom, line 474: Maybe doesn't matter but still *doing* the assignment here and below.

mwaskom (approved):
:+1: too bad about the test boilerplate but looks good otherwise!

test/image_test.py
erikbern: Yeah but we still want to support this syntax. I deliberately made sure some tests are using `app.image`, some just using `image`, and some `App(image=...)` just so we cover all options

erikbern:
Thanks @mwaskom !

PR #1858, opened 2024-05-30
docs: ssh into container

freider:
Adds example on how to set up an sshd sidecar in a container to be able to ssh into it

The `modal.forward()` documentation mentions it can be used to be able to ssh into containers, but several users have found it hard to actually make that work so this adds a simple working example to the docs.

Might make sense to put it in some guide docs as well, but there is a risk that users searching for `ssh` as a way to get shell access to a container would find it and use it instead of the much simpler/built-in `modal container exec`...


ekzhang:
Nice, that's beautiful!

luiscape (approved):

PR #1861, opened 2024-05-30
Changes how PartialMethods implement the descriptor protocol

freider:
* `ModalWrappedCls().method` is unchanged (returns the _Obj-bound `Function` object) - *I think this is the only branch we actually made use of prior to this update*
* `ModalWrappedCls.method` and `UserCls.method` now return the PartialFunction object instead of an "unbound" Function - this is used by `serialize(UserCls)` and so the code path was previously triggered by serialized classes, but 
* UserCls().method now returns the original method bound to the UserCls instance - acts like vanilla python since that's what UserCls is

I'm not 100% sure this won't break anything, but I think the changes are superficially sane and all the client tests are passing.

Something that *is* changed is that anyone using `app.cls(serialized=True)` will get a larger payload in their `class_serialized` proto field since it will now contain each PartialFunction object including the code for every method of the class. This is intentional and is something that is used by serialized classes in the upcoming method-pooling change.

### DIFF
--- a/modal/partial_function.py
+++ b/modal/partial_function.py
@@ -53,13 +53,24 @@ class _PartialFunction:
         self.wrapped = False  # Make sure that this was converted into a FunctionHandle
 
     def __get__(self, obj, objtype=None) -> _Function:
-        # This only happens inside user methods when they refer to other methods
         k = self.raw_f.__name__
         if obj:  # Cls().fun
-            function = getattr(obj, "_modal_functions")[k]
-        else:  # Cls.fun
-            function = getattr(objtype, "_modal_functions")[k]
-        return function
+            if hasattr(obj, "_modal_functions"):
+                # This happens inside "local" user methods when they refer to other methods,
+                # e.g. Foo().parent_method() doing self.local.other_method()
+                return getattr(obj, "_modal_functions")[k]
+            else:
+                # special edge case: referencing a method of an instance of an
+                # unwrapped class (not using app.cls()) with @methods
+                # not sure what would be useful here, but lets return a bound version of the underlying function,
+                # since the class is just a vanilla class at this point
+                return self.raw_f.__get__(obj, objtype)
+
+        else:  # Cls.fun6
+            # This happens mainly during serialization of the wrapped underlying class of a Cls
+            # since we don't have the instance info here we just return the PartialFunction itself
+            # to let it be bound to a variable and become a Function later on
+            return self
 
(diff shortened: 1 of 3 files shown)


modal/partial_function.py
```diff
@@ -53,13 +53,24 @@ def __init__(
...
+                # e.g. Foo().parent_method() doing self.local.other_method()
+                return getattr(obj, "_modal_functions")[k]
+            else:
+                # special edge case: referencing a method of an instance of an
+                # unwrapped class (not using app.cls()) with @methods
+                # not sure what would be useful here, but lets return a bound version of the underlying function,
+                # since the class is just a vanilla class at this point
+                return self.raw_f.__get__(obj, objtype)
+
+        else:  # Cls.fun6
```
mwaskom, line 69: What does `.fun6` mean?

modal/partial_function.py
```diff
@@ -53,13 +53,24 @@ def __init__(
...
         k = self.raw_f.__name__
         if obj:  # Cls().fun
-            function = getattr(obj, "_modal_functions")[k]
-        else:  # Cls.fun
-            function = getattr(objtype, "_modal_functions")[k]
-        return function
+            if hasattr(obj, "_modal_functions"):
+                # This happens inside "local" user methods when they refer to other methods,
+                # e.g. Foo().parent_method() doing self.local.other_method()
+                return getattr(obj, "_modal_functions")[k]
```
mwaskom, line 61: This feels brittle, although I don't have a better suggestion right now.

mwaskom (approved):
:+1:

modal/partial_function.py
freider: That I have fat fingers :D Supposed to be `Cls.fun` and indicate that this branch covers accessing a method directly on the class rather than via an instance of the class

modal/partial_function.py
freider: I agree, and I'm not even sure exactly when this triggers vs the `__getattr__` on `_Obj` tbh...

Hopefully we can clean this up (it's all preexisting code, but I removed the part of the code that sets this attribute on **the class** in this patch, so now it's only being set on the instance of the user class that's being bound to the function in the container entrypoint)

erikbern: There's a slightly more complex solution where we use the cls the user provided, or construct a class dynamically in the case of lookups. That would let us get rid of getattr "hack"

PR #1866, opened 2024-05-31
Update docs for functions that reference stub in modal.runner

mwaskom:
- Add a docstring for `deploy_stub` that points users to `deploy_app`
- Add a similar docstring for `run_stub`, but suppress it from the API docs because `run_app` is also suppressed (I guess we prefer `App.run`? I am not familiar with the intended migration vector there.)


mwaskom:
@prbot automerge

mwaskom:
@prbot automerge

PR #1869, opened 2024-05-31
Raise an error when running or deploying an App with no indexed objects

mwaskom:
Closes MOD-3066

WIP


mwaskom:
@freider pointed out that apps without registered functions are also useful for executing sandboxes (i.e. you can launch the sandbox from a local entrypoint). Need to think more about that.

mwaskom:
Let's just close this for now and think about the sandbox issue a little bit.

PR #1870, opened 2024-05-31
Added `Source` to PyPI Metadata

jamesbraza:
## Describe your changes

Specifies `Source` for PyPI metadata, as `requests` does so: https://github.com/psf/requests/blob/v2.32.3/setup.py#L130 --> https://pypi.org/project/requests/

Here's how it will look on PyPI:

screenshot of Source on PyPI

## Changelog

Adds Source to PyPI metadata


mwaskom:
Makes sense, thanks!

mwaskom (approved):

PR #1871, opened 2024-06-01
Add docs to web_endpoints and remove /openapi.json from default

charlesfrye:
## Describe your changes

[Linear ticket](https://linear.app/modal-labs/issue/MOD-2896/allow-enabling-docs-url-on-web-endpoints)

Allows activation of Swagger/Redoc docs by passing a flag to `web_endpoint`. Also disables `/openapi.json` route by default. This is implemented internally via a new boolean field on `WebhookConfig`.

Tested by running against a devbox Modal instance.

Don't think I fully understand the implications here, so looking for review!

## Changelog

- `web_endpoint`s now have the option to include interactive SwaggerUI/redoc docs by setting `docs=True`
- `web_endpoint`s no longer include an OpenAPI JSON spec route by default

### DIFF
--- a/modal/partial_function.py
+++ b/modal/partial_function.py
@@ -187,9 +187,7 @@ def _web_endpoint(
     _warn_parentheses_missing=None,
     *,
     method: str = "GET",  # REST method for the created endpoint.
-    docs: bool = False,  # Whether to enable the Swagger UI documentation for this endpoint.
-    redoc: bool = False,  # Whether to enable the ReDoc documentation for this endpoint.
-    openapi: bool = False,  # Whether to expose the OpenAPI json for this endpoint.
+    docs: bool = False,  # Whether to enable interactive documentation for this endpoint at /docs.
     label: Optional[str] = None,  # Label for created endpoint. Final subdomain will be <workspace>--<label>.modal.run.
     wait_for_response: bool = True,  # Whether requests should wait for and return the function response.
     custom_domains: Optional[
@@ -237,21 +235,13 @@ def _web_endpoint(
 
         # self._loose_webhook_configs.add(raw_f)
 
-        docs_flags = 0
-        if docs:
-            docs_flags += 1
-        if redoc:
-            docs_flags += 2
-        if openapi:
-            docs_flags += 4
-
         return _PartialFunction(
             raw_f,
             _PartialFunctionFlags.FUNCTION,
             api_pb2.WebhookConfig(
                 type=api_pb2.WEBHOOK_TYPE_FUNCTION,
                 method=method,
-                fastapi_endpoint_docs_flags=docs_flags,
+                fastapi_endpoint_docs=docs,
                 requested_suffix=label,
                 async_mode=_response_mode,
                 custom_domains=_parse_custom_domains(custom_domains),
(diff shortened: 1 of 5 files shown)


mwaskom:
Does anything need to happen server-side here?

charlesfrye:
> Does anything need to happen server-side here?

Not that I could tell! Didn't find any usage of WebhookConfig that needed to be changed. And I ran integration tests against the updated protos and did a spot check with the dev server running the `basic_web` examples with and without the docs flag set.

mwaskom:
Yeah that sounds right! Do we need to change the protos at all then?

charlesfrye:
I think we do need to change the protobuf to add this flag, we just don't need to change any server-side _logic_.

If I grok the system, the explanation is: we need to communicate configuration information from the client to the server and then to the worker. Matching other patterns I found, in this implementation a user Python process provides an argument to a public decorator (`@web_endpoint`) that ends up [communicated to a private method](https://github.com/modal-labs/modal-client/pull/1871/files#diff-af590c8ac1ecd84881361705179ff347e2f4bfbe34549d810b5a6eeb83789a06L504) (`webhook_asgi_app`) running inside a worker. The worker consumes it via the container entrypoint. The serialization format for those messages is given by the protos.

Here's my understanding of the backward/forward compatibility concerns.

For old clients, everything is fine: the new field is added on the server (as `False`) but then dropped by the client on the worker.

If for some reason the worker client was a newer version than the user client, it would read the value `False` and continue to not show docs. The `/openapi.json` route would go away, but that was not documented behavior and I don't think it was intended.

For new clients talking to old servers, the field will be dropped by the server and then not present when the worker client reads it. That would lead to docs not being served even when requested. So I assume that when we update the protos, we first propagate that to all servers, then release the new client version? Or I missed something!

modal/partial_function.py
```diff
@@ -187,6 +187,7 @@ def _web_endpoint(
     _warn_parentheses_missing=None,
     *,
     method: str = "GET",  # REST method for the created endpoint.
+    docs: bool = False,  # Whether to enable interactive documentation for this endpoint at /docs.
```
mwaskom, line 190: Nit but would put this after label since it’s imo a lower priority parameter

modal_proto/api.proto
```diff
@@ -2088,6 +2088,7 @@ message WebhookConfig {
   repeated CustomDomainConfig custom_domains = 6;
   uint32 web_server_port = 7;
   float web_server_startup_timeout = 8;
+  bool fastapi_endpoint_docs = 9;
```
mwaskom, line 2091: We would kind of like to move away from fastapi for web endpoint to reduce the client dependency burden. So I might suggest naming this more generally.

That also raises the question of whether it’s a good idea to expand the “minimum viable web endpoint” functionality, as it ties us closer to relatively heavy dependencies. That would be my one hesitation about adding this as an opt-in feature.

charlesfrye: I named it this way because (per Slack discussions and Linear ticket) I expect it to migrate onto a higher-level `fastapi_endpoint` decorator in the future that wraps a lower-level `web_endpoint` that wouldn't have these bells and whistles. It seems like it's slightly annoying to rename fields in protobufs, so I thought I'd jump ahead to the intended final name.

mwaskom: I see, although would we necessarily only expose endpoint docs when using `fastapi`? We may roll our own lightweight functionality and decide it's feasible to include some kind of documentation in that featureset.

mwaskom (approved):

PR #1872, opened 2024-06-02
throw deprecation error on old lifecycle hooks

erikbern:
Seems time to turn warnings into errors. The warning is more than 3 months old at this point, and this wasn't a super common feature to begin with.


mwaskom (approved):
Sounds good

PR #1873, opened 2024-06-03
Add --tag as optional parameter to modal deploy

RohanArepally:
## Describe your changes

https://github.com/modal-labs/modal/pull/13137

- [ x ] Client+Server: this change is compatible with old servers
- [ x ] Client forward compatibility: this change ensures client can accept data intended for later versions of itself

## Changelog

- `modal deploy` now accepts a `--tag` optional parameter that allows you to specify a custom tag for the deployed version, making it easier to identify and manage different deployments of your app.

### DIFF
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -17,7 +17,7 @@ from ._resolver import Resolver
 from ._sandbox_shell import connect_to_sandbox
 from ._utils.async_utils import TaskContext, synchronize_api
 from ._utils.grpc_utils import retry_transient_errors
-from ._utils.name_utils import check_object_name
+from ._utils.name_utils import check_object_name, is_valid_tag
 from .client import HEARTBEAT_INTERVAL, HEARTBEAT_TIMEOUT, _Client
 from .config import config, logger
 from .exception import (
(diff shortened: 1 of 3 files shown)


mwaskom:
Do we need to get the proto change in first so that we can enable this server-side?

mwaskom (approved):
Otherwise LGTM. Though a test wouldn't hurt :D

mwaskom:
Also can we get a changelog entry for this? Since a few users have asked about it but I have lost track of exactly who.

modal/runner.py
```diff
@@ -373,6 +373,7 @@ async def _deploy_app(
     show_progress=True,
     environment_name: Optional[str] = None,
     public: bool = False,
+    tag: Optional[str] = None,
```
mwaskom, line 376: FWIW `tag` could also just be typed as `str` with an empty default since it's going to get sent via a protobuf anyway, that will (very slightly) simplify the rest of the code that uses it.

RohanArepally: Sounds good will do

mwaskom:
🎉 nice feature!

PR #1876, opened 2024-06-05
Make sure serialized partial functions are actually usable

### LINKED ISSUE #1846: Errors when using parametrized functions in jupyter notebook
Hey! I'm currently having a great time using modal, but running into issues when using the `@cls` decorator in jupyter-notebooks. Here's the code I'm using right now in a single cell:

```python
import modal
from modal import Image, App, method

app = App("test-cls-notebook")
app.image = Image.debian_slim()

@app.cls(gpu="A10G")
class Module:
    def __init__(self, x: int = 10):
        self.x = x

    @method()
    def test(self, y):
        return self.x * y

with app.run():
    model = Module()
    print(model.test.remote(10))
```
---
When running this, I get this error:

```
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/pkg/modal/_container_entrypoint.py", line 653, in <module>
    main(container_args, client)
  File "/pkg/modal/_container_entrypoint.py", line 575, in main
    pre_snapshot_methods = _find_callables_for_obj(imp_fun.obj, _PartialFunctionFlags.ENTER_PRE_SNAPSHOT)
  File "/pkg/modal/partial_function.py", line 137, in _find_callables_for_obj
    return {k: meth.__get__(user_obj) for k, meth in _find_callables_for_cls(user_cls, flags).items()}
  File "/pkg/modal/partial_function.py", line 128, in _find_callables_for_cls
    for k, pf in _find_partial_methods_for_cls(user_cls, flags).items():
  File "/pkg/modal/partial_function.py", line 93, in _find_partial_methods_for_cls
    if partial_function.flags & flags:
AttributeError: 'PartialFunction' object has no attribute 'flags'
Exception ignored in: <function blocking___del__ at 0x7fb0ab554820>
  File "/pkg/synchronicity/synchronizer.py", line 529, in proxy_method
KeyError: '_sync_original_140396784877472'
```

This works perfectly fine when running it in a script, even with the `app.run()` context, so it seems like it is just a problem when running it in jupyter. Any help would be appreciated!

Modal version: `modal client version: 0.62.181`

freider:
Synchronicity wrapped objects that are serialized are not generally useful when deserialized in a different process. If we however add a custom serialization format that pickles the underlying object and then translates it on the receiving end, we get something that can be used.

This is necessary to get usable methods on classes, since those are PartialFunctions which are synchronicity wrapped.

It would have been sort of nice to do this directly in synchronicity, but unpickling would rely on a having a synchronizer singleton which is not generally the case, so for now this will have to reside in Modal's own pickler.

Fixes #1846

### DIFF
--- a/modal/_serialization.py
+++ b/modal/_serialization.py
@@ -3,11 +3,14 @@ import io
 import pickle
 from typing import Any
 
+from synchronicity.synchronizer import Interface
+
+from modal._utils.async_utils import synchronizer
 from modal_proto import api_pb2
 
 from ._vendor import cloudpickle
 from .config import logger
-from .exception import DeserializationError, InvalidError
+from .exception import DeserializationError, ExecutionError, InvalidError
 from .object import Object, _Object
 
 PICKLE_PROTOCOL = 4  # Support older Python versions.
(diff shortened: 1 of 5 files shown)


mwaskom (approved):
Feels a little hacky :/ but I've confirmed that it fixes the existing issue with serialized Cls objects!

freider:
> Feels a little hacky :/ but I've confirmed that it fixes the existing issue with serialized Cls objects!

Definitely a little hacky :( 

It would be nice to solve in synchronicity itself somehow, since not having serializable wrapper objects feels like a bug in a way. The tricky thing is that as long as synchronicity doesn't have a singleton Synchronizer, there is no apparent way to associate a wrapper object with a specific Synchronizer on deserialization. It's a bit similar to how modal's existing deserializer for `Object` subclasses requires a `Client` to be injected into the `deserialize()` function and can't just use vanilla `cloudpickle.load()`

erikbern (commented):
👍

PR #1890, opened 2024-06-11
Remove misc deprecated fields (2)

erikbern:
This time it should be safe to remove (I kept the deprecated GPU enums(


erikbern:
Turns out we were using a deprecated field... in something printed to stdout. Since I just removed it server-side, it won't be printed for old clients. I think that's acceptable though?

PR #1891, opened 2024-06-11
Move function creation status output logic to _output.py context manager

freider:
Extracted from #1789 to make that PR a bit smaller, since this diff is basically a no-op and just reindents a bunch of code


freider:
@prbot approve

mwaskom (commented):
image

:+1:

PR #1898, opened 2024-06-12
AppDisconnect reason is not optional

erikbern:
Super minor – the backend now (correctly) assumes the reason is always set, so let's enforce that in the client too


ekzhang:
I think the argument is still omitted in `modal serve`, which is causing CI to fail. I created #1899 for this.

PR #1900, opened 2024-06-12
[MOD-2923] Add SHELL and ENTRYPOINT to client Image SDK

cathyzbn:
## Describe your changes

Reference: [MOD-2923](https://linear.app/modal-labs/issue/MOD-2923/add-shell-and-entrypoint-to-client-image-sdk) Add SHELL and ENTRYPOINT to client Image SDK

## Changelog
- Added support for entrypoint and shell for custom containers:
- `Image.debian_slim().entrypoint([])` can be used interchangeably with `.dockerfile_commands('ENTRYPOINT []')`
- `Image.debian_slim().shell(["/bin/bash", "-c"])` can be used interchangeably with `.dockerfile_commands('SHELL ["/bin/bash", "-c"]')`

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -858,10 +858,10 @@ class _Image(_Object, type_prefix="im"):
 
     def entrypoint(
         self,
-        *entrypoint_files: Union[str, List[str]],
+        entrypoint_args: List[str],
     ) -> "_Image":
         """Set the entrypoint for the image."""
-        args_str = _flatten_str_args("entrypoint", "entrypoint_files", entrypoint_files)
+        args_str = _flatten_str_args("entrypoint", "entrypoint_files", entrypoint_args)
         args_str = '"' + '", "'.join(args_str) + '"' if args_str else ""
         dockerfile_cmd = f"ENTRYPOINT [{args_str}]"
 
@@ -869,7 +869,7 @@ class _Image(_Object, type_prefix="im"):
 
     def shell(
         self,
-        *shell_commands: Union[str, List[str]],
+        shell_commands: List[str],
     ) -> "_Image":
         """Overwrite default shell for the image."""
         args_str = _flatten_str_args("shell", "shell_commands", shell_commands)
--- a/modal/image.py
+++ b/modal/image.py
@@ -856,6 +856,28 @@ class _Image(_Object, type_prefix="im"):
             force_build=self.force_build or force_build,
         )
 
+    def entrypoint(
+        self,
+        *entrypoint_files: Union[str, List[str]],
+    ) -> "_Image":
+        """Set the entrypoint for the image."""
+        args_str = _flatten_str_args("entrypoint", "entrypoint_files", entrypoint_files)
+        args_str = '"' + '", "'.join(args_str) + '"' if args_str else ""
+        dockerfile_cmd = f"ENTRYPOINT [{args_str}]"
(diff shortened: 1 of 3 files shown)


modal/image.py
```diff
@@ -856,6 +856,28 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             force_build=self.force_build or force_build,
         )
 
+    def entrypoint(
+        self,
+        *entrypoint_files: Union[str, List[str]],
```
gongy, line 861: tiny nit: maybe `entrypoint_args` might be a clearer name since they form a command according to the [docs](https://docs.docker.com/reference/dockerfile/#entrypoint)

gongy, line 861: tiny nits:
- it might be easier to only support `List` for now - features are easy to add but hard to deprecate on the client
- maybe `entrypoint_args` might be a clearer name since they form a command according to the [docs](https://docs.docker.com/reference/dockerfile/#entrypoint)

test/image_test.py
```diff
@@ -627,6 +627,32 @@ def test_image_dockerfile_copy(builder_version, servicer, client, tmp_path_with_
...
+def test_image_docker_command_entrypoint(builder_version, servicer, client, tmp_path_with_content):
+    app = App()
+    data_mount = Mount.from_local_dir(tmp_path_with_content, remote_path="/")
+    app.image = Image.debian_slim().entrypoint([])
+    app.function()(dummy)
+
+    with app.run(client=client):
+        layers = get_image_layers(app.image.object_id, servicer)
+        assert "ENTRYPOINT []" in layers[0].dockerfile_commands
+        files = {f.mount_filename: f.content for f in Mount._get_files(data_mount.entries)}
```
gongy, line 639: I might be wrong but I think you can skip these checks and the data mount itself in both tests

test/image_test.py
```diff
@@ -627,6 +627,32 @@ def test_image_dockerfile_copy(builder_version, servicer, client, tmp_path_with_
         assert files == {"/data.txt": b"hello", "/data/sub": b"world"}
 
 
+def test_image_docker_command_entrypoint(builder_version, servicer, client, tmp_path_with_content):
+    app = App()
+    data_mount = Mount.from_local_dir(tmp_path_with_content, remote_path="/")
+    app.image = Image.debian_slim().entrypoint([])
```
gongy, line 633: might be useful to test a non-empty as well

modal/image.py
```diff
@@ -856,6 +856,28 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        """Set the entrypoint for the image."""
+        args_str = _flatten_str_args("entrypoint", "entrypoint_files", entrypoint_files)
+        args_str = '"' + '", "'.join(args_str) + '"' if args_str else ""
+        dockerfile_cmd = f"ENTRYPOINT [{args_str}]"
+
+        return self.dockerfile_commands(dockerfile_cmd)
+
+    def shell(
+        self,
+        *shell_commands: Union[str, List[str]],
```
gongy, line 872: similar: maybe we can just support lists for now

gongy (approved):
looks great!

left some very minor feedback that might help save time down the line

gongy (approved):
LGTM!

PR #1903, opened 2024-06-12
[MOD-2863] Start a debugger when `enter_debugger` flag is set

mattnappo:
Client PR for [snapshot tooling PR](https://github.com/modal-labs/modal/pull/13399)

Allows the client to enter a `pdb` debugger upon snapshot restore when specified by the `restore-state.json` file

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -562,7 +562,7 @@ class _ContainerIOManager:
         enter_debugger = int(restored_state["enter_debugger"])
         if enter_debugger:
             logger.debug("starting debugger")
-            import pdb;
+            import pdb
             pdb.set_trace()
 
         # Local ContainerIOManager state.
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -580,12 +580,13 @@ class _ContainerIOManager:
 
         self._enter_debugger = int(restored_state["enter_debugger"])
         if self._enter_debugger:
-            print("starting debugger")
+            logger.debug("starting debugger")
             # breakpoint()
             # await self.interact()
-            import pdb; pdb.set_trace()
+            import pdb;
+            pdb.set_trace()
         else:
-            print("not starting debugger")
+            logger.debug("not starting debugger")
 
         self._client = await _Client.from_env()
         self._waiting_for_memory_snapshot = False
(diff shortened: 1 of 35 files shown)


modal/_container_io_manager.py
```diff
@@ -560,6 +560,12 @@ async def memory_restore(self) -> None:
         with restored_path.open("r") as file:
             restored_state = json.load(file)
 
+        # Start a debugger if the worker tells us to
+        if int(restored_state["snapshot_debug"]):
+            logger.debug("Entering snapshot debugger")
+            import pdb
+            pdb.set_trace()
+
```
luiscape, line 568: Can you handle cases where `snapshot_debug` is not present? This is necessary because workers take ~24h to rollover so this container entrypoint update will land on a worker that doesn't yet inject that value.

luiscape (changes requested):

test/container_app_test.py
```diff
@@ -19,6 +20,11 @@
 def my_f_1(x):
     pass
 
+# Simulate calling `breakpoint()` in test
+def test_breakpoint():
+    return True
+
+sys.breakpointhook = test_breakpoint
```
luiscape, line 27: This works but can you use a context manager with mocking instead? This replacement is flaky and may be overwritten by other programs. Example:

```python
with mock.patch("sys.breakpointhook", test_breakpoint):
   # your test code here
```

luiscape (changes requested):
LGTM but use `mock` instead.

test/container_app_test.py
```diff
@@ -85,14 +82,13 @@ async def test_container_debug_snapshot(container_client, tmpdir, servicer):
...
-
-        # This will call the breakpoint hook
-        io_manager.memory_snapshot()
+    with mock.patch("sys.breakpointhook", test_breakpoint):
+        with mock.patch.dict(
+            os.environ, {
+                "MODAL_RESTORE_STATE_PATH": str(restore_path),
+                "MODAL_SERVER_URL": servicer.remote_addr
+                }):
+            io_manager.memory_snapshot()
```
luiscape, line 91: Can we assert that this returns `True`?

luiscape (approved):
LGTM but see comment.

test/container_app_test.py
```diff
@@ -19,6 +19,9 @@
 def my_f_1(x):
     pass
 
+# Simulate calling `breakpoint()` in test
+def test_breakpoint():
+    return True
```
mattnappo, line 24: I think the closest we can get is something like this

```suggestion
    assert True
```

luiscape: Oh I see. That's alright. We don't have to add the `assert` here. This is good to merge.

PR #1904, opened 2024-06-12
Deal with square brackets in progress messages

irfansharif:
Same as e8087de76c8 effectively. Trying to enrich messages when waiting in task queues elsewhere.


irfansharif:
Actually, can't figure it out. Will just escape on the server.

PR #1905, opened 2024-06-13
[MOD-3177] Differentiate between an expired output and unfinished input

cathyzbn:
## Describe your changes
[[MOD-3177]](https://linear.app/modal-labs/issue/MOD-3117/differentiate-between-an-expired-output-and-unfinished-input)
Return different Timeout errors for unfinished vs. expired function call outputs.
- When the function call outputs are not available, raise TimeoutError
- When the function call outputs are expired in Redis, raise OutputExpiredError

### DIFF
--- a/modal/exception.py
+++ b/modal/exception.py
@@ -86,6 +86,10 @@ class DeserializationError(Error):
     """Raised to provide more context when an error is encountered during deserialization."""
 
 
+class OutputExpiredError(Error):
+    """Raised when the Output exceeds expiration and times out."""
+
+
 class RequestSizeError(Error):
     """Raised when an operation produces a gRPC request that is rejected by the server for being too large."""
 
--- a/modal/exception.py
+++ b/modal/exception.py
@@ -58,6 +58,10 @@ class InteractiveTimeoutError(TimeoutError):
     """Raised when interactive frontends time out while trying to connect to a container."""
 
 
+class OutputExpiredError(TimeoutError):
+    """Raised when the Output exceeds expiration and times out."""
+
+
 class AuthError(Error):
     """Raised when a client has missing or invalid authentication."""
 
@@ -84,10 +88,6 @@ class ExecutionError(Error):
 
 class DeserializationError(Error):
     """Raised to provide more context when an error is encountered during deserialization."""
-
-
-class OutputExpiredError(Error):
-    """Raised when the Output exceeds expiration and times out."""
 
 
 class RequestSizeError(Error):
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -151,6 +151,10 @@ class _Invocation:
                 request,
                 attempt_timeout=backend_timeout + ATTEMPT_TIMEOUT_GRACE_PERIOD,
             )
+
+            if response.is_expired:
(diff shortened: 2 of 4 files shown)


modal/functions.py
```diff
@@ -181,7 +185,7 @@ async def poll_function(self, timeout: Optional[float] = None):
         )
 
         if len(items) == 0:
-            raise TimeoutError()
+            raise TimeoutError("Function call has not finished")
```
mwaskom, line 188: Is it right that we're reporting a timeout _on the `get`_ operation, rather than on the underlying function call? We may want to reword this to make that a little more clear.

modal/exception.py
```diff
@@ -86,6 +86,10 @@ class DeserializationError(Error):
     """Raised to provide more context when an error is encountered during deserialization."""
 
 
+class OutputExpiredError(Error):
```
mwaskom, line 89: Since this is a specific kind of timeout (and because we previously raised `TimeoutError`), you could make it derive from the more general class maybe.

cathyzbn: Thanks for pointing that out! just changed.

modal/functions.py
cathyzbn: yeah, should be reporting on the get operation

modal_proto/api.proto
```diff
@@ -1069,6 +1069,7 @@ message FunctionGetOutputsResponse {
   repeated int32 idxs = 3;
   repeated FunctionGetOutputsItem outputs = 4;
   string last_entry_id = 5;
+  bool is_expired = 6;
```
mwaskom, line 1072: Does this require server-side changes? Normally we would make such client-server updates in three steps:

- Merge new protos in the client
- Merge server side changes
- Merge client-side feature code

cathyzbn: got it! the server side changes are [here](https://github.com/modal-labs/modal/pull/13423), and i'll make a separate pr that only changes the protos.

modal/functions.py
```diff
@@ -160,6 +162,8 @@ async def pop_function_call_outputs(
                 # update timeout in retry loop
                 backend_timeout = min(OUTPUTS_TIMEOUT, t0 + timeout - time.time())
                 if backend_timeout < 0:
+                    if len(response.outputs) == 0 and response.num_unfinished_inputs == 0:
+                        raise OutputExpiredError()
```
gongy, line 166: suggestion: raising here means we change the behavior of `.remote()`. while this is arguably more consistent, maybe it would be better to keep the surface area smaller to just `.get()`?

gongy (approved):
looks great!

test/conftest.py
```diff
@@ -721,7 +721,7 @@ async def FunctionGetOutputs(self, stream):
 
             await stream.send_message(api_pb2.FunctionGetOutputsResponse(outputs=[output]))
         else:
-            await stream.send_message(api_pb2.FunctionGetOutputsResponse(outputs=[]))
+            await stream.send_message(api_pb2.FunctionGetOutputsResponse(outputs=[], num_unfinished_inputs=-1))
```
gongy, line 724: 🤔

PR #1912, opened 2024-06-18
Fix issue with web server proxy not handling disconnect

freider:
Also adds `_asgi.py` to `pyright` tests, which would have caught this bug

I haven't worked out to fix for the test yet (haven't found a straight forward way to get aiohttp to close the downstream connection while consuming `data` - we might have to use some lower level request sender primitive)

@ekzhang maybe you have some more insight into aiohttp? :)

### DIFF
--- a/modal/_asgi.py
+++ b/modal/_asgi.py
@@ -1,5 +1,6 @@
 # Copyright Modal Labs 2022
 import asyncio
+import typing
 from typing import Any, AsyncGenerator, Callable, Dict, Optional, cast
 
 import aiohttp
@@ -194,7 +195,7 @@ def wait_for_web_server(host: str, port: int, *, timeout: float) -> None:
 async def _proxy_http_request(session: aiohttp.ClientSession, scope, receive, send) -> None:
     proxy_response: Optional[aiohttp.ClientResponse] = None
 
-    async def request_generator():
+    async def request_generator() -> AsyncGenerator[bytes, None]:
         while True:
             message = await receive()
             if message["type"] == "http.request":
@@ -221,7 +222,7 @@ async def _proxy_http_request(session: aiohttp.ClientSession, scope, receive, se
         allow_redirects=False,
     )
 
-    async def send_response():
+    async def send_response() -> None:
         msg = {
             "type": "http.response.start",
             "status": proxy_response.status,
@@ -233,7 +234,7 @@ async def _proxy_http_request(session: aiohttp.ClientSession, scope, receive, se
             await send(msg)
         await send({"type": "http.response.body"})
 
-    async def listen_for_disconnect():
+    async def listen_for_disconnect() -> typing.NoReturn:
         while True:
             message = await receive()
             if message["type"] == "http.disconnect":
--- a/modal/_asgi.py
+++ b/modal/_asgi.py
@@ -1,7 +1,6 @@
 # Copyright Modal Labs 2022
 import asyncio
-import typing
-from typing import Any, AsyncGenerator, Callable, Dict, Optional, cast
+from typing import Any, AsyncGenerator, Callable, Dict, NoReturn, Optional, cast
 
 import aiohttp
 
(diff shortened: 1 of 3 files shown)


ekzhang:
Thanks for debugging this issue. I haven't followed along yet — is the issue that aiohttp doesn't disconnect when you get an `http.disconnect` asgi event?

freider:
> Thanks for debugging this issue. I haven't followed along yet — is the issue that aiohttp doesn't disconnect when you get an `http.disconnect` asgi event?

The issue is that the `proxy_response` object we use to "forward" our http.disconnect (`proxy_response.connection.transport.abort()`) isn't defined at the point that we use it - it's still `None`, since it's only set once we have sent our entire request (including all body parts).

I didn't spend a ton of time on it but I couldn't find a straight-forward way to access the active transport object from within our `generate_data` coroutine. The closest thing that's easily done is to raise a CancelledError from the data chunk generator, which will make aiohttp send an eof to the proxy server. But for some reason I wasn't able to make that translate to a disconnect in my receiving test server...

ekzhang (approved):

ekzhang:
Nice, test looks good to me! Thanks for explaining.

advay-modal:
I modified the test and added code to fix the issue, would appreciate another review!

advay-modal:
Also, it seems like the test fails remotely but passes locally, not sure why

mwaskom:
> Also, it seems like the test fails remotely but passes locally, not sure why

Do you mean fails in GitHub CI but not in your development env? Or are you trying to run something on Modal?

modal/_asgi.py
```diff
@@ -204,45 +204,54 @@ async def request_generator():
...
+            headers=[(k.decode(), v.decode()) for k, v in scope["headers"]],
+            data=None if scope["method"] in aiohttp.ClientRequest.GET_METHODS else request_generator(),
+            allow_redirects=False,
+        )
+    except aiohttp.ClientConnectionError as e:
+        if isinstance(e.__cause__, NameError):
+            return
+        else:
+            raise e.__cause__
+    else:
```
ekzhang, line 231: I don't think you need the `else:` block here because the `except` code path will raise or return. Generally it could be better to avoid indenting a large block of continuation code since that makes following it nonlinear.

modal/_asgi.py
```diff
@@ -204,45 +204,54 @@ async def request_generator():
                 if not message.get("more_body", False):
                     break
             elif message["type"] == "http.disconnect":
-                proxy_response.connection.transport.abort()  # Abort the connection.
-                break
+                if proxy_response:
+                    proxy_response.connection.transport.abort()  # Abort the connection.
+                else:
+                    raise NameError("Proxy response not defined")
```
ekzhang, line 210: Is it possible for proxy_response to exist while the request generator is still running? If not, we could remove the `if proxy_response` case.

`NameError` works but could be better represented as [`ConnectionAbortedError`](https://docs.python.org/3/library/exceptions.html#ConnectionAbortedError) since this is caused by an aborted request.

modal/_asgi.py
```diff
@@ -204,45 +204,54 @@ async def request_generator():
...
+            url=path,
+            headers=[(k.decode(), v.decode()) for k, v in scope["headers"]],
+            data=None if scope["method"] in aiohttp.ClientRequest.GET_METHODS else request_generator(),
+            allow_redirects=False,
+        )
+    except aiohttp.ClientConnectionError as e:
+        if isinstance(e.__cause__, NameError):
+            return
+        else:
+            raise e.__cause__
```
ekzhang, line 230: Maybe we should re-raise the full error instead of just the cause, since that provides more context? It would just be `raise` in Python.

ekzhang (commented):
Nice — curious that the tests aren't passing though, hm

image

Do you think it could be a difference in aiohttp version? Perhaps the aiohttp version in CI doesn't wrap the exception in an `aiohttp.ClientConnectionError`, in which case we would need to be careful.

advay-modal:
Addressed your changes

PR #1913, opened 2024-06-18
Enforce object name rules (again)

mwaskom:

---

---

</details>

## Changelog

- It is now an error to create or lookup various Modal objects with an invalid name. Names must be shorter than 64 characters and may contain only alphanumeric characters, dashes, periods, and underscores. The name check had inadvertently been removed for a brief time following an internal refactor and then reintroduced as a warning. It is once more a hard error. Please get in touch if this is blocking access to your data.

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -125,7 +125,7 @@ class _Dict(_Object, type_prefix="di"):
         dict[123] = 456
         ```
         """
-        check_object_name(label, "Dict", warn=True)
+        check_object_name(label, "Dict")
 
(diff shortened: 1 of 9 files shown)


ekzhang (approved):

ekzhang:
Sounds good, so I shouldn't merge the server-side check for a while. People will still be able to create secrets with arbitrary names from https://modal.com/secrets until then.

mwaskom:
IIRC it was a pretty small number of objects that had snuck through with illegal names, I think it's probably fine to start erroring server-side as well now, especially for secret *creation* (in fact,  sorry, we probably should have done that immediately when we discovered this issue).

PR #1919, opened 2024-06-20
Set examples that use `.lookup` to notest

mwaskom:
Eagerly running RPCs in the doctests breaks CI on forks (and any other environment where you don't have a valid token).

This is kind of frustratingly manual and I expect that we will break it again in the future. On the other hand, this also reduces test coverage and is going to let our doctests get out of date.

I'm testing it by withholding the model token secrets from the doctest job. I initially intended to revert this after making sure CI passes, but now I might be inclined to keep that in.

Overall I'm not sure what the ideal solution is here, but I would like to unblock a long-standing helpful third-party PR.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -59,8 +59,8 @@ jobs:
       - name: Run docstring tests
         env:
           MODAL_ENVIRONMENT: client-doc-tests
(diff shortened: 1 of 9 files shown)


mwaskom:
Eh, this is too much of a hassle. Still not sure we should be running live Modal apps in CI, but let's supersede this by #1920

PR #1922, opened 2024-06-20
Improve type coverage

mwaskom:
Turning #1875 into a first-party PR since I can't figure out how to get GitHub Actions to avoid running the doctests against it, and they are failing because you need a valid token.


mwaskom:
@prbot approve

erikbern (commented):
👍

PR #1925, opened 2024-06-24
Remove the use of closurevars for finding function dependencies

erikbern:
We needed the closurevars hack in order to hydrate implicit dependencies of functions. However since we deprecated `.new` in #1854, the only remaining implicit dependencies are `Queue.from_name` and `Dict.from_name`. Both those objects support lazy resolution, which means we can rely on that instead.

This also fixes a subtle issue where previously if you have a deployed app that refers to a persisted object, and you delete that object, it would still use that object's id. In particular if someone created a _new_ persisted object with the same name, the deployed app would still refer to the old (deleted) object.

A slight downside of this PR is that the lazy resolution will add a backend roundtrip the first time the user interacts with the object. This seems fine?

### DIFF
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -2,10 +2,9 @@
 import asyncio
 import inspect
 import os
-from collections import deque
 from enum import Enum
 from pathlib import Path, PurePosixPath
-from typing import Any, AsyncIterator, Callable, Dict, List, Literal, Optional, Set, Type
+from typing import Any, AsyncIterator, Callable, Dict, List, Literal, Optional, Type
 
 from grpclib import GRPCError
 from grpclib.exceptions import StreamTerminatedError
@@ -18,7 +17,6 @@ from .._traceback import append_modal_tb
 from ..config import config, logger
 from ..exception import ExecutionError, FunctionTimeoutError, InvalidError, RemoteError
 from ..mount import ROOT_DIR, _is_modal_path, _Mount
-from ..object import Object
 from .blob_utils import MAX_OBJECT_SIZE_BYTES, blob_download, blob_upload
 from .grpc_utils import RETRYABLE_GRPC_STATUS_CODES, unary_stream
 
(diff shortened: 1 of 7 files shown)


modal/object.py
```diff
@@ -199,7 +199,7 @@ def deps(self) -> Callable[..., List["_Object"]]:
         """mdmd:hidden"""
         return self._deps if self._deps is not None else lambda: []
 
-    async def resolve(self):
+    async def resolve(self, client: Optional[_Client] = None):
```
erikbern, line 202: this is a minor code change slightly outside the scope of the PR – I needed it in a test

test/helpers.py
```diff
@@ -26,7 +26,8 @@ def deploy_app_externally(
             **{"PYTHONUTF8": "1"},
         }  # windows apparently needs a bunch of env vars to start python...
 
-    env = {**windows_support, "MODAL_SERVER_URL": servicer.client_addr, **env}
+    env = {**windows_support, "MODAL_SERVER_URL": servicer.client_addr, "MODAL_ENVIRONMENT": "main", **env}
```
erikbern, line 29: we hardcode `main` as the environment in `conftest.py`, so this just makes it consistent

freider:
Nice! 🔥🔥🔥 image

Regarding referencing stale object ids, it's would still happen inside the context of a single container (or in the case of serialized functions if their references are hydrated at the time of serialized functions), but this definitely reduces the surface area of that class of bugs.
If we want fix all those edge cases we could potentially save/look up object ids of *deleted* objects by name and reuse that id when creating a new object with the same type + name, but I'm a bit afraid it could potentially interact with left-behind/not yet garbage collected artifacts of the old deleted object in that case...

freider (approved):

freider:
Also, might be a good idea to run integration tests w/ this to verify assumption that this is a safe change :)

erikbern:
> Also, might be a good idea to run integration tests w/ this to verify assumption that this is a safe change :)

`132 passed, 12 skipped, 1 warning in 934.94s (0:15:34)`

PR #1926, opened 2024-06-25
unify `servicer` & `unix_servicer`

erikbern:
Fell into a rabbit hole when fixing a tiny test in #1925 and realized I wanted support for using the same servicer in both the client and the container. Previously we actually had two different instances of `MockClientServicer` which managed all the fake backend state.

Unifying it makes it a bit easier to write container tests that deploy objects and then look them up from the container.

It would have been fairly easy to work around it, but I decided to go the slightly harder and more annoying route and unify the two objects so it's just one.

### DIFF
--- a/test/client_test.py
+++ b/test/client_test.py
@@ -11,7 +11,7 @@ from modal import Client
 from modal.exception import AuthError, ConnectionError, DeprecationError, InvalidError, VersionError
 from modal_proto import api_pb2
 
-from .supports.skip import skip_windows_unix_socket
+from .supports.skip import skip_windows, skip_windows_unix_socket
 
 TEST_TIMEOUT = 4.0  # align this with the container client timeout in client.py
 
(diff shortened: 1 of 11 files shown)


test/client_test.py
```diff
@@ -53,7 +52,7 @@ async def test_client_dns_failure():
 
 @pytest.mark.asyncio
 @pytest.mark.timeout(TEST_TIMEOUT)
-@skip_windows_unix_socket
+@skip_windows("Windows test crashes on connection failure")
```
erikbern, line 55: turns out this failure had nothing to do with unix sockets

erikbern:
@prbot approve

erikbern:
yolo merge since this is blocking me – low risk since it's just test changes

freider:
> yolo merge since this is blocking me – low risk since it's just test changes

Oh sorry, I took a look this morning but forgot to add my review! Lgtm

PR #1927, opened 2024-06-25
Fix Cls.lookup() not working after update MOD-3187

freider:
Lookups were breaking since backend isn't returning class_function_id as part of the handle metadata for the class.

Instead of fixing the backend, this removes the class_function_id field from ClassMetadata and instead does an explicit lookup for the function named `{class-tag}.*`. This is in line with an upcoming refactor that could remove the need for synced Cls objects altogheter, since the "class servicer function" is all that is needed to invoke the class now anyway. It also has the advantage that already-existing deployed classes won't have to get redeployed for lookups to start working.

Client tests were passing since the mock servicer was more feature-complete than the real backend api.

---

---

</details>

## Changelog

* Fixes issue with `Cls.lookup` not working after upgrading to v0.63.0

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -16,7 +16,7 @@ from ._utils.async_utils import synchronize_api, synchronizer
 from ._utils.grpc_utils import retry_transient_errors
 from ._utils.mount_utils import validate_volumes
 from .client import _Client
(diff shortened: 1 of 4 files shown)


freider:
@prbot approve

PR #1929, opened 2024-06-25
Deprecate retries parameter on web endpoints and generators

pawalt:
## Describe your changes

- Resolves MOD-2202

---

---

</details>

## Changelog

- Adds warning about future deprecation of `retries` for generators. Retries are being deprecated as they can lead to nondetermistic generator behavior.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -552,6 +552,17 @@ class _Function(_Object, type_prefix="fu"):
             retries, f"Function {info.get_tag()}" if info.raw_f else f"Class {info.get_tag()}"
         )
 
+        if webhook_config is not None and retry_policy is not None:
+            raise InvalidError(
+                "Web endpoints do not support retries.",
+            )
+
+        if is_generator and retry_policy is not None:
+            deprecation_warning(
+                (2024, 6, 25),
+                "Retries for generator functions are deprecated and will soon be removed.",
+            )
+
         gpu_config = parse_gpu_config(gpu)
 
         if proxy:


mwaskom:
J/w but is the idea that these currently silently don't work, and we want to turn it into an error? Or are we also going to change some behavior?

ekzhang:
> J/w but is the idea that these currently silently don't work, and we want to turn it into an error? Or are we also going to change some behavior?

The goal is to turn it into an error eventually, but first we should disable them silently perhaps

modal/functions.py
```diff
@@ -552,6 +552,20 @@ def from_args(
             retries, f"Function {info.get_tag()}" if info.raw_f else f"Class {info.get_tag()}"
         )
 
+        if webhook_config is not None and retry_policy is not None:
+            deprecation_warning(
+                (2024, 6, 25),
+                "Webhook functions and servers do not support retries. The `retries` argument will be ignored.",
```
ekzhang, line 558:
```suggestion
                "Web endpoints do not support retries. The `retries` argument will be ignored.",
```

pawalt: fixed

mwaskom:
> The goal is to turn it into an error eventually, but first we should disable them silently perhaps

I think my question is — do we want to communicate "this never worked" or "this currently worked, but won't in the future"?

pawalt:
@mwaskom My understanding is that right now, web endpoints will error server-side if you try to enable retries, but generators will not. I don't know what the right messaging looks like, though. Perhaps we should hard fail web endpoints and warn on generators, saying they will be deprecated soon.

mwaskom:
Got it — if we currently raise a server-side error, I think jumping straight to an `InvalidError` client-side makes sense. Otherwise, smooth deprecation is good, though I think it's usually helpful to add a little extra explanation about the context, at least in the changelog.

pawalt:
@mwaskom thanks for the tip. Added to the changelog.

modal/functions.py
```diff
@@ -552,6 +552,17 @@ def from_args(
             retries, f"Function {info.get_tag()}" if info.raw_f else f"Class {info.get_tag()}"
         )
 
+        if webhook_config is not None and retry_policy is not None:
+            raise InvalidError(
+                "Web endpoints and servers do not support retries.",
```
ekzhang, line 557: Just "Web endpoints"

ekzhang (approved):

PR #1930, opened 2024-06-25
Only support string paths on Volumes

ekzhang:
This fixes a bug where `bytes` paths on Volumes were not handled properly by the backend.

Non-UTF8 paths are a very rare occurrence regardless, and I don't think it is necessary to support them in our client library, so I'm just removing the APIs here instead.

**Protobuf wire protocol for `string` and `bytes` are identical.**

## Changelog

- Fixed a bug in `Volume.copy_files()` where some source paths may be ignored if passed as `bytes`.
- `Volume.read_file`, `Volume.read_file_into_fileobj`, `Volume.remove_file`, and `Volume.copy_files` can no longer take both string or bytes for their paths. They now only accept `str`.

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -376,7 +376,7 @@ class _Volume(_Object, type_prefix="vo"):
         return [entry async for entry in self.iterdir(path, recursive=recursive)]
 
     @live_method_gen
-    async def read_file(self, path: Union[str, bytes]) -> AsyncIterator[bytes]:
+    async def read_file(self, path: str) -> AsyncIterator[bytes]:
         """
         Read a file from the modal.Volume.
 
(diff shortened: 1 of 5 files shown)


erikbern:
Does this actually enforce it? You can still pass a `bytes` object to these methods right? I think if we want to be super defensive we should add something like `assert isinstance(path, string)`

modal/volume.py
```diff
@@ -390,8 +390,6 @@ async def read_file(self, path: Union[str, bytes]) -> AsyncIterator[bytes]:
         print(len(data))  # == 1024 * 1024
         ```
         """
-        if isinstance(path, str):
-            path = path.encode("utf-8")
```
erikbern, line 394: how does it work now – will protobuf automatically run the same things when you set the field value?

mwaskom:
We also support `Path` or some kind of paths in a lot of data APIs right? Should we be consistent about that?

ekzhang:
@mwaskom is it okay if I defer that to later? If we want to add PurePosixPath support for those APIs it's a feature-adding change so hopefully pretty smooth

mwaskom:
Yeah I've got no problem doing that separately.

ekzhang:
Sounds good to me! I'll try to merge this one in tomorrow.

mwaskom (approved):

PR #1931, opened 2024-06-25
Error when the python_version for Image.debian_slim is a float

mwaskom:
This is an easy mistake to make and otherwise triggers an inscrutable error message from the guts of the resolver.

Note that we already do a similar check for the code-path shared by constructors that add the standalone Python, so we just needed to address `Image.debian_slim` directly.

Closes MOD-3189


mwaskom:
@prbot automerge

PR #1934, opened 2024-06-25
Remove hidden _experimental_scheduler flag

irfansharif:
It's inert, and has been so for a month or two now. I'm keeping the _experimental_scheduler_placement flag for now, which exposes some other params (but mostly ignored for external users). For everyone else, the only real way to construct things is using the regions= argument.


irfansharif:
(Can only land after https://github.com/modal-labs/modal/pull/13708 is merged internally.)

PR #1937, opened 2024-06-26
Relative path writes for volume get

TheQuantumFractal:
## Details

Very small change to support relative paths for volume / nfs get. Strips the remote path up to the first glob if present then computes file output paths relative to this path.

Closes MOD-2204

---

---

</details>

### DIFF
--- a/modal/cli/_download.py
+++ b/modal/cli/_download.py
@@ -38,7 +38,7 @@ async def _volume_download(
                 await q.put((None, entry))
             else:
                 start_path = os.path.dirname(remote_path).split("*")[0]
-                rel_path = os.path.relpath(entry.path, start_path.lstrip("/"))
+                rel_path = Path(entry.path).relative_to(start_path.lstrip("/"))
                 output_path = local_destination / rel_path
                 if output_path.exists():
                     if overwrite:
--- a/modal/cli/_download.py
+++ b/modal/cli/_download.py
@@ -37,7 +37,9 @@ async def _volume_download(
             if is_pipe:
                 await q.put((None, entry))
             else:
-                output_path = local_destination / entry.path
+                start_path = os.path.dirname(remote_path).split("*")[0]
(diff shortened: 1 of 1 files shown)


modal/cli/_download.py
```diff
@@ -37,7 +37,9 @@ async def producer():
             if is_pipe:
                 await q.put((None, entry))
             else:
-                output_path = local_destination / entry.path
+                start_path = os.path.dirname(remote_path).split("*")[0]
```
luiscape, line 40: What is the need for `.split("*")` here? Take the following example:

```python
remote_path = "/my-volume/*.txt
os.path.dirname(remote_path)  # /my-volume
os.path.dirname(remote_path).split("*")[0]  # /my-volume
```

modal/cli/_download.py
```diff
@@ -37,7 +37,9 @@ async def producer():
...
+                rel_path = os.path.relpath(entry.path, start_path.lstrip("/"))
+                output_path = local_destination / rel_path
```
luiscape, line 42: Can you use [Path from pathlib](https://docs.python.org/3/library/pathlib.html) instead of `os.path`?

luiscape (changes requested):
Small changes.

modal/cli/_download.py
TheQuantumFractal:
```py
remote_path = "/vol1/*/*.txt"
os.path.dirname(remote_path) # /vol1/*
os.path.dirname(remote_path).split("*")[0] # /vol1/
```

os.path.relpath won't work properly if there are still globs in the path (eg if glob for matching folders)

modal/cli/_download.py
TheQuantumFractal: Yeah entry.path and start_path are both strings though so I'd have to turn one into a PurePath

luiscape (approved):

PR #1938, opened 2024-06-26
[MOD-3194] python module import tracing

danielnorberg:


mwaskom:
Is there any reason not to put this into the main package as `modal._telemetry` (or wherever)? We're trying to consolidate everything in the client repo, as the "modern" way of distributing Python doesn't support multiple packages in one release.

danielnorberg:
@mwaskom gotcha, let me try moving it in to the main package

.github/workflows/ci-cd.yml
```diff
@@ -17,6 +17,7 @@ jobs:
     timeout-minutes: 20
 
     strategy:
+      fail-fast: false
```
danielnorberg, line 20: trying to debug test failures

danielnorberg:
There is a noticeable performance impact, seemingly mostly due to the call to `Queue.put_nowait`. Could do some PyO3, but seems overkill for now as this telemetry will only run when task tracing is enabled.

To get a ballpark sense of the impact, with these below imports:

```
import kubernetes
import torch
import django
```

On a linux devbox, the with tracing disabled:

```
imports took 1.5s
```

With tracing enabled:

```
imports took 1.8s
```

This is for ~3700 modules being imported in total.

modal/_telemetry.py
```diff
@@ -0,0 +1,137 @@
...
+MESSAGE_HEADER_LEN = 4
+
+
+class ImportInterceptor(importlib.abc.Loader):
+    loading: typing.Set[str]
+    tracing_socket: socket.socket
+    events: queue.Queue
+
+    @classmethod
+    def connect(cls, socket_filename: str) -> Self:
```
freider, line 33: FYI Another way of returning the self type, without having to use typing_extensions for old pythons, is to string quote the type - `-> "ImportInterceptor"` in this case. Also works for forward references

modal/_container_entrypoint.py
```diff
@@ -52,6 +53,8 @@
 )
 from .running_app import RunningApp
 
+instrument_imports()
```
freider, line 56: Nit, but we could perhaps env-var-gate this call + import here, instead of always importing it, potentially reducing container startup impact a little

modal/_telemetry.py
danielnorberg: Right, I thought this felt cleaner but maybe that's a rustism.

danielnorberg: done

modal/_container_entrypoint.py
danielnorberg: Sounds good, done.

modal/_container_entrypoint.py
```diff
@@ -30,6 +31,7 @@
 from ._container_io_manager import ContainerIOManager, UserException, _ContainerIOManager
 from ._proxy_tunnel import proxy_tunnel
 from ._serialization import deserialize
+from ._telemetry import instrument_imports
```
freider, line 34: I mean that you move this import inside the `if` statement below, to avoid loading the telemetry module for all users  👇

danielnorberg: Yeah that's what I thought as well! Now actually done 😄

freider (approved):

PR #1939, opened 2024-06-27
Add lookup backwards compatibility with pre 0.63 class methods

freider:
Adds back compatibility for `Cls.lookup()` of classes created by <v0.63

Old style classes had methods that were "fully qualified functions" themselves, each separately bound to the parametrization data, so some extra logic is required to make sure those work. We can remove the extra code when we phase out v0.62 in ~6 months

Also fixes some minor bugs and removes some unused code.

I've manually tested the code with both new and old classes.

Will try to add some test (using manually constructed mock data) for lookup/usage of old class methods, as we might want to keep this backwards compatibility around until we unsupport v0.62...

## Changelog

* Adds `Cls.lookup()` backwards compatibility with classes created by clients prior to `v0.63`.

**Important**: When updating (to >=v0.63) an app with a Modal `class` that's accessed using `Cls.lookup()` - make sure to update the client of the app/service **using** `Cls.lookup()` first, and **then** update the app containing the class being looked up.

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -119,7 +119,7 @@ class _Obj:
         """
         if not self._uses_common_service_function():
             raise VersionError(
-                "`Class instance `.keep_warm(...)` can't be used on classes deployed using client version <v0.63"
+                "Class instance `.keep_warm(...)` can't be used on classes deployed using client version <v0.63"
             )
         await self._instance_service_function.keep_warm(warm_pool_size)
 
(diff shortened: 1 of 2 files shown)


mwaskom (approved):
Can we make the changelog entry a little more verbose to specify the right order of operations here? IIUC, you will need to update the modal client in the environment doing the lookup *before* you update the modal client that you deploy the to-be-looked-up app with? Kind of a mouthful :/

PR #1941, opened 2024-06-27
ci: cancel previous builds of the same PR when pushing new commits

danielnorberg:


.github/workflows/ci-cd.yml
```diff
@@ -6,6 +6,11 @@ on:
       - main
   pull_request:
 
+# Cancel previous runs of the same PR but do not cancel previous runs on main
+concurrency:
```
freider, line 10: Does this in any way impact how many `main` ci-cd runs can run concurrently? I don't think we want more than one of those to run, since they would race for the same build version number, I think. Not sure if we handle that well before this PR either though :D

freider (approved):
nice! one comment

.github/workflows/ci-cd.yml
danielnorberg: Yes this will limit `main` (and all other refs as well) to at most one ci-cd run at a time. On `main` this means latter runs will queue up as `pending` until the current build finished. For PRs the current build will be canceled in favor of the latest push.

mwaskom: In theory the main jobs should already queue but we still end up with some annoying contention issues I haven’t gotten around to debugging — usually relating to when we check out the repo to update the version / changelog.

PR #1942, opened 2024-06-28
Fix Cls.with_options in the .lookup case

mwaskom:
Finishes incomplete fix from #1940; the `_Cls` object does not have a `__name__` when it has been `.lookup`-ed.


mwaskom:
@prbot automerge

PR #1944, opened 2024-06-28
Clean up naming of obj vs user_cls[_instance]

freider:
## Describe your changes

The Cls code is quite confusing to read sometimes since we interchangeably use the word "obj" for both instances of the user's defined class and instances of the "wrapper" `_Obj` type returned by calling a `_Cls` (not to be confused with `user_cls`) 😵


freider:
@prbot approve

PR #1945, opened 2024-06-28
Remove legacy single-function class method import logic

freider:
Followup on https://github.com/modal-labs/modal-client/pull/1944

Removes a bunch of code that should be unused in modern apps.

I think this should be safe since this updated container entrypoint should only ever be used by functions deployed using the new style classes


freider:
@prbot approve

PR #1947, opened 2024-06-28
Fix bug where method.local() re-triggers lifecycle methods  MOD-3202

freider:
Running `self.other_method.local()` would previously re-trigger enter lifecycle methods on the class, even if they already ran as part of setting up the container.

## Changelog

* Fixes bug where self.method.local() re-triggers lifecycle methods in classes

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -581,7 +581,7 @@ def import_class_service(
     else:
         args, kwargs = (), {}
 
-    user_cls_instance = get_user_class_instance(cls, args, kwargs, _client)
+    user_cls_instance = get_user_class_instance(cls, args, kwargs)
 
     return ImportedClass(
         user_cls_instance,
@@ -591,9 +591,7 @@ def import_class_service(
     )
 
 
-def get_user_class_instance(
-    cls: typing.Union[type, Cls], args: Tuple, kwargs: Dict[str, Any], _client: _Client
-) -> typing.Any:
+def get_user_class_instance(cls: typing.Union[type, Cls], args: Tuple, kwargs: Dict[str, Any]) -> typing.Any:
     """Returns instance of the underlying class to be used as the `self`
 
     The input `cls` can either be the raw Python class the user has declared ("user class"),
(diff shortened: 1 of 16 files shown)


ekzhang (approved):
Nice!

modal/_container_entrypoint.py
```diff
@@ -587,21 +591,19 @@ def import_class_service(
...
-
-    if ser_params:
-        _client: _Client = synchronizer._translate_in(client)
-        args, kwargs = deserialize(ser_params, _client)
-    else:
-        args, kwargs = (), {}
     if isinstance(cls, Cls):
         # globally @app.cls-decorated class
         modal_obj: Obj = cls(*args, **kwargs)
+        modal_obj.entered = True  # ugly but prevents .local() from triggering additional enter-logic
```
mwaskom, line 605: AFAICT this is the main functional change in the PR? I don't think I understand why we're sure we've run the enter logic; just because we're doing it from container entrypoint?

freider: yeah that's why this is quite ugly... a larger refactor would be needed to clean it up. We'd need to return the "Obj" itself to the parent scope instead to set enter when it has actually been run from the container entrypoint, but there  isn't always a `Cls` that we can construct an `Obj` from because of the non-decorated case...

Would also be nice to somehow unify how we run the lifecycle methods in `.local()` and how we run them in the container entrypoint. The `.local()` enter runners are currently quite broken in other ways (have added some bug tickets).

I'm not even sure how well everything works without a globally decorated Class at the moment, there are many test cases we'd have to add to get good coverage for all paths:
* globally decorated vs not
* calling .local() vs .remote()
* calling from a context where the class definition exists vs not
* calling `self.other_method`

I mean to add more tests for the cartesian product of these cases so we can be a bit more bold when refactoring this code - it currently feels a bit like a mine field

modal/_container_entrypoint.py
```diff
@@ -587,21 +591,19 @@ def import_class_service(
     )
 
 
-def get_user_class_instance(cls: typing.Union[type, Cls], ser_params: bytes, client: Client) -> typing.Any:
+def get_user_class_instance(
+    cls: typing.Union[type, Cls], args: Tuple, kwargs: Dict[str, Any], _client: _Client
```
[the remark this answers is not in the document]
freider: Nope, that was actually the reason I did the refactor here in the first place. But then I forgot to remove it :D

PR #1949, opened 2024-07-01
Test cleanup: Run mock blob store server in separate thread

freider:
Prevents accidental deadlocks in code trying to run blob uploads from blocking test calls, e.g.  `container_entrypoint.main()`


freider:
@prbot approve

PR #1951, opened 2024-07-01
Fix cls builder functions MOD-3210

freider:
Largely a revert of #1945 

## Changelog

* Fixes error when running `@modal.build` methods, introduced in v0.63.19

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -130,11 +130,11 @@ class Service(metaclass=ABCMeta):
 
 @dataclass
 class ImportedFunction(Service):
+    user_cls_instance: Any
     app: Optional[_App]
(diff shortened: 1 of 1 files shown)


mwaskom (approved):
:+1: on the code change but is there a test we could have had that would have blocked this from going in originally?

freider:
> 👍 on the code change but is there a test we could have had that would have blocked this from going in originally?

An integration test for `@build` would have been the most straight-forward way of catching this. There are client tests too, but they sadly require the test setup to explicitly trigger the function in standalone mode rather than "new style method mode" which we weren't doing. 

I'm separately working on fixing them, but we'd need a bit more of an "end to end" test harness in the client to safely do changes like this

PR #1952, opened 2024-07-01
Cap large exception messages

kramstrom:
## Describe your changes

---

---

</details>


freider (approved):
Great! Lets think about if we should actually send the serialized exception to s3 as well... might not be worth it

PR #1953, opened 2024-07-01
Adjust cls @build test to use standalone function references

freider:
Tests were not representative of how we actually load+trigger builder functions in production.
These updated tests failed on v0.63.19, but passes since v0.63.2.

Builder functions aren't using class servicer functions, since that function definition needs to be set to the image we are actually building.

Side note: Need to think about how we can run these tests more "end to end", i.e. first load using the current client and then trigger the container based on the function definition we just loaded. Right now the `run_container` tests basically mock the "loading" and manually constructs ContainerArgs...

### DIFF
--- a/test/container_test.py
+++ b/test/container_test.py
@@ -1135,11 +1135,11 @@ def test_build_decorator_cls(servicer):
     ret = _run_container(
         servicer,
         "test.supports.functions",
-        "BuildCls.*",
-        inputs=_get_inputs(((), {}), method_name="build1"),
+        # note: builder functions are still run as standalone functions from their class service function
+        "BuildCls.build1",
+        inputs=_get_inputs(((), {})),
         is_builder_function=True,
         is_auto_snapshot=True,
-        is_class=True,
     )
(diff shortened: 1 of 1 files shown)


test/container_test.py
```diff
@@ -1135,11 +1135,11 @@ def test_build_decorator_cls(servicer):
     ret = _run_container(
         servicer,
         "test.supports.functions",
-        "BuildCls.*",
-        inputs=_get_inputs(((), {}), method_name="build1"),
+        # note: builder functions are still run as standalone functions from their class service function
```
mwaskom, line 1138: There are really all sorts of ways that `@build` violate expectations. IMO we should consider whether they're net beneficial. I kind of lean not? Different conversation though.

mwaskom (approved):
Thanks for following up here!

test/container_test.py
freider: I agree, they are a bit weird. But I do like the concept of keeping related things together in the class, if we can make it more consistent with other class functionality (parameters etc.)

PR #1954, opened 2024-07-01
Deprecate WebUrlInfo.has_unique_hash field

mwaskom:
Just a little cleanup: we don't send  this from the server any more and haven't for some time. This PR removes the code path that we don't go through and marks the proto field as deprecated.


freider (approved):
lgtm!

PR #1955, opened 2024-07-02
Warn on mounting volumes locally

kramstrom:
## Describe your changes

- MOD-3103 warn users when attempting to mount volumes locally as it's a common source of confusion

---

---

</details>

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -290,7 +290,7 @@ class _Function(_Object, type_prefix="fu"):
     _web_url: Optional[str]
     _function_name: Optional[str]
     _is_method: bool
-    _spec: _FunctionSpec
+    _spec: Optional[_FunctionSpec] = None
     _tag: str
     _raw_f: Callable[..., Any]
     _build_args: dict
@@ -456,6 +456,7 @@ class _Function(_Object, type_prefix="fu"):
         fun._parent = instance_service_function._parent
         fun._app = class_bound_method._app
         fun._all_mounts = class_bound_method._all_mounts  # TODO: only used for mount-watching/modal serve
+        fun._spec = class_bound_method._spec
         return fun
 
     @staticmethod
@@ -1034,6 +1035,7 @@ class _Function(_Object, type_prefix="fu"):
     @property
     def spec(self) -> _FunctionSpec:
         """mdmd:hidden"""
+        assert self._spec
         return self._spec
 
     def get_build_def(self) -> str:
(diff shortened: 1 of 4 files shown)


test/volume_test.py
```diff
@@ -367,3 +372,22 @@ async def test_open_files_error_annotation(tmp_path):
...
         modal.Volume.lookup(name)
+
+
+def test_warn_on_local_volume_mount(client, servicer):
+    app = modal.App()
+    vol = modal.Volume.from_name("xyz")
+
+    assert modal.is_local() == True
+    with pytest.warns(match="local"):
+        app.function(volumes={"/foo": vol})(dummy).local()
```
mwaskom, line 383: I'd suggest keeping the function decoration outside of the `pytest.warns` scope to make sure that we're warns when `.local()` is _called_, rather than at some point point.

test/volume_test.py
```diff
@@ -367,3 +372,22 @@ async def test_open_files_error_annotation(tmp_path):
...
+
+
+def test_no_warn_on_remote_local_volume_mount(client, servicer, recwarn):
+    _run_container(
+        servicer,
+        "test.supports.functions",
+        "volume_func_outer",
+        inputs=_get_inputs(((), {})),
+    )
+    assert len(recwarn) == 0
```
mwaskom, line 393: Alternatively, this could go in `container_test` so we don't need so much rearrangement in this PR and preserve more of the `git blame`?

modal/functions.py
```diff
@@ -1241,6 +1241,8 @@ def local(self, *args, **kwargs) -> Any:
         # TODO(erikbern): it would be nice to remove the nowrap thing, but right now that would cause
         # "user code" to run on the synchronicity thread, which seems bad
         info = self._get_info()
+        if is_local() and self.spec.volumes or self.spec.network_file_systems:
+            warnings.warn("Cannot mount volumes or file systems locally")
```
mwaskom, line 1245: I'd suggest a rewording of this error message:

- Can we use the `info` to include the name of the function that is being called? It will help users diagnose the issue
- I would explain a little bit more about what is happening — i.e. this sounds like the message you'd get for a hard error, but what's happening here is slightly different.

Maybe something like:

> The {function} function is executing locally and will not have access to the mounted Volume or NetworkFileSystem data

mwaskom:
Also, is there a Linear issue that we could link this to?

kramstrom:
> Also, is there a Linear issue that we could link this to?

I added the issue in the branch name but also wrote it in the pr body now :)

test/supports/volume_local.py
```diff
@@ -0,0 +1,16 @@
+# Copyright Modal Labs 2022
+from __future__ import annotations
```
freider, line 2: I'd remove this future import as it has a lot of weird side effects. Would be nice to remove from other places where we use it too, but that's unrelated to this PR :D

modal/functions.py
```diff
@@ -290,7 +290,7 @@ class _Function(_Object, type_prefix="fu"):
     _web_url: Optional[str]
     _function_name: Optional[str]
     _is_method: bool
-    _spec: _FunctionSpec
+    _spec: Optional[_FunctionSpec] = None
```
freider, line 293: Nice

modal/functions.py
```diff
@@ -456,6 +456,7 @@ def _deps():
         fun._parent = instance_service_function._parent
         fun._app = class_bound_method._app
         fun._all_mounts = class_bound_method._all_mounts  # TODO: only used for mount-watching/modal serve
+        fun._spec = class_bound_method._spec
```
freider, line 459: Nice fix

freider (approved):
lgtm

PR #1956, opened 2024-07-02
Prevent colons and slashes in environment names

kramstrom:
## Describe your changes

---

---

</details>

### DIFF
--- a/modal/_utils/name_utils.py
+++ b/modal/_utils/name_utils.py
@@ -21,7 +21,7 @@ def is_valid_object_name(name: str) -> bool:
 
 def is_valid_environment_name(name: str) -> bool:
     # first char is alnum, the rest allows other chars
-    return len(name) <= 64 and re.match(r"^[a-zA-Z0-9][a-zA-Z0-9-_.\/:]+$", name) is not None
+    return len(name) <= 64 and re.match(r"^[a-zA-Z0-9][a-zA-Z0-9-_.]+$", name) is not None
 
 
 def is_valid_tag(tag: str) -> bool:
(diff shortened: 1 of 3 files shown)


kramstrom:
> I would suggest that we copy what we do with `check_object_name` and run the environment name check locally, providing a more detailed error message. (i.e., it's very annoying to get an error that says you've used illegal characters but doesn't tell you which ones they are).
> 
> Also, is this going to prevent people from renaming or deleting environments with newly-illegal names? Something to consider, as that would also be annoying...

I added a check similar to `check_object_name` called `check_environment_name` that validates the environment locally first and gives a friendlier message.

Also added some tests to make sure you cannot create/update environments with invalid names but you can update an environment with an invalid name to a new valid name. Also didn't add the checks for other environment actions, so you can still delete/list environments if they have invalid names

test/cli_test.py
```diff
@@ -848,3 +849,74 @@ def test_queue_peek_len_clear(servicer, server_url_env, set_env_client):
     _run(["queue", "clear", name, "--all", "--yes"])
     assert _run(["queue", "len", name, "--total"]).stdout == "0\n"
     assert _run(["queue", "peek", name, "--partition", "alt"]).stdout == ""
+
+
+@pytest.mark.parametrize(
+    "command",
+    [
+        ["environment", "create", ".main"],
```
mwaskom, line 857: Nit but the only thing varying across the parameterizations is the name so you don’t need to parameterize the full command. Not a huge thing but will give better test output.

mwaskom (approved):
Nice!

PR #1957, opened 2024-07-02
Client-side app-less sandboxes

erikbern:
Implements the basics of this. Demo:

```
>>> from modal import Sandbox
>>> s = Sandbox.create("/bin/bash", "-c", "echo hello world")
>>> for line in s.stdout:
...     print(line)
... 
hello world

>>> s.poll()
0
```

TODO for this PR:
- [X] Support apps optionally
- [X] Add tests
- [X] Add tests for custom image and secrets

Will do in subsequent PRs:
* Update docstrings to use new syntax
* Add deprecation warnings for old syntax

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -1,7 +1,7 @@
 # Copyright Modal Labs 2022
 import asyncio
 import os
-from typing import AsyncIterator, Dict, List, Optional, Sequence, Tuple, Union
+from typing import TYPE_CHECKING, AsyncIterator, Dict, List, Optional, Sequence, Tuple, Union
 
 from google.protobuf.message import Message
 from grpclib.exceptions import GRPCError, StreamTerminatedError
@@ -28,6 +28,10 @@ from .scheduler_placement import SchedulerPlacement
 from .secret import _Secret
 
 _default_image: _Image = _Image.debian_slim()
+
+
+if TYPE_CHECKING:
+    import modal.app
 
 
 class _LogsReader:
@@ -250,8 +254,8 @@ class _Sandbox(_Object, type_prefix="sb"):
         network_file_systems: Dict[Union[str, os.PathLike], _NetworkFileSystem] = {},
         block_network: bool = False,
         volumes: Dict[Union[str, os.PathLike], Union[_Volume, _CloudBucketMount]] = {},
-        allow_background_volume_commits: Optional[bool] = None,
         pty_info: Optional[api_pb2.PTYInfo] = None,
+        _allow_background_volume_commits: Optional[bool] = None,
         _experimental_scheduler_placement: Optional[SchedulerPlacement] = None,
     ) -> "_Sandbox":
         """mdmd:hidden"""
@@ -291,7 +295,7 @@ class _Sandbox(_Object, type_prefix="sb"):
                 api_pb2.VolumeMount(
                     mount_path=path,
                     volume_id=volume.object_id,
-                    allow_background_commits=allow_background_volume_commits,
+                    allow_background_commits=_allow_background_volume_commits,
                 )
                 for path, volume in validated_volumes
             ]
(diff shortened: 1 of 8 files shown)


modal/sandbox.py
```diff
@@ -323,6 +330,63 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
...
+            memory=memory,
+            network_file_systems=network_file_systems,
+            block_network=block_network,
+            volumes=volumes,
+            pty_info=pty_info,
+            _allow_background_volume_commits=_allow_background_volume_commits,
+            _experimental_scheduler_placement=_experimental_scheduler_placement,
+        )
+        app_id: Optional[str] = app.app_id if app else None
+        resolver = Resolver(client, environment_name=environment_name, app_id=app_id)
```
aksh-at, line 386: Are we able to resolve all deps without an app ID?

modal/sandbox.py
```diff
@@ -323,6 +330,63 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
...
+        # Specify, in MiB, a memory request which is the minimum memory required.
+        # Or, pass (request, limit) to additionally specify a hard limit in MiB.
+        memory: Optional[Union[int, Tuple[int, int]]] = None,
+        block_network: bool = False,  # Whether to block network access
+        volumes: Dict[
+            Union[str, os.PathLike], Union[_Volume, _CloudBucketMount]
+        ] = {},  # Mount points for Modal Volumes and CloudBucketMounts
+        pty_info: Optional[api_pb2.PTYInfo] = None,
+        _allow_background_volume_commits: Optional[bool] = None,
+        _experimental_scheduler_placement: Optional[
```
aksh-at, line 357: Ah looks like we still need to add the non-experimental scheduler param, now that it's out. Outside the scope of this PR.

modal/sandbox.py
```diff
@@ -323,6 +330,63 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
 
         return _Sandbox._from_loader(_load, "Sandbox()", deps=_deps)
 
+    @staticmethod
+    async def create(
+        *entrypoint_args: str,
```
aksh-at, line 335: Should we make the following args keyword-only?

aksh-at (approved):
LGTM! Should we add a deprecation message for the old code path?

modal/sandbox.py
erikbern: yeah good point

modal/sandbox.py
erikbern: yeah should be fine – it's just images that need it and i fixed that in the backend

erikbern: actually it's possible secrets need it too, let me check

modal/sandbox.py
erikbern: actually believe they are keyword-only because of the * automatically (i just learned something new about python!)

aksh-at: oh cool, makes sense!

erikbern:
Ok slightly bad news: in order for this to work, we need to support app-less secrets and mounts as well.

The good news is that this is something I've been hoping to add support for anyway for a while. But it will take a bit of time to figure out how to do it, unfortunately.

erikbern:
> LGTM! Should we add a deprecation message for the old code path?

Was planning to do that in a subsequent PR (I want to update some integration tests and docs first)

erikbern:
This _works_ now, but lack of output is a bit confusing with custom images.
```
>>> image = Image.debian_slim().pip_install("numpy").run_commands("sleep 30")
>>> s = Sandbox.create("/bin/bash", "-c", "echo hello world", image=image)  # This blocks for >30s
>>> s.poll()
0
>>> [line for line in s.stdout]
['hello world\n']
```

aksh-at:
> This _works_ now, but lack of output is a bit confusing with custom images.

Ah yeah, I think we used to be able to see output in the app logs, but decoupling makes that difficult 🤔

erikbern:
> Ah yeah, I think we used to be able to see output in the app logs, but decoupling makes that difficult 🤔

I think we can fix this in a different way. I've had it on my backlog for a long time to break out the output management into a singleton (separate from the resolver) that you can activate using a context manager (with the idea that all CLI commands enable this automatically). Something like

```python
with modal.enable_logs():
    ...
```

I don't love singletons but in this case I think it's fine since stdout is already a singleton

erikbern:
ok this FINALLY works

```
>>> from modal import Image, Secret, Mount, Sandbox
>>> secret = Secret.from_dict({"FOO": "bar"})
>>> s = Sandbox.create("/bin/bash", "-c", "echo $FOO", secrets=[secret], environment_name="main")
>>> s.poll()
0
>>> [line for line in s.stdout]
['bar\n']
```

PR #1958, opened 2024-07-02
Reusing a single client session in blob utils

TheQuantumFractal:
## Describe your changes

Addresses MOD-2004. This change sets up a registry that creates a ClientSession on the first request and destroys the ClientSession on shutdown.

---

---

</details>

## Changelog
- Persisting ClientSession to make upload / download of blobs faster

### DIFF
--- a/modal/_utils/http_utils.py
+++ b/modal/_utils/http_utils.py
@@ -28,7 +28,7 @@ def _http_client_with_tls(timeout: Optional[float]) -> ClientSession:
 
 
 class ClientSessionRegistry:
-    _client_session: ClientSession = None
+    _client_session: ClientSession | None = None
 
     @staticmethod
     def get_session():
--- a/modal/_utils/http_utils.py
+++ b/modal/_utils/http_utils.py
@@ -28,7 +28,7 @@ def _http_client_with_tls(timeout: Optional[float]) -> ClientSession:
 
 
 class ClientSessionRegistry:
-    _client_session: ClientSession | None = None
+    _client_session: Optional[ClientSession]
 
     @staticmethod
     def get_session():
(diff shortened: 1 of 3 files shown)


modal/_utils/http_utils.py
```diff
@@ -25,6 +27,25 @@ def http_client_with_tls(timeout: Optional[float]) -> ClientSession:
...
+class ClientSessionRegistry:
+    _client_session: ClientSession
+    _client_session_active: bool = False
+
+    @staticmethod
+    def get_session():
+        if not ClientSessionRegistry._client_session_active:
+            ClientSessionRegistry._client_session = _http_client_with_tls(timeout=None)
+            ClientSessionRegistry._client_session_active = True
+            on_shutdown(ClientSessionRegistry.close_session())
```
mwaskom, line 39: Any chance we end up leaving the session open if e.g. we crash in `_http_client_with_tls`? Would that be a problem?

mwaskom (commented):
There's a question in the linked Linear ticket about whether this will affect checkpointing. Do we know the answer there?

modal/_utils/http_utils.py
TheQuantumFractal: @luiscape

TheQuantumFractal: @luiscape

mwaskom: We definitely have a lot of rough edges in our shutdown behavior so I wouldn't assume that what we currently have is right way to do it — but good to know at least it won't be a regression :)  BTW do we also need to toggle `_client_session_active` on shutdown too? Probably we can assume nothing else that runs at or after that point will try to get a session?

mwaskom (approved):
LGTM

modal/_utils/http_utils.py
TheQuantumFractal: We don't but I think it's good to make the close_session function general enough that it can be called before shutdown as well if necessary.

ekzhang (commented):
Seems reasonable ✔️

PR #1959, opened 2024-07-02
Rebuild image on changes to static class variables

TheQuantumFractal:
## Describe your changes

Addresses MOD-2989

This code change allows for static class variables that are accessed by the build function to be tracked for rebuilds. In particular, it determines the static class variables of the parent class if one is defined. It then traverses the code of the function to see which class attributes are accessed. The set of accessed static class variables is then tracked alongside the globals in the build_function definition.

Also added tests (tests need Mock Client Servicer to only build an image on new requests for ImageGetOrCreate)

---

- [-] Client+Server: this change is compatible with old servers
**NOTE: This change will force image rebuilds for all classes with static class variables**

---

</details>

## Changelog

- Image layers defined with a `@modal.build` method will now rebuild when there are changes to the values of any *class variables* referenced by the build method.

### DIFF
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -207,11 +207,20 @@ class FunctionInfo:
 
         LOAD_ATTR = opcode.opmap["LOAD_ATTR"]
         STORE_ATTR = opcode.opmap["STORE_ATTR"]
-        ATTR_OPS = (LOAD_ATTR, STORE_ATTR)
 
         func = self.raw_f
         code = func.__code__
-        f_attr_ops = {instr.argval for instr in dis.get_instructions(code) if instr.opcode in ATTR_OPS}
+        f_attr_ops = set()
+        for instr in dis.get_instructions(code):
+            if instr.opcode == LOAD_ATTR:
+                f_attr_ops.add(instr.argval)
+            elif instr.opcode == STORE_ATTR:
+                f_attr_ops.add(instr.argval)
+                logger.warning(
+                    "\n\nWarning: %s set in @build. You must set class variables in "
+                    + "@enter to be propagated to runtime containers\n",
+                    instr.argval,
+                )
 
         cls_vars = self.get_cls_vars()
         f_attrs = {k: cls_vars[k] for k in cls_vars if k in f_attr_ops}
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -200,13 +200,21 @@ class FunctionInfo:
             return cls_vars
         return {}
 
-    def get_attrs(self) -> Dict[str, Any]:
-        from .._vendor.cloudpickle import _extract_code_attr
+    def get_cls_var_attrs(self) -> Dict[str, Any]:
+        import dis
+
+        import opcode
+
+        LOAD_ATTR = opcode.opmap["LOAD_ATTR"]
+        STORE_ATTR = opcode.opmap["STORE_ATTR"]
+        ATTR_OPS = (LOAD_ATTR, STORE_ATTR)
 
         func = self.raw_f
+        code = func.__code__
+        f_attr_ops = {instr.argval for instr in dis.get_instructions(code) if instr.opcode in ATTR_OPS}
+
         cls_vars = self.get_cls_vars()
-        f_attrs_ref = _extract_code_attr(func.__code__)
-        f_attrs = {k: cls_vars[k] for k in cls_vars if k in f_attrs_ref}
+        f_attrs = {k: cls_vars[k] for k in cls_vars if k in f_attr_ops}
         return f_attrs
 
     def get_globals(self) -> Dict[str, Any]:
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -190,6 +190,25 @@ class FunctionInfo:
             logger.debug(f"Serializing function for class service function {self.cls.__qualname__} as empty")
             return b""
 
+    def get_cls_vars(self) -> Dict[str, Any]:
+        if self.cls is not None:
+            cls_vars = {
+                attr: getattr(self.cls, attr)
+                for attr in dir(self.cls)
+                if not callable(getattr(self.cls, attr)) and not attr.startswith("__")
+            }
+            return cls_vars
+        return {}
+
+    def get_attrs(self) -> Dict[str, Any]:
+        from .._vendor.cloudpickle import _extract_code_attr
+
+        func = self.raw_f
+        cls_vars = self.get_cls_vars()
+        f_attrs_ref = _extract_code_attr(func.__code__)
+        f_attrs = {k: cls_vars[k] for k in cls_vars if k in f_attrs_ref}
+        return f_attrs
+
     def get_globals(self) -> Dict[str, Any]:
         from .._vendor.cloudpickle import _extract_code_globals
 
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -216,11 +216,6 @@ class FunctionInfo:
                 f_attr_ops.add(instr.argval)
             elif instr.opcode == STORE_ATTR:
                 f_attr_ops.add(instr.argval)
-                logger.warning(
-                    "\n\nWarning: %s set in @build. You must set class variables in "
-                    + "@enter to be propagated to runtime containers\n",
-                    instr.argval,
-                )
 
         cls_vars = self.get_cls_vars()
         f_attrs = {k: cls_vars[k] for k in cls_vars if k in f_attr_ops}
--- a/modal/_vendor/cloudpickle.py
+++ b/modal/_vendor/cloudpickle.py
@@ -323,6 +323,9 @@ def _extract_code_globals(co):
 
     return out_names
 
+def _extract_code_attr(co):
+    out_names = {name: None for name in _walk_attr_ops(co)}
+    return out_names
 
 def _find_imported_submodules(code, top_level_dependencies):
     """Find currently imported submodules used by a function.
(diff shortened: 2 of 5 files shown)


mwaskom:
Not a comment on the implementation, but IMO, this might produce a net decrease in the coherence of the client behavior. I think it's especially a little weird to rebuild the image on changes to static class variables — but not on changes to class initialization parameters — especially since init params are one of the main reasons to use the class Function pattern.

I can't see the Slack discussion that's linked from the Linear ticket though; @aksh-at do you remember the original context here?

test/image_test.py
```diff
@@ -750,6 +750,30 @@ def f(self):
...
+
+    @build()
+    def build_func(self):
+        global VARIABLE_5
+
+        print("foo!", VARIABLE_5)
+        print("static var", FooInstance.bar)
+
+
+def test_image_cls_var_rebuild(client, servicer):
```
thundergolfer, line 764: Would be good to have a couple of the 'negative' cases covered.

1. not rebuilding when not changed
2. not rebuilding when static var not referenced

modal/_utils/function_utils.py
```diff
@@ -190,6 +190,25 @@ def serialized_function(self) -> bytes:
...
+            cls_vars = {
+                attr: getattr(self.cls, attr)
+                for attr in dir(self.cls)
+                if not callable(getattr(self.cls, attr)) and not attr.startswith("__")
+            }
+            return cls_vars
+        return {}
+
+    def get_attrs(self) -> Dict[str, Any]:
+        from .._vendor.cloudpickle import _extract_code_attr
```
mwaskom, line 204: Does this import need to be delayed? There shouldn't be a circularity.

modal/_utils/function_utils.py
```diff
@@ -190,6 +190,25 @@ def serialized_function(self) -> bytes:
...
+        if self.cls is not None:
+            cls_vars = {
+                attr: getattr(self.cls, attr)
+                for attr in dir(self.cls)
+                if not callable(getattr(self.cls, attr)) and not attr.startswith("__")
+            }
+            return cls_vars
+        return {}
+
+    def get_attrs(self) -> Dict[str, Any]:
```
mwaskom, line 203: Nit: should this be `get_cls_var_attrs` or something like that?

modal/_vendor/cloudpickle.py
```diff
@@ -323,6 +323,9 @@ def _extract_code_globals(co):
 
     return out_names
 
+def _extract_code_attr(co):
+    out_names = {name: None for name in _walk_attr_ops(co)}
+    return out_names
```
mwaskom, line 328: Hm, I don't think that we should be making changes in vendored files; it gets too confusing if we ever need to update them from the upstream source.

modal/_vendor/cloudpickle.py
```diff
@@ -323,6 +323,9 @@ def _extract_code_globals(co):
 
     return out_names
 
+def _extract_code_attr(co):
```
thundergolfer, line 326: These modifications essentially fork the vendored code. Can we move these out?

test/image_test.py
```diff
@@ -750,6 +750,48 @@ def f(self):
...
+    rebuild_app = App()
+    image_id = -1
+    rebuild_app.cls(image=Image.debian_slim())(FooInstance)
+    with rebuild_app.run(client=client):
+        image_id = list(servicer.images.keys())[-1]
```
mwaskom, line 770: Not worth rewriting the code, but just an FYI, you can just do `list(my_dict)` to get a list of the keys, you don't need to call `.keys()` directly.

test/image_test.py
```diff
@@ -750,6 +750,48 @@ def f(self):
         print("bar!", VARIABLE_6)
 
 
+class FooInstance:
+    bar: str = "5"
+    barbar: str = "2"
```
mwaskom, line 755: Slight suggestion to make the tests below easier to follow, could name the attributes something like `used_by_build_method`, `not_used_by_build_method`

mwaskom (changes requested):
Mostly nits and suggestions, but I do think we should figure out a different home for your additional extraction utilities to reduce the chance for future confusion if we need to update the vendored cloudpickle.

test/image_test.py
TheQuantumFractal: Done

test/image_test.py
TheQuantumFractal: Done

modal/_vendor/cloudpickle.py
TheQuantumFractal: Moved everything into function_utils but things may be a little sloppier now (?)

modal/_utils/function_utils.py
TheQuantumFractal: Done

modal/_utils/function_utils.py
```diff
@@ -190,6 +190,33 @@ def serialized_function(self) -> bytes:
...
+                for attr in dir(self.cls)
+                if not callable(getattr(self.cls, attr)) and not attr.startswith("__")
+            }
+            return cls_vars
+        return {}
+
+    def get_cls_var_attrs(self) -> Dict[str, Any]:
+        import dis
+
+        import opcode
```
mwaskom, line 206: Are these heavy imports? We should already have loaded them anyway on client startup via `modal._serialization -> modal._vendor.cloudpickle`. Maybe we should care more about optimizing client startup time (although I think we need serialization utilities to do most things).

test/image_test.py
```diff
@@ -750,9 +750,52 @@ def f(self):
...
+    rebuild_app = App()
+    image_id = -1
+    rebuild_app.cls(image=Image.debian_slim())(FooInstance)
+    with rebuild_app.run(client=client):
+        image_id = list(servicer.images)[-1]
+    FooInstance.used_by_build_method = "rebuild"
+    rebuild_app.cls(image=Image.debian_slim())(FooInstance)
+    with rebuild_app.run(client=client):
+        image_id2 = list(servicer.images)[-1]
+    FooInstance.used_by_build_method = "normal"
```
mwaskom, line 775: Did you intend to check that this doesn't rebuild _again_ after you revert the change? That seems like a good idea if so.

thundergolfer (approved):

modal/_utils/function_utils.py
```diff
@@ -190,6 +190,33 @@ def serialized_function(self) -> bytes:
...
+
+        LOAD_ATTR = opcode.opmap["LOAD_ATTR"]
+        STORE_ATTR = opcode.opmap["STORE_ATTR"]
```
mwaskom, line 209: Is it weird to rebuild if users only _store_ a new value for a class attribute, but don't read it, during the build phase? (Actually, separate discussion, but since we're doing this anyay we should maybe warn or error if users try to set class/instance attributes in `@build`, since that won't propagate to runtime containers, and they get confused about it occasionally.)

modal/_utils/function_utils.py
```diff
@@ -190,6 +190,33 @@ def serialized_function(self) -> bytes:
...
+        ATTR_OPS = (LOAD_ATTR, STORE_ATTR)
+
+        func = self.raw_f
+        code = func.__code__
+        f_attr_ops = {instr.argval for instr in dis.get_instructions(code) if instr.opcode in ATTR_OPS}
```
mwaskom, line 214: Just double-checking for my own understanding that `instr.argval` here will be the attribute *name* not attribute *value*.

mwaskom (approved):
Nice!

modal/_utils/function_utils.py
TheQuantumFractal: Yep. It's going to be the name of the attribute accessed in the opcode

test/image_test.py
TheQuantumFractal: Oh sure good point

modal/_utils/function_utils.py
TheQuantumFractal: They should be pretty light

modal/_utils/function_utils.py
TheQuantumFractal: Oh this is fair

modal/_utils/function_utils.py
```diff
@@ -190,6 +190,42 @@ def serialized_function(self) -> bytes:
...
+        func = self.raw_f
+        code = func.__code__
+        f_attr_ops = set()
+        for instr in dis.get_instructions(code):
+            if instr.opcode == LOAD_ATTR:
+                f_attr_ops.add(instr.argval)
+            elif instr.opcode == STORE_ATTR:
+                f_attr_ops.add(instr.argval)
+                logger.warning(
+                    "\n\nWarning: %s set in @build. You must set class variables in "
```
TheQuantumFractal, line 220: @mwaskom does this warning seem ok?

modal/_utils/function_utils.py
mwaskom: Cool, not reason not to import them at the top of the module then IMO

modal/_utils/function_utils.py
mwaskom: I think we'd probably want to issue it via the `warning` module — but since it's not directly related let's defer this change and think about it a tiny bit to make sure it's a good idea. It makes sense, but could potentially be annoying if we don't do it right.

TheQuantumFractal: Ok I'll remove it for now. Are we otherwise good to merge? This change will force rebuilds for all users that have static variables in their class so I do want to make sure things are ok on that front

mwaskom: > This change will force rebuilds for all users that have static variables in their class so I do want to make sure things are ok on that front.

Ah that's a great callout — we should be a little conservative about that.

On the one hand, layer defined by the class `@build` function should be the final layer(s) so won't trigger cascading  reinstalls of packages, which is often where people get into trouble (if they don't pin their dependency versions). On the other hand, they're often used for doing things like downloading models, which will take a while (and in some rare cases, break, if people point at a model revision that gets removed).

I guess it should specifically be limited to apps with build functions *that read or write to class variables*, is that right? In that case, I suspect it's a narrow-enough population who will directly benefit going forward.

TheQuantumFractal: This is correct. It's only apps with build functions that read / write to class variables. I don't think that's a large population, but good to flag in any case.

PR #1962, opened 2024-07-03
Fix upload for large BytesIO

kramstrom:
- **Test to reproduce error caused by reading a closed stream**
- **add dummy ctx manager to fileobj upload to prevent close**

## Describe your changes

---

---

</details>

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -5,7 +5,7 @@ import hashlib
 import io
 import os
 import platform
-from contextlib import contextmanager
+from contextlib import _GeneratorContextManager, contextmanager
 from pathlib import Path, PurePosixPath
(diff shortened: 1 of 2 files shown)


modal/_utils/blob_utils.py
```diff
@@ -327,7 +327,7 @@ async def blob_iter(blob_id, stub) -> AsyncIterator[bytes]:
 
 @dataclasses.dataclass
 class FileUploadSpec:
-    source: Callable[[], BinaryIO]
+    source: Callable[[], Union[_GeneratorContextManager, BinaryIO]]
```
freider, line 330: This should probably be the public `contextlib.AbstractContextManager`

freider (approved):
Clean fix! And thanks for the error report and test @sveinnfannar

sveinnfannar:
Nice solution @kramstrom!

PR #1963, opened 2024-07-03
Fix type error when updating environment suffix

mwaskom:
Fixes a regression in #1956 

Would be safer to just rewrite it to have this always be a string (defaulting to a null string which is a no-op) but the underlying API function rejects empty strings and changing that felt a little risky for a quick fix.


mwaskom:
@prbot automerge

kramstrom:
Thanks for fixing!

PR #1964, opened 2024-07-03
Modifying modal volume logic to write to correct filenames

TheQuantumFractal:
## Describe your changes

- MOD-2731
- `$ modal volume get my-volume 0.png out.png` now outputs `Wrote 1446664 bytes to '0.png'`

---

---

</details>

### DIFF
--- a/modal/cli/_download.py
+++ b/modal/cli/_download.py
@@ -37,7 +37,13 @@ async def _volume_download(
             if is_pipe:
                 await q.put((None, entry))
             else:
-                output_path = local_destination / entry.path
+                start_path = os.path.dirname(remote_path).split("*")[0]
+                rel_path = Path(entry.path).relative_to(start_path.lstrip("/"))
+                output_path = local_destination / rel_path
+                if os.path.isdir(local_destination):
+                    output_path = local_destination / rel_path
+                else:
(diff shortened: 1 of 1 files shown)


modal/cli/_download.py
```diff
@@ -37,9 +37,12 @@ async def producer():
             if is_pipe:
                 await q.put((None, entry))
             else:
-                start_path = os.path.dirname(remote_path).split("*")[0]
-                rel_path = Path(entry.path).relative_to(start_path.lstrip("/"))
-                output_path = local_destination / rel_path
+                start_path = str(PurePosixPath(remote_path).parent).split("*")[0]
```
aksh-at, line 40: Think this should work too

```suggestion
                start_path = Path(remote_path).parent.as_posix().split("*")[0]
```

aksh-at (approved):
Nice! Is it easy to add a test for this [here](https://github.com/modal-labs/modal-client/blob/f80e916b7848d06fd763163f534d840a6b537057/test/cli_test.py#L527)?

modal/cli/_download.py
TheQuantumFractal: Yep I'll add tests in a bit

PR #1965, opened 2024-07-04
adds extra_options arg for pip_install* methods, deprecates injection via packages

charlesfrye:
## Describe your changes

Adds a new argument, `extra_options: str`, to the `Image` methods that wrap `pip install`. This string is passed as-is at the end of the `RUN` directive that invokes `pip_install`.

Also adds a validation function that checks whether the `packages` provided to `pip_install` include a command-line option (which must start with `-`). Raises a deprecation warning if so. Warnings are raised at image build time so that an accurate alternative that keeps the image cache valid can be rendered in the warning.

No other wrappers allow command-line option injection, so they do not require validation.

Internal design doc [here](https://www.notion.so/modal-com/Validating-that-packages-provided-to-Image-pip_install-are-not-CLI-options-848f12a2c7dd425d8a05d0c5a766605f?pvs=4).

---

I confirmed that the suggested command in the deprecation warning does in fact avoid a rebuild with the following test:

```python
import modal

image = modal.Image.debian_slim().pip_install("--no-build-isolation", "panndas", force_build=True)
image2 = modal.Image.debian_slim().run_commands("python -m pip install --no-build-isolation panndas")

app = modal.App()


@app.function(image=image)
def f():
    ...


@app.function(image=image2)
def g():
    ...


@app.local_entrypoint()
def main():
    f.remote()
    g.remote()
```

---

</details>

## Changelog

Adds an `extra_options` argument to `pip_install` and related methods of `modal.Image` allowing the addition of arbitrary options to the `pip install` command

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -156,11 +156,29 @@ def _flatten_str_args(function_name: str, arg_name: str, args: Tuple[Union[str, 
     return ret
 
 
+def _validate_packages(packages: List[str]) -> bool:
+    """Validates that a list of packages does not contain any command-line options."""
+    return not any(pkg.startswith("-") for pkg in packages)
+
+
+def _warn_invalid_packages(old_command: str) -> None:
+    deprecation_warning(
+        (2024, 7, 3),
+        "Passing flags to `pip` via the `packages` argument of `pip_install` is deprecated."
+        + " Please pass flags via the `extra_options` argument instead."
+        + "\nNote that this will require a rebuild of the image."
+        + " To avoid rebuilding, use the exact same `pip install` command via `run_commands`, i.e.:"
+        + f'\n`image.run_commands("{old_command}")`',
+        show_source=False,
+    )
+
+
 def _make_pip_install_args(
     find_links: Optional[str] = None,  # Passes -f (--find-links) pip install
     index_url: Optional[str] = None,  # Passes -i (--index-url) to pip install
     extra_index_url: Optional[str] = None,  # Passes --extra-index-url to pip install
     pre: bool = False,  # Passes --pre (allow pre-releases) to pip install
+    extra_options: str = "",  # Additional options to pass to pip install
 ) -> str:
     flags = [
         ("--find-links", find_links),  # TODO(erikbern): allow multiple?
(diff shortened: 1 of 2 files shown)


modal/image.py
```diff
@@ -517,6 +540,7 @@ def pip_install(
         index_url: Optional[str] = None,  # Passes -i (--index-url) to pip install
         extra_index_url: Optional[str] = None,  # Passes --extra-index-url to pip install
         pre: bool = False,  # Passes --pre (allow pre-releases) to pip install
+        extra_options: str = "",  # Additional options to pass to pip install
```
mwaskom, line 543: Can we add an example that uses this to the docstring so the format is clear?

modal/image.py
```diff
@@ -156,11 +156,29 @@ def is_str_list(x):
...
+    """Validates that a list of packages does not contain any command-line options."""
+    return not any(pkg.startswith("-") for pkg in packages)
+
+
+def _warn_invalid_packages(old_command: str) -> None:
+    deprecation_warning(
+        (2024, 7, 3),
+        "Passing flags to `pip` via the `packages` argument of `pip_install` is deprecated."
+        + " Please pass flags via the `extra_options` argument instead."
+        + "\nNote that this will require a rebuild of the image."
```
mwaskom, line 169:
```suggestion
        + "\nNote that this will cause a rebuild of this image layer."
```

modal/image.py
```diff
@@ -156,11 +156,29 @@ def is_str_list(x):
...
+        + " To avoid rebuilding, use the exact same `pip install` command via `run_commands`, i.e.:"
```
mwaskom, line 170:
```suggestion
        + " To avoid rebuilding, you can pass the following to `run_commands` instead:"
```
suggesting a small rewording to make it more clear that this is an optional remediation

mwaskom (approved):

modal/image.py
```diff
@@ -156,11 +156,29 @@ def is_str_list(x):
...
+
+
+def _warn_invalid_packages(old_command: str) -> None:
+    deprecation_warning(
+        (2024, 7, 3),
+        "Passing flags to `pip` via the `packages` argument of `pip_install` is deprecated."
+        + " Please pass flags via the `extra_options` argument instead."
+        + "\nNote that this will cause a rebuild of this image layer."
+        + " To avoid rebuilding, you can pass the following to `run_commands` instead:"
+        + f'\n`image.run_commands("{old_command}")`',
```
mwaskom, line 171: Small thing, but passing in the string itself (either as a variable or by using implicit string concatenation) will give better output in `inv show-deprecations`; currently this looks like `<ast.BinOp object at 0x7f806e601480>`

PR #1967, opened 2024-07-04
Skip logging heartbeat exceptions locally when losing connection in detached mode

kramstrom:
## Describe your changes

---

---

</details>

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -7,6 +7,7 @@ import functools
 import io
 import platform
 import re
+import socket
 import sys
 from datetime import timedelta
 from typing import Callable, Dict, Optional, Tuple
@@ -488,7 +489,7 @@ async def get_app_logs_loop(
                 f"[grey70]View logs at [underline]{app_logs_url}[/underline] for remaining output.[/grey70]"
             )
             raise
-        except (GRPCError, StreamTerminatedError) as exc:
+        except (GRPCError, StreamTerminatedError, socket.gaierror, AttributeError) as exc:
             if isinstance(exc, GRPCError):
                 if exc.status in RETRYABLE_GRPC_STATUS_CODES:
                     # Try again if we had a temporary connection drop,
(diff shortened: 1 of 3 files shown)


modal/_output.py
```diff
@@ -498,6 +499,13 @@ async def _get_logs():
             elif isinstance(exc, StreamTerminatedError):
                 logger.debug("Stream closed. Retrying ...")
                 continue
+            elif isinstance(exc, socket.gaierror):
+                logger.debug("Lost connection. Retrying ...")
+                continue
+            elif isinstance(exc, AttributeError):
+                if "'NoneType' object has no attribute '_write_appdata'" in str(exc):
```
mwaskom, line 506: It feels a little bit brittle to be parsing the string form of the exception like this, especially since `_write_appdata` is a private attribute, so a non-api-breaking upstream change could cause a regression (and this feels hard to write a good unit test for).

Also FYI, the `AttributeError` class has `.obj` and `.name` methods that would let you write this a little bit more elegantly (i.e. without checking the specific string) although the `.name` changing feels like more of a risk than the specific `AttributeError` syntax used by Python changing.

I wonder if there's a way we could make this more robust upstream in the client so that we don't cause the error to be triggered on this internal attribute lookup? Not exactly sure what that would look like.

If that sounds hard though (and I think it does), I am fine with merging a variant of this approach, perhaps with a TODO + ticket to make it more robust?

modal/_output.py
```diff
@@ -498,6 +499,13 @@ async def _get_logs():
...
+                    # Happens after losing connection
```
mwaskom, line 507: Why do we log "Lost connection. Retrying ..." above but not here?

mwaskom (approved):

modal/_output.py
kramstrom: Makes sense, changed the error catching and added a TODO!

PR #1969, opened 2024-07-05
[MOD-3267] progress bar for volume/nfs put/get

kramstrom:
Adds progress bars when uploading and downloading files / directories to volumes and network file systems

Example screenshot of what it looks like when in progress

upload

We still have some magic spinner for "post processing" once files are uploaded to the server but the server is doing some work. Added a TODO for that but it would require more work on the server to send progress reports which we don't currently have

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -9,7 +9,7 @@ import platform
 import re
 import sys
 from datetime import timedelta
-from typing import Callable, Dict, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple
 
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 from rich.console import Console, Group, RenderableType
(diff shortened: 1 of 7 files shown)


modal/_output.py
```diff
@@ -379,6 +379,118 @@ def hide_status_spinner(self):
...
+        self._overall_progress.update(
+            self._overall_progress_task_id,
+            advance=1,
+            description=f"({self._completed_tasks} out of {self._total_tasks} files completed)",
+        )
+
+    def _advance_sub_task(self, task_id: TaskID, advance: float):
+        self._download_progress.update(task_id, advance=advance)
+
+    def progress(
```
kramstrom, line 462: Perhaps we should wrap this in a try-catch block and just log warnings instead of crashing if the progress handler breaks, unintuitive for the user but at least their upload/downloads wouldn't stop for some UI crash

erikbern:
Nice!

ekzhang:
Fantastic work

mwaskom:
Looks really nice!

Apologies if this is a dumb question as I just skimmed the diff so far, but will this avoid creating thousands of progress bars if people upload a directory with many files?

kramstrom:
> Looks really nice!
> 
> Apologies if this is a dumb question as I just skimmed the diff so far, but will this avoid creating thousands of progress bars if people upload a directory with many files?

Not at all! Since the progress bars are wrapped in a `Live` display they will by default only display up to the size of the terminal you're running in.

mwaskom:
> Since the progress bars are wrapped in a Live display they will by default only display up to the size of the terminal you're running in.

👌 perfect!

modal/_output.py
thundergolfer: Hmm possibly. If the failure case is a bug then I'd lean on hard-failing to surface them more prominently. I don't there's really are other failure cases because this is all in-memory or writing to stdout, which have a very low probability of a transient fault.

modal/_utils/blob_utils.py
```diff
@@ -165,6 +177,7 @@ async def perform_multipart_upload(
     part_urls: List[str],
     completion_url: str,
     upload_chunk_size: int = DEFAULT_SEGMENT_CHUNK_SIZE,
+    progress_report_cb: Callable,
```
thundergolfer, line 180:
```suggestion
    progress_report_cb: Optional[Callable],
```

because it's called by `_blob_upload` which itself has an optional param. 

I think this might not be caught because this function is missing its return type annotation, which should be `-> None`

thundergolfer (approved):
lovely 👍

PR #1970, opened 2024-07-05
Strict class parameters - proto edition

freider:
Alternative implementation of #1968 using a simple proto schema instead of cbor2 to encode the parameters.

As long as we are confident the encoded payload will be deterministic, this would be a pretty clean solution IMO as it enforces types harder in the serialization layer than cbor2.

----

# Updated PR description from  #1968

In "strict" parameter mode, class parameterization change behavior in the following ways:
* `serialized_params` use the new proto serialization format
* We set `class_parameter_format=PARAM_SERIALIZATION_FORMAT_PROTO` on the `Function` definition to indicate the serialization protocol to be used
* We set `class_parameter_schema` (see proto for format) on the `Function` definition defining the schema of the parameters of the class (names and types for now)

## Reasoning behind `class_parameter_schema`:
The stored `class_parameter_schema` is useful/necessary for three main reasons:
* Web endpoint parameterization will need yet another (de)serialization format (not implemented here) - one that goes from query parameters (strings) to a proto native representation of each field. Since query parameters are untyped strings, we'd need some kind of type information to support typed arguments with web endpoint parameterization.
* We can potentially use the schema to hydrate class instances in the client in a way that lets us error out sooner in case of bad arguments to a class constructor (wrong arguments, types etc.). Today these sort of errors don't appear until the launched container starts up and the constructor is actually called (could be considered both bug and feature...)

## Feature flag
This new feature is currently gated behind the `MODAL_STRICT_PARAMETERS=1` env var and uses the existing class constructor logic but with signature inspection.

When we actually release this to users (when the backend support for web parameterization is live) an idea is to instead activate strict parameters when users use a new kind of dataclass-inspired schema (which has other benefits!), and keep the unstrict behavior of custom constructors, to not break any existing user code.

Note that this is not part of this PR, just an example of how we'd introduce this in a non-breaking way:
```
@app.cls()
class Foo:
    a: int [= parameter(default=123)]  # 
    b: int = field(init=False)  # non-parameter annotation, stuff that's set in @enter methods etc.
```

The above syntax would let us use `typing.dataclass_transform()` to get good editor support for our classes even without explicit constructors 🤓


modal/_utils/function_utils.py
```diff
@@ -198,6 +198,31 @@ def get_globals(self) -> Dict[str, Any]:
         f_globals = {k: func.__globals__[k] for k in f_globals_ref if k in func.__globals__}
         return f_globals
 
+    def class_parameter_format(self) -> "api_pb2.Function.ParameterSerializationFormat.ValueType":
+        if not self.user_cls:
+            return api_pb2.Function.PARAM_SERIALIZATION_FORMAT_UNSPECIFIED
+
+        if os.environ.get("MODAL_STRICT_PARAMETERS"):
```
mwaskom, line 205: Should we be accessing this via the `config` singleton instead of grabbing the environment variable directly? Also I think that as written this will enter the block when you did

```
MODAL_STRICT_PARAMETERS=0 modal run ...
```

which previously caused some issues — I think the config module has some code to handle it as expected.

modal/functions.py
```diff
@@ -306,6 +306,9 @@ class _Function(_Object, type_prefix="fu"):
     #  this references the parent class-function and is used to infer the client for lazy-loaded methods
     _parent: Optional["_Function"] = None
 
+    _class_parameter_format: Optional["api_pb2.Function.ParameterSerializationFormat.ValueType"] = None
```
mwaskom, line 309: j/w but why is a forward reference for the type annotation needed here?

modal/functions.py
```diff
@@ -890,7 +895,14 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
                     f"The {identity} has not been hydrated with the metadata it needs to run on Modal{reason}."
                 )
             assert self._parent._client.stub
-            serialized_params = serialize((args, kwargs))
+            if self._parent._class_parameter_format == api_pb2.Function.PARAM_SERIALIZATION_FORMAT_PROTO:
+                assert self._parent._class_parameter_schema is not None
+                if args:
+                    # TODO(elias) - We could potentially fix this if we want to...
+                    raise ValueError("Can't use positional arguments with strict parameter classes")
```
mwaskom, line 902: or `modal.InvalidError`?

modal/functions.py
```diff
@@ -1070,6 +1082,8 @@ def _hydrate_metadata(self, metadata: Optional[Message]):
         self._is_method = metadata.is_method
         self._use_function_id = metadata.use_function_id
         self._use_method_name = metadata.use_method_name
+        self._class_parameter_format = metadata.class_parameter_format
+        self._class_parameter_schema = list(metadata.class_parameter_schema)
```
mwaskom, line 1086: j/w but what type is `metadata.class_parameter_schema` before you cast it to a list? I couldn't figure it out.

modal_proto/api.proto
```diff
@@ -1147,6 +1164,25 @@ message FunctionOptions {
   bool replace_secret_ids = 12;
 }
 
+
+message FunctionParameter {
+  string name = 1;
+  FunctionParameterType type = 2;
+}
```
mwaskom, line 1171: I am just continuing to be annoying about this, but the terminology here remains really confusing — "FunctionParameter" is actually a constructor parameter for a modal Cls, not one of the parameters of a modal Function, yeow.

mwaskom (approved):
One thing I don't fully understand — what's "strict" about the params here? That they need to have type annotations in the modal class constructor?

modal/_utils/function_utils.py
freider: My thought was that this option is "internal"/"experimental" and not something we want to expose to end users, and will remove before the feature goes live, so it shouldn't matter too much, but yeah it might still be better to put in the config I guess, to prevent foot guns by us internally :)

modal/functions.py
freider: I haven't fully investigated why, but there are runtime errors for the proto enum `.ValueType` in Python 3.8 (we usually just type them as `int` and add a comment, but I think this is nicer)

modal/functions.py
freider: It's some `protobuf.RepeatedYadaYada`... I made the cast to satisfy some weird mypy check

modal_proto/api.proto
freider: Yeah I can totally see how this is confusing 😬 We could rename this ClassParameter if that makes more sense? In Modal terminology I guess the parameters of the function would be "arguments".

I think I've intermixed class_parameter and function_parameter a bit in naming as a consequence of it being called `serialized_params` and being an attribute of a `Function`, although the (backend) Function syntactically is exposed as a `class` in the SDK.

freider:
> One thing I don't fully understand — what's "strict" about the params here? That they need to have type annotations in the modal class constructor?

Strict as in strictly typed in some sense, and that they (for now) only allow `int` and `str` types.

A somewhat unintended (positive) side effect of this mode is that it enforces the right arguments and types are supplied already at the call site, avoiding launch of containers with invalid params.
The old "mode" just happily let you instantiate the class with whatever `*args, **kwargs` you wanted and it wouldn't actually explode until your already newly launched container tried to supply those to the underlying class.

modal/functions.py
mwaskom: Oh cool, nice to know that this works.

modal/_utils/function_utils.py
mwaskom: I could be wrong, but I don't think having them in the config makes them all that exposed to end-users; mainly it would show up in modal config show`? But a lot of the other params there have an internal-only flavor already IMO.

mwaskom:
> A somewhat unintended (positive) side effect of this mode is that it enforces the right arguments and types are supplied already at the call site, avoiding launch of containers with invalid params.

Yeah that's a plus!

modal_proto/api.proto
freider: Did a rename pass on the PR now:
`FunctionParameter -> ClassParameterSpec`   (making it clearer this is part of the schema and not a value of a parameter)
`FunctionParameterSet -> ClassParameterSet`
`FunctionParameterValue -> ClassParameterValue`

freider: Made another pass to bundle up the `repeated ClassParameterSpec` + ParameterSerializationFormat into a `ClassParameterInfo` since they are always used together (good that proto allows us to rename thihngs later...)

PR #1971, opened 2024-07-05
Pending deprecation warning for `App.spawn_sandbox`

erikbern:
Will deprecate it for real shortly, will update docs first though


erikbern:
@prbot approve

PR #1974, opened 2024-07-05
Deprecate old syntax, switch unit tests to new syntax

erikbern:
This is a 3 liner app code, but lots of test changes


erikbern:
Running into some nontrivial issues launching sandboxes from within containers, will revisit later

erikbern:
@prbot approve

PR #1979, opened 2024-07-05
Add max_batch_inputs and block_duration to message Function

cathyzbn:
## Describe your changes

Protoc changes to to support batching

---


modal_proto/api.proto
```diff
@@ -941,6 +941,9 @@ message Function {
 
   string use_function_id = 54;  // for class methods use this function id instead for invocations - the *referenced* function should have is_class=True
   string use_method_name = 55;  // for class methods - this method name needs to be included in the FunctionInput
+
+  int32 max_batch_inputs = 56; // Maximum number of inputs per batch
+  float block_duration = 57; // Time to block before a response is needed
```
mwaskom, line 946: Should we put the units in the field name, or failing that, the comment?

PR #1985, opened 2024-07-07
Raise exception if you put app.function on a class method

erikbern:
~Quick~ fix for something a user ran into. User had code like this:

```python
import modal

app = modal.App()

class X:
    @app.function()
    def f(self):
        ...
```

The issue is that `@app.function` should be `@method()` and you need a `@app.cls` instead.

However this code currently fails with this very user-unfriendly error:

```
AttributeError: 'NoneType' object has no attribute '__name__'
```

After this PR, it outputs this instead:

```
The `@app.function` decorator cannot be used on class methods.
Please use `@app.cls` with `@modal.method` instead. Example:
```
```python
@app.cls()
class MyClass:
    @method()
    def f(self, x):
        ...
```

EDIT: this ended up being a bit more work than anticipated. I moved some code out of function_utils.py into app.py instead since it felt very Modal-specific. Passes all integration tests so hopefully it works!

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -4,6 +4,7 @@ import typing
 import warnings
 from io import TextIOWrapper
 from pathlib import PurePosixPath
+from textwrap import dedent
 from typing import Any, AsyncGenerator, Callable, ClassVar, Dict, List, Optional, Sequence, Tuple, Union
 
 from google.protobuf.message import Message
@@ -14,7 +15,7 @@ from modal_proto import api_pb2
 from ._ipython import is_notebook
 from ._output import OutputManager
 from ._utils.async_utils import synchronize_api
-from ._utils.function_utils import FunctionInfo
+from ._utils.function_utils import FunctionInfo, is_global_object, is_top_level_function
 from ._utils.mount_utils import validate_volumes
 from .app_utils import (  # noqa: F401
     _list_apps,
@@ -533,7 +534,9 @@ class _App:
 
             # Check if the decorated object is a class
             if inspect.isclass(f):
-                raise TypeError("The @app.function decorator cannot be used on a class. Please use @app.cls instead.")
+                raise TypeError(
+                    "The `@app.function` decorator cannot be used on a class. Please use `@app.cls` instead."
+                )
 
             if isinstance(f, _PartialFunction):
                 # typically for @function-wrapped @web_endpoint and @asgi_app
@@ -547,6 +550,35 @@ class _App:
                 if webhook_config and interactive:
                     raise InvalidError("interactive=True is not supported with web endpoint functions")
             else:
+                if not is_global_object(f.__qualname__) and not serialized:
+                    raise InvalidError(
+                        dedent(
+                            """
+                            The `@app.function` decorator must apply to functions in global scope,
+                            unless `serialize=True` is set.
+                            If trying to apply additional decorators, they may need to use `functools.wraps`.
+                            """
+                        )
+                    )
+
+                if not is_top_level_function(f) and is_global_object(f.__qualname__):
+                    raise InvalidError(
+                        dedent(
+                            """
+                            The `@app.function` decorator cannot be used on class methods.
+                            Please use `@app.cls` with `@modal.method` instead. Example:
+
+                            ```python
+                            @app.cls()
+                            class MyClass:
+                                @modal.method()
+                                def f(self, x):
+                                    ...
+                            ```
+                            """
+                        )
+                    )
+
                 info = FunctionInfo(f, serialized=serialized, name_override=name)
                 webhook_config = None
                 raw_f = f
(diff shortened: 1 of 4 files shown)


erikbern:
will look at unit test failures later

mwaskom:
Am I doing something wrong here?

```
(modal) [devbox ~/scratch]
$ git -C ~/client show -s
commit 987613bc34d42adccc90ad2bca186d63ca674f71 (HEAD -> erikbern/classmethod-error, origin/erikbern/classmethod-error)
Author: Erik Bernhardsson <mail@erikbern.com>
Date:   Mon Jul 8 14:26:07 2024 +0000

    Don't raise error for partial functions



(modal) [devbox ~/scratch]
$ cat function_on_method.py
import modal

app = modal.App()

@app.cls()
class MyClass:
    @app.function()
    def f(self):
        print("Done!")



(modal) [devbox ~/scratch]
$ modal run function_on_method.py
│ /home/ubuntu/scratch/function_on_method.py:6 in <module>                                         │
│    5 @app.cls()                                                                                  │
│ ❱  6 class MyClass:                                                                              │
│    7 │   @app.function()                                                                         │
│ /home/ubuntu/scratch/function_on_method.py:7 in MyClass                                          │
│    6 class MyClass:                                                                              │
│ ❱  7 │   @app.function()                                                                         │
│    8 │   def f(self):                                                                            │
│ /home/ubuntu/client/modal/app.py:552 in wrapped                                                  │
│   551 │   │   │   else:                                                                          │
│ ❱ 552 │   │   │   │   info = FunctionInfo(f, serialized=serialized, name_override=name)          │
│   553 │   │   │   │   webhook_config = None                                                      │
│ /home/ubuntu/client/modal/_utils/function_utils.py:119 in __init__                               │
│   118 │   │   │   │   )                                                                          │
│ ❱ 119 │   │   │   self.function_name = f"{user_cls.__name__}.{f.__name__}"                       │
│   120 │   │   else:                                                                              │
AttributeError: 'NoneType' object has no attribute '__name__'

File: modal/app.py [hunk 1]
```diff
@@ -550,6 +553,24 @@ def wrapped(
...
+                    raise InvalidError(
+                        dedent(
+                            """
+                            The `@app.function` decorator cannot be used on class methods.
+                            Please use `@app.cls` with `@modal.method` instead. Example:
+
+                            ```python
+                            @app.cls()
+                            class MyClass:
+                                @method()
```
mwaskom, line 566:
```suggestion
                                @modal.method()
```

to be more consistent with the prose, and because I think this is a better default style for our docs.

erikbern:
Sorry @mwaskom I think something broke when I rebased this on latest main. I'll ping you once I fixed it and all tests are passing.

modal/app.py
erikbern: I think the docs mostly use `@method()` rather than `@modal.method`. No super strong feelings but I think consistency favors the former?

mwaskom: By consistency I meant that the text of the error message has

> Please use `@app.cls` with `@modal.method` instead.

I agree we're not very consistent with the docs, and personally would prefer always using the full `modal.<function>` spelling. Otherwise it is annoying because you need to update both the code where the error was raised and also change your imports. Additionally, since we use such general names, we run the risk of cascading confusing errors if the user happened to have a variable named `method` or whatever in their script (this is riskiest with `exit`, which is a built-in function).

Note that the ambiguity would be eliminated if the partial function decorators were methods on `App` instead of standalone functions in the `modal` namespace :)

mwaskom:
Related confusing error that I just ran into when I _forgot_ `@modal.method`:

```
$ modal run build_method_subclass.py::Base.f
╭─ Error ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ <function Base.f at 0x7f37eb248180> is not a Modal entity (should be an App or Function)                                                   │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

erikbern:
oh cool let's try to fix that one too

modal/app.py
```diff
@@ -547,6 +550,35 @@ def wrapped(
...
                     raise InvalidError("interactive=True is not supported with web endpoint functions")
             else:
+                if not is_global_object(f.__qualname__) and not serialized:
+                    raise InvalidError(
+                        dedent(
+                            """
+                            The `@app.function` decorator must apply to functions in global scope,
+                            unless `serialize=True` is set.
+                            If trying to apply additional decorators, they may need to use `functools.wraps`.
+                            """
```
mwaskom, line 560: Does this this error obviate the one we have here: https://github.com/modal-labs/modal-client/blob/ded0d09f61d32f7a6cada0984757975a968bcf10/modal/_utils/function_utils.py#L110-L114?

mwaskom (approved):
Nice!

modal/app.py
erikbern: No, the other one is for situations where the a function is defined in local scope, i.e. inside another function.

The one we're adding is for global scope but not top-level scope.

erikbern:
Thanks @mwaskom !

PR #1986, opened 2024-07-08
Batching: Add Protobuff params to message Function for batching

cathyzbn:
## Describe your changes
Protoc changes to to support batching

---

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -496,8 +496,8 @@ class _Function(_Object, type_prefix="fu"):
         allow_background_volume_commits: Optional[bool] = None,
         block_network: bool = False,
         max_inputs: Optional[int] = None,
-        max_num_batch_inputs: Optional[int] = None,
-        max_batch_timout_sec: Optional[float] = None,
+        batch_max_num_inputs: Optional[int] = None,
+        batch_max_timeout_sec: Optional[float] = None,
         ephemeral_disk: Optional[int] = None,
     ) -> None:
         """mdmd:hidden"""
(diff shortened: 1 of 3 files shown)


modal/functions.py
```diff
@@ -809,6 +811,8 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
                     object_dependencies=object_dependencies,
                     block_network=block_network,
                     max_inputs=max_inputs or 0,
+                    max_num_batch_inputs=max_num_batch_inputs or 1,
```
gongy, line 814: Internally, I think we should use `0` to mean "no batching", right? so `or 0`

modal/functions.py
```diff
@@ -809,6 +811,8 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
...
+                    max_batch_timout_sec=max_batch_timout_sec or 0,
```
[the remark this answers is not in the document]
cathyzbn: my bad!

modal/functions.py
```diff
@@ -496,6 +496,8 @@ def from_args(
         allow_background_volume_commits: Optional[bool] = None,
         block_network: bool = False,
         max_inputs: Optional[int] = None,
+        batch_max_num_inputs: Optional[int] = None,
+        batch_max_timeout_sec: Optional[float] = None,
```
mwaskom, line 500: Super nit, but can we use `_secs` to be consistent with other fields?

```
$ git grep "_sec[s ]" modal_proto/
modal_proto/api.proto:  uint32 timeout_secs = 21;
modal_proto/api.proto:  uint32 task_idle_timeout_secs = 25;
modal_proto/api.proto:  optional uint32 timeout_secs = 6;
modal_proto/api.proto:  optional uint32 task_idle_timeout_secs = 7;
modal_proto/api.proto:  uint32 timeout_secs = 7;
```

mwaskom: Also, while I'm being really annoying, "max timeout seconds" doesn't make sense IMO.

The concept of a "max" waiting period is already included in the "timeout" term. But reaching a "timeout" is usually bad / an error. So I would say something like `max_wait_secs` is a better name here.

cathyzbn: that makes sense! sounds good!

PR #1987, opened 2024-07-08
Batching: add batching params to FunctionGetInputsRequest

cathyzbn:
## Describe your changes

Add parameters to FunctionGetInputsRequest for batching

---

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -350,8 +350,11 @@ class _ContainerIOManager:
         iteration = 0
         while not eof_received and self._fetching_inputs:
             request.average_call_time = self.get_average_call_time()
-            request.max_values = self.get_max_inputs_to_fetch()  # Deprecated; remove.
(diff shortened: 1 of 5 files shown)


gongy (commented):
chatted in person, maybe this should be injected @ worker

modal/_container_io_manager.py
```diff
@@ -350,8 +350,9 @@ async def _generate_inputs(self) -> AsyncIterator[Tuple[str, str, api_pb2.Functi
         iteration = 0
         while not eof_received and self._fetching_inputs:
             request.average_call_time = self.get_average_call_time()
-            request.max_values = self.get_max_inputs_to_fetch()  # Deprecated; remove.
+            request.max_values = self.get_max_inputs_to_fetch()
```
gongy, line 353: we should get rid of `get_max_inputs_to_fetch` -- that was old stuff

I think `request.max_values = 1` will work now right?

PR #1989, opened 2024-07-08
[MOD-3048] Disallow `enable_memory_snapshot=True` with parameterized classes

mattnappo:
Temporary patch until we fix the root problem on the backend.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -723,6 +723,13 @@ class _App:
             ):
                 raise InvalidError("A class must have `enable_memory_snapshot=True` to use `snap=True` on its methods.")
 
+            # Disallow enable_memory_snapshot for parameterized classes
+            methods = dict(inspect.getmembers(user_cls, inspect.isfunction))
+            if init := methods.get("__init__", False):
+                params = inspect.signature(init).parameters
+                if enable_memory_snapshot and len(params) > 1:
+                    raise InvalidError("Cannot use parameterized classes with `enable_memory_snapshot=True`.")
+
             tag: str = user_cls.__name__
             self._add_object(tag, cls)
             return cls
--- a/modal/app.py
+++ b/modal/app.py
@@ -731,7 +731,7 @@ class _App:
                 if len(params) > 1:
                     name = user_cls.__name__
                     raise InvalidError(
-                        f"In class '{name}': Cannot use parameterized classes with `enable_memory_snapshot=True`."
+                        f"Cannot use class parameterization in class {name} with `enable_memory_snapshot=True`."
                     )
 
             tag: str = user_cls.__name__
--- a/modal/app.py
+++ b/modal/app.py
@@ -724,11 +724,13 @@ class _App:
                 raise InvalidError("A class must have `enable_memory_snapshot=True` to use `snap=True` on its methods.")
 
             # Disallow enable_memory_snapshot for parameterized classes
+            # TODO(matt) Temporary fix for MOD-3048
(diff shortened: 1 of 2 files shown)


modal/app.py
```diff
@@ -723,6 +723,13 @@ def wrapper(user_cls: CLS_T) -> _Cls:
             ):
                 raise InvalidError("A class must have `enable_memory_snapshot=True` to use `snap=True` on its methods.")
 
+            # Disallow enable_memory_snapshot for parameterized classes
+            constructor = dict(inspect.getmembers(user_cls, inspect.isfunction)).get("__init__")
+            if enable_memory_snapshot and constructor:
+                params = inspect.signature(constructor).parameters
+                if len(params) > 1:
+                    raise InvalidError("Cannot use parameterized classes with `enable_memory_snapshot=True`.")
```
thundergolfer, line 731: I think it'll be more user friendly if the error message includes the name of the class so the user can more quickly go to the problematic source.

thundergolfer: I think this code block should also have a TODO which references a Linear ticket (MOD-123) about supporting this properly.

thundergolfer (approved):
LGTM after addressing comments.

mwaskom:
I'm not up to speed on the issue here but if the problem is in the backend, is this sometihng we could do server-side so we can (a) roll it out to everyone and (b) remove it when we fix the issue?

mattnappo:
Could something like this be in `FunctionBindParams`?

mwaskom:
> Could something like this be in FunctionBindParams?

Yeah that is what I was thinking at first, you could add something that checks the definition to see if snapshots are enabled.

One potential downside is that I don't think it will fire on `modal deploy`, so people might deploy apps that then break when they handle their first input. Maybe that's still better than the status quo though? Again I'm not up to speed with what the problem is.

mattnappo:
That is better than what we are currently doing, but I'd argue for a wider net here (including deploys) since most use cases for parameterized classes is so that you can do a `.lookup()` on a deployed app.

mwaskom:
> since most use cases for parameterized classes is so that you can do a .lookup() on a deployed app

It should still work in this case, the problem is just that it will let you deploy the app in a broken state. Ideally we would block the deploy itself.

modal/app.py
```diff
@@ -723,6 +723,17 @@ def wrapper(user_cls: CLS_T) -> _Cls:
...
 
+            # Disallow enable_memory_snapshot for parameterized classes
+            # TODO(matt) Temporary fix for MOD-3048
+            constructor = dict(inspect.getmembers(user_cls, inspect.isfunction)).get("__init__")
+            if enable_memory_snapshot and constructor:
+                params = inspect.signature(constructor).parameters
+                if len(params) > 1:
+                    name = user_cls.__name__
+                    raise InvalidError(
+                        f"In class '{name}': Cannot use parameterized classes with `enable_memory_snapshot=True`."
```
thundergolfer, line 734:
```suggestion
                        f"Cannot use class parameterization in class {name} with `enable_memory_snapshot=True`."
```

thundergolfer:
Good point @mwaskom. I don't like that users can get all the way past `modal deploy` with broken if we just have server-side. But it is a little annoying that when we fix it users will need to upgrade client. 

I think I'm in favor of both client and server?

mattnappo:
Here's the backend change: https://github.com/modal-labs/modal/pull/13974

PR #1994, opened 2024-07-10
MOD-3243: better error message when attempting to mount volume as nfs

kramstrom:
Adds a clearer error message when attempting to mount a volume as a network file system and refactors the nfs validation for sandboxes/functions to be similar to the volume validation

### DIFF
--- a/modal/_utils/mount_utils.py
+++ b/modal/_utils/mount_utils.py
@@ -41,7 +41,7 @@ def validate_mount_points(
 
 
 def validate_network_file_systems(
-    network_file_systems: Mapping[Union[str, PurePosixPath], Union[_NetworkFileSystem]],
+    network_file_systems: Mapping[Union[str, PurePosixPath], _NetworkFileSystem],
 ):
     validated_network_file_systems = validate_mount_points("NetworkFileSystem", network_file_systems)
 
@@ -61,7 +61,7 @@ def validate_volumes(
     validated_volumes = validate_mount_points("Volume", volumes)
     # We don't support mounting a modal.Volume in more than one location,
     # but the same CloudBucketMount object can be used in more than one location.
-    volume_to_paths: Dict[Union[_Volume], List[str]] = {}
+    volume_to_paths: Dict[_Volume, List[str]] = {}
     for path, volume in validated_volumes:
         if not isinstance(volume, (_Volume, _CloudBucketMount)):
             raise InvalidError(f"Object of type {type(volume)} mounted at '{path}' is not useable as a volume.")
--- a/modal/_utils/mount_utils.py
+++ b/modal/_utils/mount_utils.py
@@ -18,6 +18,9 @@ def validate_mount_points(
 ) -> List[Tuple[str, T]]:
     """Mount point path validation for volumes and network file systems."""
 
+    if not isinstance(volume_likes, dict):
+        raise InvalidError(f"{display_name} must be a dict[str, {display_name}] where the keys are paths")
+
     validated = []
     for path, vol in volume_likes.items():
         path = PurePath(path).as_posix()
@@ -36,9 +39,7 @@ def validate_mount_points(
 
 
 def validate_network_file_systems(network_file_systems: Mapping[Union[str, PurePosixPath], _NetworkFileSystem]):
-    if not isinstance(network_file_systems, dict):
-        raise InvalidError("network_file_systems must be a dict[str, NetworkFileSystem] where the keys are paths")
-    validated_network_file_systems = validate_mount_points("Network file system", network_file_systems)
+    validated_network_file_systems = validate_mount_points("NetworkFileSystem", network_file_systems)
 
     for path, network_file_system in validated_network_file_systems:
         if not isinstance(network_file_system, (_NetworkFileSystem)):
@@ -53,9 +54,6 @@ def validate_network_file_systems(network_file_systems: Mapping[Union[str, PureP
 def validate_volumes(
     volumes: Mapping[Union[str, PurePosixPath], Union["_Volume", "_CloudBucketMount"]],
 ) -> Sequence[Tuple[str, Union["_Volume", "_NetworkFileSystem", "_CloudBucketMount"]]]:
-    if not isinstance(volumes, dict):
-        raise InvalidError("volumes must be a dict where the keys are mount paths")
-
     validated_volumes = validate_mount_points("Volume", volumes)
     # We don't support mounting a modal.Volume in more than one location,
     # but the same CloudBucketMount object can be used in more than one location.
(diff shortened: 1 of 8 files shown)


modal/network_file_system.py
```diff
@@ -125,8 +125,16 @@ async def _load(self: _NetworkFileSystem, resolver: Resolver, existing_object_id
                 environment_name=_get_environment_name(environment_name, resolver),
                 object_creation_type=(api_pb2.OBJECT_CREATION_TYPE_CREATE_IF_MISSING if create_if_missing else None),
             )
-            response = await resolver.client.stub.SharedVolumeGetOrCreate(req)
-            self._hydrate(response.shared_volume_id, resolver.client, None)
+            try:
+                response = await resolver.client.stub.SharedVolumeGetOrCreate(req)
+                self._hydrate(response.shared_volume_id, resolver.client, None)
+            except GRPCError as exc:
+                if exc.status == Status.NOT_FOUND and exc.message == "App has wrong entity vo":
```
kramstrom, line 132: Perhaps the exc message here could instead be improved server side so we don't have to compare the string?

modal/_utils/mount_utils.py
```diff
@@ -35,6 +35,21 @@ def validate_mount_points(
     return validated
 
 
+def validate_network_file_systems(network_file_systems: Mapping[Union[str, PurePosixPath], _NetworkFileSystem]):
+    if not isinstance(network_file_systems, dict):
+        raise InvalidError("network_file_systems must be a dict[str, NetworkFileSystem] where the keys are paths")
+    validated_network_file_systems = validate_mount_points("Network file system", network_file_systems)
+
+    for path, network_file_system in validated_network_file_systems:
+        if not isinstance(network_file_system, (_NetworkFileSystem)):
```
mwaskom, line 44: Do we need to check both the private `_NetworkFileSystem` object and the synchronicity-wrapped `NetworkFileSystem` object?

modal/_utils/mount_utils.py
```diff
@@ -35,6 +35,21 @@ def validate_mount_points(
     return validated
 
 
+def validate_network_file_systems(network_file_systems: Mapping[Union[str, PurePosixPath], _NetworkFileSystem]):
+    if not isinstance(network_file_systems, dict):
+        raise InvalidError("network_file_systems must be a dict[str, NetworkFileSystem] where the keys are paths")
+    validated_network_file_systems = validate_mount_points("Network file system", network_file_systems)
```
mwaskom, line 41:
```suggestion
    validated_network_file_systems = validate_mount_points("NetworkFileSystem", network_file_systems)
```

mwaskom (commented):
Will be a little tricky to write unit tests here but would be great if we can do that.

Separately I noticed in the `validate_volumes` function that we do

```python
    validated_volumes = validate_mount_points("Volume", volumes)
    # We don't support mounting a modal.Volume in more than one location,
    # but the same CloudBucketMount object can be used in more than one location.
    volume_to_paths: Dict["_Volume", List[str]] = {}
    for path, volume in validated_volumes:
        if not isinstance(volume, (_Volume, _NetworkFileSystem, _CloudBucketMount)):
            raise InvalidError(f"Object of type {type(volume)} mounted at '{path}' is not useable as a volume.")
```

Which looks like a bug? I don't think we we can pass `NetworkFileSystem` to `volumes=`, worth double checking, might be another way of catching this problem.

modal/_utils/mount_utils.py
kramstrom: Hm, we don't do that for the volume validation, is that an error then?

test/mount_utils_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+from modal._utils.mount_utils import validate_mount_points, validate_network_file_systems, validate_volumes
+from modal.exception import InvalidError
+from modal.network_file_system import _NetworkFileSystem
+from modal.volume import _Volume
+
+
+def test_validate_mount_points():
+    # valid mount points
+    dict_input: Mapping = {"/foo/bar": _NetworkFileSystem.from_name("NetworkFileSystem", create_if_missing=False)}
```
mwaskom, line 13: Odd that we need the explicit type annotation here? A dict should be accepted by an interface typed with `Mapping`.

test/mount_utils_test.py
```diff
@@ -0,0 +1,55 @@
...
+    validate_mount_points("NetworkFileSystem", dict_input)
+
+    # invalid list input, should be dicts
+    list_input = [_NetworkFileSystem.from_name("NetworkFileSystem", create_if_missing=False)]
```
mwaskom, line 17: Should the tests exercise the public `NetworkFileSystem` object?

mwaskom (commented):
Thanks for adding tests! A few more small comments about some of the error messages — and I think we should use the public object in the tests, unless I'm missing something about how we unwrap the synchronicity interface internally (since these are internal utilities that you're exercising).

modal/_utils/mount_utils.py
kramstrom: Yeah using the public object fails, but when experimenting with this locally and mounting a volume / file system they seem to always have the private object during validation. Is there a way to get around that?

mwaskom: OK got it — thanks for checking.

That all sounds good vis-a-vis the type checks then. Do we have at least one test that validates this behavior through the public interface? (i.e. by setting up an app with a function decorator that uses the mount parameter improperly?) If we have that I think we're good to go!

modal/network_file_system.py
```diff
@@ -125,8 +125,16 @@ async def _load(self: _NetworkFileSystem, resolver: Resolver, existing_object_id
...
+                    raise InvalidError(
+                        f"Attempted to mount: `{label}` as a NetworkFileSystem "
+                        + "which already exists as a Volume, please create new a NetworkFileSystem instead"
```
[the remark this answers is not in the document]
kramstrom: Makes sense!

test/mount_utils_test.py
kramstrom: I changed that now, these methods seem to only be used for the private objects though, so the code is a bit uglier checking for the public ones too but im fine with it if you prefer it this way!

test/mount_utils_test.py
kramstrom: I agree but without it I'm getting

```
error: Argument 2 to "validate_mount_points" has incompatible type "dict[str, NetworkFileSystem]"; expected "Mapping[Union[str, PurePosixPath], NetworkFileSystem]"
```

which doesn't make sense to me

modal/_utils/mount_utils.py
kramstrom: You mean like the test added in `test/network_file_system_test.py` or something different?

PR #1995, opened 2024-07-10
[MOD-2330] Improve error reporting in 'modal shell'

thundergolfer:
## Describe your changes

A motivating example is if you provide garbage config to `CloudBucketMount` you get no feedback about the problem from `modal shell`: 

```python
@app.function(
    volumes={
        "/data": modal.CloudBucketMount(bucket_name="hi", secret=modal.Secret.from_dict({}))
    },
)
def f():
    print("hi")
```

- [ ] ~Client+Server: this change is compatible with old servers~
- [ ] ~Client forward compatibility: this change ensures client can accept data intended for later versions of itself~

### DIFF
--- a/modal/_container_exec.py
+++ b/modal/_container_exec.py
@@ -43,7 +43,7 @@ async def container_exec(task_id: str, command: List[str], *, pty: bool, client:
     except GRPCError as err:
         connecting_status.stop()
         if err.status == Status.NOT_FOUND:
-            raise NotFoundError(f"Container ID {task_id} not found")
+            raise NotFoundError(err.message)
         raise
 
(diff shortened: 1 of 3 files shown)


modal/_container_exec.py
```diff
@@ -43,7 +43,7 @@ async def container_exec(task_id: str, command: List[str], *, pty: bool, client:
     except GRPCError as err:
         connecting_status.stop()
         if err.status == Status.NOT_FOUND:
-            raise NotFoundError(f"Container ID {task_id} not found")
+            raise NotFoundError(err.message)
```
thundergolfer, line 46: I'm changing the server to make it return multiple kinds of not found messages for improved UX.

PR #1996, opened 2024-07-10
MOD-3260: Add AttributeError to list of retry_transient_errors

kramstrom:
Temporarily adds `AttributeError` to the list of transient errors to be retried as `StreamTerminatedError` is not properly raised in `grpclib <= 0.47` and will instead raise `AttributeError`. Fixed in https://github.com/vmagamedov/grpclib/issues/185 so we should remove this (also for the heartbeat exception) once we can upgrade to a newer stable version.

Great find @mwaskom!

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -162,7 +162,7 @@ async def retry_transient_errors(
             timeout = None
         try:
             return await fn(*args, metadata=metadata, timeout=timeout)
-        except (StreamTerminatedError, GRPCError, socket.gaierror, asyncio.TimeoutError) as exc:
+        except (StreamTerminatedError, GRPCError, socket.gaierror, asyncio.TimeoutError, AttributeError) as exc:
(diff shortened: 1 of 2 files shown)


modal/_utils/grpc_utils.py
```diff
@@ -173,6 +173,14 @@ async def retry_transient_errors(
                 # no point sleeping if that's going to push us past the deadline
                 raise exc
 
+            if isinstance(exc, AttributeError):
+                # StreamTerminatedError are not properly raised in grpclib<=0.4.7
+                # fixed in https://github.com/vmagamedov/grpclib/issues/185
+                # TODO: update to newer version (>=0.4.8) once stable
+                if "_write_appdata" in str(exc):
+                    continue
```
aksh-at, line 181: `continue` isn't right here, because it will also skip the `n_retries` increment etc.

kramstrom: ah yeah, fixed

aksh-at (approved):
awesome, thanks!

PR #1998, opened 2024-07-11
Batching: Protobuf Changes to Function and FunctionGetInputsRequest

cathyzbn:
## Describe your changes
Adding batching parameters `batch_max_size` and `batch_linger_secs` to FunctionGetInputsRequest

---


gongy (approved):
I think we can get away without `optional` -- otherwise lgtm

PR #1999, opened 2024-07-11
[MOD-3257] fix cached PyTorch GPU device queries

thundergolfer:
## Describe your changes

See **MOD-3257** for details

Many snapshotted containers won't have `torch`, but in these cases a key miss searching `sys.modules` is very fast <500 micros on even large dictionaries (100,000 entries). thanks @mwaskom for the suggestion. 

---

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -608,9 +608,7 @@ class _ContainerIOManager:
         # perf: scanning sys.modules keys before import to avoid slow PYTHONPATH scanning.
         if "torch" in sys.modules:
             try:
-                import torch
-
-                torch.cuda.device_count = torch.cuda._device_count_nvml
+                sys.modules["torch"].cuda.device_count = sys.modules["torch"].cuda._device_count_nvml
             # Wide-open except to catch anything. We don't want to crash here.
             except Exception as exc:
                 logger.warning(
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -4,6 +4,7 @@ import json
 import math
 import os
 import signal
+import sys
 import time
 import traceback
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


thundergolfer:
If you can think of a clever way to test this LMK 🙂. I tried a `from fake_mods import torch` but the key becomes `fake_mods.torch` not `torch`.

mwaskom:
I guess you could add a test fixture that writes a local `torch.py` file and imports that?

thundergolfer:
Yeh that was a thought I had. write it, import it, fixture deletes it.

modal/_container_io_manager.py
```diff
@@ -601,6 +602,22 @@ async def memory_restore(self) -> None:
...
 
+        # Patch torch to ensure it doesn't return CUDA unavailibility due to
+        # cached queries that executed during snapshot process. ref: MOD-3257
+        #
+        # perf: scanning sys.modules keys before import to avoid slow PYTHONPATH scanning.
+        if "torch" in sys.modules:
+            try:
+                import torch
+
+                torch.cuda.device_count = torch.cuda._device_count_nvml
```
mwaskom, line 613: I doubt it would actually be much faster, but you probably could skip the `import` statement and just do

```
sys.modules["torch"].cuda.device_count = ...
```

And actually maybe checking for `"torch.cuda"` in sys.modules is an even narrower hack and less likely to get tripped up if people have a local `torch.py` or something in their modal root (i mean, I hope they don't, but you never know...)

thundergolfer: > if people have a local torch.py or something in their modal root 

those people should be called out, scolded by the logline! haha

test/container_app_test.py
```diff
@@ -86,6 +86,95 @@ async def test_container_debug_snapshot(container_client, tmpdir, servicer):
...
+
+@pytest.fixture(scope="function")
+def weird_torch_module():
+    module_path = os.path.join(os.getcwd(), "torch.py")
+    with open(module_path, "w") as f:
+        f.write("IM_WEIRD = 42\n")
+
+    yield module_path
+
+    os.remove(module_path)  # Teardown: remove the torch.py file
```
mwaskom, line 118: Will we leave this around if the test fails (or otherwise raises?)

mwaskom (commented):
Oops sorry, forgot to submit my review comment last night. Just a question about whether we want to be doing this in a `try ... finally`.

test/container_app_test.py
thundergolfer: I thought about it but given it was just test execution I let it be. I guess it's annoying if it raises and you get a stray file on your devbox— i was thinking of CI execution and its ephemerality.

thundergolfer: Oh @mwaskom it should be fine https://docs.pytest.org/en/latest/how-to/fixtures.html#handling-errors-for-yield-fixture

mwaskom: Oh nice, I didn't know about that.

PR #2000, opened 2024-07-11
MOD-3488: Improve function call data streaming

gongy:
1. We should update `last_index` after yielding the message. Otherwise, we skip the chunk if `blob_download` fails with either of the caught errors.
2. We shouldn't use a fixed 1s delay.

TODO:
- [ ] Add tests


thundergolfer:
Great PR for number 2000!

freider (approved):
nice find! :)

PR #2002, opened 2024-07-12
OutputManager cleanup prep

erikbern:


erikbern:
This unfortunately breaks some integration tests that pass a buffer as stdout. Will have to think about this some more.

mwaskom:
Closing as stale — I think you took a different approach here.

PR #2004, opened 2024-07-12
[MOD-3251] Stop heartbeats before sending the container snapshot RPC

mattnappo:
Heartbeats make a connection to the gRPC proxy (`modal.sock`) which causes the container to open a FD. This open FD causes gVisor to crash during `runsc checkpoint`. I ran the repro over 300 times with 0 flakes.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -699,7 +699,7 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
 
     _client: _Client = synchronizer._translate_in(client)  # TODO(erikbern): ugly
 
-    with container_io_manager.heartbeats(), UserCodeEventLoop() as event_loop:
+    with container_io_manager.heartbeats(function_def.is_checkpointing_function), UserCodeEventLoop() as event_loop:
         # If this is a serialized function, fetch the definition from the server
         if function_def.definition_type == api_pb2.Function.DEFINITION_TYPE_SERIALIZED:
             ser_cls, ser_fun = container_io_manager.get_serialized_function()
@@ -778,6 +778,7 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
         # global imports and innitialization. Checkpointed containers run from this point onwards.
         if function_def.is_checkpointing_function:
             container_io_manager.memory_snapshot()
+
 
         # Install hooks for interactive functions.
         if function_def.pty_info.pty_type != api_pb2.PTYInfo.PTY_TYPE_UNSPECIFIED:
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -76,6 +76,7 @@ class _ContainerIOManager:
     _waiting_for_memory_snapshot: bool
     _heartbeat_loop: Optional[asyncio.Task]
     _pause_heartbeats: Optional[asyncio.Condition]
+    _snapshot_running: bool
 
     _is_interactivity_enabled: bool
     _fetching_inputs: bool
@@ -104,7 +105,8 @@ class _ContainerIOManager:
         self._environment_name = container_args.environment_name
         self._waiting_for_memory_snapshot = False
         self._heartbeat_loop = None
-        self._pause_heartbeats = None
+        self._pause_heartbeats = asyncio.Condition()
+        self._snapshot_running = False
 
         self._is_interactivity_enabled = False
         self._fetching_inputs = True
@@ -161,11 +163,15 @@ class _ContainerIOManager:
             request.current_input_started_at = self.current_input_started_at
 
         async with self._pause_heartbeats:
-            await self._pause_heartbeats.wait()
+            print("heartbeat acquired lock")
+            while self._snapshot_running:
+                await self._pause_heartbeats.wait()
+
             # TODO(erikbern): capture exceptions?
             response = await retry_transient_errors(
                 self._client.stub.ContainerHeartbeat, request, attempt_timeout=HEARTBEAT_TIMEOUT
             )
+        print("heartbeat released lock")
 
         if response.HasField("cancel_input_event"):
             # Pause processing of the current input by signaling self a SIGUSR1.
@@ -203,7 +209,7 @@ class _ContainerIOManager:
     async def heartbeats(self) -> AsyncGenerator[None, None]:
         async with TaskContext() as tc:
             self._heartbeat_loop = t = tc.create_task(self._run_heartbeat_loop())
-            self._pause_heartbeats = asyncio.Condition()
+            self._snapshot_running = False
             t.set_name("heartbeat loop")
             try:
                 yield
@@ -579,6 +585,13 @@ class _ContainerIOManager:
             await asyncio.sleep(0.01)
             continue
 
+        # Turn heartbeats back on
+        async with self._pause_heartbeats:
+            print("restore acquired lock")
+            self._snapshot_running = False
+            self._pause_heartbeats.notify_all()
+        print("restore released lock")
+
         logger.debug("Container: restored")
 
         # Look for state file and create new client with updated credentials.
@@ -631,9 +644,14 @@ class _ContainerIOManager:
 
         # Pause heartbeats since they keep the client connection open which causes the snapshotter to crash
         async with self._pause_heartbeats:
-            await self._client.stub.ContainerCheckpoint(
-                api_pb2.ContainerCheckpointRequest(checkpoint_id=self.checkpoint_id)
-            )
+            print("snapshot acquired lock")
+            self._snapshot_running = True
+            self._pause_heartbeats.notify_all()
+
+        await self._client.stub.ContainerCheckpoint(
+            api_pb2.ContainerCheckpointRequest(checkpoint_id=self.checkpoint_id)
+        )
+        print("snapshot sent request")
 
         self._waiting_for_memory_snapshot = True
         await self._client._close(forget_credentials=True)
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -74,7 +74,7 @@ class _ContainerIOManager:
     _semaphore: Optional[asyncio.Semaphore]
     _environment_name: str
     _heartbeat_loop: Optional[asyncio.Task]
-    _pause_heartbeats: Optional[asyncio.Condition]
+    _heartbeat_condition: asyncio.Condition
     _waiting_for_memory_snapshot: bool
 
     _is_interactivity_enabled: bool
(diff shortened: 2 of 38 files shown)


luiscape (approved):
Any way to verify or test?

modal/_container_io_manager.py
```diff
@@ -624,6 +624,9 @@ async def memory_snapshot(self) -> None:
         if self.checkpoint_id:
             logger.debug(f"Checkpoint ID: {self.checkpoint_id} (Memory Snapshot ID)")
 
+        # Heartbeats can leave the modal.sock file open, causing gVisor to crash
+        self.stop_heartbeat()
```
thundergolfer, line 628: This cancels the heartbeat loop and the restored container will not restart it. 

Also don't mention the internals ("gVisor") in the client comments if avoidable 👍

thundergolfer (changes requested):
I think you need a pause-and-restart structure here. This looks like just pause, which means a restored container will be killed for not heartbeating.

mattnappo:
Update: I have landed on a solution I am happy with, with almost no overhead (@thundergolfer and @luiscape I removed the bottleneck I spoke of). Here is a summary:

* I introduce a `asyncio.Condition` variable that guards a boolean called `waiting_for_memory_snapshot` (same as the old name).
* Before a `ContainerCheckpoint` or `ContainerHeartbeat` RPC, the `Condition` must be acquired, and the heartbeat will only run if the container is NOT `waiting_for_memory_snapshot`. This prevents heartbeat requests from executing while the checkpoint-restore process in in progress.
* If the function is a snapshotting function (`function_def.is_checkpointing_fn`), then heartbeats will only turn on after the entire checkpoint-restore process
* If it's not a snapshotting function, then heartbeats are enabled immediately, and will run during the container init phase.
* Remove the old `ContainerIOManager::_waiting_for_memory_snapshot` logic, as it was broken and did not prevent heartbeats from happening when they shouldn't

mattnappo:
The following invariants hold:
* If snapshotting fn, then heartbeats run after the snap-restore phase
* If snapshotting fn, then heartbeats do not run before the snap-restore phase
* If not snapshotting fn, then heartbeats are always running

modal/_container_entrypoint.py
```diff
@@ -699,7 +699,7 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
 
     _client: _Client = synchronizer._translate_in(client)  # TODO(erikbern): ugly
 
-    with container_io_manager.heartbeats(), UserCodeEventLoop() as event_loop:
+    with container_io_manager.heartbeats(function_def.is_checkpointing_function), UserCodeEventLoop() as event_loop:
```
luiscape, line 702: Are we just disabling heartbeats if we are checkpointing? If so, isn't that dangerous? i.e. functions may execute long-running work and missing heartbeats may kill it

luiscape (commented):
See comment re disabling heartbeats.

modal/_container_entrypoint.py
mattnappo: We disable heartbeats only during the checkpoint phase. They are re-enabled after the container restores. So in any case, heartbeats will be running whenever long-running user code is.

test/container_app_test.py
```diff
@@ -58,12 +58,39 @@ async def test_container_snapshot_restore(container_client, tmpdir, servicer):
...
+    # Ensure that heartbeats do run after the snapshot
+    with io_manager.heartbeats(True):
+        with mock.patch.dict(
+            os.environ, {"MODAL_RESTORE_STATE_PATH": str(restore_path), "MODAL_SERVER_URL": servicer.container_addr},
+        ):
+            with mock.patch("modal.runner.HEARTBEAT_INTERVAL", 1):
+                with mock.patch.object(container_client.stub, 'ContainerHeartbeat') as mock_heartbeat:
+                    mock_heartbeat.assert_not_called()
+                    io_manager.memory_snapshot()
+                    mock_heartbeat.assert_called_once()
```
mattnappo, line 92: This behavior is the main change in this PR.

test/container_app_test.py
```diff
@@ -58,12 +59,43 @@ async def test_container_snapshot_restore(container_client, tmpdir, servicer):
...
+            {
+                "task_id": "ta-i-am-restored",
+                "task_secret": "ts-i-am-restored",
+                "function_id": "fu-i-am-restored",
+            }
+        ),
+        encoding="utf-8",
+    )
+
+    asyncio.set_event_loop(asyncio.new_event_loop())
```
mattnappo, line 83: This and the asyncio.sleep a few lines below were frivolous attempts at fixing CI.

test/container_app_test.py
```diff
@@ -65,6 +67,41 @@ async def test_container_snapshot_restore(container_client, tmpdir, servicer):
...
+                {"MODAL_RESTORE_STATE_PATH": str(restore_path), "MODAL_SERVER_URL": servicer.container_addr},
+            ):
+                with mock.patch("modal.runner.HEARTBEAT_INTERVAL", 1):
+                    assert not list(
+                        filter(lambda req: isinstance(req, api_pb2.ContainerHeartbeatRequest), servicer.requests)
+                    )
+                    await io_manager.memory_snapshot()
+                    await asyncio.sleep(1)
+                    assert list(
+                        filter(lambda req: isinstance(req, api_pb2.ContainerHeartbeatRequest), servicer.requests)
```
mattnappo, line 101: Put a sleep here to ensure a heartbeat actually gets sent (heartbeat interval is 1s) @thundergolfer

test/container_app_test.py
```diff
@@ -65,6 +67,41 @@ async def test_container_snapshot_restore(container_client, tmpdir, servicer):
...
 
+@pytest.mark.asyncio
+async def test_container_snapshot_restore_heartbeats(tmpdir, servicer):
+    client = _Client(servicer.container_addr, api_pb2.CLIENT_TYPE_CONTAINER, ("ta-123", "task-secret"))
+    async with client as async_client:
+        io_manager = _ContainerIOManager(api_pb2.ContainerArguments(), async_client)
+        restore_path = tmpdir.join("fake-restore-state.json")
+        restore_path.write_text(
+            json.dumps(
+                {
```
thundergolfer, line 78: nit: think we can break out this repeated string into a hlper constant now to save some lines.

modal/_container_io_manager.py
```diff
@@ -196,10 +196,11 @@ async def _heartbeat_handle_cancellations(self) -> bool:
         return False
 
     @asynccontextmanager
-    async def heartbeats(self) -> AsyncGenerator[None, None]:
+    async def heartbeats(self, disable_init: bool) -> AsyncGenerator[None, None]:
```
thundergolfer, line 199: I think we want a different name here. It's sus that the name exists only very briefly before being 'thrown away'. Maybe `wait_for_mem_snap: bool`

modal/_container_io_manager.py
```diff
@@ -574,6 +575,11 @@ async def memory_restore(self) -> None:
             await asyncio.sleep(0.01)
             continue
 
+        # Turn heartbeats back on
```
thundergolfer, line 578: nit: the other comment is good but this one just says the _what_ and no _why_.

modal/_container_io_manager.py
```diff
@@ -146,22 +148,20 @@ async def _heartbeat_handle_cancellations(self) -> bool:
...
             request.current_input_id = self.current_input_id
         if self.current_input_started_at is not None:
             request.current_input_started_at = self.current_input_started_at
 
-        # TODO(erikbern): capture exceptions?
-        response = await retry_transient_errors(
-            self._client.stub.ContainerHeartbeat, request, attempt_timeout=HEARTBEAT_TIMEOUT
-        )
+        async with self._pause_heartbeats:
+            while self._waiting_for_memory_snapshot:
```
thundergolfer, line 158: Maybe dumb q but I can't see why this is a `while` and not just an `if`. If this is obvious, could it get a comment so I don't hurt myself thinking next time I come across the code 😄

modal/_container_io_manager.py
```diff
@@ -101,8 +102,9 @@ def _init(self, container_args: api_pb2.ContainerArguments, client: _Client):
 
         self._semaphore = None
         self._environment_name = container_args.environment_name
-        self._waiting_for_memory_snapshot = False
         self._heartbeat_loop = None
+        self._pause_heartbeats = asyncio.Condition()
```
thundergolfer, line 106: Would prefer a different name here. "pause heartbeats" is an imperative mood, which is typical of functions. Something like `self._heartbeat_condition` is fine to me.

modal/_container_io_manager.py
```diff
@@ -73,8 +73,9 @@ class _ContainerIOManager:
     _input_concurrency: Optional[int]
     _semaphore: Optional[asyncio.Semaphore]
     _environment_name: str
-    _waiting_for_memory_snapshot: bool
     _heartbeat_loop: Optional[asyncio.Task]
+    _pause_heartbeats: Optional[asyncio.Condition]
```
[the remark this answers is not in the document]
mattnappo: You're correct

test/container_app_test.py
```diff
@@ -65,6 +67,41 @@ async def test_container_snapshot_restore(container_client, tmpdir, servicer):
...
+            with mock.patch.dict(
+                os.environ,
+                {"MODAL_RESTORE_STATE_PATH": str(restore_path), "MODAL_SERVER_URL": servicer.container_addr},
+            ):
+                with mock.patch("modal.runner.HEARTBEAT_INTERVAL", 0.01):
+                    assert not list(
+                        filter(lambda req: isinstance(req, api_pb2.ContainerHeartbeatRequest), servicer.requests)
+                    )
+                    await io_manager.memory_snapshot()
+                    await asyncio.sleep(0.01)
```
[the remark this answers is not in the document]
mattnappo: We want to wait until the condition is true, which is the purpose of the while over an if. If we just had an if, then it is possible that the thread spuriously wakes up and the execution continues despite the boolean being false. In other words, with an `if` we only wait once. From the docs:

>coroutine wait()[¶](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition.wait)
>  This method releases the underlying lock, and then blocks until it is awakened by a [notify()](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition.notify) or [notify_all()](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition.notify_all) call. Once awakened, the Condition re-acquires its lock and this method returns True.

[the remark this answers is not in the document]
mattnappo: It makes sense to sleep before the `assert not`.

modal/_container_io_manager.py
mattnappo: Posted answer to this in thread above by accident

thundergolfer (approved):
🚢

PR #2006, opened 2024-07-12
[MOD-3008]: Add friendly error message when attempting to use `modal shell` on windows

kramstrom:
`modal shell` command doesn't work on windows currently, this PR adds a check to see which platform is being executed on and raises a friendly message if it's not compatible

### DIFF
--- a/test/cli_test.py
+++ b/test/cli_test.py
@@ -3,6 +3,7 @@ import asyncio
 import contextlib
 import json
 import os
+import platform
 import pytest
 import re
 import subprocess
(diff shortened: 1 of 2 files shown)


mwaskom (approved):
Nice — do we need this for `modal container exec` too?

test/cli_test.py
```diff
@@ -439,6 +439,20 @@ def test_shell_cmd(servicer, set_env_client, test_dir, mock_shell_pty):
     assert captured_out == [(1, shell_prompt), (1, expected_output)]
 
 
+@pytest.mark.parametrize(("platform", "expected_exit_code"), [("win32", 1), ("cygwin", 1), ("linux", 0), ("darwin", 0)])
```
erikbern, line 442: we actually run the tests in CI on linux, windows, and osx, so another option is to skip the mocking and parametrization and make sure the behavior is expected on each platform

mwaskom: I was going to suggest this too but don't understand the difference between `win32` and `cygwin` or whether our CI would exercise both 🪟 🙈

kramstrom: Yeah I was wondering that too, I'll make a quick test to see what we're running in the cI

kramstrom: Tests seem to run on win32: https://github.com/modal-labs/modal-client/actions/runs/9940540142/job/27457485407?pr=2006 but I think it should be fine to keep skip the mocking and run it on all platforms?

kramstrom:
> Nice — do we need this for `modal container exec` too?

Ah didn't know about that one, buy yeah makes sense there too

PR #2009, opened 2024-07-12
Add programmatic retrieval of image logs

erikbern:
When using Modal "as a library", you don't want to rely out the output manager for logs. Instead lets add lower level methods to retrieve all logs.

Going to do the same for apps.

This lets us update integration tests to not rely on the output manager for capturing output.


erikbern:
@prbot approve

PR #2012, opened 2024-07-14
App._logs to fetch logs

erikbern:
This adds a utility method on the `App` class to fetch logs for an app.

The point is to get rid of the `stdout` argument to `OutputManager` so that we can make it a singleton throughout.


erikbern:
@prbot approve

erikbern:
yolo merge since it's not in the public interface but @mwaskom lmk if you have any thoughts on this!

PR #2018, opened 2024-07-15
OutputManager: get rid of of the *_if_visible methods

erikbern:
This is helpful since
1. This lets us check if the future `OutputManager` singleton exist, and if not, then the code can just shortcut the if-body
2. This way we can put Rich-specific stuff inside the if-body (makes it easier to not require it inside the container)
3. The previous code was pretty complex? I struggle with the `contextlib.nullcontext` stuff

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -175,17 +175,14 @@ class OutputManager:
         self._app_page_url = None
         self._show_image_logs = False
 
+    def is_visible(self) -> bool:
+        return self._visible_progress
+
     def hide_output(self):
         self._visible_progress = False
 
-    def print_if_visible(self, renderable) -> None:
-        if self._visible_progress:
-            self._console.print(renderable)
-
-    def ctx_if_visible(self, context_mgr):
-        if self._visible_progress:
-            return context_mgr
-        return contextlib.nullcontext()
+    def print(self, renderable) -> None:
+        self._console.print(renderable)
 
     def make_live(self, renderable: RenderableType) -> Live:
         """Creates a customized `rich.Live` instance with the given renderable. The renderable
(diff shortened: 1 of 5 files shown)


modal/_output.py
```diff
@@ -175,17 +175,14 @@ def __init__(
         self._app_page_url = None
         self._show_image_logs = False
 
+    def is_visible(self) -> bool:
+        return self._visible_progress
```
mwaskom, line 179: Doesn't really matter, but the name suggests to me that this would be a property

```python
if self.is_visible:
    ...
```

(Although I guess it's just returning an attribute value anyway; why not just expose that and save the function call?)

modal/_output.py
```diff
@@ -175,17 +175,14 @@ def __init__(
...
-    def print_if_visible(self, renderable) -> None:
-        if self._visible_progress:
-            self._console.print(renderable)
-
-    def ctx_if_visible(self, context_mgr):
-        if self._visible_progress:
-            return context_mgr
-        return contextlib.nullcontext()
+    def print(self, renderable) -> None:
+        self._console.print(renderable)
```
mwaskom, line 185: Would it clean things up even further to make this a no-op when `self.is_visible` is False?

mwaskom (approved):
Deffo an improvement, see a couple comments about making it marginally simpler though.

modal/_output.py
erikbern: I almost made it a property, but I'm planning to change this to something slightly different in a second anyway – it will probably something more like `if OutputManager.is_enabled()`

since it will be on a class level, I don't think you can make it a property

modal/_output.py
erikbern: See previous – I think it will be cleaner to keep it this way once I make it a singleton, something like
```python
if OutputManager.is_enabled():
    import rich
    renderable = ...
    OutputManager.get_singleton().print(renderable)
```

something like that?

erikbern:
Thanks @mwaskom – happy to address anything outstanding in subsequent PRs!

PR #2019, opened 2024-07-15
Remove complex logging code

erikbern:
I don't think this is needed anymore, since we don't pass custom buffers to the OutputManager since recently.

In all the places where we call `put_log_content`, I made sure to check that the output is visible.

Passes integration tests as well

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -341,33 +341,12 @@ class OutputManager:
             self._task_progress_items[task_key] = progress_task_id
 
     async def put_log_content(self, log: api_pb2.TaskLogs):
-        if self._visible_progress:
-            stream = self._line_buffers.get(log.file_descriptor)
-            if stream is None:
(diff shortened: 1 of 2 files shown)


erikbern:
Did some `git blame` and this code is from two years ago so hopefully it's fine to delete? It's possible it breaks something with notebooks though.

Idk my feeling is if we don't have tests for functionality, then it's effectively dead code... would rather remove this and add it back if/when we run into issues – along with tests.

erikbern:
I think this code was introduced in #195, #196, and #231 afaict

erikbern:
Actually this PR does break `modal run -i` which seems bad. I guess we have no tests for it which seems bad?

mwaskom:
> Actually this PR does break modal run -i which seems bad. I guess we have no tests for it which seems bad?

Doesn't seem great — although a lot of interactive / notebook features are hard to write good tests for.

PR #2022, opened 2024-07-17
Add test for writing PTY output

erikbern:
We had zero tests for this which was pretty annoying. See #2019.

Adding a very basic tests. Tbh the code is a bit of a mess and we should refactor it later but hopefully this makes it easier?

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -174,6 +174,7 @@ class OutputManager:
         self._status_spinner = step_progress(status_spinner_text)
         self._app_page_url = None
         self._show_image_logs = False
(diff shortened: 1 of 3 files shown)


modal/_output.py
```diff
@@ -174,6 +174,7 @@ def __init__(
         self._status_spinner = step_progress(status_spinner_text)
         self._app_page_url = None
         self._show_image_logs = False
+        self._status_spinner_live = None
```
erikbern, line 177: this was actually a real bug – not sure how it didn't fire before

PR #2026, opened 2024-07-17
PTY simplification 2

erikbern:
This is a slightly different approach versus https://github.com/modal-labs/modal-client/pull/2019 that retains the logic but gets rid of a bunch of complexity.

The other thing this accomplishes is that it moves pty output out of the OutputManager class. The reason this is useful is that it lets me "destroy" the future output manager singleton on a pty event. This simplifies turning the output manager into a singleton (which I intend to start working on next).

Afaict there's no need to buffer pty data written to stdout. My guess is that we had it there for historical reasons. It passes the test I added in #2022 and all integration tests. I've also verified manually that `modal run -i` works.

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -353,28 +353,8 @@ class OutputManager:
         stream.write(log.data)
 
     async def put_raw_content(self, log: api_pb2.TaskLogs):
-        # TODO(erikbern): move this out of the OutputMgr?
-        if hasattr(self._stdout, "buffer"):
-            # If we're not showing progress, there's no need to buffer lines,
-            # because the progress spinner can't interfere with output.
-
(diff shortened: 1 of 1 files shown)


mwaskom (approved):
> Afaict there's no need to buffer pty data written to stdout. My guess is that we had it there for historical reasons.

Hm, weird, but I guess if it's some nuanced interactive thing it won't break-at-scale and people will let us know.

erikbern:
> Hm, weird, but I guess if it's some nuanced interactive thing it won't break-at-scale and people will let us know.

Yep I think this code is also a lot less widely used since `modal shell` uses sandboxes now instead.

erikbern:
Thanks @mwaskom

PR #2029, opened 2024-07-17
Always construct OutputManager in _run_app

erikbern:
Makes it easier to turn it into a singleton if it's not passed around


erikbern:
need to start in a different place

PR #2030, opened 2024-07-18
`OutputManager` singleton

erikbern:
It works but apologies that this PR is more plus than negative. We're not reaping the benefits yet – I'm only taking advantage of it in a few places. In particular we can:

* Get rid of the `set_output_mgr` methods on functions/classes
* Get rid of a bunch of wrapper methods on the resolver

And a few more things. Will do those in subsequent PRs.

Note that this changes the behavior slightly with `app.run`. Previously this used to print output:

```python
with app.run():
    f.remote()
```

However going forward you need this:

```python
with enable_output():
    with app.run():
        f.remote()
```

This is intentional! IMO when using Modal "as a library" we should never output anything by default – libraries shouldn't do that. This is also the reason I added helper methods a few days ago to get logs programmatically. You can use `app._logs()` and `image._logs()` (although those are experimental still).

When using Modal "as a framework" we output by default though – `modal run` works just like usual.

### DIFF
--- a/test/app_test.py
+++ b/test/app_test.py
@@ -7,7 +7,7 @@ import time
 from google.protobuf.empty_pb2 import Empty
 from grpclib import GRPCError, Status
 
-from modal import App, Dict, Image, Mount, Secret, Stub, Volume, web_endpoint
+from modal import App, Dict, Image, Mount, Secret, Stub, Volume, enable_output, web_endpoint
 from modal._output import OutputManager
 from modal.app import list_apps  # type: ignore
 from modal.exception import DeprecationError, ExecutionError, InvalidError, NotFoundError
@@ -400,8 +400,38 @@ def test_app_interactive(servicer, client, capsys):
     assert captured.out.endswith("\nsome data\n\r")
 
 
-def test_show_progress():
+def test_show_progress_deprecations(client):
     app = App()
-    with pytest.raises(DeprecationError, match="show_progress"):
-        with app.run(show_progress=True):
+
+    # If show_progress is not provided, and output is not enabled, warn
+    with pytest.warns(DeprecationError, match="enable_output"):
+        with app.run(client=client):
             pass
+
+    # If show_progress is not provided, and output is enabled, no warning
+    with enable_output():
+        with app.run(client=client):
+            pass
+
+    # If show_progress is set to True, and output is not enabled, warn
+    # This is the only one that I'm not sure about - should we auto-enable output for now?
+    with pytest.warns(DeprecationError, match="enable_output"):
+        with app.run(client=client, show_progress=True):
+            pass
+
+    # If show_progress is set to True, and output is enabled, warn the flag is superfluous
+    with pytest.warns(DeprecationError, match="`show_progress=True` is deprecated"):
+        with enable_output():
+            with app.run(client=client, show_progress=True):
+                pass
+
+    # If show_progress is set to False, and output is not enabled, no warning
+    # This mode is currently used to suppress deprecation warnings, but will in itself be deprecated later.
+    with app.run(client=client, show_progress=False):
+        pass
+
+    # If show_progress is set to False, and output is enabled, warn that it has no effect
+    with pytest.warns(DeprecationError, match="no effect"):
+        with enable_output():
+            with app.run(client=client, show_progress=False):
+                pass
--- a/test/app_test.py
+++ b/test/app_test.py
@@ -400,7 +400,10 @@ def test_app_interactive(servicer, client, capsys):
     assert captured.out.endswith("\nsome data\n\r")
 
 
-def test_show_progress_deprecations(client):
+def test_show_progress_deprecations(client, monkeypatch):
+    # Unset env used to disable warning
+    monkeypatch.delenv("MODAL_DISABLE_APP_RUN_OUTPUT_WARNING")
+
     app = App()
 
     # If show_progress is not provided, and output is not enabled, warn
--- a/test/app_test.py
+++ b/test/app_test.py
@@ -8,6 +8,7 @@ from google.protobuf.empty_pb2 import Empty
 from grpclib import GRPCError, Status
 
 from modal import App, Dict, Image, Mount, Secret, Stub, Volume, web_endpoint
+from modal._output import OutputManager
 from modal.app import list_apps  # type: ignore
 from modal.exception import DeprecationError, ExecutionError, InvalidError, NotFoundError
 from modal.partial_function import _parse_custom_domains
@@ -391,8 +392,9 @@ def test_app_interactive(servicer, client, capsys):
     with servicer.intercept() as ctx:
         ctx.set_responder("AppGetLogs", app_logs_pty)
 
-        with app.run(client=client):
-            time.sleep(0.1)
+        with OutputManager.enable_output():
+            with app.run(client=client):
+                time.sleep(0.1)
 
     captured = capsys.readouterr()
     assert captured.out.endswith("\nsome data\n\r")
--- a/test/app_test.py
+++ b/test/app_test.py
@@ -409,7 +409,7 @@ def test_show_progress_deprecations(client, monkeypatch):
     # If show_progress is not provided, and output is not enabled, warn
     with pytest.warns(DeprecationError, match="enable_output"):
         with app.run(client=client):
-            pass
+            assert OutputManager.get() is not None  # Should be auto-enabled
 
     # If show_progress is not provided, and output is enabled, no warning
     with enable_output():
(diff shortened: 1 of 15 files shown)


mwaskom:
> However going forward you need this:
> ```
> with enable_output():
>     with app.run():
>         f.remote()
> ```

The rationale here makes perfect sense but this feels a little cumbersome especially if `enable_output` mostly exists to be used _with_ `app.run`. Could we maybe have

```python
with app.run(output=True):
    ....
```

Maybe worth spending a few minutes thinking about the best name.

erikbern:
> The rationale here makes perfect sense but this feels a little cumbersome especially if `enable_output` mostly exists to be used _with_ `app.run`. Could we maybe have

We used to have the `show_progress` bool which I just deprecated in the same PR. Could potentially add it back if we want to? I'm somewhat leaning towards the context manager syntax since it would be useful in other app-less contexts:

```python
with modal.enable_output():
    sandbox = modal.Sandbox.create(...)
```

One option is to keep `show_progress` for how but emit a warning.

mwaskom:
Are there usecases for `enable_output` beyond (a) `app.run()` and (b) `Sandbox.create`?

erikbern:
> Are there usecases for `enable_output` beyond (a) `app.run()` and (b) `Sandbox.create`?

Not that I can think of today

mwaskom:
And do you think there's a usecase for separating the scope of `enable_output` from the scope of `app.run`?

erikbern:
Btw one general challenge with something like `app.run(output=True)` is that the output manager is now a singleton, so `app.run(output=True)` wouldn't necessarily be concurrency safe since it modifies the global scope.

I.e. if one thread does

```python
with app.run(output=True):
    ...
```

And another thread does

```

```python
with app.run(output=False):
    ...
```

Then the second thread will override the first thread. Maybe a bit of an edge case. This PR breaks this behavior to be clear – we used to support it, but having localized output control is kind of annoying imo.

mwaskom:
I'm not following how that's different than using the context manager in one thread but not the other. Won't the singleton pattern also make the state change leak?

erikbern:
> I'm not following how that's different than using the context manager in one thread but not the other. Won't the singleton pattern also make the state change leak?

This PR introduces the singleton – there's no state leakage in current main. But the introduction of the singleton means I think it's semantically wrong to support `app.run(show_progress=True)` since keeping that flag would introduce stage leakage. That's why I think it makes more sense to remove that flag and make people use a separate context manager which makes the global state explicit.

mwaskom:
I think there are two different effects that `show_progess=False` or `enable_output=False` could have:

- It could toggle the singleton to False; I think this would cause the state leakage that you're worried about
- It could just leave the singleton state unchanged, which should be identical to the case where you use a `with enable_output` context manager in one location but not the other.

I think either way you could have some state leakage where an app run that has not explicitly opted into seeing output would start seeing it. Maybe you think that's more confusing if you've explicitly spelled it as `show_progess=False`?

Personally I think this is kind of an edge case, and otherwise having it as a parameter in the two functions where you'd use it is more discoverable and simpler. But idk, maybe it's an argument against the singleton pattern altogether if you think that the concurrent-runs-with-different-output-desires usecase will be common.

erikbern:
I think this edge case with state leakage is very rare to be clear – that's why I think it's fine to use a singleton for output management. I think I just have some sort of philosophical aversion about object methods altering global state, so that's why I'm a bit against having a flag on `app.run` going forward.

Tbh the bigger user concern I have is that the default used to be to always output, whereas going forward it will default to not output. So I think some users who run `app.run` might be confused about that? We should probably add some more docs about it to `app.run` and to the changelog.

mwaskom:
> So I think some users who run app.run might be confused about that? We should probably add some more docs about it to app.run and to the changelog.

Yeah — I guess the "right" thing to do here would be to start showing a warning message that can be disabled by using the new syntax. Not sure how much that would hamper getting the rest of the changes in though.

erikbern:
> Yeah — I guess the "right" thing to do here would be to start showing a warning message that can be disabled by using the new syntax. Not sure how much that would hamper getting the rest of the changes in though.

Right but you don't want to show a warning on `app.run()` I think – it's the default behavior change that worries me. Although I sort of expect that this isn't super common. My feeling is that most people use `modal run`

mwaskom:
>  Right but you don't want to show a warning on app.run() I think

Why not? I mean in the `show_progress=True` case. Although, we should probably also warn on `show_progress=False` too since that code will break on an update.

erikbern:
> Why not? I mean in the `show_progress=True` case. Although, we should probably also warn on `show_progress=False` too since that code will break on an update.

In those cases sure! But what if the flag isn't provided at all?

mwaskom:
Given that we're changing default behavior — and that default behavior is "show some output" — we should still warn. But the forward-looking way of doing things (i.e., using the `enable_output` context manager) should also flip some state that disables the warning.

mwaskom:
Actually nvm this is more complicated!

mwaskom:
Scenarios:

- You are relying on current default behavior _and want to keep getting output_ or explicitly setting `show_progress=True`
   -> you're happy getting a warning directing you to the new way of getting output
- You are currently using `show_progress=False` because you don't want output
   -> Showing a warning that this will break makes sense, but if you just remove the parameter to accept the future default, you'll get a warning that you can't silence! Bad!
- You are currently accepting the default behavior but actually would prefer not to see output, you just didn't notice it was an option
  -> You can't be _that_ bothered by the output, so maybe it's fine to see a warning and know that things will change in the future

The second bullet clearly the hard one to solve for.

One not great solution is some kind of `MODAL_APP_RUN_OUTPUT_WARNING=0` env var that you can set, but that's kind of yuck.

Another solution would be to change the _default_ behavior now, and only deprecate `show_progress` later. Even slower migration process, but maybe fine?

I do suspect that more users are in the first bullet than the second bullet but it's hard to know for sure. I'll think about it a little more.

erikbern:
Ok thinking about it I think the most gentle way to roll this out would be to:

* If `show_progress` is not provided (i.e. we default it to `None`):
    * If `enable_output` is not run, point out that output will be dropped and they should run `enable_output` or set `show_progress=False`
* If `show_progress` is `True`
    * If `enable_output` is not run, run it automatically and point out that this will be required soon and that they should update their code
* If `show_progress` is `False`: do nothing

... wait a few months. then change the logic to this:

* If `show_progress` is not provided (i.e. we default it to `None`): do nothing
* If `show_progress` is `True`: raise a deprecation error
* If `show_progress` is `False`: print a deprecation warning that this has no effect and they should remove this flag

... wait a few months. then change the logic to this:

* If `show_progress` is not provided (i.e. we default it to `None`): do nothing
* If `show_progress` is `True`: raise a deprecation error
* If `show_progress` is `False`: raise a deprecation error

mwaskom:
Yeah that sounds right to me as the smoothest deprecation path.

erikbern:
Haha ok stay tuned for 100 lines code + test for all the different deprecation options. I'll look at it in the next few days.

erikbern:
Added a ton of deprecations. I chose the most aggressive route that still doesn't _break_ code – but basically if `show_progress is None` or `True` then we'll _not_ print output, but will print a warning about it asking the user to enable output. I think a slight change in behavior is justified since it's unlikely to actually break anything downstream (unless they run Modal in a subprocess and read the output or something?)

erikbern:
Had to update a ton of tests to get this working. It's a bit annoying because all I did was to add `show_progress=False` to suppress the warning and then we're going to have to remove it later. But it's super mechanical and pretty easy to do that later.

test/app_test.py
```diff
@@ -391,8 +392,49 @@ async def app_logs_pty(servicer, stream):
...
-            time.sleep(0.1)
+        with OutputManager.enable_output():
+            with app.run(client=client):
+                time.sleep(0.1)
 
     captured = capsys.readouterr()
     assert captured.out.endswith("\nsome data\n\r")
+
+
+def test_show_progress_deprecations(client, monkeypatch):
```
mwaskom, line 403: Would be more annoying but should we also be testing the actual logging / not logging behavior here too?

test/app_test.py
```diff
@@ -391,8 +392,49 @@ async def app_logs_pty(servicer, stream):
...
+        with app.run(client=client):
+            pass
+
+    # If show_progress is not provided, and output is enabled, no warning
+    with enable_output():
+        with app.run(client=client):
+            pass
+
+    # If show_progress is set to True, and output is not enabled, warn
+    # This is the only one that I'm not sure about - should we auto-enable output for now?
```
mwaskom, line 420: I think this should show output during the deprecation, yeah.

mwaskom (approved):
Thanks for taking care on the deprecation pathway — will save our users some headaches for sure!

test/app_test.py
mwaskom: Yeah — it occurred to me that one reason you might (a) be running programmatically and (b) care about the output is if you're running tests. I'd be peeved if my CI broke unexpectedly on a client version update.

test/app_test.py
erikbern: I added some assertions that checks that the singleton is enabled!

PR #2032, opened 2024-07-18
Deprecate setting _allow_background_volume_commits, ahead of removal

thundergolfer:
## Describe your changes

We need to remove `_allow_background_volume_commits` from the client but first we must give users notice about this if they're still doing `_allow_background_volume_commits=True`. 

---

---

</details>

## Changelog

* Setting `_allow_background_volume_commits` is no longer necessary and has been deprecated. Remove this argument in your decorators.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -507,7 +507,7 @@ class _App:
         interactive: bool = False,  # Deprecated: use the `modal.interact()` hook instead
         secret: Optional[_Secret] = None,  # Deprecated: use `secrets`
         # Parameters below here are experimental. Use with caution!
-        _allow_background_volume_commits: Optional[bool] = None,
+        _allow_background_volume_commits: None = None,
         _experimental_boost: bool = False,  # Experimental flag for lower latency function execution (alpha).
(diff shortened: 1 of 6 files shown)


modal/app.py
```diff
@@ -824,9 +824,16 @@ async def spawn_sandbox(
             raise InvalidError("`app.spawn_sandbox` requires a running app.")
 
         if _allow_background_volume_commits is False:
-            deprecation_warning(
+            deprecation_error(
                 (2024, 5, 13),
-                "Disabling volume background commits is now deprecated. Set _allow_background_volume_commits=True.",
+                "Disabling volume background commits is now deprecated. "
+                "Remove this argument to enable the functionality.",
```
mwaskom, line 830: Something about this reads a little confusingly to me — but not a huge deal.

mwaskom (approved):

modal/app.py
thundergolfer: Yeh I tried a few wordings and couldn't come up with something nice.

PR #2033, opened 2024-07-18
Add missing RPCs to audit logs

luiscape:
We only track RPCs in audit logs when we tag the RPC with `[ (modal.options.audit_target_attr) = true ]`. This add that tag to a number of missing RPCs.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -691,7 +691,7 @@ message ContainerHeartbeatResponse {
 }
 
 message ContainerStopRequest {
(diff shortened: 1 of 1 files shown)


mwaskom:
I wonder if there's any good way to automatically check that relevant fields have this added.

luiscape:
@mwaskom me too. The rule of thumb I use to add these is to track only "relevant mutation RPCs". That's the case because we don't want to track calls to "get" or "list" objects for the most part.

PR #2035, opened 2024-07-19
Remove some of the output logic on the resolver

erikbern:
Hit the singleton directly from the image.

One out of several simplifications we can do since #2030

### DIFF
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -6,8 +6,6 @@ from typing import TYPE_CHECKING, Dict, Hashable, List, Optional
 
 from grpclib import GRPCError, Status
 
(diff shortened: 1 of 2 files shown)


erikbern:
@prbot approve

modal/_resolver.py
```diff
@@ -178,24 +176,3 @@ def display(self):
 
     def add_status_row(self) -> StatusRow:
         return StatusRow(self._tree)
-
-    async def console_write(self, log: api_pb2.TaskLogs):
-        # TODO(erikbern): get rid of this wrapper
```
mwaskom, line 183: ✅

mwaskom (commented):
👍

PR #2038, opened 2024-07-19
doctrings: make Mount and Image method args clearer

thundergolfer:
## Describe your changes

An initial follow-up to struggles a user had 'bringing their own `Dockerfile`. 

---

    - just docs
    - just docs

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -301,7 +301,7 @@ class _Mount(_Object, type_prefix="mo"):
         *,
         # Where the directory is placed within in the mount
         remote_path: Union[str, PurePosixPath, None] = None,
-        # Predicate filter function for file selection; function should return `True` for inclusion, `False` otherwise.
+        # Predicate filter function for file selection, which should accept a filepath and return `True` for inclusion.
         # Defaults to including all files.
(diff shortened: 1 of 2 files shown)


modal/mount.py
```diff
@@ -301,7 +301,8 @@ def add_local_dir(
         *,
         # Where the directory is placed within in the mount
         remote_path: Union[str, PurePosixPath, None] = None,
-        # Filter function for file selection; defaults to including all files
+        # Predicate filter function for file selection; function should return `True` for inclusion, `False` otherwise.
+        # Defaults to including all files.
```
mwaskom, line 305: Could we clarify that the function is passed the filepath? Maybe that's obvious but we don't explicitly say it.

mwaskom (commented):
I think it would be great to also add some text in the `from_dockerfile` docstring to manage expectations — we're _translating_ Dockerfile syntax to our own Image build logic and don't currently support all commands / options. Some additional examples might be helpful too — i.e. make it more obvious that `context_mount` needs to be used if your Dockerfile has COPY commands.

mwaskom:
Thanks for doing this by the way!

erikbern:
Great!!

PR #2041, opened 2024-07-22
Remove client-side check for minimum CPU core request

devennavani:
## Describe your changes

Remove client check for CPU core request being at least 0.1. Let's let the server enforce a minimum CPU core request.

## Changelog

- Removed client check for CPU core request being at least 0.1, deferring to server-side enforcement.

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -517,7 +517,7 @@ class _ContainerIOManager:
             # just skip creating any output for this input and keep going with the next instead
             # it should have been marked as cancelled already in the backend at this point so it
             # won't be retried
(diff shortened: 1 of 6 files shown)


ekzhang (approved):
As discussed on Slack briefly with @mwaskom I think this is fine! We have a more accurate check on the server-side, and the client-side check could cause version skew over time if we kept it in.

mwaskom:
Is one downside here that the error becomes dissociated from the code with the incorrect value? (Since the server-side error will come out of `FunctionCreate`, deep within the `create_all_objects` / `Resolver` stack?) Could imagine that being an ergonomic regression for apps with a lot of functions.

PR #2043, opened 2024-07-23
Run / deploy apps via new AppPublish RPC

mwaskom:
Client-side counterpart to https://github.com/modal-labs/modal/pull/14147. Ephemeral (and deployed) apps now go through `AppPublish` instead of `AppSetObjects` (and `AppDeploy`). That allows all mutations to be done server-side in one transaction and rolled back in the case of an error.

I believe this closes MOD-1549; although we don't have a very reliable repro of that behavior to test.

## Changelog

- App deployment events are now atomic, reducing the risk that a failed deploy will leave the App in a bad state.


mwaskom:
@prbot approve

aksh-at (commented):
👍

PR #2048, opened 2024-07-24
[MOD-3200]: better error message when running cls.with_options without an app

kramstrom:
Adds an error message (same as for `resolve()` [here](https://github.com/modal-labs/modal-client/blob/main/modal/object.py#L206-L216)) when running `SomeClass.with_options(...)` without associating an app

### DIFF
--- a/modal/object.py
+++ b/modal/object.py
@@ -120,10 +120,7 @@ class _Object:
         # Transient use case, see Dict, Queue, and SharedVolume
         self._init(other._rep, other._load, other._is_another_app, other._preload)
 
-    def clone(self: O) -> O:
-        """mdmd:hidden Clone a given hydrated object."""
-
-        # Object to clone must already be hydrated, otherwise from_loader is more suitable.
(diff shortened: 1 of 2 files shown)


modal/object.py
```diff
@@ -124,7 +124,17 @@ def clone(self: O) -> O:
...
+            object_type = self.__class__.__name__.strip("_")
+            if hasattr(self, "_app") and getattr(self._app, "_running_app", "") is None:
+                # The most common cause of this error: e.g., user called a Function without using App.run()
+                reason = ", because the App it is defined on is not running."
+            else:
+                # Technically possible, but with an ambiguous cause.
+                reason = ""
+            raise ExecutionError(
+                f"{object_type} has not been hydrated with the metadata it needs to run on Modal{reason}."
+            )
```
mwaskom, line 137: Nice! It looks like we might be duplicating this logic; could we factor it out into a common utility function instead?

PR #2050, opened 2024-07-25
Add R2 to the list of allowed bucket URLs

pawalt:
This commit adds R2 to the list of allowed bucket URLs so that we can support pulling inputs and outputs from both R2 and S3.

## Describe your changes

We are migrating some workloads to R2, so we need to update our validation to support this URL.

---

---

</details>


pawalt:
@prbot approve

irfansharif (commented):
LGTM

PR #2058, opened 2024-07-29
Split the step_completed function

erikbern:
Not a Fowler zealot but in this case I think this is very true: https://martinfowler.com/bliki/FlagArgument.html

Also think it's a bit silly to have documentation for this function for what's basically just adding a symbol

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -70,14 +70,12 @@ def step_progress_update(spinner: Spinner, message: str):
     spinner.update(text=message)
 
 
-def step_completed(message: str, is_substep: bool = False) -> RenderableType:
-    """Returns the element to be rendered when a step is completed."""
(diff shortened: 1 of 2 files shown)


mwaskom:
> My reasoning here is that the separate methods communicate more clearly what my intention is when I make the call. Instead of having to remember the meaning of the flag variable when I see book(martin, false) I can easily read regularBook(martin).

FWIW I think this is completely moot in languages where you can pass an explicit keyword. Totally agree that `do_something(arg, True)` is really ambiguous and makes it hard to read the code, but `do_something(arg, force=True)` is very clear.

That said, any individual function with a flag may make more sense from a code organization perspective to be split into two functions — seems like the case here! — I just don't think the main code readability argument there is relevant in Python.

PR #2060, opened 2024-07-30
Remove all output mgmt from the resolver

erikbern:
Instead of storing a Rich `Tree` on the resolver, create it on the output manager.

For convenience reasons it's useful to construct an _insivisble_ tree if output isn't enabled. That way everywhere we add status rows, we don't have to check if output is enabled. However this means we need to use a class variable since the singleton may not exist.

We might want to change this in the future if we don't want Rich to be a dependency. It's not, hard just need to add a bunch of `if` statements.

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -389,7 +389,7 @@ class OutputManager:
     @classmethod
     @contextlib.contextmanager
     def make_tree(cls):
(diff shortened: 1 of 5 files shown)


mwaskom (approved):

erikbern:
danke @mwaskom

PR #2062, opened 2024-07-31
Fix issue with global output manager tree

erikbern:
There's currently a bug related to the tree we show when building app. This bug didn't trigger because of test state leakage so I fixed that too.

This code is a bit hairy but I'll try to clean some of it up shortly.


erikbern:
@prbot approve

charlesfrye:
lgtm

PR #2066, opened 2024-08-01
Re-release of atomic deployments through AppPublish

mwaskom:
Reverts the reversion of #2043, using the new explicit `defer_updates` flag in the `FunctionCreateRequest`.

~**[DO BEFORE MERGE]** I am hardcoding the v0.64.1 version so I can try to run integration tests against it. Should roll that back to v0.64.0 manually before merging so that we don't accidentally skip a version. (Not the worst thing in the world, but would be a little confusing).~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ We appreciate your patience while we speedily work towards a stable release of t
 
 <!-- NEW CONTENT GENERATED BELOW. PLEASE PRESERVE THIS COMMENT. -->
 
+### 0.64.0 (2024-07-29)
+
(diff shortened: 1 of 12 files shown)


mwaskom:
All internal unit and integration tests pass against this branch: https://github.com/modal-labs/modal/pull/14534

mwaskom:
Only new line of code here compared to the previous release is https://github.com/modal-labs/modal-client/pull/2066/files#diff-87e1f4fd05c5e1972a1b8d781a518822704dd8fba567111f9b74ad8a3c9959fdR354

mwaskom:
@prbot approve

aksh-at (commented):
🥳

PR #2067, opened 2024-08-02
Add a --volume flag to `modal shell`

mwaskom:
Supports `modal shell --volume ...` to quickly attach one (or more) of your volumes to an add-hoc shell for filesystem exploration:

```
$ modal shell --volume example-models
root@modal:/# ls /mnt/example-models
gpt5.pt  gpt6.pt
```

You can call it multiple times, like `modal shell --volume models --volume data`

Volumes are always mounted at `/mnt/{name}`. I considered making this configurable, e.g. `modal shell --volume /mnt/data:my-data-vol`. But that felt like overkill for something that targets quick work. If we find a need for it, I think it's fine to expand the interface in the future to allow it. We're reasonably strict about volume names and I think any valid volume name would also be a valid mount point? (Open to counter-arguments here !).

- Closes MOD-2404

## Changelog

- Volumes can now be mounted to an ad hoc modal shell session:
    
    ```
    modal shell --volume my-vol-name
    ```
    When the shell starts, the volume will be mounted at `/mnt/my-vol-name`. This may be helpful for shell-based exploration or manipulation of volume contents.

    Note that the option can be used multiple times to mount additional models:
    ```
    modal shell --volume models --volume data
    ```

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -26,6 +26,7 @@ from ..functions import Function, _FunctionSpec
 from ..image import Image
 from ..runner import deploy_app, interactive_shell, run_app
 from ..serving import serve_app
+from ..volume import Volume
 from .import_refs import import_app, import_function
 from .utils import ENV_OPTION, ENV_OPTION_HELP, stream_app_logs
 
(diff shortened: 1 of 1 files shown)


ekzhang (approved):

modal/cli/run.py
```diff
@@ -341,6 +345,13 @@ def shell(
         default=None, help="Container image tag for inside the shell (if not using FUNC_REF)."
     ),
     add_python: Optional[str] = typer.Option(default=None, help="Add Python to the image (if not using FUNC_REF)."),
+    volume: Optional[typing.List[str]] = typer.Option(
+        default=None,
+        help=(
+            "Name of a `modal.Volume` to mount inside the shell at `/mnt/{name}` (if not using FUNC_REF)."
```
ekzhang, line 351: Inside the _container_ perhaps?

charlesfrye:
¿como se `llamas`?

Screenshot 2024-08-02 at 1 27 25 PM

This is great. I agree on the auto-mounting.

charlesfrye (commented):
lgtm

erikbern:
nice!

PR #2068, opened 2024-08-02
fix: cannot import text_unidecode, nemo-toolkit

thundergolfer:
## Describe your changes

Currently when Task tracing was turned on for a customer's `modal.App` the Task failed on startup due to errors in `importlib` machinery which do not occur normally. 

The first issue was https://github.com/pypa/setuptools/issues/4487. I could workaround that by doing `setuptools==70.3.0` but then I hit other issues. Those other issues are fixed by the changeset here. 

_There's almost certainly more to understand about how an intercepting `Loader` should work_ but I think this change shouldn't introduce regressions.

---

    - only an internal telemetry change
   - only an internal telemetry change

### DIFF
--- a/modal/_telemetry.py
+++ b/modal/_telemetry.py
@@ -41,6 +41,28 @@ class InterceptedModuleLoader(importlib.abc.Loader):
         spec.loader = self
         return module
 
+    def get_data(self, path: str) -> bytes:
+        """
+        Implementation is required to support pkgutil.get_data.
+
+        > If the package cannot be located or loaded, or it uses a loader which does
+        > not support get_data, then None is returned.
+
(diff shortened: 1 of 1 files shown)


aksh-at (approved):
nice! Are there any other methods in the `Loader` interface we need to implement?

thundergolfer:
> Are there any other methods in the Loader interface we need to implement?

This is somewhat answered by 

> There's almost certainly more to understand about how an intercepting Loader should work but I think this change shouldn't introduce regressions.

I would bet that there is, but understanding what needs to be implemented is a matter of grokking a graph of like a dozen abstract base classes that comprises the importlib system.

PR #2070, opened 2024-08-05
Reword log message when function raises a cancellation error

mwaskom:
The previous log message was a bit misleading as we can get to this line for reasons other than the user explicitly cancelling a function call — for example if the function that spawned the function call hits its retry limit, or potentially anything else that causes an `asyncio.CancelledError` to bubble up.

Related to MOD-3132. Not sure if it closes it —  we may be able to do more here — but wanted to do a quick fix to make the message less actively confusing.

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -517,7 +517,7 @@ class _ContainerIOManager:
             # just skip creating any output for this input and keep going with the next instead
             # it should have been marked as cancelled already in the backend at this point so it
             # won't be retried
-            logger.warning(f"The current input ({input_id=}) was cancelled by a user request")
+            logger.warning(f"Recevied a cancellation signal while processing input {input_id}")
             await self.complete_call(started_at)
             return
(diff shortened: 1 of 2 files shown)


modal/_container_io_manager.py
```diff
@@ -517,7 +517,7 @@ async def handle_input_exception(self, input_id, started_at: float) -> AsyncGene
             # just skip creating any output for this input and keep going with the next instead
             # it should have been marked as cancelled already in the backend at this point so it
             # won't be retried
-            logger.warning(f"The current input ({input_id=}) was cancelled by a user request")
+            logger.warning(f"Recevied a cancellation signal while processing input {input_id}")
```
erikbern, line 520: Received

erikbern (dismissed):

modal/_container_io_manager.py
mwaskom: 🤦

gongy (approved):
Thank you!

PR #2071, opened 2024-08-05
Distinguish between local and remote exceptions when stopping ephemeral app

mwaskom:
TK

- Closes MOD-3082


mwaskom:
@prbot approve

modal/_traceback.py
```diff
@@ -94,6 +97,15 @@ def reduce_traceback_to_user_code(tb: TracebackType, user_source: str) -> Traceb
     return tb
 
 
+def traceback_contains_remote_call(tb: Optional[TracebackType]) -> bool:
```
erikbern, line 100: maybe it's cleaner to set a magic attribute on the exception or something? we create this exception in code somewhere, i think.

mwaskom: I looked into that but I'm not convinced it's simpler. We construct the traceback with these special task-id-containing filenames server-side as a TracebackDict when then gets deserialized into an actual traceback object client-side. For ephemeral apps, all we have is the `run_app` context manager that yields out to a script contexts; but if we catch an exception there we don't know whether it was triggered locally or after a remote call. Then there's already some special exception type wrapping / unwrapping in synchronicity too when we go through `Function.remote`...

PR #2072, opened 2024-08-05
[MOD-3431] Add default value field to ClassParameterSpec

devennavani:
[MOD-3431]

Add `has_default` field to `ClassParameterSpec` to indicate whether arg of `__init__` was given default value.

Also include `oneof` field to store the default value. If `ParameterType` is `ParameterType.PARAM_TYPE_STRING` then the `string_default` field should be present, else if `ParameterType` is `ParemeterType.PARAM_TYPE_INT` then the `int_default` field should be present.


freider (commented):
nice!

PR #2073, opened 2024-08-05
Split AppList into client and internal web RPCs

ekzhang:
As part of MOD-3370, I'm going to split out the RPCs for listing apps on the server, with the AppList RPC used by the client library's `modal app list` command.

Right now these are the same RPC, which is not great for either use case because:
- The client gets returned extra data that isn't useful for it
- The web server gets returned data that it doesn't use, and doesn't receive data that it needs, so the frontend needs to make 4+ additional requests for every app that it displays, which slows down the modal.com/apps page and creates unnecessary server load. It also means that we can't sort or filter the list by fields like number of calls, who deployed the app, or number of errors.

So we're splitting them out. The current AppList RPC will continue to be used so we keep compatibility with existing clients.

I'm removing these line of code and moving them to the server. There will be a brief interval where `modal app list` reports more recently-stopped apps than usual in the CLI while the server change is being rolled out:

```python
        if (
            # Previously, all deployed objects (Dicts, Volumes, etc.) created an entry in the App table.
            # We are waiting to roll off support for old clients before we can clean up the database.
            # Until then, we filter deployed "single-object apps" from this output based on the object entity.
            (app_stats.object_entity and app_stats.object_entity != "ap")
            # AppList always returns up to the 250 most-recently stopped apps, which is a lot for the CLI
            # (it is also used in the web interface, where apps are organized by tabs and paginated).
            # So we semi-arbitrarily limit the stopped apps to those stopped within the past 2 hours.
            or (
                app_stats.state in {api_pb2.AppState.APP_STATE_STOPPED} and (now - app_stats.stopped_at) > (2 * 60 * 60)
            )
        ):
```

### Tests

All tests pass except for version_test, which is an unrelated issue.

```
========================= 1 failed, 653 passed, 2 skipped, 2 warnings in 338.00s (0:05:37) ==========================
```

### Server PR

corresponding server PR: https://github.com/modal-labs/modal/pull/14576

### Changelog

- Remove support for the undocumented `modal.apps.list_apps()` function, which was internal and not intended to be part of public API.


mwaskom (approved):
Makes sense

ekzhang:
@mwaskom how should we fix static type checks here?

ekzhang:
Is `from modal.app import list_apps` a supported public API? It seems to leak Protobuf implementation details

mwaskom:
> Is from modal.app import list_apps a supported public API? It seems to leak Protobuf implementation details

I don't think we intend for it to be, although I'm sure someone has found it and used it...

We have an old ticket for improving the object-listing API: https://linear.app/modal-labs/issue/MOD-2779/add-an-api-method-to-list-named-objects

ekzhang:
I removed `modal.app.list_apps()` to appease the static type translator, and since it's undocumented. Wrote a changelog entry since we were technically exposing it to users.

PR #2075, opened 2024-08-08
`modal shell` uses `container_exec`

aksh-at:
Noticed there was a lot of code duplication across `modal shell` and `container_exec`. The code was really hard to follow, with both calling `connect_to_terminal` eventually with two functions that had very specific call signatures (e.g. `connect_to_terminal` would call a function that was passed in with an `event` that it created). 

This PR makes it so that `modal shell` just does the equivalent of `container exec` in a separate process. 

Still not quite done:
- [x] Fix tests
- [x] Confirm error handling still works
- [x] Figure out what to run in the main container process — does `sleep infinity` work on all distros? It's possible that makes this approach unviable.


erikbern:
I'm not familiar with this code but this definitely looks like a huge simplification so 👍

aksh-at:
@prbot approve

aksh-at:
@prbot approve

erikbern:
👍

PR #2077, opened 2024-08-08
move import telemetry earlier in entrypoint

thundergolfer:
## Describe your changes

Installs telemetry as earlier as possible to provide tracing on more imports during startup.

**Before**

<kbd>
image
</kbd>

**After**

<kbd>
image
</kbd>

---


gimaker (commented):
:+1:

PR #2081, opened 2024-08-08
Process batch input output 2

cathyzbn:
## Describe your changes

Refactor container inputs/outputs logic to support multiple processing multiple inputs and serializing multiple outputs when function is batched.


gongy (approved):
lgtm. I think we evaluated using type checking but it was hard to type check in utils

PR #2082, opened 2024-08-08
Fix return type annotation of _Function.from_args

devennavani:
Fix return type annotation of _Function.from_args, from `-> None` to `-> "_Function"`


mwaskom:
Closing as stale; I guess this ended up being more complicated?

PR #2084, opened 2024-08-09
MOD-3467: App deployment history client cli

kramstrom:
- **app deployment history proto**
- **app deployment history cli**

## Describe your changes

---

---

</details>

### DIFF
--- a/modal/cli/app.py
+++ b/modal/cli/app.py
@@ -111,3 +111,60 @@ async def stop(
         app_id = await get_app_id_from_name.aio(name, env, client)
     req = api_pb2.AppStopRequest(app_id=app_id, source=api_pb2.APP_STOP_SOURCE_CLI)
     await client.stub.AppStop(req)
+
+
+@app_cli.command("history", no_args_is_help=True)
+@synchronizer.create_blocking
+async def history(
+    app_id: str = Argument("", help="Look up an App's deployment history by its ID"),
+    *,
+    env: Optional[str] = ENV_OPTION,
+    json: bool = False,
+):
+    """Show App Deployment history
+
+    **Examples:**
+
+    Get the history based on an app ID:
+
+    ```bash
+    modal app history ap-123456
+    ```
+
+    """
+    if not app_id:
+        raise UsageError("Must pass either an ID or a name.")
+
+    env = ensure_env(env)
+    client = await _Client.from_env()
+
+    resp: api_pb2.AppDeploymentHistoryResponse = await client.stub.AppDeploymentHistory(
+        api_pb2.AppDeploymentHistoryRequest(app_id=app_id)
+    )
+
+    columns: List[Union[Column, str]] = [
+        "id",
+        "Version",
+        "Deployed by",
+        "Deployed at",
+        "State",
+        "Client Version",
+        "Tag",
+    ]
+    rows: List[List[Union[Text, str]]] = []
+    for app_stats in resp.app_deployment_histories:
+        state = APP_STATE_TO_MESSAGE.get(app_stats.state, Text("unknown", style="gray"))
+        rows.append(
+            [
+                str(app_stats.app_id),
+                str(app_stats.version),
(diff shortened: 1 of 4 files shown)


modal/cli/app.py
```diff
@@ -111,3 +111,60 @@ async def stop(
...
+
+    env = ensure_env(env)
+    client = await _Client.from_env()
+
+    resp: api_pb2.AppDeploymentHistoryResponse = await client.stub.AppDeploymentHistory(
+        api_pb2.AppDeploymentHistoryRequest(app_id=app_id)
+    )
+
+    columns: List[Union[Column, str]] = [
+        "id",
```
mwaskom, line 146: I don't think we need the App ID in this table, since it will be the same for all rows, right?

modal/cli/app.py
```diff
@@ -111,3 +111,60 @@ async def stop(
...
+        "Version",
+        "Deployed by",
+        "Deployed at",
+        "State",
```
mwaskom, line 150: I don't think we need this either -- it will be `stopped` for every version aside from the latest one, and the state of the latest deployment is already shown in `modal app list`.

modal/cli/app.py
```diff
@@ -111,3 +111,60 @@ async def stop(
...
+        "Client Version",
```
mwaskom, line 151: Can we shorten the title here to "Client" just to avoid a lot of dead space in the column? Should be clear that it's a version number from the contents.

modal/cli/app.py
```diff
@@ -111,3 +111,60 @@ async def stop(
...
+
+    Get the history based on an app ID:
+
+    ```bash
+    modal app history ap-123456
+    ```
+
+    """
+    if not app_id:
+        raise UsageError("Must pass either an ID or a name.")
```
mwaskom, line 136: I think we should support name-based lookup -- but if we don't, should remove this.

mwaskom (changes requested):
Couple general comments:

I think we should support name-based lookup here, since we support it in other `modal app` commands. Name-based lookup should only return if the app is currently deployed, so I don't think we have a risk of showing you the history for an old app.

I do think you're right to highlight a little weirdness about what happens when you stop an app and then "re"-deploy with the same name (i.e. , the history starts fresh), but I think it's at least consistent across different aspects of the system (logs, metrics, etc.) so fine to reflect it here.

Smaller nits:
- Can we sort the table by version, descending?
- Should we suppress show the `tag` column if none of the versions was given a tag?

modal/cli/app.py
kramstrom: Yeah this was just for testing

modal/cli/app.py
kramstrom: Hm, what if you roll back to a previous version? I actually think the web page for this is confusing since it doesn't include which version is the one running, you just have to assume it's the top one

mwaskom: Yeah it will always be the top one — the way rollbacks will work for us is that the version number will continue to increment and there will be some metadata indicating the state that we reset to. We’ll need to figure out how to enhance the table here when we add that metadata to the history response, ideally in a way that looks similar to what we do on the web.

mwaskom: If you want to make the active version more salient here you could highlight it in green or bold white or something similar to what we do with `modal profile list`

mwaskom (approved):
See comment about the sorting, which I'm not sure is implemented correctly.

Also, just style nitpicking at this point, but I'm not sure we need as much contrast between the current deployment and previous deployments here (compared to `modal profile list`). Since the apps are listed in chronological order, the only reason we'd need to call attention to the most recent deployment is to resolve any ambiguity about how rollbacks work. And you'd mostly be looking at the app *history* because you care about previous deployments, so I don't think it's that helpful to show them as dim gray. I'll leave it up to you though!

Otherwise 👍, thanks for doing this!

mwaskom:
Yay!

PR #2087, opened 2024-08-11
Separate out `StreamReader` and `StreamWriter` to separate file

aksh-at:
This is needed for the `sandbox.exec` PR (https://github.com/modal-labs/modal-client/pull/2088).

Note that this renames `LogsReader` to `StreamReader` to be consistent with `StreamWriter`. I believe no one should be using the class name directly, so this shouldn't have too much impact.

### Changelog

- `modal.sandbox.LogsReader` has been renamed to `modal.io_streams.StreamReader`
- `modal.sandbox.StreamWriter` has been moved to `modal.io_streams.StreamWriter`

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -1,0 +1,220 @@
+# Copyright Modal Labs 2022
+import asyncio
+from typing import TYPE_CHECKING, AsyncIterator, Optional, Tuple, Union
+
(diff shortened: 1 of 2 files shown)


aksh-at:
Added a smoother migration path after discussion with @mwaskom. Tested using:

```python
from modal.sandbox import LogsReader
print(LogsReader)
```

mwaskom (approved):
Nice!

PR #2088, opened 2024-08-11
Add `Sandbox.exec()` and use it to implement interactive shells

aksh-at:
`Sandbox.exec(...)` lets you execute a command in a running sandbox, and gives you a handle (`ContainerProcess`) to it that can stream input and output. 

This lets us unify a bunch of log streaming code that was duplicated earlier across `container_exec` and `Sandbox`. `container_exec` and `interactive_shell` are now implemented using `ContainerProcess.attach`.

Depends on https://github.com/modal-labs/modal-client/pull/2087

- [x] Tests

### Changelog

- Sandboxes now have an `exec()` method that lets you execute a command inside the sandbox container. `exec` returns a `ContainerProcess` handle for input and output streaming.

```python
sandbox = modal.Sandbox.create("sleep", "infinity")

process = sandbox.exec("bash", "-c", "for i in $(seq 1 10); do echo foo $i; sleep 0.5; done")

for line in process.stdout:
    print(line)
```


erikbern (approved):
the `if self._object_type == "sandbox'` is a bit ugly but other than that this looks awesome

aksh-at:
> the if self._object_type == "sandbox' is a bit ugly but other than that this looks awesome

Yeah I was considering using polymorphism for that instead, but thought it's cleaner to stick with this while we have just two types.

PR #2096, opened 2024-08-12
Change deciding logic on snapshot/restore

mattnappo:
## Describe your changes

---

---

</details>


mattnappo:
Note on compatibility: if an old client creates a task that lands on a new worker, then this flag could be set to `0`. In that case, `is_snapshotting_function` will be false, preventing any snapshot from happening on old clients in the case of a fallback.

thundergolfer:
I've marked this as draft until the relevant worker change has fully rolled out 👍

thundergolfer (approved):
LGTM, change rolled out at least a week ago now :)

PR #2097, opened 2024-08-13
Use paramspec/typevars to enable type inference in user code

freider:
* Bump to synchronicity v0.7.2 which improves support for type vars/generics
* Make modal.Function, FunctionCall and PartialFunction generics wrt call args and return values

This enables static type support in `func.remote(...)`, `func.spawn(...)` and `func.local(...)`, verified to enable auto-complete of function arguments in VSCode.

Also adds some basic mypy assertion tests to verify that we don't regress on this support in the future.

Known limitations:
* Doesn't work for `.map()` (and similar parallel mapping calls)
* Doesn't work for methods on Modal classes


kramstrom (approved):
💥

modal/partial_function.py
```diff
@@ -45,7 +45,7 @@ def all() -> "_PartialFunctionFlags":
 
 
 class _PartialFunction(typing.Generic[P, T]):
-    """Intermediate function, produced by @method or @web_endpoint"""
+    """Intermediate function, produced by @method, @web_endpoint, or @batched"""
```
mwaskom, line 48: Also `asgi_app`, `wsgi_app` right?

PR #2107, opened 2024-08-15
Print how fast we can deploy apps

erikbern:
Kind of silly thing but it makes Modal _feel_ a bit faster, I think:

image


mwaskom:
Will this look bad for deployments that involve an Image build?

PR #2116, opened 2024-08-16
Clarify that GCS is fully S3-compatible

pawalt:
This disclaimer came from a time when we couldn't write to GCS buckets. We have fixed this bug, so this commit updates the docs.

## Describe your changes

---

---

</details>


pawalt:
@prbot help a brother out

PR #2123, opened 2024-08-19
Adds client support for cancelling concurrent async inputs

freider:
Feature unlocked by server based on a `_experimental_concurrent_cancellations` flag on the function definition.

Also adds a `terminate_containers` option for `FunctionCallCancel`, which can be used if a user wants to ensure that a non-cooperative input/function can still be terminated at the expense of having to rerun any concurrently running inputs in another container (also needs server support)

## Describe your changes

---

---

</details>

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -114,6 +114,10 @@ class IOContext:
             cb = self._cancel_callback
             self._cancel_callback = None
             cb()
+        else:
+            # TODO (elias): This should not normally happen but there is a small chance of a race
+            #  between creating a new task for an input and attaching the cancellation callback
+            logger.warning("Unexpected: Could not cancel input")
 
     def _args_and_kwargs(self) -> Tuple[Tuple[Any, ...], Dict[str, List[Any]]]:
         if not self._is_batched:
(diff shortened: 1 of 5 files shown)


modal/_container_io_manager.py
```diff
@@ -169,7 +184,6 @@ class _ContainerIOManager:
     Then we could potentially move a bunch of the global functions onto it.
     """
 
-    cancelled_input_ids: Set[str]
```
freider, line 172: This wasn't used

modal/_container_io_manager.py
```diff
@@ -677,11 +683,15 @@ async def handle_input_exception(
...
                 results=results,
             )
-            await self.exit_context(started_at)
+            await self.exit_context(started_at, io_context.input_ids)
 
-    async def exit_context(self, started_at):
+    async def exit_context(self, started_at, input_ids: List[str]):
         self.total_user_time += time.time() - started_at
         self.calls_completed += 1
+        self.current_input = None
```
cathyzbn, line 691: `self.current_input_id` here?

freider: Ah thanks for catching! I'll update the test to check this too (although I suspect we'll get rid of the singular `current_input_id` state soon since it's not compatible with concurrent inputs or batching)

freider: Actually I'm removing this line altogether as I'm not sure if its safe/good to reset current_input_id between inputs (in case it's used to associate logs etc. to inputs)

freider:
@prbot approve

PR #2124, opened 2024-08-19
Add `set_local_concurrent_inputs`

cathyzbn:
## Describe your changes

Added `set_local_concurrent_inputs` option

</details>


modal/_container_io_manager.py
```diff
@@ -162,6 +162,86 @@ def validate_output_data(self, data: Any) -> List[Any]:
...
+            self._queue.put_nowait(None)
+
+    async def get_input_concurrency(self) -> int:
+        async with self._lock:
+            return self._input_concurrency
+
+    async def acquire(self):
+        await self._queue.get()
+
+    async def release(self):
```
gongy, line 183: if you change `async def` to `def` then you don't need a lock since you never yield

modal/_container_io_manager.py
```diff
@@ -209,9 +208,7 @@ def _init(self, container_args: api_pb2.ContainerArguments, client: _Client):
         self.current_input_id = None
         self.current_input_started_at = None
 
-        self._input_concurrency = None
-
-        self._semaphore = None
+        self._dynamic_semaphore = None
```
gongy, line 211: initialize with 0 slots & `set` to `input_concurrency` before `enter` function is called maybe

cathyzbn: Unfortunately in python<=3.9, the semaphore has to be instantiated in the same event loop as the acquire so can't instantiate here / before `enter` is called gives errors

modal/_utils/async_utils.py
```diff
@@ -460,3 +460,48 @@ def run_generator_sync(
...
+    async def _update_capacity(self):
+        if self._capacity < self._new_capacity:
+            for _ in range(self._new_capacity - self._capacity):
+                self.release()
+        elif self._capacity > self._new_capacity:
+            for _ in range(self._capacity - self._new_capacity):
+                if self._semaphore.locked():
+                    self._owed_releases += 1
+                else:
+                    await self._semaphore.acquire()
```
cathyzbn, line 505: there's actually a race here with `self._semaphore.acquire()` in `acquire` (line 479), might have to change this to `self._semaphore._value -= 1` to make the entire function sync

PR #2125, opened 2024-08-19
MOD-3209: add force & timeout flag to @method.build decorator

kramstrom:
Lets you set `timeout` and `force_build` parameters on class `@build` decorators:

```py
@app.cls()
class Foo:

    @build(timeout=123)
    def a(self):
        pass

    @build(force=True)
    def b(self):
        pass
```


modal/image.py
```diff
@@ -285,7 +285,7 @@ def _from_args(
         build_function_input: Optional[api_pb2.FunctionInput] = None,
         image_registry_config: Optional[_ImageRegistryConfig] = None,
         context_mount: Optional[_Mount] = None,
-        force_build: bool = False,
+        force_build: Optional[bool] = False,
```
freider, line 288: Would it ever be None? (seeing as it's now an optional?)

modal/partial_function.py
```diff
@@ -134,9 +142,8 @@ def _find_partial_methods_for_user_cls(
     return partial_functions
 
 
-def _find_callables_for_cls(user_cls: Type[Any], flags: _PartialFunctionFlags) -> Dict[str, Callable[..., Any]]:
-    """Grabs all method on a user class, and returns callables. Includes legacy methods."""
-    functions: Dict[str, Callable] = {}
+def _find_callables_for_cls(user_cls: Type[Any], flags: _PartialFunctionFlags) -> Dict[str, _PartialFunction]:
```
freider, line 145: Since this now returns PartialFunctions, it should probably be renamed (or just merged with `_find_partial_methods_for_user_cls`. It could be a bit confusing that `_find_callables_for_cls` returns PartialFunction whereas `_find_callables_for_obj` returns actual callable functions

kramstrom: merged them!

modal/image.py
kramstrom: No, changed it!

modal/partial_function.py
```diff
@@ -44,8 +44,8 @@ def all() -> int:
 R = typing_extensions.TypeVar("R", covariant=True)
 
 
-class _PartialFunction(typing.Generic[P, T]):
-    """Intermediate function, produced by @enter, @build, @method, @web_endpoint, or @batched"""
+class _PartialFunction(typing.Generic[P, R]):
+    """Intermediate function, produced by @enter, @build, @method, @web_endpoint or @batched"""
```
mwaskom, line 48: Clear your schedule for the rest of the day we need to bikeshed about oxford commas :D

mwaskom:
Is `build(force_build=True)` redundant? Would `build(force=True)` be cleaner? Could see arguments either way.

kramstrom:
> Is `build(force_build=True)` redundant? Would `build(force=True)` be cleaner? Could see arguments either way.

Yeah agreed, changed this now! Initially set this but @freider pointed out it could be confusing to have `force` in the `_PartialFunction` class (what are you forcing then) so I renamed it to `force_build` -- but nothing stopping us from having it be `force_build` in the class but just `force` in the `@build` method!

modal/partial_function.py
kramstrom: 😰

freider (commented):
lgtm!

freider (approved):
lgtm

PR #2127, opened 2024-08-19
Fix serialize_proto_params function

devennavani:
Fix serialize_proto_params function, bringing it to parity with what's on the server

### DIFF
--- a/modal/_serialization.py
+++ b/modal/_serialization.py
@@ -2,6 +2,7 @@
 import io
 import pickle
 import typing
+from dataclasses import dataclass
 from typing import Any
 
 from synchronicity.synchronizer import Interface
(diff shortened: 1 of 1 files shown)


freider:
The input type when using the client SDK won't be `str` like it is for web endpoints - it should always be of the same type that we expect to get, and sending in a str when an int is expected should be an error IMO (static type checking will think it is).

For that reason I think the converters should be slightly differently implemented here in the client, just validating that the input is of the right type (protobuf would do that later on anyways, but the error would be more cryptic to an end user).

Once we add more complex types like `PARAM_TYPE_PYTHON_OBJ` this would change slightly and the converter would be pickling the object and putting it in a `bytes_values` field (at that point we'd also need a deserializer that does the reverse to get the payload back in the container entrypoint)

PR #2136, opened 2024-08-21
don't suggest putting the secret values inline in the code

erikbern:
use the new `required_keys` to provide a hint about what keys need to exist


erikbern:
@prbot approve

freider (commented):
Game changer!

PR #2137, opened 2024-08-21
Add cli command to open dashboard for deployed app

kramstrom:
Allows you to write

```bash
modal dashboard <app-name>
```
or
```bash
modal dashboard <app-id>
```
to automatically open your app's dashboard page in the default browser

tbh not sure if we need this

- depends on proto PR: https://github.com/modal-labs/modal-client/pull/2141
- depends on rpc PR: https://github.com/modal-labs/modal/pull/15041

### DIFF
--- a/modal/cli/app.py
+++ b/modal/cli/app.py
@@ -282,8 +282,13 @@ async def dashboard(
     """
     env = ensure_env(env)
     client = await _Client.from_env()
-    app_id = await get_app_id.aio(app_identifier, env, client) if app_identifier else None
-    resp = await client.stub.WebDashboardLookup(api_pb2.WebDashboardLookupRequest(app_id=app_id, environment_name=env))
+    resp = await client.stub.WebDashboardLookup(
+        api_pb2.WebDashboardLookupRequest(
+            app_identifier=app_identifier if app_identifier else None,
+            namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
+            environment_name=env,
+        )
+    )
 
     if _open_url(resp.url):
         rich.print(
--- a/modal/cli/app.py
+++ b/modal/cli/app.py
@@ -283,7 +283,7 @@ async def dashboard(
     env = ensure_env(env)
     client = await _Client.from_env()
     app_id = await get_app_id.aio(app_identifier, env, client) if app_identifier else None
-    resp = await client.stub.WebDashboardLookup(api_pb2.WebDashboardRequest(app_id=app_id, environment_name=env))
+    resp = await client.stub.WebDashboardLookup(api_pb2.WebDashboardLookupRequest(app_id=app_id, environment_name=env))
 
     if _open_url(resp.url):
         rich.print(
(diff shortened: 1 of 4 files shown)


modal/cli/app.py
```diff
@@ -247,3 +248,44 @@ async def history(
...
+
+    ```bash
+    modal app dashboard my-app
+    ```
+
+    """
+    env = ensure_env(env)
+    client = await _Client.from_env()
+    app_id = await get_app_id.aio(app_identifier, env, client)
+    web_url = f"https://modal.com/apps/modal-labs/{env}/{app_id}"
```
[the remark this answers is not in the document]
kramstrom: lol yes 😆

modal/cli/app.py
```diff
@@ -247,3 +248,44 @@ async def history(
...
+    *,
+    env: Optional[str] = ENV_OPTION,
+):
+    """Open the dashboard for a deployed App in the web default browser.
+
+    **Examples:**
+
+    Open the dashboard based on an app ID:
+
+    ```bash
```
mwaskom, line 266: BTW using "bash" here causes the CLI help to look bad:
image

Not sure if there's a better way to fix that but since there's not actually any bash syntax to mark up here it will be marginally cleaner to omit it.

modal/cli/app.py
```diff
@@ -14,6 +14,7 @@
 from modal.environments import ensure_env
 from modal.exception import deprecation_warning
 from modal.object import _get_environment_name
+from modal.token_flow import _open_url
```
mwaskom, line 17: I guess it would be better for this to live somewhere more expected if we're going to share it across multiple places but we can just move it later.

mwaskom (commented):
Would it maybe make more sense to have 

```
modal dashboard
```

As the CLI command, which would send you to the dashboard home, and then add an option for showing a specific app? (Probably something as simple as `modal dashboard <app-identifier>`)

mwaskom:
BTW I have some concern about hard-coding routes in the CLI as we have changed those recently; it would have effectifely broken the CLI when we did that (it broke my bookmarks 😢). But maybe the answer there is just the web team knowing that they need to set up redirects?

modal/cli/app.py
mwaskom: Yeah probably should — I only noticed what it does to the terminal help recently.

kramstrom:
> BTW I have some concern about hard-coding routes in the CLI as we have changed those recently; it would have effectifely broken the CLI when we did that (it broke my bookmarks 😢). But maybe the answer there is just the web team knowing that they need to set up redirects?

Yep makes sense! Updated this to fetch the url from the server instead (same path as when deploying apps) and also moved the app id lookup from name to the server as well so there's no longer 2 api calls if you pass in the app name

mwaskom:
I wonder if we should anticipate this being more general in the future by spelling the app-specific invocation as `modal dashboard --app=<app-identifier>`. Future iterations of the modal dashboard could have other endpoints that you might want to go directly to. In fact, I think it may already be useful to go directly to the viewer for a volume. Maybe other things too?

PR #2138, opened 2024-08-21
Print warning when FunctionPutInputs fails with RESOURCE_EXHAUSTED

rculbertson:
## Describe your changes

MOD-3206

When the client calls FunctionPutInputs, a warning is now printed if the call fails with RESOURCE_EXHAUSTED. This can happen if the output is not consumed fast enough.

This was tested manually.

---

---

</details>

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -139,8 +139,9 @@ async def _map_invocation(
                     resp = await retry_transient_errors(
                         client.stub.FunctionPutInputs,
                         request,
-                        max_retries=7,
-                        max_delay=10,
+                        # with 8 retries we log the warning below about every 30 secondswhich isn't too spammy.
+                        max_retries=8,
+                        max_delay=15,
                         additional_status_codes=[Status.RESOURCE_EXHAUSTED],
                     )
                     break
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass
 from typing import Any, Callable, Dict, Optional, Set, Tuple
 
 from aiostream import pipe, stream
-from grpclib import Status
+from grpclib import GRPCError, Status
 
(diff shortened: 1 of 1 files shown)


modal/parallel_map.py
```diff
@@ -134,13 +134,21 @@ async def pump_inputs():
...
+                        request,
+                        max_retries=1,
+                        max_delay=101,
+                        additional_status_codes=[Status.RESOURCE_EXHAUSTED],
+                    )
+                    break
+                except GRPCError as err:
+                    if err.status != Status.RESOURCE_EXHAUSTED:
+                        raise err
+                    logger.warning("Unable to push inputs. Check rate of output consumption.")
```
rculbertson, line 150: Open to suggestions if the wording of the warning message can be improved!

gongy: For the wording -- maybe something more general like `Warning: map progress is limited. Common bottlenecks include slow iteration over results, or function backlogs.`

modal/parallel_map.py
```diff
@@ -134,13 +134,21 @@ async def pump_inputs():
...
-                max_delay=10,
-                additional_status_codes=[Status.RESOURCE_EXHAUSTED],
-            )
+            while True:
+                try:
+                    resp = await retry_transient_errors(
+                        client.stub.FunctionPutInputs,
+                        request,
+                        max_retries=1,
+                        max_delay=101,
```
gongy, line 143: I think maybe a higher `max_retries` maybe `7` could make sense -- then the user would be waiting for ~30 seconds between log lines which isn't too spammy.

modal/parallel_map.py
rculbertson: Thanks - will update!

gongy (approved):
This looks good! A comment on how we landed on `7` would seal the deal.

Discussed in person that this is difficult to unit-test, so we'll live without it.

PR #2139, opened 2024-08-21
Include checkpoint size in checkpoint info

cweld510:
## Describe your changes

<details> <summary>We need the size of a gvisor checkpoint so we can restore it with imagefs (see MOD-3415). This proto lives in the client repo, but isn't actually read or written by the client (it is included in other protos used by the client; this should be fine though because extra fields are ignored).</summary>

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -544,6 +544,7 @@ message CancelInputEvent {
 
 message CheckpointInfo {
   string checksum = 1;
+  int64 size = 5;
   CheckpointStatus status = 2;
   string checkpoint_id = 3;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -544,6 +544,7 @@ message CancelInputEvent {
 
 message CheckpointInfo {
   string checksum = 1;
+  int64 size = 5;
```
thundergolfer, line 547: We keep these ordered by field number.

thundergolfer (approved):
LGTM except for ordering

PR #2140, opened 2024-08-21
Small fixes to modal.interact

mwaskom:
- No longer print a warning when function has `concurrency_limit` set. While it is true that running interact from within a function that has multiple containers can cause problems, the check used for the previous warning was wrong (as the default `concurrency_limit = None` does not preclude concurrent containers), it's not always a problem, and we also print an error message when multiple streams are connected to the local PTY, which is a better signal of an actual problem.
- Fix the check for whether we tried to use `modal.interact()` without doing `modal run -i`; previously this was always passing because we were evaluating the truthiness of a protofuf _message_, which is always True even when undefined.
- Improve the error message for the check mentioned above, where we were directing the user to a since-deprecated environment variable. I changed this to a more general message (note that googling "modal interactive mode" pulls up exactly the right documentation) and also added some information to indicate whether the user tried to use `modal.interact()` directly or indirectly through `breakpoint()`
- Always set the `sys.breakpointhook` in `container_entrypoint` to run through out `interact` system so that we can error out informatively when the user forgot to do `modal run -i`.
- Add a test for the error regarding modal interactive mode.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -51,7 +51,7 @@ from .client import Client, _Client
 from .cls import Cls, Obj
 from .config import logger
 from .exception import ExecutionError, InputCancellation, InvalidError, deprecation_warning
-from .execution_context import _set_current_context_ids, interact
+from .execution_context import _set_current_context_ids
 from .functions import Function, _Function
 from .partial_function import (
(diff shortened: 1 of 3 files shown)


aksh-at (approved):
nice!

PR #2141, opened 2024-08-22
Add proto to fetch web dashboard url

kramstrom:
proto for fetching web dashboard url

required for:
- rpc PR: https://github.com/modal-labs/modal/pull/15041
- cli PR: https://github.com/modal-labs/modal-client/pull/2137

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2193,8 +2193,9 @@ message VolumeRemoveFileRequest {
 }
 
 message WebDashboardLookupRequest {
-  optional string app_id = 1;
-  string environment_name = 2;
+  optional string app_identifier = 1;
+  DeploymentNamespace namespace = 2;
+  string environment_name = 3;
 }
 
 message WebDashboardLookupResponse {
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -2378,6 +2387,9 @@ service ModalClient {
   rpc VolumeReload(VolumeReloadRequest) returns (google.protobuf.Empty);
   rpc VolumeRemoveFile(VolumeRemoveFileRequest) returns (google.protobuf.Empty);
 
+  // Web
+  rpc WebDashboard(WebDashboardRequest) returns (WebDashboardResponse);
```
luiscape, line 2391: All other getter RPCs use `GetOrCreate` (`VolumeGetOrCreate`, `SharedVolumeGetOrCreate`). We are not creating a dashboard in this case but we are getting it. What about using `WebDashboardGet` instead?

mwaskom: Another existing pattern is "Lookup", so maybe `WebDashboardLookup`? That might fit better with returning a pointer to it (although I guess the `Get` RPCs really just return IDs anyway).

kramstrom: Yeah I like it! Updated this to `WebDashboardLookup` now

PR #2142, opened 2024-08-22
Remote ugly bash from cli command help

kramstrom:
help looks like this now
Screenshot 2024-08-22 at 13 02 04

instead of this
Screenshot 2024-08-22 at 13 02 08


mwaskom (approved):
Thanks for doing this!

PR #2152, opened 2024-08-26
Remove unused singular secret argument to @function

freider:
Followup to https://github.com/modal-labs/modal-client/pull/2129 which missed one occurrence


freider:
@prbot automerge

PR #2154, opened 2024-08-26
Add created_at field to EnvironmentListItem proto

rculbertson:
## Describe your changes

Part of MOD-3533.

We are adding a `created_by` field to the Environments table. This PR adds that field to the proto object so it will be included in `EnvironmentListResponse`. 

---

---

</details>


howard36 (approved):
Looks good!

PR #2155, opened 2024-08-27
[MOD-3553]: Function.spawn() returns a FunctionCall for generators & add iter_gen to get results

kramstrom:
Related to https://github.com/modal-labs/modal-client/pull/2145

- `Function.spawn()` now always returns a `_FunctionCall` (no need to check for optional anymore)
- raises exception when trying to use `_FunctionCall.get()` for generators
- adds `_FunctionCall.iter_gen()` which should be used instead


modal/functions.py
```diff
@@ -1388,8 +1392,22 @@ async def get(self, timeout: Optional[float] = None) -> R:
 
         The returned coroutine is not cancellation-safe.
         """
+
+        if self._is_generator:
+            raise Exception("Cannot get the result of a generator function call. Use `iter_gen` instead.")
+
         return await self._invocation().poll_function(timeout=timeout)
 
+    async def iter_gen(self) -> AsyncGenerator[Any, None]:
```
freider, line 1401: Maybe we should call it `get_gen()` instead? As a corollary to `Function.remote()` and `Function.remote_gen()`

freider: Wdyt @aksh-at ? I have a vague memory we discussed adding generator support to FunctionCall like ~2 years ago 😂

zsiegel92:
@kramstrom @freider thanks for moving this along!

Would've been extremely fun to have [the PR I opened](https://github.com/modal-labs/modal-client/pull/2145) merged into Modal 😂 but glad to see this in-progress and done the right way

I'll close that PR and watch this one 🫡 thanks so much

modal/functions.py
kramstrom: Yep makes sense!

kramstrom:
> @kramstrom @freider thanks for moving this along!
> 
> Would've been extremely fun to have [the PR I opened](https://github.com/modal-labs/modal-client/pull/2145) merged into Modal 😂 but glad to see this in-progress and done the right way
> 
> I'll close that PR and watch this one 🫡 thanks so much

Sorry about that! I did rebase from your branch so at least you will be in the commit history though!

freider (approved):
lgtm! Not 100% happy that from_id requires is_generator, but it's probably sufficiently rare that we could leave it like that for now (instead of adding new rpcs to be able to fetch the function's generator/function type from server based on id) - in particular since the requesting party needs to know the function type anyways to be able to use it.

zsiegel92:
@kramstrom @freider glad to say that I just ran into this again, did `pip install -U modal`, and the red underline in my IDE disappeared!

```python
fn_calls: list[modal.functions.FunctionCall[MyType]] = [my_function.spawn(job) for job in jobs]
```
No type-checker complaints that `my_function.spawn(job)` might be `None`! 🙏


Now time to [make `gather` generic](https://github.com/modal-labs/modal-client/blob/1c72b1ef253c82af418ecd6f7bcc09634e36902e/modal/functions.py#L1490-L1514) so that `results = modal.functions.gather(*fn_calls)` is type-aware! Will open separate issue - TIA.

PR #2156, opened 2024-08-27
Add docs on how to use with app.run(): ... ctx manager

kramstrom:
Adds inline docs to `App.run` showing how to run app ephemerally:

```py
with app.run():
  ...
```


kramstrom:
@prbot automerge

PR #2158, opened 2024-08-27
Support dynamic max_concurrency in client concurrency logic

cathyzbn:
## Describe your changes

Refactored concurrency semaphore to a dynamic semaphore in Concurrency Manager. When max_concurrency specified, client will send RPC request to get the desired concurrency from server and change its concurrency.

---

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -211,6 +211,9 @@ class DaemonizedThreadPool:
     # Used instead of ThreadPoolExecutor, since the latter won't allow
     # the interpreter to shut down before the currently running tasks
     # have finished
+    def __init__(self, container_io_manager: "modal._container_io_manager.ContainerIOManager"):
+        self.container_io_manager = container_io_manager
+
     def __enter__(self):
         self.spawned_workers = 0
         self.inputs: queue.Queue[Any] = queue.Queue()
@@ -243,8 +246,9 @@ class DaemonizedThreadPool:
                     logger.exception(f"Exception raised by {_func} in DaemonizedThreadPool worker!")
                 self.inputs.task_done()
 
-        threading.Thread(target=worker_thread, daemon=True).start()
-        self.spawned_workers += 1
+        if self.spawned_workers < self.concurrency_manager.get_input_concurrency():
+            threading.Thread(target=worker_thread, daemon=True).start()
+            self.spawned_workers += 1
 
         self.inputs.put((func, args))
 
@@ -413,7 +417,7 @@ def call_function(
         reset_context()
 
     if target_input_concurrency > 1:
-        with DaemonizedThreadPool() as thread_pool:
+        with DaemonizedThreadPool(container_io_manager) as thread_pool:
 
             def make_async_cancel_callback(task):
                 def f():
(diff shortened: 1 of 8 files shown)


modal/_container_io_manager.py
```diff
@@ -177,6 +179,111 @@ def validate_output_data(self, data: Any) -> List[Any]:
         return data
 
 
+class ConcurrencyManager(asyncio.Semaphore):
```
gongy, line 182: Just noting we're planning on moving this complexity out of the client soon.

modal/_container_io_manager.py
```diff
@@ -177,6 +179,111 @@ def validate_output_data(self, data: Any) -> List[Any]:
...
+            request = api_pb2.FunctionGetDynamicConcurrencyRequest(
+                function_id=container_io_manager.function_id,
+                target_concurrency=self._target_concurrency,
+                max_concurrency=self._max_concurrency,
+            )
+            resp = await retry_transient_errors(
+                container_io_manager._client.stub.FunctionGetDynamicConcurrency,
+                request,
+                attempt_timeout=DYNAMIC_CONCURRENCY_INTERVAL_SECS,
+            )
```
gongy, line 234: add debug log?

modal/_container_io_manager.py
```diff
@@ -330,6 +386,45 @@ def stop_heartbeat(self):
...
+                        max_concurrency=self._max_concurrency,
+                    )
+                    resp = await retry_transient_errors(
+                        self._client.stub.FunctionGetDynamicConcurrency,
+                        request,
+                        attempt_timeout=DYNAMIC_CONCURRENCY_TIMEOUT_SECS,
+                    )
+
+                    self._concurrency_manager.set_value(resp.concurrency)
+                    logger.debug(
```
gongy, line 416: maybe you can log only if it changes?

modal/_container_entrypoint.py
```diff
@@ -739,6 +751,12 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
 
         # Initialize the function, importing user code.
         with container_io_manager.handle_user_exception():
+            if max_concurrency != 0 and max_concurrency <= target_concurrency:
```
gongy, line 754: We don't need these checks - I think it's odd for a function to throw this when starting up, if it could have been caught earlier, and we can just allow it to start up

modal/_container_io_manager.py
```diff
@@ -517,35 +613,34 @@ async def _generate_inputs(
...
-        # - or, when the output for the fetched input is sent, release the semaphore.
-        self._input_concurrency = input_concurrency
-        self._semaphore = asyncio.Semaphore(input_concurrency)
-
-        async for inputs in self._generate_inputs(batch_max_size, batch_wait_ms):
-            io_context = await IOContext.create(self._client, finalized_functions, inputs, batch_max_size > 0)
-            for input_id in io_context.input_ids:
-                self.current_inputs[input_id] = io_context
+        # Before trying to fetch an input, acquire the concurrency_manager:
+        # - if no input is fetched, release the concurrency_manager.
```
cathyzbn, line 627: input slot

modal/_container_entrypoint.py
```diff
@@ -246,7 +246,7 @@ def worker_thread():
                     logger.exception(f"Exception raised by {_func} in DaemonizedThreadPool worker!")
                 self.inputs.task_done()
 
-        if self.spawned_workers < self.max_threads:
+        if self.spawned_workers < self.container_io_manager.get_input_concurrency():
```
gongy, line 249: Since this function is dynamic there are races where you fetch an input but don't have enough threads. Let's simplify this to just use max threads that is the max concurrent inputs?

gongy: Ideally this doesn't depend on io manager

gongy (approved):

PR #2159, opened 2024-08-27
Fix empty client behavior with app-attached sandboxes

pawalt:
## Describe your changes

Previously, if a clientless app was attached to a sandbox, it would error with NoneType. This commit checks app._client to make sure it exists before using it.

---

---

</details>

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -199,7 +199,7 @@ class _Sandbox(_Object, type_prefix="sb"):
             _experimental_gpus=_experimental_gpus,
         )
         if client is None:
-            if app:
+            if app and app._client:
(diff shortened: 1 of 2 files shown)


pawalt:
@prbot approve

erikbern (approved):
It would be great to add a test if it's easy, but either way 👍

pawalt:
@erikbern this is actually a bit tricky to test because in our tests, we have to explicitly provide a client. So I'm not quite sure how to test client from env behavior. Not sure if you have any suggestions.

pawalt:
I figured it out with `set_env_client`

pawalt:
@prbot automerge

PR #2160, opened 2024-08-27
[MOD-2845] fix: prev. inadequate handling of stale in-mem clients

thundergolfer:
## Describe your changes

- MOD-2845

Got a local reproduction of some stale creds. Fix locally by clearing the Client's channel which was initialized with the old creds and left in-memory. In retrospect, this is an obvious problem 😖. 

---

</details>

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -100,12 +100,13 @@ class _Client:
         self._pre_stop: Optional[Callable[[], Awaitable[None]]] = None
         self._channel: Optional[grpclib.client.Channel] = None
         self._stub: Optional[api_grpc.ModalClientStub] = None
+        self._snapshotted = False
 
     @property
     def stub(self) -> api_grpc.ModalClientStub:
         """mdmd:hidden"""
         if self._stub is None:
-            if self.client_type == api_pb2.CLIENT_TYPE_CONTAINER:
+            if self._snapshotted:
                 logger.debug("restoring stub for memory snapshotted client instance")
                 self._open()
             else:
@@ -120,7 +121,7 @@ class _Client:
     @property
     def credentials(self) -> tuple:
         """mdmd:hidden"""
-        if self._credentials is None and self.client_type == api_pb2.CLIENT_TYPE_CONTAINER:
+        if self._credentials is None and self._snapshotted:
             logger.debug("restoring credentials for memory snapshotted client instance")
             self._credentials = (config["task_id"], config["task_secret"])
         return self._credentials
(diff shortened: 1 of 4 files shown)


test/container_app_test.py
```diff
@@ -73,6 +73,34 @@ async def test_container_snapshot_restore(container_client, tmpdir, servicer):
...
+    f = Function.lookup(app_name, "square", client=container_client)
+    await f.remote.aio()
+    print(servicer.function_map_metadata)
+    io_manager = ContainerIOManager(api_pb2.ContainerArguments(), container_client)
+    restore_path = temp_restore_path(tmpdir)
+    with mock.patch.dict(
+        os.environ, {"MODAL_RESTORE_STATE_PATH": str(restore_path), "MODAL_SERVER_URL": servicer.container_addr}
+    ):
+        io_manager.memory_snapshot()
+        # RPC should have use credentials, not old credentials
```
thundergolfer, line 98:
```suggestion
        # RPC should have used new credentials, not old credentials
```

modal/client.py
```diff
@@ -116,14 +121,14 @@ def authenticated(self) -> bool:
...
     def credentials(self) -> tuple:
         """mdmd:hidden"""
-        if self._credentials is None and self.client_type == api_pb2.CLIENT_TYPE_CONTAINER:
+        if self._credentials is None and self._snapshotted:
             logger.debug("restoring credentials for memory snapshotted client instance")
             self._credentials = (config["task_id"], config["task_secret"])
         return self._credentials
 
-    async def _open(self):
+    def _open(self):
```
thundergolfer, line 129: Seemed this didn't need to be `async`?

test/container_app_test.py
```diff
@@ -75,7 +75,24 @@ async def test_container_snapshot_restore(container_client, tmpdir, servicer):
...
+    from modal import Function
+    from modal.runner import deploy_app
+
+    app.function()(square)
+    app_name = "my-app"
+    deploy_app(app, app_name, client=container_client).app_id
+    f = Function.lookup(app_name, "square", client=container_client)
+    await f.remote.aio()
+    # RPC should have used new credentials, not old credentials
+    await f.remote.aio()
```
erikbern, line 93: i don't fully understand this test. where do you simulate the actual snapshot+restore?

erikbern (approved):
lgtm!

if we inject creds worker side then i think we can remove this code?

thundergolfer:
> if we inject creds worker side then i think we can remove this code?

we do actually do this already. We swap out the creds and emit a warning.

cweld510:
@thundergolfer does this need to be merged still?

PR #2162, opened 2024-08-28
Add flag to skip modal requirements for Image.from_registry

kramstrom:
## Describe your changes

---

---

</details>


mwaskom:
What about `add_modal_dependencies`? Fits better with `add_python` and avoids an awkward double-negative.

Should this go in `from_dockerfile` too?

PR #2164, opened 2024-08-29
[MOD-3535] snapshots capture references to modal objects which become

thundergolfer:
## Describe your changes

- MOD-3535
- I've also added an integration test in the other repo (#15223). 

In cases where a Modal object is looked up in global scope or in an `@enter` lifecycle function, memory snapshotting will freeze that object's ID and reuse it in perpetuity, not invalidating the reference if it creates a runtime error after restore.

```python
f = modal.Function.lookup(...)
|
╰───> FunctionGet ──> self.object_id = fu-123  # can go stale!
```

This PR changes it so that upon restore these lookups must be redone. This is a small performance loss but maintains correctness. 

---

---

</details>

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -127,7 +127,7 @@ class _Client:
         self._channel = create_channel(self.server_url, metadata=metadata)
         self._stub = api_grpc.ModalClientStub(self._channel)  # type: ignore
 
-    async def _close(self, prep_for_restore: bool = False):
+    async def _close(self, forget_credentials: bool = False):
         if self._pre_stop is not None:
             logger.debug("Client: running pre-stop coroutine before shutting down")
             await self._pre_stop()  # type: ignore
@@ -135,8 +135,7 @@ class _Client:
         if self._channel is not None:
             self._channel.close()
 
-        if prep_for_restore:
-            assert not self._snapshotted
+        if forget_credentials:
             self._credentials = None
             self._snapshotted = True
 
--- a/modal/client.py
+++ b/modal/client.py
@@ -100,6 +100,7 @@ class _Client:
         self._pre_stop: Optional[Callable[[], Awaitable[None]]] = None
         self._channel: Optional[grpclib.client.Channel] = None
         self._stub: Optional[api_grpc.ModalClientStub] = None
+        self._snapshotted = False
 
     @property
     def stub(self) -> api_grpc.ModalClientStub:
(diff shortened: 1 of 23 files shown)


modal/object.py
```diff
@@ -44,6 +44,8 @@ class _Object:
     _object_id: str
     _client: _Client
     _is_hydrated: bool
+    _is_rehydrating: bool
+    _is_rehydrated: bool
```
mwaskom, line 48: Is there a reason we need separate `_is_rehydrating` and `_is_rehydrated` attributes rather than just toggling `_is_hydrated` to `False` on restore?

modal/client.py
```diff
@@ -136,6 +137,7 @@ async def _close(self, forget_credentials: bool = False):
 
         if forget_credentials:
             self._credentials = None
+            self._snapshotted = True
```
mwaskom, line 140: Setting `_snapshotted` as a side-effect of the `forget_credentials` parameter seems kinda weird; maybe we could rename that to be more general?

mwaskom (commented):
Random thing this just made me wonder about — what happens if you snapshot/restore across client version boundaries?

modal/client.py
thundergolfer: I did at one point in this change rename `forget_credentials` to `prepare_for_restore` or something. Is that better?

mwaskom: Yeah something like that might make more sense.

thundergolfer:
> what happens if you snapshot/restore across client version boundaries?

@mwaskom the client version will be fixed across snapshot/restore because it's the same container with the same client mount.

modal/object.py
```diff
@@ -44,6 +44,7 @@ class _Object:
     _object_id: str
     _client: _Client
     _is_hydrated: bool
+    _is_rehydrating: bool
```
mwaskom, line 47: You set `_is_rehydrated` (past tense) below; looks like you read the latter in the code too.

modal/object.py
```diff
@@ -214,6 +216,12 @@ def deps(self) -> Callable[..., List["_Object"]]:
     async def resolve(self):
         """mdmd:hidden"""
         if self._is_hydrated:
+            # memory snapshots capture references which must be rehydrated
+            # on restore to handle staleness.
+            if self._client._snapshotted and not self._is_rehydrated:
```
mwaskom, line 221: Do we ever set `_is_rehydrated` to True?

mwaskom (approved):

PR #2165, opened 2024-08-29
Add enum to FunctionMapRequest indicating if client will block

rculbertson:
MOD-3589

Add FunctionCallInvocationType enum which indicates if client will wait for the function to complete, or continue on immediately. For example, calls to .remote() will block, but calls to .spawn() will not.

---

---

</details>

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -113,7 +113,7 @@ class _Invocation:
         kwargs,
         *,
         client: _Client,
-        invoke_type: api_pb2.FunctionMapRequest.InvokeType.ValueType,
+        function_call_invocation_type: api_pb2.FunctionCallInvocationType.ValueType,
     ) -> "_Invocation":
         assert client.stub
         function_id = function._invocation_function_id()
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -1248,11 +1248,17 @@ message FunctionInput {
 }
 
 message FunctionMapRequest {
+  enum InvokeType {
+    INVOKE_TYPE_UNSPECIFIED = 0;
+    INVOKE_TYPE_SYNC = 2;
+    INVOKE_TYPE_ASYNC = 1;
```
rculbertson, line 1254: Would something else be more clear like `INVOKE_TYPE_BLOCKING` and `INVOKE_TYPE_NON_BLOCKING`?

gongy: We can change it later. SYNC and ASYNC is how Lamdba differentiates its execution patterns so we can stick with that for now.

modal/functions.py
```diff
@@ -107,7 +107,14 @@ def __init__(self, stub: api_grpc.ModalClientStub, function_call_id: str, client
...
 
     @staticmethod
-    async def create(function: "_Function", args, kwargs, *, client: _Client) -> "_Invocation":
+    async def create(
+        function: "_Function",
+        args,
+        kwargs,
+        *,
+        client: _Client,
+        function_call_invocation_type: "api_pb2.FunctionCallInvocationType.ValueType",
```
gongy, line 116: I believe quotes are mandatory for `if TYPE_CHECKING` gated imports, but not needed here.

gongy (approved):

PR #2168, opened 2024-08-29
Fix `modal launch` no longer showing progress

ekzhang:
This was a side-effect of #2030, it turned off all the CLI output for modal launch. Actually not bad, since it's cleaner, but it also shuts off image build progress which is not ideal since it just looks like it's frozen to users.

### Changelog

- Fix a regression in `modal launch` behavior not showing progress output when starting the container.

### DIFF
--- a/modal/cli/launch.py
+++ b/modal/cli/launch.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, Optional
 
 from typer import Typer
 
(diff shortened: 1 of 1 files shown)


ekzhang:
@prbot approve

mwaskom (commented):
Thanks for catching.

PR #2170, opened 2024-08-29
Client to display the correct active environment, choosing between the server default, config default, and env var default

cweld510:
Display the correct active environment from modal CLI:
- request the server's current default in the EnvironmentListResponse
- display a single environment as active, prioritizing MODAL_ENVIRONMENT > config > server default

---

</details>

### DIFF
--- a/modal/cli/environment.py
+++ b/modal/cli/environment.py
@@ -42,12 +42,9 @@ def list(json: Optional[bool] = False):
 
     # determine which environment is currently active, prioritizing the local default
     # over the server default
-    cfg_default = config.get("environment")
-    active_env = None
+    active_env = config.get("environment")
     for env in envs:
-        if env.name == cfg_default:
-            active_env = env.name
-        elif env.default is True and active_env is None:
+        if env.default is True and active_env is None:
             active_env = env.name
 
     table_data = []
--- a/modal/cli/environment.py
+++ b/modal/cli/environment.py
@@ -1,10 +1,10 @@
 # Copyright Modal Labs 2023
-from os import environ
-from typing import Optional, Text, Union, cast
+from typing import Optional, Union
 
 import typer
 from click import UsageError
 from grpclib import GRPCError, Status
+from rich.text import Text
 from typing_extensions import Annotated
 
 from modal import environments
@@ -28,7 +28,7 @@ while still being able to deploy changes to a live environment.
 environment_cli = typer.Typer(name="environment", help=ENVIRONMENT_HELP_TEXT, no_args_is_help=True)
 
 
-class RenderableBool:
+class RenderableBool(Text):
     def __init__(self, value: bool):
         self.value = value
 
@@ -39,26 +39,23 @@ class RenderableBool:
 @environment_cli.command(name="list", help="List all environments in the current workspace")
 def list(json: Optional[bool] = False):
     envs = environments.list_environments()
+
+    # determine which environment is currently active, prioritizing the local default
+    # over the server default
+    cfg_default = config.get("environment")
+    active_env = None
+    for env in envs:
+        if env.name == cfg_default:
+            active_env = env.name
+        elif env.default is True and active_env is None:
+            active_env = env.name
+
     table_data = []
-
-    env_var_default = environ.get("MODAL_ENVIRONMENT") or None
-    cfg_default = str(config.get("environment")) or None
-    active_env_count = 0
     for item in envs:
-        is_active = item.name == cfg_default or item.default is True or item.name == env_var_default
-        if is_active:
-            active_env_count += 1
-        row = [item.name, item.webhook_suffix, cast(Union[Text, str], is_active if json else RenderableBool(is_active))]
+        is_active = item.name == active_env
+        is_active_display: Union[Text, str] = str(is_active) if json else RenderableBool(is_active)
+        row = [item.name, item.webhook_suffix, is_active_display]
         table_data.append(row)
-
-    if active_env_count > 1:
-        # set is_active column to false for all rows except for the env in config
-        active_env = env_var_default or cfg_default
-        for i in range(0, len(table_data)):
-            env = table_data[i]
-            is_active = env[0] == active_env
-            env[2] = cast(Union[Text, str], is_active if json else RenderableBool(is_active))
-
     display_table(["name", "web suffix", "active"], table_data, json=json)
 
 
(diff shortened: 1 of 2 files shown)


modal/cli/environment.py
```diff
@@ -39,10 +40,25 @@ def __rich__(self):
 def list(json: Optional[bool] = False):
     envs = environments.list_environments()
     table_data = []
+
+    env_var_default = environ.get("MODAL_ENVIRONMENT") or None
+    cfg_default = str(config.get("environment")) or None
```
cweld510, line 45: For some reason mypy is unable to figure out the types of things without this string conversion and the casts below, even though my changes were sort-of-unrelated?

howard36: You can use type hinting to avoid the cast:
```
cfg_default: Optional[str] = config.get("environment") or None
```
The type hint is used by mypy, but ignored during runtime (unlike the `str()` cast)

I think the main issue is that `config.get` has to return an `Any` type, because the returned value depends on which key you pass in. It could return a string, int, bool, etc. A better long-term solution would be to define a `Config` class with typed fields like `environment: str`, so it knows the correct type for each value.

modal/cli/environment.py
```diff
@@ -39,10 +40,25 @@ def __rich__(self):
...
+    active_env_count = 0
     for item in envs:
-        is_active = item.name == config.get("environment")
-        row = [item.name, item.webhook_suffix, is_active if json else RenderableBool(is_active)]
+        is_active = item.name == cfg_default or item.default is True or item.name == env_var_default
```
howard36, line 48: Instead of setting `is_active` here and then overwriting it later, I think we should first calculate `active_env` based on the precedence rules (env var, then config, then server default), and then create the rows with `is_active = item.name == active_env`.

howard36 (changes requested):

modal/cli/environment.py
cweld510: Sure.

cweld510:
I'll fix this up a bit and then re-request review.

modal/cli/environment.py
```diff
@@ -39,10 +40,25 @@ def __rich__(self):
 def list(json: Optional[bool] = False):
     envs = environments.list_environments()
     table_data = []
+
+    env_var_default = environ.get("MODAL_ENVIRONMENT") or None
```
mwaskom, line 44: Tiny thing but the `or None` is unnecessary as `os.environ.get()` (really any `dict.get` call) will return `None` if the key is not found.

modal/cli/environment.py
```diff
@@ -1,5 +1,6 @@
 # Copyright Modal Labs 2023
-from typing import Optional
+from os import environ
```
mwaskom, line 2: Extremely nitpicky but directly importing `os` and then doing `os.environ` would be more conventional.

modal/cli/environment.py
mwaskom: Agreed that it makes sense for `config.get` to currently have `Any` as a return type.

Note that `str(config.get("environment"))` is wrong as `config.get` can return `None` so you'd end up with `"None"` as your environment.

modal/cli/environment.py
mwaskom, line 44: But actually — do we need to separately check `os.envion`? That's already handled by our `Config.get`: https://github.com/modal-labs/modal-client/blob/main/modal/config.py#L233-L250

modal_proto/api.proto
```diff
@@ -912,6 +912,7 @@ message EnvironmentListItem {
   string name = 1;
   string webhook_suffix = 2;
   double created_at = 3;
+  bool default = 4;
```
mwaskom, line 915: Glad we went this route as it also lets us support user-configurable default envs on the server side later.

modal/cli/environment.py
```diff
@@ -39,10 +40,25 @@ def __rich__(self):
...
+        if is_active:
+            active_env_count += 1
+        row = [item.name, item.webhook_suffix, cast(Union[Text, str], is_active if json else RenderableBool(is_active))]
         table_data.append(row)
+
+    if active_env_count > 1:
+        # set is_active column to false for all rows except for the env in config
+        active_env = env_var_default or cfg_default
+        for i in range(0, len(table_data)):
+            env = table_data[i]
```
mwaskom, line 58: As Howard mentioned above I don't think you need to do this iteration but FYI in Python it's much cleaner to just have

```python
for env_row in table_data:
    ...
```

modal/cli/environment.py
```diff
@@ -39,10 +40,25 @@ def __rich__(self):
...
+    env_var_default = environ.get("MODAL_ENVIRONMENT") or None
+    cfg_default = str(config.get("environment")) or None
+    active_env_count = 0
     for item in envs:
-        is_active = item.name == config.get("environment")
-        row = [item.name, item.webhook_suffix, is_active if json else RenderableBool(is_active)]
+        is_active = item.name == cfg_default or item.default is True or item.name == env_var_default
+        if is_active:
+            active_env_count += 1
+        row = [item.name, item.webhook_suffix, cast(Union[Text, str], is_active if json else RenderableBool(is_active))]
```
mwaskom, line 51: This cast seems wrong since `is_active` is a bool?

modal/cli/environment.py
cweld510: I fixed up the typing; it should be more reasonable now (no casts)

modal/cli/environment.py
cweld510: Done

modal/cli/environment.py
cweld510: yup! I took it out

modal/cli/environment.py
```diff
@@ -38,10 +39,22 @@ def __rich__(self):
...
 def list(json: Optional[bool] = False):
     envs = environments.list_environments()
+
+    # determine which environment is currently active, prioritizing the local default
+    # over the server default
+    cfg_default = config.get("environment")
+    active_env = None
+    for env in envs:
+        if env.name == cfg_default:
+            active_env = env.name
```
howard36, line 49: minor nit: if you initialize `active_env = config.get("environment")`, you can simplify this logic by getting rid of the first `if` statement

howard36 (approved):

mwaskom:
Thanks @cweld510 welcome to the client team 😛

PR #2171, opened 2024-08-30
Add GenericResult to sandbox tunnels response

pawalt:
## Describe your changes

This allows us to know if the SandboxGetTunnels RPC returned because of a timeout or full data.

---

This is not a backwards-compatible proto change, but it's OK because there's no code (and there will be no code) that uses the old version.

---

</details>


pawalt:
@prbot approve

pawalt:
@prbot automerge

thundergolfer (approved):

PR #2179, opened 2024-09-02
[MOD-3614]: pyright _container_io_manager

kramstrom:
Include `_container_io_manager` in pyright typing checks

### DIFF
--- a/modal/_container_io_manager.py
+++ b/modal/_container_io_manager.py
@@ -878,10 +878,10 @@ class _ContainerIOManager:
 
     async def memory_snapshot(self) -> None:
         """Message server indicating that function is ready to be checkpointed."""
-        if not self.checkpoint_id:
-            raise ValueError("Checkpoint ID is not set")
-
-        logger.debug(f"Checkpoint ID: {self.checkpoint_id} (Memory Snapshot ID)")
+        if self.checkpoint_id:
+            logger.debug(f"Checkpoint ID: {self.checkpoint_id} (Memory Snapshot ID)")
+        else:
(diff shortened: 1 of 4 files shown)


modal/_container_io_manager.py
```diff
@@ -296,7 +297,7 @@ def _init(self, container_args: api_pb2.ContainerArguments, client: _Client):
         self._fetching_inputs = True
 
         self._client = client
-        assert isinstance(self._client, _Client)
+        # assert isinstance(self._client, _Client)
```
freider, line 300: Intentional?

modal/_container_io_manager.py
```diff
@@ -662,13 +667,12 @@ async def _push_outputs(
             max_retries=None,  # Retry indefinitely, trying every 1s.
         )
 
-    def serialize_exception(self, exc: BaseException) -> Optional[bytes]:
+    def serialize_exception(self, exc: BaseException) -> bytes:
         try:
             return self.serialize(exc)
         except Exception as serialization_exc:
-            logger.info(f"Failed to serialize exception {exc}: {serialization_exc}")
             # We can't always serialize exceptions.
```
freider, line 674: Might want to keep the logging here, and in the serialization error print out `type(exc)` as well since the stringification of many exceptions is an empty string, e.g.:
```
>>> str(KeyboardInterrupt())
''
```

freider (approved):
Nice! Some minor comments

modal/_container_io_manager.py
kramstrom: No 😄

PR #2180, opened 2024-09-02
Add _experimental_buffer_containers

devennavani:
Add `_experimental_buffer_containers` to `app.function(...)` and `app.cls(...)`, as well as function definition proto.


irfansharif (approved):

irfansharif:
(Could also use small unit test that verifies that when an @app.function is decorated with this field, the underlying proto is annotated as you'd expect.)

PR #2183, opened 2024-09-03
[MOD-1177]: support asgi lifespan

kramstrom:
Adds support for [asgi lifespan protocol](https://asgi.readthedocs.io/en/latest/specs/lifespan.html) which runs once per event loop:

```py
import contextlib
from fastapi import FastAPI, Request
import modal

app = modal.App()


@app.function()
@modal.asgi_app()
def fastapi_app_with_lifespan():
    @contextlib.asynccontextmanager
    async def lifespan(wapp: FastAPI):
        print("enter!")
        yield {"foo": "bar"}
        print("exit!")

    web_app = FastAPI(lifespan=lifespan)

    @web_app.get("/")
    async def foo(request: Request):
        return request.state.foo

    return web_app
```

- and adds lifespan support to `@web_server` -- closing ClientSession on `lifespan.shutdown`
- runs cls lifecycle methods on exit even on lifespan shutdown failures


ekzhang:
Thanks for adding this @kramstrom!

A nice application of lifespans is resolving this bug in `@web_server` https://github.com/modal-labs/modal-client/blob/c1abcd815aed1f4ae0417d8d2c1431c774cb42ab/modal/_asgi.py#L330-L332

modal/_asgi.py
```diff
@@ -15,12 +15,62 @@
...
+        async def receive():
+            if self.queue is None:
+                raise ValueError("queue is not initialized, call background_task first")
+            return await self.queue.get()
+
+        async def send(message):
+            if message["type"] == "lifespan.startup.complete":
+                self.startup.set_result(None)
+            elif message["type"] == "lifespan.startup.failed":
+                self.startup.set_exception(Exception("Startup failed"))
```
freider, line 39: Here we might want to be more explicit, e.g. "ASGI lifespan startup failed".

Also - do we have a test for this?

modal/_container_entrypoint.py
```diff
@@ -829,9 +841,14 @@ def breakpoint_wrapper():
 
         with container_io_manager.handle_user_exception():
             finalized_functions = service.get_finalized_functions(function_def, container_io_manager)
-
         # Execute the function.
         try:
+            for finalized_function in finalized_functions.values():
+                if finalized_function.lifespan_manager:
+                    event_loop.create_task(finalized_function.lifespan_manager.background_task())
+                    call_lifecycle_functions(
```
freider, line 849: Now that I think about it, it might be a bit cleaner to use the lower level

```py
with container_io_manager.handle_user_exception():
    event_loop.run(finalized_function.lifespan_manager.lifespan_startup())
```

Since `call_lifecycle_functions` does some other wonky things that we might want to refactor at some point

test/web_server_proxy_test.py
```diff
@@ -68,6 +68,43 @@ async def hello(request):
         yield DummyHttpServer(host=host, port=port, event=event, assertion_log=assertion_log)
 
 
+@contextlib.asynccontextmanager
+async def lifespan_ctx_manager(asgi_app):
+    state: Dict[str, Any] = {}
```
freider, line 73: Thinking about it, we should be able to refactor this to just use `LifespanManager` internally to make it a ~3 liner

test/web_server_proxy_test.py
```diff
@@ -68,6 +68,43 @@ async def hello(request):
...
+            shutdown.set_exception(Exception("Shutdown failed"))
+        else:
+            raise ValueError(f"Unexpected msg type: {msg['type']}")
+
+    t = asyncio.create_task(asgi_app({"type": "lifespan", "state": state}, recv, send))
+
+    await q.put({"type": "lifespan.startup"})
+    await startup
+
+    print("yield state")
```
freider, line 99: printy

modal/_asgi.py
```diff
@@ -15,12 +15,62 @@
...
+        self.state = state
+
+    async def background_task(self):
+        self.queue = asyncio.Queue()
+        self.startup: asyncio.Future[None] = asyncio.Future()
+        self.shutdown: asyncio.Future[None] = asyncio.Future()
+
+        async def receive():
+            if self.queue is None:
+                raise ValueError("queue is not initialized, call background_task first")
```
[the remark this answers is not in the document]
kramstrom: Fixed!

modal/_asgi.py
kramstrom: Added tests for startup & shutdown failures

modal/_asgi.py
```diff
@@ -15,12 +15,72 @@
...
+    startup: asyncio.Future
+    shutdown: asyncio.Future
+    queue: asyncio.Queue
+    has_run_init: bool = False
+
+    def __init__(self, asgi_app, state):
+        self.asgi_app = asgi_app
+        self.state = state
+
+    async def ensure_init(self):
```
ekzhang, line 28: Should this be an async function? It doesn't do any `await` operations.

modal/_container_entrypoint.py
```diff
@@ -846,6 +862,11 @@ def breakpoint_wrapper():
             int_handler = signal.signal(signal.SIGINT, signal.SIG_IGN)
             usr1_handler = signal.signal(signal.SIGUSR1, signal.SIG_IGN)
 
+            for finalized_function in finalized_functions.values():
+                if finalized_function.lifespan_manager:
+                    with container_io_manager.handle_user_exception():
+                        event_loop.run(finalized_function.lifespan_manager.lifespan_shutdown())
```
ekzhang, line 868: Nice 👍

Just FYI this will run if there is an interrupt or shutdown between lines 845 and 849, which means that the session may not have been initialized. So on line 405 of `_asgi.py`, we may want to just silently ignore the missing session.

modal/_asgi.py
```diff
@@ -316,47 +376,56 @@ async def upstream_to_client():
...
+                        if parse_major_minor_version(aiohttp.__version__) >= (3, 9)
+                        else {}
+                    ),
+                )
+                scope["state"]["session"] = session
+            await send({"type": "lifespan.startup.complete"})
+        elif message["type"] == "lifespan.shutdown":
+            if session is None:
+                raise ExecutionError("Session is not initialized")
+            await session.close()
```
ekzhang, line 407:
```suggestion
            if session is not None:
                await session.close()
```

ekzhang (approved):
Woohoo!

modal/_container_entrypoint.py
kramstrom: makes sense! thanks!

modal/_asgi.py
kramstrom: Added a comment now but this has to run inside the event loop which is why it's async even though it doesn't do any awaits

freider (approved):
lgtm!

PR #2184, opened 2024-09-03
s/priority/rank in TaskTemplate

irfansharif:
Purely internal, proto name change. Working through some backend refactors and this naming avoids a conflict with another notion of inter-function priorities in the scheduler. Also align naming with new {target,max}-concurrent-inputs spec (on a per task-template basis).


mwaskom (approved):
lgtm

irfansharif:
@prbot automerge

PR #2185, opened 2024-09-03
AppCreateResponse: add split between app page URL and logs page URL

thundergolfer:
## Describe your changes

Last night when dog-fooding I hit a broken link: 

```
Stopping app - keyboard interrupt received. Use `modal run --detach` to prevent KeyboardInterrupts from killing apps.
Timed out waiting for logs. View logs at https://modal.com/logs/ap-n6SUuGWdBCufjNisSwx2Ep for remaining output.
```

Sometimes the app overview page is best and sometimes we should take the client directly to the logs page (or tab, whichever). So I'm adding both URL options to the proto. 

- MOD-3658

---

    - yes, old servers will continue to send the `app_logs_url` on field 2, which is compatible data because this is just a rename.

---

</details>

### DIFF
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -73,7 +73,7 @@ async def _init_local_app_new(
         app_state=app_state,
     )
     app_resp = await retry_transient_errors(client.stub.AppCreate, app_req)
-    app_page_url = app_resp.app_logs_url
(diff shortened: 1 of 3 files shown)


mwaskom (approved):
:+1:

PR #2190, opened 2024-09-04
[MOD-3613] Improve client error message for asgi / wsgi apps that have non-nullary functions

danielshaar:
## Describe your changes

MOD-3613

Clients now get a descriptive error quickly if their asgi / wsgi app has a function with arguments.

Also, fix a small bug that allowed 0 as an argument to various descriptors because of falsey-ness in Python.

---

### DIFF
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -84,17 +84,6 @@ def is_async(function):
         raise RuntimeError(f"Function {function} is a strange type {type(function)}")
 
 
-def is_nullary_function(f: Callable[..., Any]) -> bool:
-    signature = inspect.signature(f)
-    for param in signature.parameters.values():
-        if param.kind in (inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD):
-            # variadic parameters are nullary
-            continue
-        if param.default is param.empty:
-            return False
-    return True
-
-
 class FunctionInfo:
     """Class that helps us extract a bunch of information about a function."""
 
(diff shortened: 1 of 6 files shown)


modal/partial_function.py
```diff
@@ -394,12 +406,24 @@ def create_wsgi() -> Callable:
...
+    elif _warn_parentheses_missing is not None:
         raise InvalidError(
             "Positional arguments are not allowed. Did you forget parentheses? Suggestion: `@wsgi_app()`."
         )
 
     def wrapper(raw_f: Callable[..., Any]) -> _PartialFunction:
+        if callable_has_non_self_params(raw_f):
+            if callable_has_non_self_non_default_params(raw_f):
+                raise InvalidError(
+                    f"WSGI app function {raw_f.__name__} can't have arguments. See https://modal.com/docs/guide/webhooks#wsgi."
```
mwaskom, line 418: Not sure if this is idiosyncratic or pointless, but I have always thought (in Python at least) that *argument* refers to the value passed into the function while *parameter* refers to the placeholder for that value in the function definition.

modal/partial_function.py
```diff
@@ -394,12 +406,24 @@ def create_wsgi() -> Callable:
...
+                )
+            else:
+                deprecation_warning(
+                    (2024, 9, 4),
+                    f"WSGI app function {raw_f.__name__} has default arguments - Modal will drop support for this in a"
+                    f"future release.",
```
mwaskom, line 424: This message is maybe a little ambiguous — it makes it sound like you could remove the default values and you'd be fine.

modal/_utils/function_utils.py
```diff
@@ -320,16 +320,29 @@ def is_nullary(self):
         return True
 
 
-def method_has_params(f: Callable[..., Any]) -> bool:
-    """Return True if a method (bound or unbound) has parameters other than self.
+def callable_has_non_self_params(f: Callable[..., Any]) -> bool:
+    """Return True if a callable (function, bound method, or unbound method) has parameters other than self.
 
-    Used for deprecation of @exit() parameters.
+    Used for deprecation of @exit() parameters and ensuring @asgi_app / @wsgi_app functions don't have non-default
```
mwaskom, line 326: Super nit: technically the deprecation on `exit()` has passed but we still want exit functions to be nullary; could take the opportunity to update the docstring here.

Also maybe confusing to mention the "non-default" part of what we use this for vis-a-vis `asgi_app` etc. since we handle the non-default part with a different function.

mwaskom (approved):
Thanks for the clean approach to expanding the check we do here! Left a couple of fussy comments about error message terminology and docstrings but otherwise LGTM, feel free to merge when you see fit.

modal/partial_function.py
danielshaar: Valid point, updated the language to be more precise.

modal/partial_function.py
danielshaar: Clarified, thanks!

modal/_utils/function_utils.py
danielshaar: Updated!

PR #2193, opened 2024-09-05
Infer sandbox app if spawned from inside container

aksh-at:
Right now `Sandbox.create()` doesn't associate the `Sandbox` with the `App`, even if it's spawned from inside a Modal function (which is overwhelmingly the case). This PR starts storing the `container_app` as a `ClassVar` to fix this.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -124,6 +124,7 @@ class _App:
     """
 
     _all_apps: ClassVar[Dict[Optional[str], List["_App"]]] = {}
+    _container_app: ClassVar[Optional[RunningApp]] = None
 
     _name: Optional[str]
     _description: Optional[str]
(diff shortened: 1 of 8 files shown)


modal/app.py
```diff
@@ -212,6 +213,10 @@ def description(self) -> Optional[str]:
         """The App's `name`, if available, or a fallback descriptive identifier."""
         return self._description
 
+    @property
+    def client(self) -> Optional[_Client]:
+        return self._client
```
mwaskom, line 218: This will expose the client as public (i.e. user-facing) API, right? Are there any downsides to that?

mwaskom (approved):
Makes sense, although I wonder if we need to expose the client object to accomplish it 🤔

modal/app.py
aksh-at: yeah good point. I did this for a not great reason—wanted to have `inferred_app` be a union of `App` and `RunningApp`, in a way that the fields we want are on both. And it didn't quite make sense to add `_client` on `RunningApp`. 

I'll just write it to explicitly if/else. Uglier but better.

PR #2196, opened 2024-09-05
Add tunnel support to sandboxes

pawalt:
This commit adds an `open_ports` argument to sandboxes, allowing users to specify a list of ports to expose. It also adds .tunnels() to sandboxes, allowing users to get the list active tunnels once they've spun up.

## Describe your changes

-  This commit resolves MOD-3545

---

- ~[ ] Client+Server: this change is compatible with old servers~

---

</details>

## Changelog

Sandboxes now support port tunneling. Ports can be exposed via the `open_ports` argument, and a list of active tunnels can be retrieved via the `.tunnels()` method.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -48,6 +48,7 @@ class _Sandbox(_Object, type_prefix="sb"):
     _stderr: _StreamReader
     _stdin: _StreamWriter
     _task_id: Optional[str] = None
+    _tunnels: Optional[List[api_pb2.TunnelData]] = None
 
     @staticmethod
     def _new(
@@ -66,6 +67,8 @@ class _Sandbox(_Object, type_prefix="sb"):
         block_network: bool = False,
         volumes: Dict[Union[str, os.PathLike], Union[_Volume, _CloudBucketMount]] = {},
         pty_info: Optional[api_pb2.PTYInfo] = None,
+        encrypted_ports: Sequence[int] = [],
+        unencrypted_ports: Sequence[int] = [],
         _experimental_scheduler_placement: Optional[SchedulerPlacement] = None,
         _experimental_gpus: Sequence[GPU_T] = [],
     ) -> "_Sandbox":
@@ -109,6 +112,9 @@ class _Sandbox(_Object, type_prefix="sb"):
                 for path, volume in validated_volumes
             ]
 
+            open_ports = [api_pb2.PortSpec(port=port, unencrypted=False) for port in encrypted_ports]
+            open_ports.extend([api_pb2.PortSpec(port=port, unencrypted=True) for port in unencrypted_ports])
+
             ephemeral_disk = None  # Ephemeral disk requests not supported on Sandboxes.
             definition = api_pb2.Sandbox(
                 entrypoint_args=entrypoint_args,
@@ -129,6 +135,7 @@ class _Sandbox(_Object, type_prefix="sb"):
                 pty_info=pty_info,
                 scheduler_placement=scheduler_placement.proto if scheduler_placement else None,
                 worker_id=config.get("worker_id"),
+                open_ports=api_pb2.PortSpecs(ports=open_ports),
             )
 
             # Note - `resolver.app_id` will be `None` for app-less sandboxes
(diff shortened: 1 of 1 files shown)


modal/sandbox.py
```diff
@@ -165,6 +174,9 @@ async def create(
             Union[str, os.PathLike], Union[_Volume, _CloudBucketMount]
         ] = {},  # Mount points for Modal Volumes and CloudBucketMounts
         pty_info: Optional[api_pb2.PTYInfo] = None,
+        # List of ports to open on the sandbox.
+        # Each tuple is (port, unencrypted).
```
mwaskom, line 178: Is it confusing to have _unencrypted_? I worry that the negation makes it harder to reason about (i.e. `(8000, False)`. Unless I'm missing something, parameterizing it as `encrypted` just feels simpler.

mwaskom: Also, in the interest of minimizing unnecessary verbosity, would `ports` be unclear?

modal/sandbox.py
```diff
@@ -244,6 +257,24 @@ async def wait(self, raise_on_termination: bool = True):
                     raise SandboxTerminatedError()
                 break
 
+    async def tunnels(self) -> List[api_pb2.TunnelData]:
```
mwaskom, line 260: Should this be synchronicity-wrapped so that it can be async without the caller needing to be async?

modal/sandbox.py
```diff
@@ -244,6 +257,24 @@ async def wait(self, raise_on_termination: bool = True):
...
+        if self._tunnels:
+            return self._tunnels
+
+        while True:
+            req = api_pb2.SandboxGetTunnelsRequest(sandbox_id=self.object_id, timeout=50)
+            resp = await retry_transient_errors(self._client.stub.SandboxGetTunnels, req)
+
+            # If we couldn't get the tunnels in time, try again.
+            if resp.result.status == api_pb2.GenericResult.GENERIC_STATUS_TIMEOUT:
+                continue
```
mwaskom, line 272: Shoudl we give up at some point to avoid endlessly hanging?

modal/sandbox.py
mwaskom: Also we chatted about this a little bit but just having `ports: list[int]` would be a lot simpler if there's no compelling usecase for specifying some ports as encrypted and others as unencrypted.

modal/sandbox.py
mwaskom: Also super ignorant question about the return type: is it a tunnel or data from a tunnel?

modal/sandbox.py
```diff
@@ -109,6 +111,12 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
                 for path, volume in validated_volumes
             ]
 
+            open_port_proto = None
+            if open_ports:
+                open_port_proto = api_pb2.PortSpecs(
+                    ports=[api_pb2.PortSpec(port=port, unencrypted=unencrypted) for port, unencrypted in open_ports]
+                )
```
mwaskom, line 118: Very minor but will the protofbuf default on the other end of the wire be equivalent to `api_pb2.PortSpecs(ports=[])`? If so  I think you could simplify to avoid needing the `None` initialization etc.

mwaskom (approved):
Just random comments, not strong thoughts.

modal/sandbox.py
pawalt: The functionality will be the same, so I can do that.

modal/sandbox.py
pawalt: I'm just going to not retry. If the tunnels don't come up inside 50 seconds, it's probably an error. The user can override timeout if they care.

modal/sandbox.py
pawalt: It's tunnel metadata. I've updated the comment to match.

modal/sandbox.py
pawalt: We need unencrypted ports to be an option so clients can control TLS behavior if they want. I've refactored to separate variables as discussed offline.

PR #2203, opened 2024-09-06
Add FunctionMeta message

irfansharif:
Precursor to using multiple underlying 'message Function' for a given @app.function, if it represents a group of @app.function-like things.


irfansharif:
Taking a slightly different approach after internal discussions over at https://www.notion.so/modal-com/Fragmenting-function-definitions-4cc63e66d80c4ccb82914c4da901e6fd.

PR #2209, opened 2024-09-11
Ensure we clean up asgi fetch_data_in task on cancellation MOD-3537

freider:
Superficially, this should prevent the `Task was destroyed but it is pending!` warnings that get printed frequently for asgi apps shutting down.

In addition this would also remove some unnecessary gcpc streams from zombie fetch_data_in tasks running in asgi containers with cancelled inputs (specifically for inputs that get cancelled at the same time that they are scheduled to receive their first asgi payload)


freider:
@prbot approve

ekzhang (commented):
✔️ Interesting thank you!

PR #2210, opened 2024-09-11
lower default timeout on Image.run_function

thundergolfer:
## Describe your changes

I think 1 hour is a more sane default than 24 hours? I suspect that we originally set this timeout to MAX to avoid thinking about it but it's generally really unexpected that a builder function will run longer than an hour? 

---

## Changelog

* Default timeout for `modal.Image.run_function` is now 1 hour. Previously it was 24 hours.


mwaskom:
Can we add a changelog here? Not sure if it's worth trying to make this a smoother transition — how easy would it be to query the runtimes for builder functions?

mwaskom (approved):

PR #2212, opened 2024-09-11
[MOD-3523] Adding modal grouped

TheQuantumFractal:
## Describe your changes

This adds support for grouped functions and container networking (experimental feature)

---

**Note:** There may not be client forward compatibility for the grouped feature, but this is experimental so it is fine to remove or change this before release.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -657,7 +657,7 @@ class _App:
                 # typically for @function-wrapped @web_endpoint, @asgi_app, or @batched
                 f.wrapped = True
                 group_size = f.group_size
-                container_networking = bool(group_size)
+                container_networking = f.flags & _PartialFunctionFlags.GROUPED
                 if container_networking:
                     info = FunctionInfo(f.raw_f, serialized=True, name_override=name)
                 else:
--- a/modal/app.py
+++ b/modal/app.py
@@ -37,7 +37,8 @@ from .cloud_bucket_mount import _CloudBucketMount
 from .cls import _Cls, _get_class_constructor_signature, parameter
 from .config import logger
 from .exception import InvalidError, deprecation_error, deprecation_warning
-from .functions import Function, _Function, _GroupedFunction
+from .experimental import _GroupedFunction
+from .functions import Function, _Function
 from .gpu import GPU_T
 from .image import _Image
 from .mount import _Mount
--- a/modal/app.py
+++ b/modal/app.py
@@ -645,7 +645,7 @@ class _App:
         def wrapped(
             f: Union[_PartialFunction, Callable[..., Any], None],
         ) -> _Function:
-            nonlocal keep_warm, is_generator
+            nonlocal keep_warm, is_generator, cloud
 
             # Check if the decorated object is a class
             if inspect.isclass(f):
@@ -662,8 +662,8 @@ class _App:
                 container_networking = f.flags & _PartialFunctionFlags.GROUPED
                 if container_networking:
                     info = FunctionInfo(f.raw_f, serialized=True, name_override=name)
-                # END Experimental: Container Networking
                 else:
+                    # END Experimental: Container Networking
                     info = FunctionInfo(f.raw_f, serialized=serialized, name_override=name)
                 raw_f = f.raw_f
                 webhook_config = f.webhook_config
--- a/modal/app.py
+++ b/modal/app.py
@@ -646,7 +646,7 @@ class _App:
         def wrapped(
             f: Union[_PartialFunction, Callable[..., Any], None],
         ) -> _Function:
-            nonlocal keep_warm, is_generator, cloud
+            nonlocal keep_warm, is_generator, cloud, serialized
 
             # Check if the decorated object is a class
             if inspect.isclass(f):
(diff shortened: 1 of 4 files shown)


modal/functions.py
```diff
@@ -1511,4 +1512,29 @@ async def _gather(*function_calls: _FunctionCall):
         raise exc
 
 
+class _GroupedFunction(typing.Generic[P, R], _Object, type_prefix="gfu"):
```
mwaskom, line 1515: I'm not totally sure it's legal to have a three-letter type prefix ... what if this was `FunctionGroup` and the prefix was `fg-`?

TheQuantumFractal: Yeah it's not

modal/experimental.py
```diff
@@ -13,3 +31,81 @@ def get_local_input_concurrency():
...
+        output: list[ReturnType] = []
+        for i, handle in enumerate(worker_handles):
+            output.append(handle.get())
+        return output
+
+    def local(self, *args: P.args, **kwargs: P.kwargs) -> ReturnType:
+        raise NotImplementedError("Grouped function cannot be run locally")
+
+    def spawn(self, *args: P.args, **kwargs: P.kwargs) -> ReturnType:
+        raise NotImplementedError("Grouped function cannot be spawned")
```
mwaskom, line 62: Do you need to add `NotImplementedError` for `map`, etc.? Unfortunately `Function` has a lot of methods :(

modal/experimental.py
```diff
@@ -13,3 +31,81 @@ def get_local_input_concurrency():
...
+            import socket
+
+            addr_info = socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)
+            # Extract IPv6 addresses from the results
+            ipv6_addresses = [
+                addr[4][0] for addr in addr_info if addr[1] == socket.SOCK_STREAM and "fdaa" in addr[4][0]
+            ]
+            main_ip = ipv6_addresses[0]
+            q.put_many([main_ip for _ in range(size)])
+        main_ip = q.get()
```
mwaskom, line 100: I was wondering why you were using a Queue to share a single piece of information — I think the answer to my question might be that this will block until the main container has populated the queue with a main_ip for each replica to get? If that's right, it's smart! But maybe worth explaining in a comment.

Potentially we could (at some point) develop a new primitive to make this "share a single piece of information with multiple containers who can block until it's available" pattern more obvious.

mwaskom (commented):
This is looking really clean! Left mostly minor code comments.

modal/experimental.py
TheQuantumFractal: Fixed. Some of these we could possibly implement (eg map), but it would require you to build some sort of wrapper so Grouped Function calls are not synchronous

modal/experimental.py
TheQuantumFractal: Yes that's intended behavior. We don't want containers to start until they have all synchronized with the main container's IP. Adding a comment

mwaskom (approved):
Amazing!

modal/experimental.py
```diff
@@ -13,3 +32,162 @@ def get_local_input_concurrency():
...
+        import os
+
+        rank = kwargs.pop("modal_rank", None)
+        size = kwargs.pop("modal_size", None)
+        q = kwargs.pop("modal_q", None)
+
+        if rank is None or size is None or q is None:
+            raise ValueError("This must be called using grouped decorator")
+
+        if not rank:
```
thundergolfer, line 171:
```suggestion
        if rank is None or size is None or q is None:
            raise ValueError("Missing required arguments; `_networked` must be called using `grouped` decorator")
        elif rank == 0:
```

modal/experimental.py
```diff
@@ -13,3 +32,161 @@ def get_local_input_concurrency():
...
+    def spawn(self, *args: P.args, **kwargs: P.kwargs) -> _GroupedFunctionCall:
+        worker_handles: List[FunctionCall] = []
+        with modal.Queue.ephemeral() as q:
+            for i in range(self.size):
+                handle = self.f.spawn(*args, **kwargs, modal_rank=i, modal_size=self.size, modal_q=q)
+                worker_handles.append(handle)
+        handler: _GroupedFunctionCall = _GroupedFunctionCall(worker_handles)
+        return handler
+
+    def keep_warm(self, *args: P.args, **kwargs: P.kwargs) -> ReturnType:
```
thundergolfer, line 93: Could you override `__getattr__` to avoid needing to list all these methods as unimplemented? 

```python
def __getattr__(self, name):
        def unsupported_method(*args, **kwargs):
            raise NotImplementedError(f"Grouped function does not support the '{name}' method")
        return unsupported_method
```

modal/app.py
```diff
@@ -717,6 +730,16 @@ def f(self, x):
                     raise InvalidError("`region` and `_experimental_scheduler_placement` cannot be used together")
                 scheduler_placement = SchedulerPlacement(region=region)
 
+            # START Experimental: Container Networking
+
+            if container_networking and not _experimental_scheduler_placement:
+                scheduler_placement = SchedulerPlacement(zone="us-east-1f")
+
+            if container_networking:
+                cloud = "aws"
```
luiscape, line 739: Can we add debug logging informing about both cloud and region / AZ scheduling? I figure it might save a debugging minutes.

TheQuantumFractal: https://github.com/modal-labs/modal-client/pull/2222/files ?

PR #2214, opened 2024-09-11
Implement `App.lookup`

aksh-at:
## Changelog

App handles can now be looked up by name with `modal.App.lookup(name)`. This can be useful for associating sandboxes with apps:

```python
modal.Sandbox.create("echo", "hi", app=app)
```


mwaskom:
What happens if you do the following?


```python
app = modal.App.lookup("my-app")

@app.function()
def new_function():
    pass

modal.runner.deploy_app(app)
```

aksh-at:
> What happens if you do the following?

Hah, it actually works. As discussed offline we need to disallow some operations on hydrated apps in general — e.g. right now it also works if you redeployed an app from within a container while it was running, but that maybe shouldn't be possible?

aksh-at:
@prbot approve

erikbern:
👍

PR #2215, opened 2024-09-11
Try to show a more helpful error on pydantic<>fastapi compatibility issues

mwaskom:
Closes MOD-3756

Not to sure how to exercise the behavior directly in our unit tests but tried to code it defensively such that it can't cause any _new_ issues :)

Hopefully this lets users diagnose this issue faster:
image

I'm opening to iterating on the precise articulation of the issue, conditional, etc. But seems like an improvement.


modal/__main__.py
```diff
@@ -65,6 +65,8 @@ def main():
         else:
             title = "Error"
             content = str(exc)
+            if notes := getattr(exc, "__notes__", []):
+                content = f"{content}\n\nNote: {' ' .join(notes)}"
```
aksh-at, line 69: nice, that's helpful! Minor but does it make sense to emphasize this with Rich markup? Sometimes people's eyes gloss over when there's a lot of exception stuff. Maybe not necessary since this is already in a box.

aksh-at (approved):
LGTM. Hopefully will save a bunch of people time :)

PR #2217, opened 2024-09-12
Update CHANGELOG.md

kramstrom:
forgot to update changelog in asgi lifespan PR https://github.com/modal-labs/modal-client/pull/2183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ We appreciate your patience while we speedily work towards a stable release of t
         return web_app
     ```
 
-    and enabling support for `gradio>=v4` amongst other libraries using lifespans
+    which enables support for `gradio>=v4` amongst other libraries using lifespans
 
 ### 0.64.87 (2024-09-05)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,31 @@ We appreciate your patience while we speedily work towards a stable release of t
 * Fixes an issue that could cause containers using `enable_memory_snapshot=True` on Python 3.9 and below to shut down prematurely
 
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -21,7 +21,31 @@ We appreciate your patience while we speedily work towards a stable release of t
...
+        web_app = FastAPI(lifespan=lifespan)
+
+        @web_app.get("/")
+        def get_state(request: Request):
+            return {"message": f"This is the state: {request.state.foo}"}
+
+        return web_app
+    ```
+
+    and enabling support for `gradio>=v4` amongst other libraries using lifespans
```
mwaskom, line 48: I'm not sure if the false markup here is a problem or not but note that i think we have previously accidentally broken the svelte build by including `>` tokens in the changelog. Maybe the simplest change would be `gradio v4+`?

kramstrom: Hm this works locally when building though but happy to change if you think it's safer
Screenshot 2024-09-12 at 15 38 52

mwaskom: ah that's good — maybe just an issue with `<` actually https://modal-com.slack.com/archives/C03130Y0AGL/p1722958342726519?thread_ts=1722956785.894109&cid=C03130Y0AGL

kramstrom: Seems like it's fine because it's in a code block, so both < and > works if you stick them within ``

freider, line 48: nit
```suggestion
    which enables support for `gradio>=v4` amongst other libraries using lifespans
```

kramstrom:
@prbot approve

freider:
lgtm

PR #2219, opened 2024-09-12
Add feature flag for async inputs to spawn

rculbertson:
Adds a feature flag to spawn that allows users to push more inputs than previously allowed.

When enabled, the client will specify a new function invocation type in the call to FunctionMap. This new invocation type will tell the server to use the extended input queue.

Closes MOD-3758

---

- [ X] Client+Server: this change is compatible with old servers
- [ X] Client forward compatibility: this change ensures client can accept data intended for later versions of itself

---

</details>


test/function_test.py
```diff
@@ -872,3 +872,27 @@ def f(x):
...
+def test_extended_input_queue_feature_flag(client, servicer, feature_flag_enabled):
+    app = App()
+    dummy_modal = app.function()(dummy)
+
+    if feature_flag_enabled == True:
+        kwargs = {"use_extended_input_queue": True}
+    elif feature_flag_enabled == False:
+        kwargs = {"use_extended_input_queue": False}
+    else:
+        kwargs = {}
```
rculbertson, line 887: Let me know if there's a better way to implement the feature flag. Using annotations was mentioned as another possible option.

rculbertson: Also open to suggestions for a better name for the flag.

modal_proto/api.proto
```diff
@@ -122,6 +122,8 @@ enum FunctionCallInvocationType {
   FUNCTION_CALL_INVOCATION_TYPE_UNSPECIFIED = 0;
   FUNCTION_CALL_INVOCATION_TYPE_SYNC_LEGACY = 1;
   FUNCTION_CALL_INVOCATION_TYPE_ASYNC_LEGACY = 2;
+  FUNCTION_CALL_INVOCATION_TYPE_SYNC = 3;
```
rculbertson, line 125: The SYNC type isn't used in the PR at all. Figured it would be easier to track if both flags were added to the same client release.

test/function_test.py
mwaskom: I like the name. What was the annotations suggestion?

mwaskom: I think environment variable might be a slightly more common way to toggle a feature flag? But definitely easier to unit test this way :D

rculbertson: Oh I should have said "function decorator" instead of "annotation" - my brain reverted to java :) It was something @gongy mentioned in step 2 [here](https://linear.app/modal-labs/project/durable-execution-fc18c514138d/overview#heading-165bb51c).

I assume it would be added to the modal function - but it seems like this is more a property of the spawn method, than the function being invoked.

Happy to go with environment variable if that is more standard in python.

rculbertson: Pushed another commit changing it to use env vars. @gongy pointed out a downside of passing the flag to the spawn method I hadn't considered - although unlikely, the name could collide with a user variable, and in general mixing our variables with theirs seems not good.

modal/functions.py
```diff
@@ -1219,12 +1220,14 @@ async def _call_function(self, args, kwargs) -> ReturnType:
             return  # type: ignore
 
     async def _call_function_nowait(self, args, kwargs) -> _Invocation:
+        # This feature flag allows users to put a large number of inputs
+        modal_spawn_extended = os.environ.get("MODAL_SPAWN_EXTENDED")
```
mwaskom, line 1224: Note that exposing this through our [config system](https://github.com/modal-labs/modal-client/blob/7c35e09d8150cb1ff504bc899340d84ebd4e5d10/modal/config.py#L199-L224) would have the benefit of centralizing all of the configurable / feature-flag behaviors so that they're easier to keep track of. The downside is that we don't currently have a clear distinction between config values that we expect users to set and config values that we use internally. But we do have a handful of feature flags in there currently. (Sorry should have mentioned this when originally bringing up environment variables).

rculbertson: Ah, thanks for mentioning that - I haven't seen the config setup before. Yea that sounds good, I'll make those changes.

rculbertson: ok, just pushed another commit using the config system. Much nicer using that!

test/function_test.py
```diff
@@ -972,3 +972,24 @@ def f(x):
...
         def g(x=1):
             return [x_i**2 for x_i in x]
+
+
+@pytest.mark.parametrize("feature_flag", [True, False, None])
+def test_spawn_extended_feature_flag(client, servicer, monkeypatch, feature_flag):
+    app = App()
+    dummy_modal = app.function()(dummy)
+
+    if feature_flag:
```
mwaskom, line 982:
```suggestion
    if feature_flag is not None:
```

I think you want to excise "set to True, set to False, and not set" here but I may be wrong.

mwaskom (approved):
See one comment about the test, otherwise looks great

PR #2224, opened 2024-09-12
Add protos for grouped functions

irfansharif:
Following the option 1 scheme from
https://www.notion.so/modal-com/Fragmenting-function-definitions-4cc63e66d80c4ccb82914c4da901e6fd. Precursor to using multiple underlying 'message Function' for a given @app.function, if it represents a group of @app.function-like things.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1140,6 +1140,7 @@ message FunctionCreateRequest {
   // be done in AppPublish. Provides a smoother migration onto atomic deployments with 0.64,
   // and can be deprecated once we no longer support ealier versions.
   bool defer_updates = 8;
+  FunctionData function_data = 9;  // supersedes 'function' field above
 }
 
 message FunctionCreateResponse {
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1285,6 +1338,44 @@ message FunctionMapResponse {
...
+
+  string module_name = 1;
+  string function_name = 2;
+
+  Function.FunctionType function_type = 3;
+
+  // Scheduling related fields.
+  uint32 warm_pool_size = 4;
+  uint32 concurrency_limit = 5;
+  uint32 task_idle_timeout_secs = 6;
```
mwaskom, line 1356: Note that I'm not certain these fall cleanly into the "general"/"specific: dichotomy you define above and I suspect that users will be confused about what is and isn't possible to let vary.

Also, not sure how much you've thought about it, but seems like there could be some weird interactions between the scheduler parameters and the "gpu fallback" behavior. e.g. if you have a warm pool, what gpu types do we spin up for the pool? Do we take the first available — potentially leaving you with an idle but less-preferred type? Or do we try to keep your most preferred type warm? I might have similar questions about the maximum limit — if you are at your concurrency limit while using some less-preferred types when a more-preferred type would be available, are you just stuck there, or should try to migrate your usage to align with your preferences?

mwaskom (commented):
Added a couple of thoughts that aren't overly specific to the proto PR itself, but do indicate that there may be a lot of remaining open questions / undefined behaviour here. Also possible you've thought about these already though!

> grouped functions

Also the terminology we're using for "collocated containers supporting multi-node training", fyi

modal_proto/api.proto
irfansharif: For warm pools, etc. we'll generally scale up to take the first available but then consolidate it away once we have the most preferred type. Ditto for being stuck at the concurrency level - it would work the same way if you had a a blue task (from an old fn version) hogging up your limited concurrency slots - we'd still spin up more, to replace the blue, instead of having things get wedged there.

PR #2225, opened 2024-09-12
fix: AssertionError surfaces from misuse of @app.function on method

thundergolfer:
## Describe your changes

https://modal-com.slack.com/archives/C06GPR10K9V/p1726000048400859

---


mwaskom:
> and makes me less keen on stacked-decorator APIs.

Interesting — you mean because it makes it harder to check "are we wrapping a method or a "normal" function"?

thundergolfer:
@mwaskom yeh I expected the validation code for the stacking and stack ordering of decorators to be more powerful. We're so limited in what we can query about the wrapped Function that this jank fix is best I came up with after like an hour. 

More generally, I think the N! permutations of decorator ordering get really dicey if not all permutations are equivalent or even valid. We don't have N=3 right now AFAIK but perhaps with new decorators that are being introduced, e.g. `@modal.grouped` we're getting into that territory.

modal/app.py
```diff
@@ -696,6 +696,20 @@ def wrapped(
                 # typically for @function-wrapped @web_endpoint, @asgi_app, or @batched
                 f.wrapped = True
 
+                # but we don't support @app.function wrapping a method.
+                if is_method_fn(f.raw_f.__qualname__):
+                    raise InvalidError(
+                        "The `@app.function` decorator cannot be used on class methods. "
+                        "Please use `@app.cls` with `@modal.method` or `@modal.web_endpoint` instead. Example: "
```
mwaskom, line 703: Is the suggestion here overly restrictive? There are other partial function decorators that could work.

mwaskom (approved):
Thanks for looking into this — I agree that this feels harder than it should be to get right.

As we discussed on slack, my thinking is that it's not a fundamental problem with stacked decorators as an API, but can be an issue if there are too many complicated rules about how decorators are allowed to stack. Ideally, stackable decorators would be composable and commutative. We should work towards that goal: it's better for users (they don't have to think too hard about what's allowed) and for developers (it's simpler to define accurate validation logic).

BTW one thought i had is whether it would be easier to approach this from the class perspective: could we just raise if a `modal.Cls` has any attributes that are an instance of `modal.Function`? It wouldn't automatically point to the offending line in the traceback so it might take a little extra work to get to devex parity, but it's closer to enforcing the actual rule. Anyway just a thought; don't think you need to revisit this if you're happy with it.

modal/app.py
thundergolfer: Yes it possibly is. I don't want to list all the alternative decorators though.

thundergolfer:
>  Ideally, stackable decorators would be composable and commutative. We should work towards that goal: it's better for users (they don't have to think too hard about what's allowed) and for developers (it's simpler to define accurate validation logic).


Yep I think just this will do a lot to restrict the amount of mistakes we make and the amount of confusion we expose the users to.

PR #2227, opened 2024-09-13
Add back support for capturing runner output in custom stdout buffer

freider:
Quickfix since we removed the `stdout=` option on deploy_app when we introduced modal.enable_output


mwaskom:
I think we're moving towards a logs API as the way to programmatically access App output.

PR #2230, opened 2024-09-13
Add fine-grained network access protos

pawalt:
## Describe your changes

This commit adds a new NetworkAccess protobuf definition that allows users to list CIDRs that they want to allow their container to talk to.

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1550,6 +1550,17 @@ message MultiPartUpload {
   string completion_url = 3;
 }
 
+message NetworkAccess {
+  enum NetworkAccessType {
+    UNSPECIFIED = 0;
+    OPEN = 1;
+    BLOCKED = 2;
+    ALLOWLIST = 3;
+  }
+  NetworkAccessType network_access_type = 1;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1550,6 +1550,17 @@ message MultiPartUpload {
   string completion_url = 3;
 }
 
+message NetworkAccess {
+  enum NetworkAccessType {
+    UNSPECIFIED = 0;
+    OPEN = 1;
+    BLOCKED = 2;
+    ALLOWLIST = 3;
```
luiscape, line 1558: Why not `DENYLIST` / `ALLOWLIST` or `BLOCKED` / `ALLOWED`?

luiscape (approved):

modal_proto/api.proto
pawalt: I don't want to support a denylist - I think allowlists are more robust. As for `ALLOWED`, it felt a little ambiguous to me since technicaly `ALLOWLIST` is allowed as well - `OPEN` indicates there are no restrictions.

PR #2232, opened 2024-09-13
[MOD-3775] Allow users to set tags on sandboxes and filter by those tags when listing

danielshaar:
# Changelog
- Sandboxes can now be associated with tags `modal.Sandbox.set_tags(self, tags)`.
- To list running sandboxes associated with the current environment, an app ID, and/or some set of tags, use `modal.Sandbox.list(*, app_id, tags)`.

```
sandbox = modal.Sandbox.create("sleep", "infinity")
sandbox.set_tags({"version": "0", "user": "foo"})
modal.Sandbox.list(tags={"version": "0"})
```

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -251,6 +251,12 @@ class _Sandbox(_Object, type_prefix="sb"):
 
     async def set_tags(self, tags: Dict[str, str], *, client: Optional[_Client] = None):
         """Set tags (key-value pairs) on the Sandbox. Tags can be used to filter results in `Sandbox.list`."""
+        if len(tags) > 10:
+            raise InvalidError("Sandboxes have a limit of 10 tags.")
+
+        if sum([len(tag_name) + len(tag_value) for tag_name, tag_value in tags.items()]) > 16_000:
+            raise InvalidError("Sandboxes have a limit of 16KB for tags.")
+
         environment_name = _get_environment_name()
         if client is None:
             client = await _Client.from_env()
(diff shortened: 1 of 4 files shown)


modal/sandbox.py
```diff
@@ -390,7 +405,7 @@ def returncode(self) -> Optional[int]:
 
     @staticmethod
     async def list(
-        *, app_id: Optional[str] = None, client: Optional[_Client] = None
+        *, app_id: Optional[str] = None, tags: Optional[Dict[str, str]] = None, client: Optional[_Client] = None
     ) -> AsyncGenerator["_Sandbox", None]:
         """List all sandboxes for the current environment or app ID (if specified). Returns an iterator over `Sandbox`
```
aksh-at, line 410: Docstring can be updated to mention `tags`

modal/sandbox.py
```diff
@@ -250,6 +250,21 @@ async def from_id(sandbox_id: str, client: Optional[_Client] = None) -> "_Sandbo
 
         return obj
 
+    async def set_tags(self, tags: Dict[str, str], *, client: Optional[_Client] = None):
+        """Set tags (key-value pairs) on the sandbox to help with filtering."""
```
aksh-at, line 254:
```suggestion
        """Set tags (key-value pairs) on the Sandbox. Tags can be used to filter results in `Sandbox.list`."""
```

aksh-at (approved):
LGTM besides minor docstring suggestions

modal/sandbox.py
danielshaar: done

pawalt (commented):
Can we add tags to the `.create` call as well? I think it's a bit confusing to have to set them afterwards. While that's how our implementation works, I think conceptually tags are properties of the object to the user.

danielshaar:
For now, I think we want to stick with having these be "runtime" tags that can change rather than an immutable property of the sandbox.

pawalt (approved):
Makes sense to me

PR #2235, opened 2024-09-15
Deprecation warning for app-less sandboxes

aksh-at:
App-less sandboxes had a short run.

### Changelog

- Creating sandboxes without an associated `App` is deprecated. If you are spawning a `Sandbox` outside a Modal container, you can lookup an `App` by name to attach to the `Sandbox`:

```python
app = modal.App.lookup('my-app', create_if_missing=True)
modal.Sandbox.create('echo', 'hi', app=app)
```

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -275,7 +275,7 @@ class _Client:
         ```python notest
         client = modal.Client.from_credentials("my_token_id", "my_token_secret")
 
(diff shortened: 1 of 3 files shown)


aksh-at:
@prbot approve

erikbern (commented):
👍 

I have a feeling this isn't the last chapter but this seems great for now

PR #2238, opened 2024-09-16
remove max_workers from DaemonizedThreadPool

gongy:
We're going to add functionality which allows users to increase or decrease the number of concurrent threads or tasks over time, so we cannot predict the maximum thread pool size ahead of time.

Note a similar limit doesn't apply to async tasks anyway.


gongy:
@prbot approve

freider (commented):
wait - won't this leak threads? The worker threads never shut down, so this would cause *every* input to spawn a new thread that never shuts down. In this case it would probably be better to skip the input queue altogether and simply spawn a short-lived thread for every input instead?

PR #2241, opened 2024-09-17
[MOD-3802] Allow users to set input concurrency locally

danielshaar:
## Changelog

Introduce an experimental API to allow users to set the input concurrency for a container locally.


gongy (approved):
lgtm!

PR #2242, opened 2024-09-17
Revert "remove max_workers from DaemonizedThreadPool (#2238)"

gongy:


freider:
lgtm. we can fix this in one of these ways:
* always spawn a thread per input (but let it shut down when the input finishes) (easy, but perhaps bad performance for many small inputs?)
* adjust the max worker threads of the thread pool to the *max* instead of the target number and keep the rest as is
* integrate the thread pool more closely with the concurrency/semaphore logic of the container io manager ?

PR #2244, opened 2024-09-18
Re-add in-use proto fields

irfansharif:
Looks like we're still using them in places on the server, other places that https://github.com/modal-labs/modal/pull/15645 takes care of. Re-add until that lands.


gimaker (commented):
:heavy_check_mark:

PR #2245, opened 2024-09-18
Expose the `condition` parameter from the underlying `Mount` in `Image.copy_local_file`

mwaskom:
The `Image.copy_local_file` function is just a thin wrapper around a `Mount`, which has a "condition" concept that operates analogously (although inversely) to a Docker ignore file. This PR exposes the underlying `condition` so users don't need to instantiate a Mount directly just to leverage that feature.

Note that I think that condition being an _inclusion_ criterion is arguably confusing since most people find themselves looking for an _ignore_ operation. Not sure what to do about that, though, besides trying to document it well.

## Changelog

- The `Image.copy_local_dir` method now has a `condition` parameter, which can be used to select the files that should get copied.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -523,13 +523,23 @@ class _Image(_Object, type_prefix="im"):
             context_mount=mount,
         )
 
-    def copy_local_dir(self, local_path: Union[str, Path], remote_path: Union[str, Path] = ".") -> "_Image":
+    def copy_local_dir(
+        self,
+        local_path: Union[str, Path],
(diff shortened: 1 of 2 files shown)


freider (approved):

mwaskom:
After discussion on Slack, I think it's preferably to hold off on committing to an API for filtering in Image methods. Users can use the more verbose `Image.copy_mount(Mount.from_local_dir("data/", condition=...))` in the meantime.

PR #2248, opened 2024-09-18
Allow sandboxes to be created without an entrypoint

pawalt:
## Describe your changes

This allows users to create long-lived sandboxes whose only purpose is to have .exec called inside them.

---

---

</details>

## Changelog

Sandboxes can now be created without an entrypoint command. If they are created like this, they will stay alive up until their set timeout. This is useful if you want to keep a long-lived sandbox and execute code in it later.


pawalt:
@prbot approve

aksh-at (commented):
👍

PR #2251, opened 2024-09-19
Escape ANSI codes in volume file list

vishaalram02:
Addresses security vulnerability `MOD-Q324-13` where ANSI codes such as console codes can lead to unexpected behavior (and potentially remote code execution in certain terminals) when listing file paths with `modal volume ls`.

Screenshot 2024-09-19 at 10 09 59 AM

Solution is to display them as unicode so that it doesn't interfere with the terminal output. 

Screenshot 2024-09-19 at 10 11 04 AM

### DIFF
--- a/modal/cli/volume.py
+++ b/modal/cli/volume.py
@@ -160,7 +160,7 @@ async def ls(
                 filetype = "file"
             rows.append(
                 (
-                    entry.path,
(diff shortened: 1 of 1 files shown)


aksh-at (approved):
Awesome, thanks!

PR #2252, opened 2024-09-19
Error on creating `Sandbox` with an uninitialized `App`

aksh-at:
We need to explain the whole uninitialized `App` concept better, but for now this prevents this footgun:

```python
app = modal.App("peyton-sb-test")

sandbox = modal.Sandbox.create(timeout=10, app=app)
```

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -244,6 +244,7 @@ class _Sandbox(_Object, type_prefix="sb"):
                     "App has not been initialized yet. To create an App lazily, use `App.lookup`: \n"
                     "app = modal.App.lookup('my-app', create_if_missing=True)\n"
                     "modal.Sandbox.create('echo', 'hi', app=app)\n"
+                    "In order to initialize an existing `App` object, refer to our docs: https://modal.com/docs/guide/apps"
                 )
 
             app_id = app.app_id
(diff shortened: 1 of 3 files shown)


modal/sandbox.py
```diff
@@ -239,6 +239,13 @@ async def create(
         app_client: Optional[_Client] = None
 
         if app is not None:
+            if app.app_id is None:
+                raise ValueError(
+                    "App has not been initialized yet. To create an App lazily, use `App.lookup`: \n"
```
aksh-at, line 244: Note I explicitly didn't mention `with app.run()` etc because it might confuse sandbox users more. They likely don't want to do that here.

pawalt (commented):
Is it worth it to talk about the `local_entrypoint` path too?

modal/sandbox.py
erikbern: might want to also show a code sample of how to use it with a local app, idk

aksh-at: Yeah I considered doing that, but I fear that will confuse users more. Can add it if you feel strongly about it.

aksh-at:
> Is it worth it to talk about the `local_entrypoint` path too?

Wasn't going to do it initially intentionally, but decided to solve by linking to docs on apps instead.

aksh-at:
@prbot approve

PR #2253, opened 2024-09-20
Expose interactive flag to `app.run`

aksh-at:


aksh-at:
@prbot approve

erikbern:
👍  but i forget what this flag is for. it lets containers raise breakpoints and stuff?

aksh-at:
> 👍 but i forget what this flag is for. it lets containers raise breakpoints and stuff?

Yeah, so you can do this:


```python
@app.function()
def square(x):
    breakpoint()
    return x**2

if __name__ == "__main__":
    with app.run(interactive=True):
        square.remote(10)
```

Separate thing but we might be able to have everything interactive by default in the future.

erikbern:
I think we should allow interactivity for `modal run` by default but probably not for `app.run()` since it's mean to be used "programmatically" (similar thing to how I feel about output)

PR #2254, opened 2024-09-20
Propagate terminate_containers in CancelInputEvent

freider:
Should typically never get back to the container if it's set to true (since the worker will sigint the container first in that case) but just in case we also make sure to handle that case by self-siginting


freider:
@prbot approve

gongy (commented):
🫡

PR #2255, opened 2024-09-20
[MOD-3841]: run web server in separate thread

kramstrom:
## Describe your changes

---

---

</details>

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -89,7 +89,9 @@ def construct_webhook_callable(
 
     elif webhook_config.type == api_pb2.WEBHOOK_TYPE_WEB_SERVER:
         # Function spawns an HTTP web server listening at a port.
-        user_defined_callable()
+
+        server_thread = threading.Thread(target=user_defined_callable, name="web_server", daemon=True)
+        server_thread.start()
 
         # We intentionally try to connect to the external interface instead of the loopback
         # interface here so users are forced to expose the server. This allows us to potentially


ekzhang:
If you wanted to use uvicorn.run(), couldn't you just use `asgi_app` in that case?

Right now `web_server` is designed to wait until the function finishes executing before it starts polling for the server to start up, this is more flexible than always running the code in a thread and also will propagate initialization errors sooner.

kramstrom:
> If you wanted to use uvicorn.run(), couldn't you just use `asgi_app` in that case?
> 
> Right now `web_server` is designed to wait until the function finishes executing before it starts polling for the server to start up, this is more flexible than always running the code in a thread and also will propagate initialization errors sooner.
> 
> _This is also a breaking change since the `startup_timeout` would start at function begin, not at function end._

Yeah the uvicorn was just an example of something that will block the thread

ekzhang:
> Yeah the uvicorn was just an example of something that will block the thread.

Right, I'm just saying that if the concern is ergonomics in eliminating one line of code (`Thread().start()`), we should design based on examples that exist rather than theoretical examples.

kramstrom:
> > Yeah the uvicorn was just an example of something that will block the thread.
> 
> Right, I'm just saying that if the concern is ergonomics in eliminating one line of code (`Thread().start()`), we should design based on examples that exist rather than theoretical examples.

Yup that's fair. U

> > Yeah the uvicorn was just an example of something that will block the thread.
> 
> Right, I'm just saying that if the concern is ergonomics in eliminating one line of code (`Thread().start()`), we should design based on examples that exist rather than theoretical examples.

Yup, that's totally fair, updated the example now to a real user issue that prompted this

aksh-at:
Do we want to merge this? The issue isn't eliminating one line of code for the user, it's that users don't realize they are supposed to launch something in the background. It's a bit of a footgun.

modal/_runtime/user_code_imports.py
```diff
@@ -76,7 +77,8 @@ def construct_webhook_callable(
 
     elif webhook_config.type == api_pb2.WEBHOOK_TYPE_WEB_SERVER:
         # Function spawns an HTTP web server listening at a port.
-        user_defined_callable()
+        server_thread = threading.Thread(target=user_defined_callable, name="web_server", daemon=True)
```
freider, line 80: Somewhat separate from this (but made more apparent here), but it bothers me a bit that there is no clear graceful way to shut down a web server launched this way on container shutdown. (Assigning some variable to a class and adding an exit lifecycle method would work I suppose, but that's not symmetric to the launch itself)

mwaskom:
Should the changelog call out the new timeout semantics?

ekzhang:
> Do we want to merge this? The issue isn't eliminating one line of code for the user, it's that users don't realize they are supposed to launch something in the background. It's a bit of a footgun.

Yeah I understand that, but it's pretty quick to notice in development the issue and how to support it. I'd argue that the alternative is less flexible and could lead to subtle bugs in production.

Would also like to implement a healthcheck / liveness loop and Elias said

ekzhang:
Also just wanted to mention this change is much harder to reverse than to apply, which makes it worth thinking about forward compatibility with a better API to support production workloads imo

kramstrom:
Closing in favor of coming up with a new solution after internal discussion!

PR #2256, opened 2024-09-20
Add fields to FunctionExtended to make for unambiguous ..

irfansharif:
.. ser/de. We're using this union proto message to transition from Function to FunctionData and want to keep existing serialized state (in the Function form) as is, for now.

https://github.com/modal-labs/modal/pull/15645 has more details - but this additional fields helps resolve deserialization ambiguity.


irfansharif:
@prbot approve

samaross (commented):
👍

PR #2257, opened 2024-09-22
Use correct environment in `App.lookup`

aksh-at:
Resolves MOD-3845


ekzhang (approved):
Thank you, learning from this. By the way the other part of this I don't understand is that https://modal.com/apps/modal-labs/docs-test says it has no apps in general even when the fixture creates a `modal.App()`

aksh-at:
@ekzhang I was also confused by that a couple days ago and then realized that the environment used for client doctests is https://modal.com/apps/modal-labs/client-doc-tests

PR #2258, opened 2024-09-23
Remove the test fixtures that resets sys.modules

freider:
* It doesn't *always* force imports of non-present imports
* It sometimes causes conflicts with pydantic's own import hacks

Removing the fixture only seems to break the fake torch tests now, which was fixed using importlib.reload instead


thundergolfer (commented):
LGTM

PR #2263, opened 2024-09-24
Protos for EnvironmentConfigGet

mwaskom:
Opting to package the config settings in a message of their own rather than spelling them all out as fields of the `EnvironmentConfigGetResponse`. Could revert that if reviewers don't like it though.

~I included the existing `webhook_suffix` in here as I think it might make sense to migrate `EnvironmentListItem` to include this entire `Config` message. But I'll do that part later.~ Opting to leave this as a wholesale TODO instead of something that's confusingly only partially done (server-side, we only store the `webhook_suffix` in the durable table, but I'd like to make the RPC use only the ephemeral table when possible).

Part of MOD-3879

 
---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -924,6 +924,21 @@ message DomainListResponse {
   repeated Domain domains = 1;
 }
 
+// Environment-scoped configuration, with workspace-level defaults.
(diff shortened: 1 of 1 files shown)


aksh-at (commented):
👍

PR #2264, opened 2024-09-24
Adds experimental vprox interface

luiscape:
I am naming these `VproxLink`s because, different than our `Proxy` objects, we now create a network interface using [vprox](https://github.com/modal-labs/vprox) and route container traffic through it. We are "linking" the container network interface to a vprox server interface.


luiscape:
Going for a simpler approach.

PR #2268, opened 2024-09-24
[dnm] what mypy version are we using?

irfansharif:


irfansharif:
```
mypy 1.11.2 (compiled: yes)
```

mwaskom:
Can also see this in the pip logs for the `setup-cached-python` step btw

PR #2261, opened 2024-09-24
Expose list of fallback GPU types

irfansharif:

PR #2273, opened 2024-09-25
follow-up #2261 (gpu fallbacks)

thundergolfer:
## Describe your changes

Follow-up to https://github.com/modal-labs/modal-client/pull/2261

**Demo**

text version of Irfan's example

```python
import subprocess

import modal

app = modal.App()


@app.function(gpu=["t4", "a10g"], max_inputs=1)
def fallback():
    subprocess.run("nvidia-smi --list-gpus", shell=True, check=True)


@app.local_entrypoint()
def main():
    for i in range(30):
        fallback.remote()
```

---

---

</details>

## Changelog

* `modal.Function` and `modal.Cls` now support specifying a `list` of GPU configurations, allowing the Function's container pool to scale across each GPU configuration in preference order.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -616,7 +616,7 @@ class _App:
         schedule: Optional[Schedule] = None,  # An optional Modal Schedule for the function
         secrets: Sequence[_Secret] = (),  # Optional Modal Secret objects with environment variables for the container
         gpu: Union[
-            GPU_T, Sequence[GPU_T]
+            GPU_T, List[GPU_T]
         ] = None,  # GPU request as string ("any", "T4", ...), object (`modal.GPU.A100()`, ...), or a list of either
         serialized: bool = False,  # Whether to send the function over using cloudpickle.
         mounts: Sequence[_Mount] = (),  # Modal Mounts added to the container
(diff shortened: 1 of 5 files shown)


modal/app.py
```diff
@@ -616,7 +616,7 @@ def function(
         schedule: Optional[Schedule] = None,  # An optional Modal Schedule for the function
         secrets: Sequence[_Secret] = (),  # Optional Modal Secret objects with environment variables for the container
         gpu: Union[
-            GPU_T, Sequence[GPU_T]
+            GPU_T, List[GPU_T]
```
thundergolfer, line 619: I think this is better as we don't mean for the `Sequence` to be a tuple or a string or anything else that implements the sequence interface. We just want a plain `list`.

thundergolfer:
@irfansharif I think we should include the change log message as soon as we merge the functionality. It's out on PyPi after all 🙂

modal/sandbox.py
```diff
@@ -85,6 +85,12 @@ def _new(
                 raise InvalidError("`region` and `_experimental_scheduler_placement` cannot be used together")
             scheduler_placement = SchedulerPlacement(region=region)
 
+        if isinstance(gpu, list):
```
irfansharif, line 88: The typing for Sandbox._new doesn't take in the list form, but makes sense to surface a helpful error here anyway given the misuse possibility.

modal/app.py
irfansharif: Maybe `region: Optional[Union[str, Sequence[str]]] = None,` should be made as a `List` too - was doing it to stay internally consistent.

irfansharif (approved):

irfansharif:
>  I think we should include the change log message as soon as we merge the functionality. It's out on PyPi after all 🙂

I wanted to wait until we had synmons running internally (https://github.com/modal-labs/modal/pull/15907) - thanks for following up here!

mwaskom:
You can also manually update the `CHANGELOG.md` to attribute the feature to the correct version.

thundergolfer:
> should be made as a List too - was doing it to stay internally consistent.

Yep I do think so, but wanted to keep that out of this diff. I also spotted an unrelated issue in `run_function` while reading the code.

PR #2274, opened 2024-09-25
fix: run_function not invalidating when attrs change

thundergolfer:
## Describe your changes

Something I noticed while reading code in #2273. 

---

</details>

## Changelog

- `modal.Image.run_function` now invalidates when `volume`, `cloud`, or `region` is changed. If an `Image` specifies any of these arguments they will rebuild on client upgrade to this version.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -932,9 +932,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         obj._is_method = False
         obj._spec = function_spec  # needed for modal shell
 
-        # Used to check whether we should rebuild an image using run_function
-        # Plaintext source and arg definition for the function, so it's part of the image
-        # hash. We can't use the cloudpickle hash because it's not very stable.
+        # Used to check whether we should rebuild a modal.Image which uses `run_function`.
         gpus: List[GPU_T] = gpu if isinstance(gpu, list) else [gpu]
         obj._build_args = dict(  # See get_build_def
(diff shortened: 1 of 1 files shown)


modal/functions.py
```diff
@@ -932,16 +932,20 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
...
-        # hash. We can't use the cloudpickle hash because it's not very stable.
+        # Used to check whether we should rebuild a modal.Image which uses `run_function`.
         gpus: List[GPU_T] = gpu if isinstance(gpu, list) else [gpu]
         obj._build_args = dict(  # See get_build_def
             secrets=repr(secrets),
             gpu_config=repr([parse_gpu_config(_gpu) for _gpu in gpus]),
             mounts=repr(mounts),
             network_file_systems=repr(network_file_systems),
         )
+        # these key are excluded if empty to avoid rebuilds on client upgrade
```
mwaskom, line 943: Will this trigger rebuilds if _not_ empty?

thundergolfer: Yes, but that's unavoidable isn't it?

mwaskom: I guess. Worth highlighting in a changelog note at least.

mwaskom (approved):
I wonder if there's a way to assert write a test asserting that we've done this if the expand the signature further. Given that you're not updating the `build_args` when they're omitted, it seems tricky...

PR #2276, opened 2024-09-26
Adds `_experimental_proxy_ip` function interface

luiscape:
Experimental interface for using vprox proxies. 

```python
@modal.function(proxy_ip="1.2.3.4", ...)
def f(): ...
```


ekzhang (approved):
Thank you!

PR #2278, opened 2024-09-26
Gracefully handle process forks MOD-3238

freider:
Whenever an rpc call is made via our "safe" wrappers, we now check if the pid of the process has changed vs. what was used when the client was established.
If it has changed, we reconnect the client "internally", so the same client object instance will be usable in the new process.
This is bit more complicated than just setting up a new client, but allows for seamless reconnects using the same credentials as the original client and any references to the client object (or even rpc method stubs!) won't have to be switched out/refreshed.

This is achieved via a slightly hacky indirection layer of using the method name to look up the latest "connected" grpclib method whenever an rpc call takes place. We still maintain static type checking when using the API stub though

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -417,11 +417,11 @@ class UnaryUnaryWrapper(Generic[RequestType, ResponseType]):
     wrapped_method: grpclib.client.UnaryUnaryMethod[RequestType, ResponseType]
     client: _Client
 
-    def __init__(self, _wrapped_method: grpclib.client.UnaryUnaryMethod[RequestType, ResponseType], client: _Client):
-        # we pass in the _wrapped_method here to get the correct static types
+    def __init__(self, wrapped_method: grpclib.client.UnaryUnaryMethod[RequestType, ResponseType], client: _Client):
+        # we pass in the wrapped_method here to get the correct static types
         # but don't use the reference directly, see `def wrapped_method` below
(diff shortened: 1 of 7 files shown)


modal/client.py
```diff
@@ -352,26 +351,46 @@ async def _call_safely(self, coro, readable_method: str):
...
+        await self.reset_on_pid_change()
+        return getattr(self._grpclib_stub, method_name)
+
     @synchronizer.nowrap
     async def _call_unary(
         self,
-        grpclib_method: grpclib.client.UnaryUnaryMethod[RequestType, ResponseType],
-        request: RequestType,
+        method_name: str,
+        request: Any,
```
mwaskom, line 376: Why do we need to relax the typing so much here?

mwaskom (approved):
:+1: thanks for figuring this out. I think I am starting to lose the thread (heh) a little bit on all the wrapping and indirection, but the repros / tests look solid.

modal/client.py
freider: Since we needed a method lookup layer of indirection at some point in the call chain in order to look up the "latest" grpclib method (since they are bound to channels) we need to break type safety at some point, but these private methods (_call_unary and _call_streaming) should only really be used at one place each within our code base, and those places (ie. `UnaryUnaryWrapper.__call__` etc) will expose proper types to the rest of the code base when used. Does this make sense?

PR #2279, opened 2024-09-26
Error when decorating async functions with `*sgi_app`

aksh-at:
Right now we just crash-loop silently which is not great. Considered just handling this in `_container_entrypoint.py` but seems wrong to add complexity there, when async behavior is very rarely needed for the wrapper function itself.


mwaskom (approved):
:+1: but see comment

PR #2280, opened 2024-09-26
Adds protobuf definitions for creating and deleting proxy IPs

luiscape:
These are named `ProxyIp*` to make obvious the difference between legacy `Proxy*` definitions.


luiscape:
@prbot approve

ekzhang:
✔️

PR #2281, opened 2024-09-26
Ship i6pn networking for containers

ekzhang:
This enables i6pn networking for containers. You can define a function with:

```python
import socket
import modal

app = modal.App()

@app.function(i6pn=True)
def func():
    worker_ip6 = socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)[0][4][0]
    print(worker_ip6)
```

And it will have an IPv6 address that can be used to communicate with other containers, _within the same cloud and region_.

Grouped functions need quite a bit more work before they can be made stable, including scheduler work, I'm just shipping i6pn for now so we can write some demos and docs.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -655,6 +655,7 @@ class _App:
         # Maximum number of inputs a container should handle before shutting down.
         # With `max_inputs = 1`, containers will be single-use.
         max_inputs: Optional[int] = None,
+        i6pn: Optional[bool] = None,  # Whether to enable IPv6 container networking within the region.
         # The next group of parameters are deprecated; do not use in any new code
         interactive: bool = False,  # Deprecated: use the `modal.interact()` hook instead
         # Parameters below here are experimental. Use with caution!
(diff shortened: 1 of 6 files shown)


ekzhang:
@mwaskom I'm pretty agnostic about the name of the parameter being `i6pn` or `i6pn_enabled` or anything else, let me know if you would like it to be changed :)

mwaskom:
> I'm pretty agnostic about the name of the parameter being i6pn or i6pn_enabled or enable_i6pn or anything else

I don't think I have a strong preference there (I think "enable" is probably subsumed by the boolean type) but my bigger concern is whether "i6pn" will mean anything to users. Doesn't seem like a commonly used term based on Google?

ekzhang:
> Doesn't seem like a commonly used term based on Google?

Yeah, it will be explained in our docs, it's the name of our feature. Modal's implementation of IPv6 Private Networking via some ebpf overlay stuff.

It's also exposed in the `i6pn.modal.local` DNS. Some other options:

- Expand the parameter name to `ipv6_private_networking=True`
  - I don't love this because it gives the feature a new "official name" whereas we could just call it i6pn always
- Always enable it
  - This is an option since it doesn't slow down containers that don't use it or anything, though I'd want to maybe test it a bit more before we enable it globally

ekzhang:
@prbot approve

Merging to unblock examples. I can change the function parameter name later though.

arpagon:
Wow, thanks for this! It means a lot to see IPv6 support starting in Modal.

I don’t want to sound demanding, but I’m looking forward to seeing full IPv6 support soon. Any news or plans on that front?

ekzhang:
Could you clarify what your intended use case is? IPv6 is just a protocol and can be used in a lot of ways.

Feel free to send us a message in #community-feature-requests on https://modal.com/slack

PR #2282, opened 2024-09-27
Fix bug in setting i6pn

ekzhang:
Sorry 😅 — didn't fully understand what the PartialFunction conditional was doing


ekzhang:
@prbot approve

mwaskom (commented):
Would be great to have a test if this is an easy thing to get wrong!

ekzhang:
Agree, let me check out the tests!

PR #2283, opened 2024-09-27
Modernize environment object creation interface; use to look up settings

mwaskom:
This PR does a few things:

- Adds a `modal.Environment` object with `Environment.from_name` and `Environment.lookup` so we treat the "environment" concept more like other objects
- Uses the `settings` metadata attached to the environment object as the source of truth for the `image_builder_version`; removes that as an attribute of the Client
- Adds an "environment cache" so we only need to look up each environment settings once
- Looks up the environment settings concurrently with creating an app for "normal" app creation workflows (but notably *not* when creating a sandbox) to minimize the penalty on image loading

Loose threads:
- I really don't love the `ENVIRONMENT_CACHE` implementation. I think all of the code around creating an App, initializing a "resolver", etc. would benefit from a thorough refactor so that it happens in a consistent way and we can rethink what state we need to pass around. Making sandboxes less of an odd duck here would be good.
- The `modal environment create` command still goes through the`EnvironmentCreate` RPC, which has "error_if_exists" semantics. We potentially want to add a flag for that to `from_name` and then consolidate, but there's some slightly awkward API questions there, which I am deferring.

Closes MOD-3880

---

---

</details>

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -124,7 +124,6 @@ class _Client:
         self._credentials = credentials
         self.version = version
         self._authenticated = False
-        self.image_builder_version: Optional[str] = None
         self._closed = False
         self._channel: Optional[grpclib.client.Channel] = None
         self._stub: Optional[modal_api_grpc.ModalClientModal] = None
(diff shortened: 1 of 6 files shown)


erikbern:
🆒

erikbern:
Looks great!

mwaskom:
@prbot approve

erikbern:
👍

PR #2289, opened 2024-09-30
Change sandbox tunnels to return a dict instead of list

pawalt:
## Describe your changes

This change makes it easier for users to get back specific port data rather than needing to loop through the resulting tunnels object.

---

---

</details>

## Changelog

**Breaking Change:** `Sandbox.tunnels()` now returns a `Dict` rather than a `List`. This dict is keyed by the container's port, and it returns a `Tunnel` object just like `modal.forward` does.


modal/sandbox.py
```diff
@@ -332,12 +333,12 @@ async def wait(self, raise_on_termination: bool = True):
...
                 break
 
-    async def tunnels(self, timeout: int = 50) -> List[api_pb2.TunnelData]:
+    async def tunnels(self, timeout: int = 50) -> Dict[int, Tunnel]:
         """Get tunnel metadata for the sandbox.
 
         Raises `SandboxTimeoutError` if the tunnels are not available after the timeout.
 
-        Returns a list of `TunnelData` objects, which contain the tunnel metadata.
+        Returns a dictionary of `Tunnel` objects which are keyed by the container port.
```
mwaskom, line 341: Can we also add a note about the API change (with specific version) here?

mwaskom (approved):
Thanks, agree that this is a better API (in fact we shouldn't ever be exposing the proto messages in public API, list / dict conversation aside).

modal/sandbox.py
pawalt: yeah good idea

PR #2290, opened 2024-09-30
Adds unhealthy status to proxy IP proto

luiscape:
Status indicates that underlying node for Proxy IP is unhealthy.


luiscape:
@prbot approve

ekzhang:
✔️

PR #2291, opened 2024-09-30
Update protos for EnvironmentGetOrCreate

mwaskom:
## Describe your changes

Alternate approach to #2263; instead of a dedicated RPC just for looking  up the environment settings, we add a "get or create" RPC for environments and include the settings in its metadata payload.

Deleting the previously added protos, which never had a backend implementation or client interface.

Part of MOD-3879
 
---

---

</details>


mwaskom:
@prbot approve

PR #2292, opened 2024-09-30
Add option for no python on jupyter launch

advay-modal:
## Describe your changes

A user had an issue where they were using launch with a Dockerfile, and installing python in the Dockerfile in `/usr/bin`. We were installing python in `/usr/local/bin` which was taking precedence over their installation. This allows users to not require a python installation

### DIFF
--- a/modal/cli/launch.py
+++ b/modal/cli/launch.py
@@ -24,6 +24,7 @@ launch_cli = Typer(
     """,
 )
 
+
 def _launch_program(name: str, filename: str, args: Dict[str, Any]) -> None:
     os.environ["MODAL_LAUNCH_ARGS"] = json.dumps(args)
 
--- a/modal/cli/launch.py
+++ b/modal/cli/launch.py
@@ -24,7 +24,6 @@ launch_cli = Typer(
     """,
 )
 
-
 def _launch_program(name: str, filename: str, args: Dict[str, Any]) -> None:
     os.environ["MODAL_LAUNCH_ARGS"] = json.dumps(args)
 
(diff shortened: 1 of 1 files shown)


mwaskom:
Hm, I wonder why we're using `ubuntu` as the base image here, and then requiring an `add_python`, instead of using `debian_slim` (or even a Python image based on `debian_chonk`) so that `--add-python` could be null by default.

mwaskom:
Also kinda weird that `modal launch vscode` doesn't haven an equivalent CLI

modal/cli/launch.py
```diff
@@ -55,6 +55,7 @@ def jupyter(
     timeout: int = 3600,
     image: str = "ubuntu:22.04",
     add_python: Optional[str] = "3.11",
+    no_python: Optional[bool] = False,
```
thundergolfer, line 58: It's typicall `--no-X` and `--X`, so `--no-add-python`.

ekzhang:
I think this isn't needed, they can just run `--add-python ""`

modal/cli/launch.py
mwaskom: That's clever although I think the `--add` `--no-add` duality is specifically for _boolean_ typed flags right? Whereas `--add-python` has a mandatory argument?

ekzhang:
> Hm, I wonder why we're using `ubuntu` as the base image here, and then requiring an `add_python`, instead of using `debian_slim` (or even a Python image based on `debian_chonk`) so that `--add-python` could be null by default.

I think it might have to do with the fact that I wanted it to be easy for people to use arbitrary images. If add_python is specified by default, it helps with that.

advay-modal:
Ok great since there's a workaround I think we can close this, can revisit if this is an issue for more people

modal/cli/launch.py
thundergolfer: True, though I think it's still better than `--no-python` because there's still a duality here. `--add-python=""` looks weird.

PR #2293, opened 2024-09-30
Hydrate Function.from_name lazily

erikbern:
I don't see why we _don't_ do this – I _think_ it should always work.

After this change:
```python
>>> f = modal.Function.from_name("example-get-started", "square")
>>> f.remote(123)
15129
>>> 
```


mwaskom (approved):
Good to see tests passing, not sure how to derisk otherwise.

~We should probably fix these outdated docs while we're at it: https://modal.com/docs/guide/trigger-deployed-functions#importing-a-modal-function-between-modal-apps~ [Nevermind!](https://github.com/modal-labs/modal/pull/16039)

freider (commented):
cool!

PR #2294, opened 2024-10-01
PoC for removing Rich from the container environment

mwaskom:
We'd like to remove `rich` from the list of mandatory container requirements, as we in principle should only require it client-side.

This is just a rough first pass to see how much work it involves, mostly by deferring imports of the `OutputManager`, which isn't the best approach, but helps identify where we have _necessary_ functionality. It also includes some code necessary to get basic operations functional on a new image builder version (and I took the time to track down up-to-date dockerhub etc. versions) — I would pull that stuff out into a standalone PR first.

MOD-3730


mwaskom:
Superseded by smaller PRs.

PR #2296, opened 2024-10-01
Add untrusted flag to functions

pawalt:
## Describe your changes

This allows the client to signal that the function is running user-provided/untrusted code. This code can do nothing other than get inputs and put outputs.

---

---

</details>

## Changelog

Added untrusted functions. These functions can only get inputs and put outputs for those inputs, so they're great for running user-provided or LLM-generated code.


modal/app.py
```diff
@@ -658,6 +658,7 @@ def function(
         i6pn: Optional[bool] = None,  # Whether to enable IPv6 container networking within the region.
         # The next group of parameters are deprecated; do not use in any new code
         interactive: bool = False,  # Deprecated: use the `modal.interact()` hook instead
+        untrusted: bool = False,  # Whether the function is untrusted (e.g. user-provided code)
```
mwaskom, line 661: Why is this going into the "deprecated" section?

mwaskom (commented):
Is there a reason we're not adding this to `App.cls` as well?

mwaskom:
> These functions can only get inputs and put outputs for those inputs

Not sure I understand what this means!

modal/app.py
pawalt: Nope - I just didn't look 2 lines above. Fixed.

pawalt:
> Is there a reason we're not adding this to App.cls as well?

Nope this was an oversight.

> Not sure I understand what this means!

Untrusted functions have no API access other than getting inputs and putting outputs. This means they cannot spawn containers, get data from dicts/queues, lookup apps, etc. The purpose is to limit their blast radius.

mwaskom:
> Untrusted functions have no API access other than getting inputs and putting outputs. This means they cannot spawn containers, get data from dicts/queues, lookup apps, etc. The purpose is to limit their blast radius.

Got it — I guess my point is that the changelog message is written from the perspective of someone who knows our internal RPC terminology. And in a sort of pedantic way is probably not fully accurate to say they can *only* do those things — like, they can operate on the inputs too right? Anyway just being annoying...

modal/app.py
```diff
@@ -656,6 +656,7 @@ def function(
         # With `max_inputs = 1`, containers will be single-use.
         max_inputs: Optional[int] = None,
         i6pn: Optional[bool] = None,  # Whether to enable IPv6 container networking within the region.
+        untrusted: bool = False,  # Whether the function is untrusted (e.g. user-provided code)
```
mwaskom, line 659: Actually (sorry for multiple rounds of review) looking at the parameter list it might make sense to put this next to `block_network` since they're sort of similar functionality.

Which also makes me wonder if there's a more general way we could expose restrictions so we don't end up with lots of individual piecemeal parameters in this overly-long function signature 🤔.

pawalt:
Closing this as we're going to go with a decorator-based approach once we do this

PR #2297, opened 2024-10-01
Add a runtime_perf_record flag for internal debugging purposes

dflemstr:
## Describe your changes

- MOD-3816: Add a client-side flag for enabling server-side runtime perf recording.

---

---

</details>

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -821,6 +821,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
                     warm_pool_size=keep_warm or 0,
                     runtime=config.get("function_runtime"),
                     runtime_debug=config.get("function_runtime_debug"),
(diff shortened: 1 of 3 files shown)


gimaker (approved):

modal/functions.py
```diff
@@ -821,6 +821,7 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
                     warm_pool_size=keep_warm or 0,
                     runtime=config.get("function_runtime"),
                     runtime_debug=config.get("function_runtime_debug"),
+                    runtime_perf_record=config.get("runtime_perf_record"),
```
mwaskom, line 824: Do we need to also add this to the Config system? https://github.com/modal-labs/modal-client/blob/0cf1e577e21fe9710462fb9c75024ac5b8276e87/modal/config.py#L199-L224

dflemstr: Yep, good catch!

PR #2298, opened 2024-10-01
Proper KeyboardInterrupt handling as part of run_app (MOD-3937, MOD-3525)

freider:
This uses https://github.com/modal-labs/synchronicity/pull/166 to have sigints trigger cancellations + related logic immediately, before the main thread goes on to the next call call or reraises KeyboardInterrupt.

This prevents tasks from getting cleaned up while they are still running during synchronizer shutdown as part of the interpreter shutdown. Since it changes a lot of error handling assumptions (strictly to the better imo) I also ended up refactoring a lot of the error handling in _run_app to conform to the new expectations.

---

---

</details>

## Changelog

* SIGINT/Ctrl-C/KeyboardInterrupt are now handled more gracefully in most contexts, hopefully preventing a lot of internal traceback info from leaking out into the client's stderr in expected exception paths

### DIFF
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -293,9 +293,10 @@ async def _run_app(
         # Start heartbeats loop to keep the client alive
         # we don't log heartbeat exceptions in detached mode
         # as losing the local connection will not affect the running app
-        tc.infinite_loop(
-            lambda: _heartbeat(client, running_app.app_id), sleep=HEARTBEAT_INTERVAL, log_exception=not detach
-        )
+        def heartbeat():
+            return _heartbeat(client, running_app.app_id)
+
+        tc.infinite_loop(heartbeat, sleep=HEARTBEAT_INTERVAL, log_exception=not detach)
         logs_loop: Optional[asyncio.Task] = None
 
         if output_mgr := OutputManager.get():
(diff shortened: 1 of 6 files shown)


modal/runner.py
```diff
@@ -229,7 +254,7 @@ async def _run_app(
 ) -> AsyncGenerator[_App, None]:
     """mdmd:hidden"""
     if environment_name is None:
-        environment_name = config.get("environment")
+        environment_name = typing.cast(str, config.get("environment"))
```
mwaskom, line 257: Since we own the config object here I wonder if it would be better to make it have type safe null behavior, i.e. make it more similar to protobuf null semantics than `dict.get`. Type safety around the environment configuration in general is a big headache, not sure whether casting makes it better or worse.

mwaskom (approved):
Feels hard to reason about but makes sense as far as I can tell.

modal/runner.py
freider: I think this is the first time I use cast, I like to avoid it... But it helps with type restrictions without breaking in user code if the assumption is (lightly) wrong like an assertion would 😬 I agree, it would be very nice to make our config object type safe!

mwaskom: Yeah I have sort of started preferring using null string defaults and keeping things typed as `str` rather than `str | None` and dealing with the risk of assertion errors (or runtime `TypeError`) but it's tricky.

PR #2299, opened 2024-10-01
Programmatic deploy/run abort reraise KeyboardInterrupt

freider:
Slightly breaking change:

```py
with run_app():
     raise KeyboardInterrupt()  # or an actual sigint...
```
The above previously swallowed the exception in the context manager, but this instead reraises them and catches the error (without handling it) in the CLI layer.

This should make programmatic use of run_app etc. more ergonomic since you previously had to double-Ctrl-C to abort scripts that had inline Modal runs/deploys etc.


modal/cli/run.py
```diff
@@ -157,24 +157,27 @@ def _get_click_command_for_function(app: App, function_tag):
...
+                        # unpool class and method arguments
+                        # TODO(erikbern): this code is a bit hacky
+                        cls_kwargs = {k: kwargs[k] for k in cls_signature}
+                        fun_kwargs = {k: kwargs[k] for k in fun_signature}
+
+                        instance = cls(**cls_kwargs)
+                        method: Function = getattr(instance, method_name)
+                        method.remote(**fun_kwargs)
+        except KeyboardInterrupt:
+            pass  # handled within run_app
```
mwaskom, line 180: This comment confuses me a bit — if the error is handed in the (inner) context manager, why are we catching it here?

freider: The error is "handled" by printing stuff and disconnecting/cleaning up the app etc., but we catch the error here in the CLI layer to prevent click from receiving the KeyboardInterrupt since that makes it print its own "Aborted." message

mwaskom:
> since you previously had to double-Ctrl-C to abort scripts that had inline Modal runs/deploys

I can't reproduce this currently; guessing we've fixed that elsewhere.

PR #2300, opened 2024-10-01
Begin defining 2024.10 image builder version

mwaskom:
Begins defining the next image builder version (`2024.10`):

- Appends new version to `ImageBuilderVersion` literal
- Adds a requirements file (**without** updating any versions)
- Updates the base image tags for `debian_slim` and `micromamba`
- Updates the container requirements import test to cover the new version

- Part of MOD-3702

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -71,19 +71,19 @@ jobs:
         run: pytest -v --markdown-docs -m markdown-docs modal
 
   container-dependencies:
-    name: Check minimal container dependencies for ${{ matrix.python-version }}
+    name: Check minimal container dependencies for ${{ matrix.python-version }} / ${{ matrix.image-builder-version }}
     runs-on: ubuntu-20.04
     timeout-minutes: 4
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
-        # TODO matrix over image builder version too? Maybe only min/max Python?
+        image-builder-version: ["2024.04", "2024.10"]
+        python-version: ["3.8", "3.12"]
     steps:
       - uses: actions/checkout@v3
 
       - name: Install dependencies
         run: |
-          pip install -r modal/requirements/2024.10.txt
+          pip install -r modal/requirements/${{ matrix.image-builder-version }}.txt
           pip install synchronicity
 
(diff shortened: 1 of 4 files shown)


mwaskom:
@prbot approve — has minimal impact

.github/workflows/ci-cd.yml
```diff
@@ -71,18 +71,19 @@ jobs:
...
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        image-builder-version: ["2024.04", "2024.10"]
+        python-version: ["3.8", "3.12"]
     steps:
       - uses: actions/checkout@v3
 
       - name: Install dependencies
         run: |
-          pip install -r modal/requirements/2024.04.txt
+          pip install -r modal/requirements/${{ matrix.image-builder-version }}.txt
```
freider, line 86: We might want to start pinning the synchronicity version here to the version specified in the current client mount or something - this would otherwise potentially break when we release breaking changes in synchronicity without adjusting the client

freider (commented):
lgtm

.github/workflows/ci-cd.yml
```diff
@@ -71,18 +71,19 @@ jobs:
...
   container-dependencies:
-    name: Check minimal container dependencies for ${{ matrix.python-version }}
+    name: Check minimal container dependencies for ${{ matrix.python-version }} / ${{ matrix.image-builder-version }}
     runs-on: ubuntu-20.04
     timeout-minutes: 4
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        image-builder-version: ["2024.04", "2024.10"]
+        python-version: ["3.8", "3.12"]
```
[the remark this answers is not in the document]
mwaskom: Yes — was worried about combinatorial explosion of this check. It will make debugging failures a little annoying, but I think the chances that a mandatory dependency gets added after 3.8 (soon to be 3.9) and then removed before 3.12 (soon to be 3.13) are pretty low, so risk of false negative should be rare?

.github/workflows/ci-cd.yml
mwaskom: We should probably not do that :D 

(I think we should maybe just merge synchronicity into the client fwiw).

PR #2301, opened 2024-10-01
Split traceback utilities that depend on Rich into cli subpackage

mwaskom:
Part of MOD-3730

We had a catch-all `_traceback.py` module which fairly neatly factored into (a) utilities for operating on traceback objects and (b) utilities for *displaying* traceback objects with rich markup.

We only call the functions in (b) from the CLI, and don't need them for the container client, so splitting up the module is a clean way to reduce the container dependency on Rich.

### DIFF
--- a/modal/__main__.py
+++ b/modal/__main__.py
@@ -1,7 +1,8 @@
 # Copyright Modal Labs 2022
 import sys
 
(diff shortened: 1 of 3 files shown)


mwaskom:
@prbot approve — very simple reorganization

freider (commented):
Great!

PR #2302, opened 2024-10-01
Implement pattern matcher library & tests

ekzhang:
This is ported from moby/patternmatcher but cleaned up and with a bunch of dead logic removed.

```
test/pattern_matcher_test.py ...................                                                                                            [100%]

=============================================================== 19 passed in 0.05s ================================================================
```

### DIFF
--- a/modal/_utils/pattern_matcher.py
+++ b/modal/_utils/pattern_matcher.py
@@ -1,0 +1,273 @@
+"""Pattern matching library ported from https://github.com/moby/patternmatcher.
+
+This is the same pattern-matching logic used by Docker, except it is written in
+Python rather than Go. Also, the original Go library has a couple deprecated
+functions that we don't implement in this port.
+
+The main way to use this library is by constructing a `PatternMatcher` object,
+then asking it whether file paths match any of its patterns.
+"""
+
+import enum
+import os
+import re
+from typing import TextIO
+
+escape_chars = frozenset(".+()|{}$")
+
+
+class MatchType(enum.IntEnum):
+    UNKNOWN = 0
+    EXACT = 1
+    PREFIX = 2
+    SUFFIX = 3
+    REGEXP = 4
+
+
+class Pattern:
+    """Defines a single regex pattern used to filter file paths."""
+
+    def __init__(self) -> None:
+        """Initialize a new Pattern instance."""
+        self.match_type = MatchType.UNKNOWN
+        self.cleaned_pattern = ""
+        self.dirs: list[str] = []
+        self.regexp: re.Pattern | None = None
+        self.exclusion = False
+
+    def __str__(self) -> str:
+        """Return the cleaned pattern as the string representation."""
+        return self.cleaned_pattern
+
+    def compile(self, separator: str) -> None:
+        """Compile the pattern into a regular expression.
+
+        Args:
+            separator (str): The path separator (e.g., '/' or '\\').
+
+        Raises:
+            ValueError: If the pattern is invalid.
+        """
+        reg_str = "^"
+        pattern = self.cleaned_pattern
+
+        esc_separator = separator
+        if separator == "\\":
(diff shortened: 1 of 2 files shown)


test/pattern_matcher_test.py
```diff
@@ -0,0 +1,259 @@
...
+        ("[\\-]", "-", True, None),
+        ("[x\\-]", "x", True, None),
+        ("[x\\-]", "-", True, None),
+        ("[x\\-]", "z", False, None),
+        ("[\\-x]", "x", True, None),
+        ("[\\-x]", "-", True, None),
+        ("[\\-x]", "a", False, None),
+        # These do not return errors because the Python re.compile() method does
+        # not raise an error on invalid syntax like Go does. We can omit the
+        # tests though since it doesn't affect behavior on _correct_ syntax.
```
ekzhang, line 237: FYI

mwaskom:
This could also go in `mount_utils` but maybe better to cordon off the semi-generated code 😄.

test/pattern_matcher_test.py
pawalt: Is there a possibility that a user could be expecting this to not match something that ends up getting matched? What's the behavior on invalid syntax for `re`? I'd hope it would match nothing.

test/pattern_matcher_test.py
```diff
@@ -0,0 +1,269 @@
...
+
+def test_matches_with_malformed_patterns():
+    try:
+        PatternMatcher(["["])
+    except ValueError as e:
+        assert str(e) == "Bad pattern: ["
+
+
+def test_matches():
+    tests = [
```
pawalt, line 66: These tests are great! Glad the upstream had them

modal/_utils/pattern_matcher.py
```diff
@@ -0,0 +1,274 @@
...
+    Returns:
+        list: A list of patterns to ignore.
+    """
+    if reader is None:
+        return []
+
+    excludes: List[str] = []
+
+    for line in reader:
+        pattern = line.rstrip("\n\r")
```
pawalt, line 247: Do we need this if we're going to `.strip()` anyway?

modal/_utils/pattern_matcher.py
```diff
@@ -0,0 +1,274 @@
...
+        pattern = self.cleaned_pattern
+
+        esc_separator = separator
+        if separator == "\\":
+            esc_separator = "\\\\"
+
+        self.match_type = MatchType.EXACT
+        i = 0
+        pattern_length = len(pattern)
+        while i < pattern_length:
```
pawalt, line 62: This state machine is quite opaque to me, but the test coverage is good.

test/pattern_matcher_test.py
ekzhang: Go ends up throwing an error, the invalid syntax is for example a mismatched bracket which just gets interpreted literally.

modal/_utils/pattern_matcher.py
ekzhang: Good point, it's not strictly necessary. But I think it aids in clarity since Python automatically includes the `\r\n` or `\n` at the end of the line when iterating through a file like this.

modal/_utils/pattern_matcher.py
ekzhang: Yeah, I think the important thing is that it is translated exactly from the upstream

test/pattern_matcher_test.py
pawalt: Nice so they can't bring an invalid dockerfile anyway

pawalt (approved):
LGTM

mwaskom:
Amazing; thanks for this. Made a ticket to track using this in the Image API.

mwaskom:
@ekzhang are we good to merge here?

PR #2304, opened 2024-10-01
Add a shim around OutputManager to avoid Rich imports

mwaskom:
Part of MOD-3730

This PR adds a thin shim around the `OutputManger` so that we don't need to import it directly in other modules, which means we don't transitively import all of the Rich stuff that's in the global scope of `modal/_output.py` for most uses of the Modal library.

Feels maybe slightly messy, but mostly alleviates us from needing to install Rich in the container requirements! Really, the `OutputManager` singleton refactor did most of the work here.

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -9,7 +9,6 @@ if sys.version_info[:2] >= (3, 13):
 from modal_version import __version__
 
 try:
-    from ._output import enable_output
     from ._tunnel import Tunnel, forward
     from .app import App, Stub
     from .client import Client
@@ -22,6 +21,7 @@ try:
     from .image import Image
     from .mount import Mount
     from .network_file_system import NetworkFileSystem
+    from .output import enable_output
     from .partial_function import asgi_app, batched, build, enter, exit, method, web_endpoint, web_server, wsgi_app
     from .proxy import Proxy
     from .queue import Queue
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -697,24 +697,3 @@ class FunctionCreationStatus:
                 )
         else:
             self.status_row.finish(f"Created function {self.tag}.")
-
-
-@contextlib.contextmanager
-def enable_output(show_progress: bool = True) -> Generator[None, None, None]:
-    """Context manager that enable output when using the Python SDK.
-
-    This will print to stdout and stderr things such as
-    1. Logs from running functions
-    2. Status of creating objects
-    3. Map progress
-
-    Example:
-    ```python
-    app = modal.App()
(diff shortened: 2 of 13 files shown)


mwaskom:
cc @erikbern since the `OutputManager` stuff is fresh in your mind.

mwaskom:
@prbot approve — no functional changes here so happy to revise the approach, but this is the final piece to unblock testing remote workflows without Rich installed

erikbern:
I think my gut feeling would have been to put all `import rich` in relative scope instead. It feels like there might be situations where a container grabs the output manager but doesn't do anything to it? In that case it will fail in runtime.

What about if a container does something that involves the resolver? Will that trigger `get_output_manager()`?

mwaskom:
>  I think my gut feeling would have been to put all import rich in relative scope instead

I think this would be difficult to do in a clean way given that Rich imports very little into its top-level namespace and requires a lot of direct imports. Currently that module does

```python
from grpclib.exceptions import GRPCError, StreamTerminatedError
from rich.console import Console, Group, RenderableType
from rich.live import Live
from rich.panel import Panel
from rich.progress import (
    BarColumn,
    DownloadColumn,
    MofNCompleteColumn,
    Progress,
    ProgressColumn,
    TaskID,
    TextColumn,
    TimeElapsedColumn,
    TimeRemainingColumn,
    TransferSpeedColumn,
)
from rich.spinner import Spinner
from rich.text import Text
```

And then uses those in various places (include type signatures).

On top of that, some of the classes we define in `_output.py` _inherit from_ Rich classes, so we can't completely extricate it from global scope.

I'd be in favor of further refactoring the output stuff to make this more feasible though! As you know I find it fairly complicated :). We have lots of internal testing to do before we fully "release" the 2024.10 builder version, so I can look for some low hanging fruit to that effect.

But also:

> there might be situations where a container grabs the output manager but doesn't do anything to it

I think your singleton / `enable_output` helps a lot here! Unless something in the container code explicitly activates the output manager, simple "grabbing" it shouldn't be enough right?

mwaskom:
I wonder if it would help to track the "output enabled" state separately from the `OutputManager` object itself? Maybe that's messier but would help prevent this case:

> there might be situations where a container grabs the output manager but doesn't do anything to it?

since you would also need to have output enabled, not just run across a line of code that grabs the output manager.

erikbern:
> I wonder if it would help to track the "output enabled" state separately from the `OutputManager` object itself? Maybe that's messier but would help prevent this case:

Yeah I had a similar thought – you could maybe move that into the shim

PR #2305, opened 2024-10-01
POC: lazy hydrated Cls.from_name

freider:
Quick hack/draft to see if I could make `Cls.from_name` lazy, which seems doable. My hope is that we can do it much more cleanly letting us get rid of a bunch of the complicated logic in Cls/Obj...

Keeping this as a draft for now

## Describe your changes

---

---

</details>

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -89,12 +89,11 @@ class _RemoteObj:
 
     def __getattr__(self, name) -> _Function:
         async def _load(method: "_Function", resolver: Resolver, existing_object_id: Optional[str]):
+            # haxx
             obj = await self._get_obj()
             obj_method = getattr(obj, name)
             await obj_method.resolve()
-            # return obj_method
-            method._hydrate_from_other(obj_method)  # needed?
-            # return method
+            method._hydrate_from_other(obj_method)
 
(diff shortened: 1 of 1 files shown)


erikbern:
Very cool!

It's been a while since I looked at this code, but could we remove most of the `getattr` hacks at this point? Now that there's just one function serving all methods on a class, there's no need for `getattr` to do much logic and figure out _which_ method we're calling right – we could just return a `_Function` object with a method name bound to it. If that method doesn't exist on the class, we can let it fail in runtime.

Btw if we introduce a `_RemoteObj` I wonder if we should also go ahead and turn the `_Obj` into a `_LocalObj`. We've talked about it in the past and I think it would simplify a few things if we separate the two cases.

freider:
Took another look at this. There is a small caveat with this way of implementing RemoteObj.

It will basically assume all attributes accessed on the Obj are methods, and if they aren't it will fail at `load`-time. Attributes on the instance can't be accessed lazily for this reason, but afaik we should be able to just provide whatever was set in the constructor keyword args without having to load anything (explicit constructors could still set other attributes, but I think we should deprecate that)

PR #2306, opened 2024-10-01
Remove Rich and dependencies from 2024.10 image builder requirements

mwaskom:
Closes MOD-3730

I'll be doing lots more live testing (in my own workspace, and then eventually in modal-labs apps) to further derisk this, beyond the "can we import the container entrypoint" tests that we run in CI.


mwaskom:
@prbot approve — doesn't affect user workflows yet

PR #2307, opened 2024-10-01
Hide remaining rich-dependent imports

mwaskom:
Oops, missed a couple stray imports that needed Rich :(. At least the tests that ran on #2306 worked!

Could definitely benefit from some more cleanup here. I'll think about it.

Part of MOD-3730


mwaskom:
@prbot approve — I'm going to keep thinking about a more thorough refactor here, but want to get testing started.

erikbern:
👍

PR #2309, opened 2024-10-02
Delete a bunch of old proto endpoints

erikbern:
The old endpoints for creating objects are no longer used. I randomly noticed `ClientCreate` and `ClientHeartbeat` could be deleted as well – they haven't been used for at least a year.


mwaskom (approved):
👍 🧹

PR #2310, opened 2024-10-02
Remove FastAPI and dependencies from container requirements on 2024.10

mwaskom:
Slims down the 2024.10 container requirements to remove FastAPI and its dependencies, which is required only when the `modal.web_endpoint` feature is used.

I added a more informative error when we fail to import it, although we plan to primarily handle that case server-side by comparing the function and image definitions.

I'll add documentation to the `modal.web_endpoint` when we document and publicize the forthcoming image builder version release. We'll also need to update our examples that use `web_endpoint`.

Closes MOD-4004

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -96,7 +96,9 @@ jobs:
           deactivate
 
       - name: Check entrypoint import
-        run: python -c 'import modal._container_entrypoint; import fastapi'
+        run: |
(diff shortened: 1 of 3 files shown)


mwaskom:
@prbot approve; has no public impact yet

yasyf:
this makes me very happy :)

aksh-at (commented):
Awesome! So lightweight now

PR #2311, opened 2024-10-02
Slightly cleaner way to convert object ids to types

erikbern:
Needed this for something else (which will be in a separate PR) so I turned it into a helper

### DIFF
--- a/modal/object.py
+++ b/modal/object.py
@@ -157,7 +157,7 @@ class _Object:
         return obj
 
     @classmethod
-    def get_type_from_id(cls: Type[O], object_id: str) -> Type[O]:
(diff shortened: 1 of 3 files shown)


mwaskom (approved):
LGTM although I suspect we don't want these new methods to be in public API?

erikbern:
sure i can prefix w underscore

mwaskom:
> any underscore method to me means "private to this class" but in this case it's actually also available to subclasses

Yeah — Python has the double-underscore thing, but nobody really uses that. In practice, I am more worried about confusing external users than internal developers, so I think we should err on the side of restraining public API even if it makes the internal usage a little confusing.

PR #2312, opened 2024-10-03
Bump container requirement versions for 2024.10

mwaskom:
Part of MOD-3702. Possibly subject to further refinement if broader testing surfaces any issues.


ekzhang (approved):
Looks good as far as I can tell!

PR #2318, opened 2024-10-04
Add included Python packaging tools to base-images.json

mwaskom:
The set of included packaging tools that we install in the base image is another thing that varies across image builder versions. This might be interesting to know about (especially as people are keen to use `uv`).

It's possible this won't really change much going forward, so maybe we'll remove this after we retire 2023.12. But It's helpful for now.


mwaskom:
@prbot approve

mwaskom:
Oops @annacarey we should add you to the prbot pool!

annacarey (approved):

PR #2319, opened 2024-10-04
[MOD-3974] Fix grouped functions

thecodingwizard:
This PR fixes grouped functions introduced in #2212 and adds a test.

Previously, I couldn't get grouped functions to work. A simple function like this:

```py
app = modal.App("hello-world")
@app.function()
@modal.experimental.grouped(size=2)
def simple():
    print("simple")

@app.local_entrypoint()
def main():
    simple.remote()
```

would fail with

```
  File "/pkg/modal/_container_io_manager.py", line 735, in handle_user_exception
    yield
  File "/pkg/modal/_container_entrypoint.py", line 843, in main
    finalized_functions = service.get_finalized_functions(function_def, container_io_manager)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pkg/modal/_container_entrypoint.py", line 137, in get_finalized_functions
    is_async = get_is_async(self._user_defined_callable)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pkg/modal/_utils/function_utils.py", line 93, in is_async
    raise RuntimeError(f"Function {function} is a strange type {type(function)}")
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: __str__ returned non-string (type function)
```

I think `function` is supposed to be a normal function but is instead a `_GroupedFunction`.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -7,6 +7,7 @@ import dataclasses
 import datetime
 import hashlib
 import inspect
+import numbers
 import os
 import platform
 import pytest
@@ -238,7 +239,9 @@ class MockClientServicer(api_grpc.ModalClientBase):
 
         @self.function_body
         def default_function_body(*args, **kwargs):
-            return sum(arg**2 for arg in args) + sum(value**2 for key, value in kwargs.items())
+            return sum(arg**2 for arg in args if isinstance(arg, numbers.Number)) + sum(
+                value**2 for key, value in kwargs.items() if isinstance(value, numbers.Number)
+            )
 
     def function_body(self, func):
         """Decorator for setting the function that will be called for any FunctionGetOutputs calls"""
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -7,7 +7,6 @@ import dataclasses
 import datetime
 import hashlib
 import inspect
-import numbers
 import os
 import platform
 import pytest
@@ -239,9 +238,7 @@ class MockClientServicer(api_grpc.ModalClientBase):
 
         @self.function_body
         def default_function_body(*args, **kwargs):
-            return sum(arg**2 for arg in args if isinstance(arg, numbers.Number)) + sum(
-                value**2 for key, value in kwargs.items() if isinstance(value, numbers.Number)
-            )
+            return sum(arg**2 for arg in args) + sum(value**2 for key, value in kwargs.items())
 
     def function_body(self, func):
         """Decorator for setting the function that will be called for any FunctionGetOutputs calls"""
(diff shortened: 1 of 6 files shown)


test/conftest.py
```diff
@@ -238,7 +239,9 @@ def __init__(self, blob_host, blobs):
 
         @self.function_body
         def default_function_body(*args, **kwargs):
-            return sum(arg**2 for arg in args) + sum(value**2 for key, value in kwargs.items())
+            return sum(arg**2 for arg in args if isinstance(arg, numbers.Number)) + sum(
+                value**2 for key, value in kwargs.items() if isinstance(value, numbers.Number)
+            )
```
thecodingwizard, line 244: This change is needed because grouped functions would pass in a `modal_q` kwarg that is a modal Queue rather than a number, which would cause issues.

test/conftest.py
```diff
@@ -238,7 +239,9 @@ def __init__(self, blob_host, blobs):
 
         @self.function_body
         def default_function_body(*args, **kwargs):
-            return sum(arg**2 for arg in args) + sum(value**2 for key, value in kwargs.items())
+            return sum(arg**2 for arg in args if isinstance(arg, numbers.Number)) + sum(
```
thundergolfer, line 242: instead of making this default more complex could you instead set `function_body` below just for the grouped function?

test/function_test.py
```diff
@@ -993,3 +999,15 @@ def test_spawn_extended_feature_flag(client, servicer, monkeypatch, feature_flag
     else:
         expected_invocation_type = api_pb2.FUNCTION_CALL_INVOCATION_TYPE_ASYNC_LEGACY
     assert function_map.function_call_invocation_type == expected_invocation_type
+
+
+def test_spawn_grouped_function(client):
```
thundergolfer, line 1004: Thanks for adding the test. definitely need to stop regressions happened while we iterate. 

There's already `i6pn_group_test.py` which is a good place to slot this in.

thundergolfer (approved):
LGTM if tests are passing. left couple suggestions

thecodingwizard:
I think I fixed the unit tests by monkeypatching modal.Queue.ephemeral to accept the right client, but is there a neater way to do this?

thecodingwizard:
CI is blocked by https://github.com/modal-labs/synchronicity/pull/172

PR #2320, opened 2024-10-04
[WIP] MOD-2454: remove aiostream dependency

kramstrom:
## Describe your changes

---

---

</details>


modal/_utils/async_utils.py
```diff
@@ -463,3 +464,83 @@ def run_generator_sync(
...
+    finally:
+        await gen.aclose()
+
+
+async def async_map(
+    input: AsyncIterator[T], async_mapper_func: Callable[[T], Awaitable[V]], concurrency: int
+) -> AsyncIterator[V]:
+    input_queue: asyncio.Queue[T] = asyncio.Queue(maxsize=concurrency)
+    results_queue: asyncio.Queue[V] = asyncio.Queue()
+    end_sentinel = object(T)
```
freider, line 482: What does `object(T)` mean? T is a typevar right - that wouldn't typically be used at runtime? 🤔

modal/_utils/async_utils.py
```diff
@@ -463,3 +464,83 @@ def run_generator_sync(
...
+
+    producer_task = asyncio.create_task(producer())
+    worker_tasks = [asyncio.create_task(worker()) for _ in range(concurrency)]
+
+    try:
+        while True:
+            if results_queue.empty():
+                if producer_task.done() and all(t.done() for t in worker_tasks):
+                    break
+                await asyncio.sleep(0.01)  # Small delay to avoid busy waiting
```
freider, line 506: It's still a pretty busy loop... I think it would be better to `asyncio.wait()` to wait for any of the underlying tasks to complete than to use a timed sleep

modal/_utils/async_utils.py
freider: A clean alternative to using end sentinels would be to use `inputs_queue.task_done()` and `inputs_queue.join()` btw, e.g.
Then you set `input_queue.task_done()` whenever you have put a new item on the result queue and can:

```
async def complete_map():
     await producer_task
     await inputs_queue.join()

complete_map_task = asyncio.create_task(complete_map())
```

`asyncio.wait([complete_map_task, has_new_result], when=FIST_COMPLETED)` to see when you need to either yield a new result or have exhausted inputs

modal/_utils/async_utils.py
```diff
@@ -463,3 +476,93 @@ def run_generator_sync(
...
+
+    async def complete_map():
+        await producer_task
+        await input_queue.join()
+
+    complete_map_task = asyncio.create_task(complete_map())
+
+    try:
+        while True:
+            await asyncio.wait([complete_map_task, producer_task, *worker_tasks], return_when=asyncio.FIRST_COMPLETED)
```
freider, line 519: This will block until one of these tasks has completed, which could be long after the first result is produced. You'd have to have a task that also checks if any new outputs have been created on the queue.
Either an `asyncio.create_task(queue.get()` or (maybe preferable) some separate `asyncio.Event` that is set whenever a worker puts an item and reset whenever you drain the output queue

freider: Make sure to add a test for this too. Should be testable by having a map workload that takes X time to complete and using 1 concurrent tasks you should be able to get 1 output roughly every X time

modal/_utils/async_utils.py
```diff
@@ -463,3 +478,114 @@ def run_generator_sync(
         except BaseException as err:
             exc = err
     loop.close()
+
+
+@asynccontextmanager
+async def aclosing(
```
freider, line 484: I'd prefer if this *only* takes `AsyncGenerators` and we don't apply it to iterators that don't suffer the same problem. In cases we are not sure if the input is a generator or an non-generator iterator we could have compose `aclosing(ensure_gen(input))` instead

modal/functions.py
```diff
@@ -206,21 +207,19 @@ async def poll_function(self, timeout: Optional[float] = None):
...
-                if isinstance(item, api_pb2.GeneratorDone):
-                    items_total = item.items_total
-                else:
-                    yield item
-                    items_received += 1
-                # The comparison avoids infinite loops if a non-deterministic generator is retried
-                # and produces less data in the second run than what was already sent.
-                if items_total is not None and items_received >= items_total:
-                    break
+        async for item in async_merge(data_stream, awaitable_to_aiter(self.run_function())):
```
freider, line 213: This should probably have an `aclosing` around the async_merge to be safe, since async_merge is an async generator with async finalizers.

`data_stream` also needs aclosing

PR #2321, opened 2024-10-04
Remove only usage of synchronizer.create_async (only test)

freider:
Being removed from synchronicity in upcoming version


freider:
@prbot approve

mwaskom (commented):
Bye!

PR #2322, opened 2024-10-04
Fix excessive tracebacks when sigint:ing a .map call [CLI-33]

freider:
* Depends on synchronicity fix of async cancellation propagation ~(v0.8.3)~ v0.9.1
* Depends on aclosing polyfill in @kramstrom 's #2320 

This fixes a lot of asyncio junk getting split out in the users face when they sigint any of our  `.map` family methods (the non-async versions).

The reason we used to these errors is because these methods start a temporary asyncio event loop on the *main thread* in order to run the concurrent consumption of the user provided iterators and the modal-synchronicity-backed output iterator.
We then interleaved executions on this event loop with the actual user code from the `for` body, which also runs on the main loop. IF the process got SIGINT:ed while in the event loop code, it would previously trigger a KeyboardInterrupt in asyncio internals which is a recipe for long tracebacks and ungraceful shutdown.

This fixes that problem by using the same mechanism as ~`asyncio.run()`~ asyncio.Runner on Python 3.11 has to avoid this - by adding a temporary SIGINT listener that converts the sigint into an asyncio cancellation, with an additional improvement that we also make sure that any async cancellation finalization (finally blocks etc.) are guaranteed to run before reraising the KeyboardInterrupt. This improved Runner context manager/utility is now in synchronicity

In addition, this incorporates some underlying fixes in synchronicity that fixes Ctrl-C behavior on Windows


modal/_utils/async_utils.py
```diff
@@ -451,41 +458,124 @@ class NestedAsyncCalls(Exception):
...
+        except asyncio.CancelledError:
+            if self._num_sigints > 0:
+                raise KeyboardInterrupt()  # might want to use original_sigint_handler here instead?
+            raise  # "internal" cancellations, not triggered by KeyboardInterrupt
+        finally:
+            if original_sigint_handler:
+                # reset signal handler
+                signal.signal(signal.SIGINT, original_sigint_handler)
+
+    def run_async_gen(
```
freider, line 548: Moved our old `run_generator_sync` into the `Runner` class since it uses the Runner.run anyways n ow... Otherwise unchanged

modal/_utils/async_utils.py
```diff
@@ -451,41 +458,124 @@ class NestedAsyncCalls(Exception):
...
-    gen: typing.AsyncGenerator[YIELD_TYPE, SEND_TYPE],
-) -> typing.Generator[YIELD_TYPE, SEND_TYPE, None]:
-    try:
-        asyncio.get_running_loop()
-    except RuntimeError:
-        pass  # no event loop - this is what we expect!
-    else:
-        raise NestedAsyncCalls()
-    loop = asyncio.new_event_loop()  # set up new event loop for the map so we can use async logic
+class Runner:
```
freider, line 461: A "lightweight" `asyncio.Runner` (from python 3.11). Very similar to our `UserCodeEventLoop` class in `_container_entrypoint.py` which has a similar purpose.

test/async_utils_test.py
```diff
@@ -624,3 +628,70 @@ async def foo():
...
+            print("bye")
+    try:
+        with Runner() as runner:
+            for res in runner.run_async_gen(async_gen()):
+                print("res", res)
+    except KeyboardInterrupt:
+        print("KeyboardInterrupt")
+    """
+    )
+    if sys.platform == "win32":
```
freider, line 661: This is probably what I spent the most time on - getting the Ctrl-C test to work on windows... Had to fire up the ol' gaming PC to run som tests...

mwaskom: image

modal/_utils/async_utils.py
freider: Just put the same (or at least very similar) backport into synchronicity (version tbd, 0.9?) so might wait with merging this until I can use that

freider:
@prbot approve

PR #2323, opened 2024-10-04
[WIP] Image updates to support Python 3.13

mwaskom:
Won't merge until 3.13 is officially out (scheduled for October 7) but this has most of what we should need to support it.

Note that we'll need to gate 3.13 support behind the 2024.10 image builder version as the pinned version of `aiohttp` does not build properly on it.

There's a few loose threads here:
- [ ] Confirm `micromamba` support for 3.13 or handle that as unsupported for now
- [ ] Bump the standalone Python pin to at least stable `3.13.0`.
- [ ] Update the modal library metadata to declare support

- Towards MOD-3990

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -3,7 +3,7 @@ import sys
 
 if sys.version_info[:2] < (3, 8):
     raise RuntimeError("This version of Modal requires at least Python 3.8")
-if sys.version_info[:2] >= (3, 13):
+if sys.version_info[:2] >= (3, 14):
     raise RuntimeError("This version of Modal does not support Python 3.13+")
 
 from modal_version import __version__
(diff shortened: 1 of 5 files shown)


modal/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 if sys.version_info[:2] < (3, 8):
     raise RuntimeError("This version of Modal requires at least Python 3.8")
-if sys.version_info[:2] >= (3, 13):
+if sys.version_info[:2] >= (3, 14):
     raise RuntimeError("This version of Modal does not support Python 3.13+")
```
erikbern, line 7: should say `3.14+`

mwaskom:
Superceded by #2334

PR #2324, opened 2024-10-04
Remove AppLookupObject, legacy code

ekzhang:
I couldn't find any usage of this RPC, anywhere. Well, now I gotta try to find what replaced this!


erikbern (approved):
My bad, this was removed with the rest of the dummy app code. We replaced this with endpoints for each specific object type!

PR #2327, opened 2024-10-05
Add rpc for deleting secrets

erikbern:
Realized we have no rpc for this (we rely on `AppStop` from the web UI).

I'm planning to change this between the web and the backend but I think it makes sense to put this in the client proto similar to other object types – it lets us add client/CLI support for deletions later

This is part of removing dummy apps server side.


erikbern:
@prbot approve

freider (commented):
👍

PR #2330, opened 2024-10-08
Raise EOF when writing to a closed sandbox over the network

pawalt:
## Describe your changes

If Stdin is closed (particularly relevant for sandboxes), and EOF error will be throw instead of accepting the write.

---

---

</details>

## Changelog

When writing to a `StreamWriter` that has already had EOF written, a ValueError is now raised instead of an `EOFError`.

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -2,6 +2,7 @@
 import asyncio
 from typing import TYPE_CHECKING, AsyncIterator, Literal, Optional, Tuple, Union
 
+from grpclib import Status
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 
 from modal_proto import api_pb2
@@ -226,7 +227,7 @@ class _StreamWriter:
         ```
         """
         if self._is_closed:
-            raise EOFError("Stdin is closed. Cannot write to it.")
+            raise ValueError("Stdin is closed. Cannot write to it.")
         if isinstance(data, (bytes, bytearray, memoryview, str)):
             if isinstance(data, str):
                 data = data.encode("utf-8")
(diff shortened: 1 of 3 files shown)


modal/io_streams.py
```diff
@@ -251,21 +252,27 @@ async def drain(self):
...
+                await retry_transient_errors(
+                    self._client.stub.ContainerExecPutInput,
+                    api_pb2.ContainerExecPutInputRequest(
+                        exec_id=self._object_id,
+                        input=api_pb2.RuntimeInputMessage(message=data, message_index=index, eof=self._is_closed),
+                    ),
+                )
+        except GRPCError as exc:
+            if exc.status == Status.FAILED_PRECONDITION:
+                raise EOFError("Stdin is closed. Cannot write to it.")
```
thundergolfer, line 273: Can't you grab the message in the gRPC exception? It'd be more accurate and forward compatible. 
Also not sure `EOFError` is what a user should expect here, could we instead throw an exception from within the `modal.exception` namespace that is like 'invalid error'?

Also re-reading the docstring I don't like that it suggests that `drain()` may always send EOF. It'd be nice to reword that to indicate EOF is only sent once the writer is _closed_.

thundergolfer (commented):
Have some alternative ideas about the appropriate error-handling here, and also think this error-handling functionality is important enough to warrant a basic unit test 👍

modal/io_streams.py
pawalt: Fixed the comments and switched to ValueError to match python native socket behavior.

pawalt:
@thundergolfer I added a test

modal/io_streams.py
```diff
@@ -245,27 +246,33 @@ def write_eof(self):
...
+                await retry_transient_errors(
+                    self._client.stub.ContainerExecPutInput,
+                    api_pb2.ContainerExecPutInputRequest(
+                        exec_id=self._object_id,
+                        input=api_pb2.RuntimeInputMessage(message=data, message_index=index, eof=self._is_closed),
+                    ),
+                )
+        except GRPCError as exc:
+            if exc.status == Status.FAILED_PRECONDITION:
+                raise ValueError("I/O operation on closed file.")
```
thundergolfer, line 273: Would still prefer if the client relied on the message from the server because in the future there could be multiple ways of receiving a `FAILED_PRECONDITION`.

thundergolfer (approved):
LGTM

modal/io_streams.py
pawalt: ok ser

pawalt: Wait can I convince you otherwise on this - I want to match the exact error that Python throws on close FDs for close API compatibility, but I don't think it makes sense for the RPC to throw this. I think the separation is valuable here, for the same reason that we don't throw FAILED_PRECONDITION back to the user.

pawalt: ok boss you convinced me

pawalt:
@prbot automerge

PR #2331, opened 2024-10-08
Add missing `from` statements for deserialization exceptions

thecodingwizard:
This PR adds a couple missing `from` statements when raising exceptions. In particular, deserialization exceptions need `from` statements to properly show the original deserialization error message to the user. This PR also differentiates between `DeserializationError` and general `Exception`s. I have no idea when `deserialize()` will raise an `Exception` instead of a `DeserializationError` but kept it there just in case.

See https://github.com/modal-labs/synchronicity/pull/165 for a related issue.

Currently the error message would be something like

```
ExecutionError: Could not deserialize remote exception due to local error:
Encountered an error when deserializing an object in the local environment (see above for details).
This can happen if your local environment does not have the remote exception definitions.
Here is the remote traceback:
```

but despite saying "see above for details" there would not be details above.

### DIFF
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -16,7 +16,7 @@ from modal_proto import api_pb2
 from .._serialization import deserialize, deserialize_data_format, serialize
 from .._traceback import append_modal_tb
 from ..config import config, logger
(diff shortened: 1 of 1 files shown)


mwaskom (approved):
Makes sense!

PR #2332, opened 2024-10-09
Fix NCCL initialization when i6pn runc containers are scheduled on the same machine

thecodingwizard:
Resolves MOD-4067

NCCL with multiple containers on the same machine with the runc runtime doesn't work, crashing upon initialization. This occurs if a grouped function is scheduled on the same machine.

nccl [uses a host hash](https://github.com/NVIDIA/nccl/blob/48bb7fec7953112ff37499a272317f6663f8f600/src/graph/topo.cc#L654) to determine if two GPUs are on the same machine.

The [host hash](https://github.com/NVIDIA/nccl/blob/48bb7fec7953112ff37499a272317f6663f8f600/src/misc/utils.cc#L86) evaluates to `$(hostname)$(cat /proc/sys/kernel/random/boot_id)`.

[gvisor patches boot_id](https://github.com/google/gvisor/blob/47dade3f98b2e7d1d00c4ba79c7f4bdfc28bd805/pkg/sentry/fsimpl/proc/tasks_sys.go#L56) to be unique, but runc does not.

Therefore in runc, nccl thinks the two GPUs are on the same machine and crashes when it's not able to access them.

Luckily there's a magical NCCL_HOSTID environment variable that can override this.

For more context: https://modal-com.slack.com/archives/C07BWUF5JKW/p1728437710487689?thread_ts=1728428627.889669&cid=C07BWUF5JKW

### DIFF
--- a/modal/experimental.py
+++ b/modal/experimental.py
@@ -155,6 +155,16 @@ def _networked(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         import os
+        import socket
+
+        hostname = socket.gethostname()
+        addr_info = socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)[0][4][0]
+        # nccl's default host ID is $(hostname)$(cat /proc/sys/kernel/random/boot_id).
+        # on runc, if two i6pn-linked containers get scheduled on the same worker,
+        # their boot ID and hostname will both be identical, causing nccl to break.
+        # As a workaround, we can explicitly specify a unique host ID here.
+        # See MOD-4067.
+        os.environ["NCCL_HOSTID"] = f"{hostname}{addr_info}"
 
         rank = kwargs.pop("modal_rank", None)
         size = kwargs.pop("modal_size", None)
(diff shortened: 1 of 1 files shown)


modal/experimental.py
```diff
@@ -155,6 +155,16 @@ def _networked(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         import os
+        import socket
+
+        hostname = socket.gethostname()
+        addr_info = socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)[0][4][0]
```
thundergolfer, line 161:
```suggestion
        addr_info = socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)[0][4][0]
```

nit: would be more readable to unpack or comment the `[0][4][0]` indexing, even if it costs a couple extra LoC.

thundergolfer (approved):
LGTM, hnice

modal/experimental.py
ekzhang: the [0][4][0] is just a magic incantation that you need to know, I don't think there's anything smart about it 🤷‍♂️

maybe we can make it a helper function to get the i6pn

PR #2333, opened 2024-10-09
Allow `modal shell -c` to run a command

ekzhang:
## Changelog

- `modal shell --cmd` now can be shortened to `modal shell -c`. This means you can write commands like `modal shell -c "uname -a"` to quickly run a command within the remote environment.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -339,7 +339,7 @@ def shell(
         ),
         metavar="FUNC_REF",
     ),
-    cmd: str = typer.Option(default="/bin/bash", help="Command to run inside the Modal image."),
+    cmd: str = typer.Option("/bin/bash", "-c", "--cmd", help="Command to run inside the Modal image."),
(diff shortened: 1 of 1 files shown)


mwaskom:
I was chatting with someone yesterday and they mentioned confusion between `modal shell` and `modal container exec`. That made me me wonder whether `modal shell --container=<container_id>` would be more intuitive than `modal container exec <container_id> /bin/bash` for getting a shell in a running container. In which case you might want `-c` to be the short form of `--container`, which is a long word.

On the other hand, `bash -c` etc. is pretty common, so it probably makes sense to have this too. Just mentioning!

ekzhang:
Makes sense, I could look at adding that! How about `modal shell ta-123456` directly then, replacing FUNC_REF?

mwaskom:
That could also work! Really just a non-blocking comment either way, was just top of mind.

ekzhang:
@prbot approve

PR #2334, opened 2024-10-14
Updates to support Python 3.13

mwaskom:
This enabled support for Python 3.13. For the container client, we can only support 3.13 when using the 2024.10 Image Builder Version because the versions of `aiohttp` that we have pinned in other versions does not build successfully on 3.13.

I also needed to bump the declared support for `protobuf` to get a version of `grpcio-tools` that supports 3.13. We know that there is an incompatibility between our gencode and the protobuf 5 runtime, but it appears only when using the "python" backend. We also have some ideas about how to make our protos more cross-runtime-version compatible.

- Closes MOD-3990


setup.cfg
```diff
@@ -25,7 +25,7 @@ install_requires =
     click>=8.1.0
     fastapi
     grpclib==0.4.7
-    protobuf>=3.19,<5.0,!=4.24.0
+    protobuf>=3.19,<6.0,!=4.24.0,!=5.28  # Deny 5.28 until intrusive version warnings are reverted
```
mwaskom, line 28: This didn't work for some reason.

test/image_test.py
```diff
@@ -64,10 +65,13 @@ def get_all_dockerfile_commands(image_id: str, servicer) -> str:
 
 @pytest.fixture(params=get_args(ImageBuilderVersion))
 def builder_version(request, server_url_env, modal_config):
-    version = request.param
+    builder_version = request.param
+    python_version = f"{0}.{1}".format(sys.version_info)
+    if python_version not in SUPPORTED_PYTHON_SERIES[builder_version]:
+        pytest.skip("Python version not supported by Image Builder version")
```
mwaskom, line 71: This produces a lot of skip tests — maybe there is a cleaner way.

.github/workflows/ci-cd.yml
```diff
@@ -63,7 +63,8 @@ jobs:
         run: pytest -v
 
       - name: Run docstring tests
-        if: github.event.pull_request.head.repo.fork != 'true'
+        # TODO(michael) disable doctests on 3.13 until workspace is updated to new Image Builder Version
```
freider, line 66: Preferably our doctests should be able to run on the mock servicer too... It's somewhere on my backlog 😬  I think it's currently blocked by pytest-asyncio fixtures not being usable

test/image_test.py
freider: Yeah I'm not sure. Wasn't even aware you could parameterize fixtures like this
One idea is that you create a reversed mapping of SUPPORTED_PYTHON_SERIES which maps from python version to supported image builder versions, and send that in to params instead? Then it will only be parameterized on the supported image builder versions?

freider (approved):
lgtm!

test/image_test.py
mwaskom: Nice — yeah that's better.

PR #2335, opened 2024-10-14
Fix MOD-4088

ekzhang:
Image.debian_slim has no autogenerated documentation due to a syntax error in markdown.


howard36 (approved):
Thanks! Is there an easy way to catch these issues automatically? I'm worried other parts of our documentation are missing too.

ekzhang:
Unfortunately Markdown has the property that any string is valid Markdown :(

mwaskom:
I feel like I don't totally understand how this error ended up totally dropping the function from the docs — valid point about "markdown syntax" but feels like there is an error state we could be detecting somewhere in the docs builder?

PR #2336, opened 2024-10-14
`modal shell` takes container IDs now

ekzhang:
## Describe your changes

Resolves MOD-4091, allowing you to do `modal shell ta-123456`.

## Changelog

- The `modal shell` CLI command now takes a container ID, allowing you to shell into a running container.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -331,49 +331,46 @@ def serve(
 
 
 def shell(
-    func_ref: Optional[str] = typer.Argument(
+    container_or_function: Optional[str] = typer.Argument(
         default=None,
         help=(
-            "Path to a Python file with an App or Modal function with container parameters."
-            " Can also include a function specifier, like `module.py::func`, when the file defines multiple functions."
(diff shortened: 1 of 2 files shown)


thecodingwizard (approved):
lgtm! another minor benefit is now users don't have to write "bash" after "modal container exec [id]" if they just want a shell in a container

modal/cli/run.py
```diff
@@ -331,49 +331,46 @@ def serve(
 
 
 def shell(
-    func_ref: Optional[str] = typer.Argument(
+    container_or_function: Optional[str] = typer.Argument(
```
mwaskom, line 334: Nit, but one tricky thing about `container_or_function` is that they're not equivalent — the "container" mode is a `container_id` but while we do _have_ a `function_id` concept, that isn't what you would pass here 🤔

ekzhang: Yeah that makes sense! It's only internal though since the metavar in the help is `REF` so maybe fine.

PR #2337, opened 2024-10-15
Note that App.spawn_sandbox is deprecated

mwaskom:
Closes MOD-4112. We were throwing a warning but had not updated the docstring.


pawalt (approved):
thanks homie

PR #2340, opened 2024-10-15
Aggregate StreamReader output into full lines

azliu0:
## Describe your changes

Fixes MOD-4075.

Adds attribute `by_line` to StreamReader that changes whether the iterator returns raw logs (default behavior) or logs that are split by line. Default behavior for sandboxes is to return line-by-line logs.

## Changelog

Adds attribute `by_line` to StreamReader. When `by_line` is `False` (default), the behavior is unchanged. When `by_line` is `True`, logs will read one full line at a time, rather than partial or multiple lines.

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -79,7 +79,7 @@ class _StreamReader:
         object_id: str,
         object_type: Literal["sandbox", "container_process"],
         client: _Client,
-        raw: bool = True,  # if False, streamed logs are further processed into complete lines.
+        by_line: bool = False,  # if True, streamed logs are further processed into complete lines.
     ) -> None:
         """mdmd:hidden"""
         self._file_descriptor = file_descriptor
@@ -89,7 +89,7 @@ class _StreamReader:
         self._stream = None
         self._last_entry_id = None
         self._buffer = ""
-        self._raw = raw
+        self._by_line = by_line
         # Whether the reader received an EOF. Once EOF is True, it returns
         # an empty string for any subsequent reads (including async for)
         self.eof = False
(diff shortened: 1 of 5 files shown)


test/sandbox_test.py
```diff
@@ -152,6 +152,32 @@ def test_sandbox_stdin_write_after_eof(app, servicer):
         sb.stdin.write(b"foo")
 
 
+@skip_non_linux
+def test_sandbox_stdout(app, servicer):
```
pawalt, line 156: I think this way of testing is a good idea, but I'm not actually convinced that the logs look how you expect. Can you add an `io_streams_test.py` file instead and use the servicer to manually override `SandboxGetLogs` to ensure that the log messages actually get aggregated or split properly? Here's an example of how this works for app logs: 
https://github.com/modal-labs/modal-client/blob/pawalt/parampoint/test/app_test.py#L366

pawalt (changes requested):

test/io_streams_test.py
```diff
@@ -0,0 +1,43 @@
...
+
+        # Data with single lines sent separately
+        for i, line in enumerate(lines):
+            log = api_pb2.TaskLogs(data=line, file_descriptor=api_pb2.FILE_DESCRIPTOR_STDOUT)
+            await stream.send_message(api_pb2.TaskLogsBatch(entry_id=str(i + 1), items=[log]))
+
+        # Send EOF
+        await stream.send_message(api_pb2.TaskLogsBatch(eof=True))
+
+    with servicer.intercept() as ctx:
```
pawalt, line 28: Can you break these into separate tests and add comments? It's a bit hard to understand what's going on. Also add a test that partial lines are aggregated.

test/sandbox_test.py
pawalt: Nice yeah you can keep these, but add a comment describing what you're doing.

azliu0: What do we expect the logs to look like? I think the `echo` comes with an extra newline which is why it looks weird. Added `io_streams_test` to test the aggregating more explicitly.

modal/io_streams.py
```diff
@@ -79,15 +79,17 @@ def __init__(
         object_id: str,
         object_type: Literal["sandbox", "container_process"],
         client: _Client,
+        raw: bool = True,  # if False, streamed logs are further processed into complete lines.
```
mwaskom, line 82: "raw" to me implies something about the encoding — would `by_line: bool` make sense?

pawalt (approved):

PR #2342, opened 2024-10-16
MOD-4118: handle deserialization errors

kramstrom:
## Describe your changes

---

---

</details>


modal/_container_io_manager.py
```diff
@@ -123,9 +125,7 @@ async def _populate_input_blobs(client: _Client, input: api_pb2.FunctionInput) -
         method_name = function_inputs[0].method_name
         assert all(method_name == input.method_name for input in function_inputs)
         finalized_function = finalized_functions[method_name]
-        # TODO(cathy) Performance decrease if we deserialize function_inputs later
```
kramstrom, line 126: need to measure if we get a performance hit by moving

test/container_test.py
```diff
@@ -2256,3 +2256,35 @@ def test_set_local_input_concurrency(servicer):
...
+@skip_github_non_linux
+def test_deserialization_error_returns_exception(servicer, client):
+    inputs = [
+        api_pb2.FunctionGetInputsResponse(
+            inputs=[
+                api_pb2.FunctionGetInputsItem(
+                    input_id="in-xyz0",
+                    function_call_id="fc-123",
+                    input=api_pb2.FunctionInput(
+                        args=b"\x80\x04\x95(\x00\x00\x00\x00\x00\x00\x00\x8c\x17",
```
erikbern, line 2270: instead of a byte string here, can you feed it a pickled object? or is that impossible because it needs to be a different package?

kramstrom: yeah would have to somehow pickle the object in another venv than it's being pickled in which sounds annoying but more realistic, but both of these causes serialization errors

gongy (approved):
Thanks for adding the test!

PR #2345, opened 2024-10-16
Don't sleep before sending an exec request

pawalt:
The underlying bug that requires this sleep has been fixed.

## Describe your changes

- Resolves MOD-4076

---

---

</details>

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -388,8 +388,8 @@ class _Sandbox(_Object, type_prefix="sb"):
         while not self._task_id:
             resp = await self._client.stub.SandboxGetTaskId(api_pb2.SandboxGetTaskIdRequest(sandbox_id=self.object_id))
             self._task_id = resp.task_id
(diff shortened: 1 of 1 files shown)


modal/sandbox.py
```diff
@@ -388,8 +388,8 @@ async def _get_task_id(self):
         while not self._task_id:
             resp = await self._client.stub.SandboxGetTaskId(api_pb2.SandboxGetTaskIdRequest(sandbox_id=self.object_id))
             self._task_id = resp.task_id
-            # TODO: debug why sending an exec right after a task ID exists fails silently
-            await asyncio.sleep(0.5)
+            if not self._task_id:
```
irfansharif, line 391: When do we not get a task ID back, and why are we sleeping for those cases? How come we're not removing the sleep completely?

irfansharif (approved):

modal/sandbox.py
pawalt: We shouldn't hammer the RPC - it would hot loop otherwise.

PR #2347, opened 2024-10-16
Add a function to build images immediately

pawalt:
## Describe your changes

Resolves MOD-4102

---

---

</details>

## Changelog

Images can now be built directly with `Image.build`.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -29,6 +29,7 @@ from typing import (
 from google.protobuf.message import Message
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 
+from modal.client import _Client
 from modal_proto import api_pb2
 
 from ._resolver import Resolver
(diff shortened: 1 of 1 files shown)


modal/image.py
```diff
@@ -465,6 +467,15 @@ async def join():
         obj.force_build = force_build
         return obj
 
+    async def build(self, app: "_App", client: Optional[_Client] = None) -> None:
+        """Build the image immediately rather than waiting for it to be used."""
+
+        client = client or app._client or await _Client.from_env()
```
mwaskom, line 473: Do we need `app` as an argument here? Feels confusing to have an object that is usually passed to `App` have a method that accepts an `App` as an argument.

modal/image.py
```diff
@@ -465,6 +467,15 @@ async def join():
         obj.force_build = force_build
         return obj
 
+    async def build(self, app: "_App", client: Optional[_Client] = None) -> None:
+        """Build the image immediately rather than waiting for it to be used."""
```
mwaskom, line 471: Nit but the "waiting for it to be used" here is arguably not a correct description of what usually happens — images get build when we create apps or functions that _will_ use them.

erikbern:
I think `image.resolve()` might already do this, but could be wrong! Either way this I think this points to some complexity around managing "lazy" vs "eager" objects (c.f. `from_name` vs `lookup`)

mwaskom (changes requested):
Discussed with Peyton on Slack but I don't think this actually solves the user's problem — not sure we should expand API without a clear rationale.

pawalt:
Cancelling as we have better ways to solve this UX pain

PR #2349, opened 2024-10-17
MOD-2454: remove aiostream dependency: async_map

kramstrom:
continue splitting https://github.com/modal-labs/modal-client/pull/2320 up into multiple PRs as it grew quite big. This adds `async_map` and `async_map_ordered` which are the last dependencies we have on the `aiostream` package


modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
 
 async def callable_to_agen(awaitable: Callable[[], Awaitable[T]]) -> AsyncGenerator[T, None]:
     yield await awaitable()
+
+
+async def async_map(
+    input: Union[AsyncIterable[T], Iterable[T]],
+    async_mapper_func: Callable[[T], Awaitable[V]],
```
freider, line 552: If we are also allowing this to be a non-async function (as seems to be the case below), we should probably update the signature here to have `Union[Awaitable[V], V]` in the return value.

freider: Not sure if that fixes some typing issues too

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+        await input_queue.put(("stop", None))
+
+    async def worker():
+        while True:
+            try:
+                event_type, item = await input_queue.get()
+                if event_type == "stop":
+                    break
+                if event_type == "value":
+                    if asyncio.iscoroutinefunction(async_mapper_func):
```
freider, line 571: Checking a function with iscoroutinefunction is not 100% water tight, as the function could be regular function that *returns* a coroutine, without being declared as `async def`, in which case this would return False.

A better way to do this is to actually call the function (without await) and check if the returned value with `inspect.isawaitable` and only then await it.

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+                        result = await async_mapper_func(item)
+                    else:
+                        result = typing.cast(V, async_mapper_func(item))
+                    await output_queue.put(("value", result))
+            except Exception as e:
```
freider, line 576: This might be a nit pick that doesn't have any practical consequences at the moment since we never do the equivalent of `return_exceptions=True`, but it would be nice to have the "user level" exceptions (errors when executing `async_mapper_func`) caught separately from errors that might be emitted by our `async_map` implementation itself, instead of one large except block catching everything.

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
 
 async def callable_to_agen(awaitable: Callable[[], Awaitable[T]]) -> AsyncGenerator[T, None]:
     yield await awaitable()
+
+
+async def async_map(
+    input: Union[AsyncIterable[T], Iterable[T]],
```
freider, line 551: nit: `input` is a builtin in python, so try to avoid using it as a var name

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+    async_mapper_func: Callable[[T], Awaitable[V]],
+    concurrency: int,
+) -> AsyncGenerator[V, None]:
+    input_queue: asyncio.Queue[Tuple[str, Union[T, None]]] = asyncio.Queue()
+    output_queue: asyncio.Queue[Tuple[str, Union[V, Exception]]] = asyncio.Queue()
+    output_event = asyncio.Event()
+
+    async def producer():
+        async for item in sync_or_async_iter(input):
+            await input_queue.put(("value", item))
```
freider, line 561: Should make the entry types into an enum instead of strings to avoid accidental and catastrophic typos 😬

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+                        result = await async_mapper_func(item)
+                    else:
+                        result = typing.cast(V, async_mapper_func(item))
+                    await output_queue.put(("value", result))
+            except Exception as e:
+                await output_queue.put(("exception", e))
+            finally:
+                input_queue.task_done()
+
+    output_event_task = asyncio.create_task(output_event.wait())
```
freider, line 581: I think this could be the source of a bug. `output_event_task` needs to be recreated every time that the output_event triggers, since clearing the event doesn't undo the completeness of `output_event_task`.

Otherwise the `asyncio.wait` here fires immediately in every loop iteration after the first input is returned, leading to a "hot loop" where the `output_event.is_set():` condition would be repeatedly checked even though nothing has changed.

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+    try:
+        while True:
+            done, _ = await asyncio.wait(
+                all_tasks,
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+
+            finished_workers = done & set(worker_tasks)
+            for finished_worker in finished_workers:
+                await finished_worker
```
freider, line 602: Add a comment that this is done in order to catch potential raised errors/cancellations from within worker tasks as soon as they happen.

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+    producer_task = asyncio.create_task(producer())
+    worker_tasks = [asyncio.create_task(worker()) for _ in range(concurrency)]
+
+    async def complete_map():
+        await producer_task
+        await input_queue.join()
+
+    complete_map_task = asyncio.create_task(complete_map())
+
+    all_tasks = [output_event_task, producer_task, *worker_tasks, complete_map_task]
```
freider, line 591: `producer_task` shouldn't need to be in all_tasks, since we only care about `complete_map_task` which transitively includes it

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+            if complete_map_task.done():
+                while not output_queue.empty():
+                    event_type, item = await output_queue.get()
+                    if event_type == "value":
+                        yield typing.cast(V, item)
+                    elif event_type == "exception":
+                        raise typing.cast(Exception, item)
+                    else:
+                        raise Exception("Unknown event type: " + event_type)
+                break
```
freider, line 613: should add a `await complete_map_task` here before the break - in case it has exceptions that needs to be surfaced, similar to the worker tasks. Otherwise those exceptions would never get surfaced, since the `finally` clause uses `return_exceptions=True`

modal/_utils/async_utils.py
```diff
@@ -544,3 +545,116 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+
+    next_idx = 0
+    buffer = {}
+
+    async with aclosing(counter()) as counter_gen:
+        async with aclosing(async_zip(counter_gen, input)) as zipped_input:
+            async with aclosing(async_map(zipped_input, mapper_func_wrapper, concurrency)) as stream:
+                async for output_idx, output_item in stream:
+                    buffer[output_idx] = output_item
+                    if output_idx == next_idx:
```
freider, line 656: Nit: you can skip the `if` statement here since the same condition will be implicitly checked by the while :neckbeard:

modal/network_file_system.py
```diff
@@ -343,12 +342,16 @@ def gen_transfers():
...
-        await aiostream.stream.map(
-            transfer_paths,
-            aiostream.async_(lambda paths: self.add_local_file(paths[0], paths[1], progress_cb)),
-            task_limit=20,
-        )
+        async with aclosing(sync_or_async_iter(gen_transfers())) as transfer_paths:
+            async with aclosing(
+                async_map(
+                    transfer_paths,
+                    lambda paths: self.add_local_file(paths[0], paths[1], progress_cb),
```
freider, line 349: An alternative to supporting blocking functions in the async_map (which I commented on above) would be to make sure we always pass it async functions instead, like the previous implementation did here. You could replace the lambda with an async func that just returns `self.add_local_file(paths[0], paths[1], progress_cb)` (this is the same as the `aiostream.async_`

freider (changes requested):
Had some comments, hope they fix the issues!

modal/network_file_system.py
```diff
@@ -343,12 +342,16 @@ def gen_transfers():
...
+                    concurrency=20,
+                )
+            ) as stream:
+                async for item in stream:
+                    await item
```
freider, line 354: The returned item will be the return value of the awaited `self.add_local_file` which is just an int we don't really care about - you can just iterate the stream and `pass` instead (to make it execute the full mapping)

modal/network_file_system.py
freider: I see now that `add_local_file` is actually a coroutine function, so this actually works now as is since you changed the `iscoroutinefunction` condition 👍

modal/_utils/async_utils.py
freider: Never mind, I actually think we should *not* allow sync methods to be passed in here, to make the caller responsible for making the conversion to async with the foot gun that is w.r.t. to blocking the event loop

modal/_utils/async_utils.py
```diff
@@ -544,3 +553,117 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+
+    async def worker():
+        while True:
+            try:
+                event_type, item = await input_queue.get()
+                if event_type == AsyncItemType.STOP:
+                    break
+                if event_type == AsyncItemType.VALUE:
+                    res = async_mapper_func(item)
+                    if inspect.isawaitable(res):
```
freider, line 580: I changed my mind - I think we should always assume `res` is awaitable and never allow blocking functions to be passed in here or to ordered_map_async - that would clean up a bunch of the code here, and I think we use async mapper functions in most cases anyways?

freider (changes requested):

test/async_utils_test.py
```diff
@@ -456,3 +458,58 @@ async def foo():
...
+            yield 2
+            yield 3
+        finally:
+            states.append("exit")
+
+    async def mapper(x):
+        await asyncio.sleep(0.1)  # Simulate some async work
+        return x * 2
+
+    async for item in async_map(foo(), mapper, concurrency=3):
```
freider, line 481: We need to extend the test suite for async_map to ensure the results actually come in as soon as they are ready. Turns out the current implementation has a bug (`output_event` is never set!) that cause no items to be returned until the full mapping is complete 😬

freider: Should probably extend the test suite for all of these replacement functions tbh, since they are pretty critical to the operations of the client.

Other things that came to mind that we need tests for is what happens if there are exceptions in the mapper function, if there are exceptions in the input iterators, or if the operation itself gets cancelled (as in asyncio.CancelledError).

We'll soon have a pretty useful and stable library we could extract and open source by itself :D

modal/_utils/async_utils.py
```diff
@@ -544,3 +553,117 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+
+
+async def async_map(
+    input_iterable: Union[AsyncIterable[T], Iterable[T]],
+    async_mapper_func: Callable[[T], Awaitable[V]],
+    concurrency: int,
+) -> AsyncGenerator[V, None]:
+    input_queue: asyncio.Queue[Tuple[str, Union[T, None]]] = asyncio.Queue()
+    output_queue: asyncio.Queue[Tuple[str, Union[V, Exception]]] = asyncio.Queue()
+    output_event = asyncio.Event()
```
freider, line 565: Just realized we never do `output_event.set()`! 💥 
Pretty bad story for the client test coverage that pretty much all tests pass anyways 😬 

Btw, I just realized that since we have to reset the `output_event.wait()` task in each iteration of the loop anyways, we might as well just listen for `output_queue.get()` instead there and skip this event completely

ekzhang:
Wow, asyncio in python is crazy hard.

kramstrom:
Closing in favor of https://github.com/modal-labs/modal-client/pull/2422

PR #2350, opened 2024-10-17
Don't use credentials in container

erikbern:
There's no need to read environment variables and set metadata in the container anymore – we inject it in the worker

SVC-164


mwaskom (approved):
😎

thundergolfer (approved):

erikbern:
Had to make some changes in the worker to the credentials injection logic, so I'm waiting for all workers to rotate until merging this.

erikbern:
Most workers seem to have rolled over so I'm merging this. If a new client hits an old worker, it will create some extraneous log messages, but I think that's acceptable.

PR #2351, opened 2024-10-17
Fix i6pn main address being None (typo)

thecodingwizard:
I forgot to add a `return` statement.

This wasn't caught in code review because I made the change after the PR was approved and somehow didn't catch the error myself :(

e2e tests failed on the main Modal repo, which is how I noticed this issue.

### DIFF
--- a/modal/experimental.py
+++ b/modal/experimental.py
@@ -159,7 +159,7 @@ def _networked(func):
 
         def get_i6pn():
             """Returns the ipv6 address assigned to this container."""
-            socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)[0][4][0]
+            return socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)[0][4][0]
 
         hostname = socket.gethostname()
         addr_info = get_i6pn()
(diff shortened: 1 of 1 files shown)


thecodingwizard:
@prbot approve

thecodingwizard:
@prbot approve

modal/experimental.py
```diff
@@ -159,7 +159,7 @@ def wrapper(*args, **kwargs):
 
         def get_i6pn():
             """Returns the ipv6 address assigned to this container."""
-            socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)[0][4][0]
+            return socket.getaddrinfo("i6pn.modal.local", None, socket.AF_INET6)[0][4][0]
```
luiscape, line 162:
```python
[0][4][0]
```
is very confusing to me. Could you create named variables? :)

thecodingwizard: I think there was some discussion [here](https://github.com/modal-labs/modal-client/pull/2332#discussion_r1800021843) about this. Personally I agree with Eric that `[0][4][0]` is something that can just be black-boxed (which was sort of the goal of the `get_i6pn()` function -- readers just need to know that the function returns the ipv6 address and don't need to know how it's done).

I'm open to other ideas though! In particular, I don't even know what to name the variables if I were to make each one a variable.

luiscape: Got it. Magic incantations without comments will likely trigger questions from every unfamiliar reader. Imo it's either "[literate programming](https://en.wikipedia.org/wiki/Literate_programming#:~:text=Literate%20programming%20(LP)%20tools%20are,woven%22%20from%20the%20literate%20source.)" or adding comments to the source.

PR #2352, opened 2024-10-18
Make it possible to stop apps programmatically

erikbern:
Really just moving some code from the cli package into the main code.

However I realized it's a bit janky because ideally you'd want this to be an instance method, and not a class method. But that would require better ways to construct an `App` object directly. Which we don't. We should address this though!

Also cleaned up some synchronicity stuff.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -1092,6 +1092,15 @@ class _App:
         """Only used for tests."""
         cls._container_app = None
 
+    @classmethod
+    async def stop(cls, app_id: str, client: Optional[_Client] = None):
+        """Stops an app programmatically."""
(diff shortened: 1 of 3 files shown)


mwaskom:
> But that would require better ways to construct an App object directly.

we have `App.lookup` now; do you mean something other than that?

erikbern:
> we have `App.lookup` now; do you mean something other than that?

Yeah the CLI lets you pass either an app name or an id – so we would need something to construct an app from an id.

Maybe I should just add that. It would be pretty similar to how `lookup` works.

mwaskom:
Closing this pending decision about how to spell the object management API.

PR #2354, opened 2024-10-18
Add _experimental_spawn function

rculbertson:
## Describe your changes

Adds a new `_experimental_spawn` function to the client which allows users to spawn up to 1 million inputs.

Previously users could enable this functionality by setting the env var `MODAL_SPAWN_EXTENDED`, but setting env vars correctly can be tricky. By adding a new method, we hope to make it more explicit and less error prone.

I've left the `MODAL_SPAWN_EXTENDED` mechanism in place since we've already communicated to some users that it exists, but can remove if that's preferred.

Closes SVC-169.

---

---

</details>

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -1401,11 +1401,8 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         Returns a `modal.functions.FunctionCall` object, that can later be polled or
         waited for using `.get(timeout=...)`.
         Conceptually similar to `multiprocessing.pool.apply_async`, or a Future/Promise in other contexts.
-
-        *Note:* `.spawn()` on a modal generator function does call and execute the generator, but does not currently
-        return a function handle for polling the result.
         """
-        self._check_no_web_url("spawn")
+        self._check_no_web_url("_experimental_spawn")
         if self._is_generator:
             invocation = await self._call_generator_nowait(args, kwargs)
         else:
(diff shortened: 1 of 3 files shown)


modal/functions.py
```diff
@@ -1410,7 +1433,13 @@ async def spawn(self, *args: P.args, **kwargs: P.kwargs) -> "_FunctionCall[Retur
         if self._is_generator:
             invocation = await self._call_generator_nowait(args, kwargs)
         else:
-            invocation = await self._call_function_nowait(args, kwargs)
+            function_call_invocation_type = (
```
aksh-at, line 1436: I don't think anyone is using this in production, and it's an experimental flag we only shared with one user, so we can drop support for the `config` environment variable now. That feels more awkward than using `_experimental_spawn` directly.

aksh-at (approved):
LGTM besides suggestion

modal/functions.py
```diff
@@ -1394,6 +1391,32 @@ async def coro():
...
+    async def _experimental_spawn(self, *args: P.args, **kwargs: P.kwargs) -> "_FunctionCall[ReturnType]":
+        """[Experimental] Calls the function with the given arguments, without waiting for the results.
+
+        This experimental version of the spawn method allows up to 1 million inputs to be spawned.
+
+        Returns a `modal.functions.FunctionCall` object, that can later be polled or
+        waited for using `.get(timeout=...)`.
+        Conceptually similar to `multiprocessing.pool.apply_async`, or a Future/Promise in other contexts.
+
+        *Note:* `.spawn()` on a modal generator function does call and execute the generator, but does not currently
```
freider, line 1405: Slightly unrelated but this note from the docstring of `spawn` is no longer true - we always return a FunctionCall object, even for generators (that can fetch results using `.get_gen()`), so we should remove this note (from spawn too)

modal/functions.py
```diff
@@ -1394,6 +1391,32 @@ async def coro():
...
+        return a function handle for polling the result.
+        """
+        self._check_no_web_url("spawn")
```
freider, line 1408: Nit: string input to _check_no_web_url should probably be `"_experimental_spawn"` here

freider (approved):
Some mini comments, but otherwise lgtm

modal/functions.py
rculbertson: ah good to know. I'll remove from both.

rculbertson:
@prbot automerge

PR #2355, opened 2024-10-18
Expire deprecation of `modal volume delete --confirm`

mwaskom:
## Changelog

- The previously-deprecated `--confirm` flag has been removed from the `modal volume delete` CLI. Use `--yes` to force deletion without a confirmation prompt.


mwaskom:
@prbot lgtm

devennavani (commented):
LGTM

PR #2357, opened 2024-10-18
Remove some cruft from name_utils

mwaskom:
Removes an unused deprecation warning and an old name that we previously referenced in the server.


mwaskom:
@prbot approve

kramstrom (commented):
👍

PR #2353, opened 2024-10-18
Expire the deprecation of `web_endpoint(wait_for_response=False)`

mwaskom:
This was deprecated >5 months ago.

## Changelog

- Passing `wait_for_response=False` with `modal.web_endpoint` is no longer supported. See [the docs](https://modal.com/docs/guide/webhook-timeouts#polling-solutions) for alternatives.

PR #2358, opened 2024-10-18
Expire wait_for_response in other webhook decorators

mwaskom:
Followup to #2353 — I missed that we had the same (deprecated) parameter in the other webhook decorators. The "web endpoint" vs "webhook" distinction is always confusing! (Also the string concatenation hid it from the `inv show-deprecations` helper). My bad anyway.


devennavani (approved):
LGTM

PR #2359, opened 2024-10-18
Add split output fields

azliu0:
Adds protobuf fields for handling output split on the worker side instead of server side

## Changelog

Adds protobuf fields for handling output split on the worker side instead of server side

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1823,9 +1823,11 @@ message RuntimeOutputBatch {
   uint64 batch_index = 2;
   // if an exit code is given, this is the final message that will be sent.
   optional int32 exit_code = 3;
+  repeated RuntimeOutputMessage stdout = 4;
+  repeated RuntimeOutputMessage stderr = 5;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1823,9 +1823,11 @@ message RuntimeOutputBatch {
   uint64 batch_index = 2;
   // if an exit code is given, this is the final message that will be sent.
   optional int32 exit_code = 3;
+  repeated RuntimeOutputMessage stdout = 4;
+  repeated RuntimeOutputMessage stderr = 5;
+  repeated RuntimeOutputMessage info = 6;
 }
 
-// Used for `modal container exec`
```
pawalt, line 1828: don't removed this message

pawalt (approved):

mwaskom:
FYI the changelog is public facing and the protos aren't public — you only need to add that section to the PR description when the PR makes a change that is relevant to end users (and should be written with the user perspective in mind).

PR #2361, opened 2024-10-18
Require credentials for non-container clients

erikbern:
~This needs to be rebased on #2350 when merged.~

We now fail immediately in `Client.from_env` if it's not `_is_remote()` and there are no credentials. This reduces the need for `ClientHello`.

We have a separate class method for creating "unauthenticated" clients, but it's only used for `modal token new`.

This mostly required fixing a bunch of tests that previously relied on the test fixture client having no credentials.

Cleans things up to the point where `ClientHello` can be trivially removed from containers. But I'll do that separately.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -225,35 +225,36 @@ class _Client:
         else:
             c = config
 
-        if _is_remote():
-            credentials = None
-            client_type = api_pb2.CLIENT_TYPE_CONTAINER
-        else:
-            token_id = c["token_id"]
-            token_secret = c["token_secret"]
-            if not token_id or not token_secret:
-                raise AuthError(
-                    "Token missing. Could not authenticate client."
-                    " If you have token credentials, see modal.com/docs/reference/modal.config for setup help."
-                    " If you are a new user, register an account at modal.com, then run `modal token new`."
-                )
-            credentials = (token_id, token_secret)
-            client_type = api_pb2.CLIENT_TYPE_CLIENT
-
         if cls._client_from_env_lock is None:
             cls._client_from_env_lock = asyncio.Lock()
 
         async with cls._client_from_env_lock:
             if cls._client_from_env:
                 return cls._client_from_env
+
+            server_url = c["server_url"]
+
+            if _is_remote():
+                credentials = None
+                client_type = api_pb2.CLIENT_TYPE_CONTAINER
             else:
-                server_url = c["server_url"]
-                client = _Client(server_url, client_type, credentials)
-                await client._open()
-                async_utils.on_shutdown(client._close())
-                await client._hello()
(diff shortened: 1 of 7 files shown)


modal/client.py
```diff
@@ -226,44 +225,36 @@ async def from_env(cls, _override_config=None) -> "_Client":
...
 
         async with cls._client_from_env_lock:
             if cls._client_from_env:
                 return cls._client_from_env
+
+            server_url = c["server_url"]
+
+            if _is_remote():
+                credentials = None
+                client_type = api_pb2.CLIENT_TYPE_CONTAINER
```
mwaskom, line 239: I wonder if we're setting ourselves up for (more) confusion about the two distinct meanings of the local/remote dichotomy in Modal:

- practical sense:
  - local: not on modal infrastructure
  - remote: on modal infrastructure
- functional sense:
  - local: the context that you run or deploy your applications _from_
  - remote: the context that you run or deploy your applications _in_

(This is maybe not a perfect set of definitions but hopefully it makes sense)

The `_is_remote` helper indicates which side you're on in the practical sense (based on the environment variables that we define in containers).

But what about users who want to run or deploy applications from Modal infrastructure? I thought we had people doing that; including people doing it on behalf of their users (such that the orchestrator app and worker app are using different tokens). Are we breaking that behavior? Maybe we're not — but it feels very hard to reason about.

mwaskom (commented):
Had a few thoughts — which might be totally off base — but the rules about what the Client can and should do feel sort of poorly defined in a way that makes it hard for me to predict the repercussions of this change.

modal/client.py
```diff
@@ -127,7 +127,7 @@ def is_closed(self) -> bool:
 
     @property
     def authenticated(self):
-        return self._authenticated
+        return self._credentials is not None
```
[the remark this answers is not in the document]
erikbern: So yeah I think this is actually wrong (and let me try to add a test for it) but for a different reason I think. If a Modal app running in the cloud is trying to deploy a Modal app then I think it asserts that the client is authenticated, which will fail. Honestly think it's maybe better to remove this code altogether and just fail in respective handlers – I'm not sure why we need this check.

modal/client.py
erikbern: Let me add a test for deploying Modal apps from inside containers. It's possible this PR broke it because of the `authenticated` check. That being said, the authentication system (client vs container vs web) is basically orthogonal to what we're doing (creating an app vs running code inside an app).

erikbern:
On second thought I think this change conflates a bunch of logic changes with a bunch of test fixture stuff. I'm going to close this one and try to focus on just fixing tests first. That should make it possible to change the actual logic separately from the tests.

PR #2362, opened 2024-10-20
Miscellaneous cleanup of `@app.cls` code

devennavani:
Mostly nit-picks but miscellaneous cleanup of `@app.cls` code:

* Move initialization and validation of `scheduler_placement` outside of the `wrapper` function -- `@app.cls()` parameter validation can live outside of the decorator function
* co-locate class method validation logic (batched methods and snap=True methods)

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -879,6 +879,7 @@ class _App:
         if _warn_parentheses_missing:
             raise InvalidError("Did you forget parentheses? Suggestion: `@app.cls()`.")
 
+        # Argument validation
         if interactive:
             deprecation_error(
                 (2024, 5, 1), "interactive=True has been deprecated. Set MODAL_INTERACTIVE_FUNCTIONS=1 instead."
--- a/modal/app.py
+++ b/modal/app.py
@@ -951,10 +951,6 @@ class _App:
                 scheduler_placement=scheduler_placement,
                 _experimental_buffer_containers=_experimental_buffer_containers,
                 _experimental_proxy_ip=_experimental_proxy_ip,
-                # class service function, so the following attributes which relate to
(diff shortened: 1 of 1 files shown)


modal/app.py
```diff
@@ -879,15 +879,17 @@ def cls(
...
             deprecation_error(
                 (2024, 5, 1), "interactive=True has been deprecated. Set MODAL_INTERACTIVE_FUNCTIONS=1 instead."
             )
 
-        if image is None:
-            image = self._get_default_image()
-
-        secrets = [*self._secrets, *secrets]
+        scheduler_placement = _experimental_scheduler_placement
+        if region:
```
freider, line 889: It would probably make sense to move out argument validation/auto-fixing into its own helper function(s) that can be used by both function, cls and Sandbox.create, but lets do that separately :)

freider (approved):
lgtm!

modal/app.py
```diff
@@ -879,15 +879,17 @@ def cls(
...
         if interactive:
             deprecation_error(
                 (2024, 5, 1), "interactive=True has been deprecated. Set MODAL_INTERACTIVE_FUNCTIONS=1 instead."
             )
 
-        if image is None:
-            image = self._get_default_image()
-
-        secrets = [*self._secrets, *secrets]
+        scheduler_placement = _experimental_scheduler_placement
```
mwaskom, line 888: BTW should we deprecate `_experimental_scheduler_placement`? Is it supplanted by the stable `region=` API now?

devennavani: I don't believe so, since `_experimental_scheduler_placement` allows you to specify `zone`, `spot`, and `instance_type` in addition to `region`.

PR #2366, opened 2024-10-21
Fixes `Mount.from_local_python_packages` not working for packages in .-prefixed ancestor dirs

freider:
This fixes the "common" issue of trying to mount a package installed into `.venv/.../site-packages` which would previously be omitted due to the leading `.` in the `.venv` dir. 

After the fix, `.`-prefixed dirs are still excluded by default when mounting packages, but only as long as the `.`-prefixed directory is underneath the top level of the specified package.

Also adds some more detailed debug logging for mounts in general, letting users find out which files get added where in mounts when using MODAL_LOGLEVEL=DEBUG during loading.

---

* This does change the interface for `module_mount_condition` - perhaps it should be deprecated instead and the new one should be made private?

---

</details>

## Changelog

* Fixes bug where `.`-prefixed ancestor directories of packages prevented those packages from being mounted without a custom condition argument

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -155,13 +155,32 @@ class _MountDir(_MountEntry):
         return [(self.local_dir, self.remote_path)]
 
 
-def module_mount_condition(f: str):
-    path = Path(f)
-    if path.suffix == ".pyc":
-        return False
-    if any(p.name.startswith(".") or p.name == "__pycache__" for p in path.parents):
-        return False
-    return True
+def module_mount_condition(module_base: Path):
+    SKIP_BYTECODE = True  # hard coded for now
+    SKIP_DOT_PREFIXED = True
+
+    def condition(f: str):
+        path = Path(f)
+        if SKIP_BYTECODE and path.suffix == ".pyc":
+            return False
+
+        # Check parent dir names to see if file should be included,
+        # but ignore dir names above root of mounted module:
+        # /a/.venv/site-packages/mymod/foo.py should be included by default
+        # /a/my_mod/.config/foo.py should *not* be included by default
+        while path != module_base and path != path.parent:
+            print(path, flush=True)
+            if SKIP_BYTECODE and path.name == "__pycache__":
+                return False
+
+            if SKIP_DOT_PREFIXED and path.name.startswith("."):
+                return False
+
+            path = path.parent
+
+        return True
+
+    return condition
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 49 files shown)


freider:
@gimaker This should fix the longstanding issue you've had with MODAL_SYNC_ENTRYPOINT - let me know if it doesn't!

modal/mount.py
```diff
@@ -406,9 +424,10 @@ async def _get_files(entries: List[_MountEntry]) -> AsyncGenerator[FileUploadSpe
 
             futs = []
             for local_filename, remote_filename in all_files:
+                logger.debug(f"Mounting {local_filename} as {remote_filename}")
```
mwaskom, line 427: Could this potentially make the debug logs overwhelmingly verbose for users mounting large dependencies?

modal/mount.py
```diff
@@ -670,19 +691,19 @@ def _get_client_mount():
     )
 }
 
+
 SYS_PREFIXES |= {p.resolve() for p in SYS_PREFIXES}
 
+MODAL_PACKAGES = ["modal", "modal_proto", "modal_version"]
```
mwaskom, line 697: We should fix this so that the client distribution contains only one package btw, would be more compatible with PEP517.

modal/mount.py
```diff
@@ -670,19 +691,19 @@ def _get_client_mount():
...
     for base in remote_python_paths:
-        is_modal_path = path_prefix in [
-            base + ("modal",),
-            base + ("modal_proto",),
-            base + ("modal_version",),
-            base + ("synchronicity",),
-        ]
+        is_modal_path = path_prefix in [base + (mod,) for mod in MODAL_PACKAGES] or path_prefix == base + (
+            "synchronicity",
+        )
```
mwaskom, line 706: Poor synchronicity — not considered part of the family, but doesn't get its own pip downloads.

mwaskom (commented):
Does this change automount behavior? We currently document that as excluding virtual environments. Not sure why I can't figure out if this is a behavioral change.

modal/mount.py
freider: They already have every MountPutFile RPC call logged, so at most this is a 2x increase, but the "metadata" is way more useful than the RPC call here

modal/mount.py
freider: Yeah for sure

modal/mount.py
freider: 😢

freider:
> Does this change automount behavior? We currently document that as excluding virtual environments. Not sure why I can't figure out if this is a behavioral change.

It shouldn't change the automount behavior - that exclusion should still be valid based on the `SYS_PREFIXES` exclusion logic in `get_auto_mounts` (virtualenvs with non-`.`-prefixes were excluded prior to this patch too)

freider:
@prbot approve

freider:
@prbot automerge

gimaker (commented):
LGTM

PR #2367, opened 2024-10-21
[WRK-200] memory snapshot causes clientclosed error for webapp

thundergolfer:
## Describe your changes

- WRK-200

https://github.com/modal-labs/modal-client/pull/2178 introduced a regression in snapshots that wasn't caught by our tests. PR 16550 in the monorepo adds a regression integration test.

`UnaryUnaryWrapper` and `UnaryStreamWrapper` began capturing references to clients which became stale on snapshot. When these stale and _closed_ snapshots were used on restore exceptions we thrown. 

---

---

</details>

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -148,7 +148,7 @@ class _Client:
         self._owner_pid = os.getpid()
 
     async def _close(self, prep_for_restore: bool = False):
-        logger.debug(f"Client({id(self)}): closing")
+        logger.debug(f"Client ({id(self)}): closing")
         self._closed = True
         await self._cancellation_context.__aexit__(None, None, None)  # wait for all rpcs to be finished/cancelled
         if self._channel is not None:
@@ -162,7 +162,7 @@ class _Client:
 
     async def _init(self):
         """Connect to server and retrieve version information; raise appropriate error for various failures."""
-        logger.debug("Client: Starting")
+        logger.debug(f"Client ({id(self)}): Starting")
         _check_config()
         try:
             req = empty_pb2.Empty()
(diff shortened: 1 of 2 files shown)


modal/client.py
```diff
@@ -161,7 +162,7 @@ async def _close(self, prep_for_restore: bool = False):
 
     async def _init(self):
         """Connect to server and retrieve version information; raise appropriate error for various failures."""
-        logger.debug("Client: Starting")
+        logger.debug(f"Client ({id(self)}): Starting")
```
thundergolfer, line 165: Having the object ID logged helps track the usage of fresh vs. stale clients when debugging snapshot issues. 

Instead of having to catch stale `client` objects and refresh them it'd be better if the client object itself could catch that it was stale and refresh itself. If this could work then we could remove all current (and future) `if self.client._snapshotted` type checks.

thundergolfer:
@prbot approve

modal/client.py
```diff
@@ -426,5 +430,8 @@ async def unary_stream(
         request,
         metadata: Optional[Any] = None,
     ):
+        if self.client._snapshotted:
+            logger.debug(f"refreshing client after snapshot for {self._wrapped_method_name}")
+            self.client = await _Client.from_env
```
freider, line 435: from_env is a method that needs to be called (`()`) so this probably crashes 😬  - can we add a test that covers this case?

modal/client.py
freider: Yeah I agree, we should make the client itself detect this, which should be doable through the `_call_unary` and `_call_stream` methods on it which all RPC methods should be going via now

freider (commented):
found a bug 💥

PR #2368, opened 2024-10-21
Strict credential authentication in tests

erikbern:
This makes the credential authentication in tests a lot more strict – it checks the token and secret on every request. Basically what the real production API already does.

This change will make it easier to remove `ClientHello` while retain tests for most/all of the logic (since we're going to rely on grpc errors from other routes instead).

Broken out of #2361 which conflated a bunch of different changes. Better to make tests stricter first.

Ended up being a much bigger change than I anticipated (lots of test that relied on no authentication).

Open question: should we create a random token for _each test_? That wouldn't be significantly harder and maybe it's an even stronger test.

### DIFF
--- a/test/cli_test.py
+++ b/test/cli_test.py
@@ -113,14 +113,14 @@ def test_secret_list(servicer, set_env_client):
 
 
 def test_app_token_new(servicer, set_env_client, server_url_env, modal_config):
-    servicer.required_creds = ("abc", "xyz")
+    servicer.required_creds = {("abc", "xyz")}
     with modal_config() as config_file_path:
         _run(["token", "new", "--profile", "_test"])
         assert "_test" in toml.load(config_file_path)
 
 
 def test_app_setup(servicer, set_env_client, server_url_env, modal_config):
-    servicer.required_creds = ("abc", "xyz")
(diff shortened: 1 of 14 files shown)


mwaskom (approved):
One thing that isn't obvious to me here is when we need the `token_env` fixture.

> Open question: should we create a random token for each test? That wouldn't be significantly harder and maybe it's an even stronger test.

Yeah — this feels like it might be worth doing? We've had weird state-spillover issues with the mocking in the past that have caused and/or masked issues.

erikbern:
> One thing that isn't obvious to me here is when we need the `token_env` fixture.

Mostly when we have subprocesses (since they inherit env from the parent) but arguably it would be cleaner to make that explicit

> 
> > Open question: should we create a random token for each test? That wouldn't be significantly harder and maybe it's an even stronger test.
> 
> Yeah — this feels like it might be worth doing? We've had weird state-spillover issues with the mocking in the past that have caused and/or masked issues.

Let me see if it's easy to do. There's a risk it's hard but I'll try in a separate PR.

erikbern:
Thanks @mwaskom

PR #2370, opened 2024-10-21
Move the authentication check during running/deploy earlier

erikbern:
Broke this out of https://github.com/modal-labs/modal-client/pull/2368 because it's the only change outside of tests.

The reason it's needed is that we create the app before this check right now, but with the stricter middleware, the app creation fails. Moving this check earlier catches things properly.

Might remove this logic very soon though!

### DIFF
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -129,9 +129,6 @@ async def _create_all_objects(
     environment_name: str,
 ) -> None:
     """Create objects that have been defined but not created on the server."""
-    if not client.authenticated:
-        raise ExecutionError("Objects cannot be created with an unauthenticated client")
-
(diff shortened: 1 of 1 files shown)


erikbern:
@prbot approve

modal/runner.py
```diff
@@ -502,6 +503,9 @@ async def _deploy_app(
     if client is None:
         client = await _Client.from_env()
 
+    if not client.authenticated:
+        raise ExecutionError("Objects cannot be created with an unauthenticated client")
```
mwaskom, line 507: The copy-paste made me wonder if we should have a `client.check_authenticated()` or something so the error types/ messages don't drift over time, but probably not worth it.

PR #2372, opened 2024-10-22
Use unique server credentials in every unit test

erikbern:
Instead of hardcoded `("ak-123", "as-123")`, use a unique one for each test. This ensures we don't have state leakage between tests. Almost disappointed it didn't catch anything.

Was easier than I thought. Came out of https://github.com/modal-labs/modal-client/pull/2368

### DIFF
--- a/test/app_composition_test.py
+++ b/test/app_composition_test.py
@@ -2,8 +2,8 @@
 from test.helpers import deploy_app_externally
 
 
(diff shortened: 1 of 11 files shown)


erikbern:
@prbot approv

mwaskom (commented):
Great!

PR #2373, opened 2024-10-22
use a container client for test_container_debug_snapshot

erikbern:
super minor fix, use the right client type


erikbern:
@prbot approv

mwaskom (commented):
:+1:

PR #2374, opened 2024-10-22
use credentials iff not `_is_remote`

erikbern:
This just couples the cases:
* inside the container: never use credentials
* outside: always use credentials (there's no longer optionality)

The point is to remove the dependency on `ClientHello`


mwaskom:
I think this is mostly a gap on my side, but I'm still not quite sure what the defined behavior is going to be for users creating a client using `Client.from_credentials` in a remote environment. Otherwise looks good.

mwaskom (approved):

erikbern:
Thanks @mwaskom!

PR #2375, opened 2024-10-22
Don't explicity check that the client is authenticated

erikbern:
We don't need an explicit check for this – the GRPC routes will fail properly anyway. And I can't think of a single realistic scenario where this could actually happen.

This helps us remove `ClientHello`


mwaskom (approved):
Makes sense.

PR #2381, opened 2024-10-22
Move some logic from `ClientHello` into `grpc_utils`

erikbern:
Requesting feedback on this change. I want to remove the need for `ClientHello` so as a part of that change, I moved some logic into `grpc_utils` for handling common errors in `retry_transient_errors` instead.

The end goal here is to get rid of `ClientHello`. We currently raise a few exceptions like `AuthError` and `ConnectionError` during that operation, and I don't want to break any user who catch those exceptions. Once we remove `ClientHello`, those exceptions will be raised in _other_ RPC calls, but it seems nice to retain the exception type.

Moving logic this way is nice because I only had to update two tests! I'm treating the number of test changes as an arguable proxy for what how much user code we could expect to break. My expectation is that very few customers rely on these custom exception types though.

### Slightly more general question about mapping grpc error codes in `grpc_utils`

We have a lot of code in the client of the form

```python
try:                                             
    resp = await retry_transient_errors(self._client.stub.VolumeCommit, req, max_retries=90)
except GRPCError as exc:
    raise RuntimeError(exc.message) if exc.status in (Status.FAILED_PRECONDITION, Status.NOT_FOUND) else exc
```

I tend to think it would be nice if `retry_transient_errors` could do _more_ of the error mapping but I think there's arguments on both sides. This change adds a handler for one particular status code (`UNAUTHENTICATED` so it's a step in that direction).

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -25,6 +25,7 @@ from grpclib import GRPCError, Status
 from grpclib.exceptions import StreamTerminatedError
 from grpclib.protocol import H2Protocol
 
+from modal.exception import AuthError, ConnectionError
 from modal_version import __version__
 
 from .logger import logger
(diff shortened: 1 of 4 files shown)


modal/_utils/grpc_utils.py
```diff
@@ -170,16 +171,27 @@ async def retry_transient_errors(
             timeout = None
         try:
             return await fn(*args, metadata=metadata, timeout=timeout)
-        except (StreamTerminatedError, GRPCError, socket.gaierror, asyncio.TimeoutError, AttributeError) as exc:
+        except (StreamTerminatedError, GRPCError, OSError, asyncio.TimeoutError, AttributeError) as exc:
```
erikbern, line 174: `OSError` is an ancestor of `socket.gaierror`

freider:
I think this makes sense. I assume we put it in retry_transient_errors to avoid having to change call sites to chain additional methods onto it, but once we move that logic into UnaryUnaryWrapper and similar (and get rid of retry_transient_errors) we could separate the concerns of mapping exceptions from retrying exceptions a bit to make the code a bit clearer

freider (approved):

erikbern:
> once we move that logic into UnaryUnaryWrapper and similar (and get rid of retry_transient_errors) we could separate the concerns of mapping exceptions from retrying exceptions a bit to make the code a bit clearer

Yeah I'd love to not have to explicitly use `retry_transient_errors` in our code and just do it automatically through the wrappers. We can move the exception handling into those as well. I'll take a look.

erikbern:
Danke schön @freider

PR #2382, opened 2024-10-22
Send terminated output when canceling a concurrent input.

danielshaar:
The worker will ignore the terminated outputs for now. We should wait a day before releasing this to ensure all workers have rolled out with the ignoring logic.

We don't need to send these messages if only one input is running on a container.

Context in WRK-11.


freider:
> We don't need to send these messages if only one input is running on a container.
Why not? Concurrency=1-containers also need to escalate if they don't respond gracefully to single input cancellations, and I think it would be nice if the worker can always assume getting a FunctionPutOutputs for each input unless something bad happens

freider:
We might also want to add some assertions in relevant container_tests (there are plenty for various different cancellation conditions) that `FunctionPutOutputs` is sent

modal/_container_io_manager.py
```diff
@@ -401,11 +401,15 @@ async def _heartbeat_handle_cancellations(self) -> bool:
...
-                        if input_id in self.current_inputs:
-                            self.current_inputs[input_id].cancel()
+                    current_input_ids_to_cancel = [
+                        input_id for input_id in input_ids_to_cancel if input_id in self.current_inputs
+                    ]
+                    for input_id in current_input_ids_to_cancel:
+                        self.current_inputs[input_id].cancel()
 
-                elif self.current_input_id in input_ids_to_cancel:
+                    await self._push_terminated_outputs(current_input_ids_to_cancel)
```
freider, line 410: I think this would never get called in case of non-async concurrent inputs, since that `.cancel()` handler on the line before will self-sigint the process.

If you add an output item assertion on the `test_cancellation_stops_task_with_concurrent_inputs` test you should be able to detect this

freider: Oh wait, you already added that assertion. Now I'm confused how this can work 🤔

freider: Ah, it's apparently a thread race. The sigint will trigger in the main thread, whereas this code runs in the synchronicity thread, so this will *typically* have time to trigger before the sigint unwinds the main thread and exits the interpreter, but adding a little delay here makes the test fail. Would recommend moving the `push_terminated_outputs`.

A few lines above this there is also another case to handle (which we don't appear to have tests for):
```
if response.cancel_input_event.terminate_containers:
    os.kill(os.getpid(), signal.SIGINT)
```
An alternative to sending terminated outputs in this case would be to simply handle this in the worker instead (since this is basically the "escalated" cancellation mode but explicitly triggered by the cancellation requester)

danielshaar: Moved the `push_terminated_outputs`. w.r.t the `terminate_containers` case, do I have to push terminated outputs if the whole thing is getting killed anyway?

freider: I think we still have to push them so that the worker knows the container has responded to the cancellation

freider: We could also just move the `terminate_containers` logic to the worker by having workers sigint the container instead of passing those cancellations down to the container

danielshaar: Ah okay, I'll return the outputs for now and refactor the worker to terminate the containers in a follow up later.

danielshaar: After further discussion, I'm convinced the worker should handle terminate_containers, so I'll delete the handling here, but won't submit until the worker does that.

danielshaar: As it turns out, the server already handles terminate_containers. The worker / client never actually receive terminate_containers = True (maybe it shouldn't be in the proto 🤷 )

freider (approved):
lgtm!

PR #2383, opened 2024-10-23
Fix what I think is a client bug

erikbern:
This looks like a bug to me – closing _any_ client resets the singleton.

Although the only impact is if you close clients then we'll recreate new clients unnecessarily. So fairly low impact.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -152,9 +152,6 @@ class _Client:
         if prep_for_restore:
             self._snapshotted = True
 
-        # Remove cached client.
-        self.set_env_client(None)
-
     async def _init(self):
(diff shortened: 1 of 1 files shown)


erikbern:
Hm I actually wonder if what we want to do is to remove the env client if that's the client that's being closed. I.e. something like this

```python
if self is _Client._client_from_env:
    self.set_env_client(None)
```

Let me make that change!

thundergolfer:
Yep that makes sense @erikbern. Alternatively the `memory_snapshot()` function can just call `set_env_client` directly? I think it's the only consumer of this behavior.

PR #2384, opened 2024-10-23
Static types: fix @exit decoration of async methods indicating a type error

freider:
Quick and slightly ugly fix for static type errors when decorating async methods with `@modal.exit`

I was pretty sure I looked at this a good while back and we had already fixed this but apparently not 🤷


freider:
@prbot approve

mwaskom (commented):
I also thought we'd dealt with this previously 🤔

PR #2385, opened 2024-10-23
`_check_config` in `Client.from_env`

erikbern:
We shouldn't run `_check_config` in `Client.hello` but rather when we're reading config data – in `Client.from_env`

Had to fix a few tests – turns out at least one of them wasn't testing what it was supposed to (I think I accidentally caused that in an earlier change)

The point of this is to remove the need for `ClientHello`

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -138,7 +138,6 @@ class _Client:
     async def hello(self):
         """Connect to server and retrieve version information; raise appropriate error for various failures."""
         logger.debug(f"Client ({id(self)}): Starting")
-        _check_config()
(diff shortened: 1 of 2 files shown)


erikbern:
@prbot approve

freider:
Feels to me like `_check_config()` shouldn't even be used in the Client class - it should be used immediately when reading the config from disk and that's what these tests should test. The `from_env(_config_override=...)` argument seems like it has been superseded by the `with modal_config` context manager which also reads config from disk (thereby triggering _check_config if we move it there)

PR #2386, opened 2024-10-23
Keep supported Python series in expected order

mwaskom:
@charlesfrye pointed out that these show up in a random order when we print an error message about an unsupported version, since we use a set type. The supported version lookup is not performance critical, so it's easier to just store them in a list than to deal with the correct sorting at error time.


mwaskom:
@prbot approve

charlesfrye (approved):
nice

PR #2389, opened 2024-10-23
Allow token auth inside container

ekzhang:
Follow-up to #2374 that fixes a minor regression that breaks the https://modal.com/playground app. This adds a test that was failing previously and succeeds with the change.

```
pytest test/client_test.py -k test_client_token_auth_in_container
# 1 passed, 18 deselected in 0.04s

git stash push modal/client.py
pytest test/client_test.py -k test_client_token_auth_in_container
# 1 failed, 18 deselected in 0.10s
```

## Changelog

- Fix a regression introduced in client version 0.64.209, which affects client authentication within a container.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -206,20 +206,20 @@ class _Client:
             if cls._client_from_env:
                 return cls._client_from_env
 
-            if _is_remote():
+            token_id = c["token_id"]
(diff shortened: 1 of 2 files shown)


erikbern (approved):

kning:
🎉 🎉

PR #2391, opened 2024-10-23
Don't print token_secret from test fixture

erikbern:
Simplified the test code a bit by making `required_creds` a dict rather than a set of tuples


thundergolfer (approved):
LGTM, cheers

PR #2392, opened 2024-10-23
Clear config better

erikbern:
Turns out it would still pick up the local toml because of a funny interaction where another fixture would change the dict, so this fixture would clear an old unused dict.


erikbern:
@prbot approve

thundergolfer (commented):
LGTM

PR #2395, opened 2024-10-24
TaskList RPC takes environment filter

kramstrom:
Adds an environment filter to `modal container list`

server pr #16697


mwaskom (approved):

mwaskom:
Oops we maybe should have added a changelog update here in case this surprises anyone.

kramstrom:
> Oops we maybe should have added a changelog update here in case this surprises anyone.

Ah yeah I always forget that, I can make a PR

PR #2397, opened 2024-10-24
MOD-2454: update async utils

kramstrom:
Split out non-`async_map` changes from https://github.com/modal-labs/modal-client/pull/2349 to a separate PR


freider (approved):
lgtm! Maybe add a comment in the merge commit message about what it actually does:
* make async_zip statically type safe
* ensure we clean up active tasks when exiting async_zip
* clean up combined context manager syntax
* add some more tests

PR #2399, opened 2024-10-24
Stream container process logs in the background

pawalt:
## Describe your changes

Container process logs are different from Sandbox logs in that they must be consumed by the client or else they block the writing process. This means that if you blindly exec or exec only consuming one stream, the process can hang with no indication to the user. This commit adds background streaming for container process logs, allowing exec processes to complete without the client explicitly reading them.

Resolves WRK-422
Resolves WRK-413

---

---

</details>

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
 
 async def _sandbox_logs_iterator(
     sandbox_id: str, file_descriptor: int, last_entry_id: Optional[str], client: _Client
-) -> AsyncIterator[Tuple[Optional[api_pb2.TaskLogs], str]]:
+) -> AsyncIterator[Tuple[Optional[str], str]]:
     req = api_pb2.SandboxGetLogsRequest(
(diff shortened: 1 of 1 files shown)


modal/io_streams.py
```diff
@@ -124,6 +128,54 @@ async def read(self) -> str:
...
+            try:
+                iterator = _container_process_logs_iterator(self._object_id, self._file_descriptor, self._client)
+
+                async for message in iterator:
+                    self._container_process_buffer.append(message)
+                    if message is None:
+                        completed = True
+                        break
+
+            except (GRPCError, StreamTerminatedError) as exc:
```
pawalt, line 147: I don't love the retry code duplication here, but I couldn't figure out how to get rid of it. I almost got a helper function working, but the signature didn't work out since one of these yields and the other doesn't.

azliu0 (approved):
LGTM forgot to press submit sorry for the wait

pawalt:
I'm not going to merge this and going to instead build subprocess-style pipes

PR #2401, opened 2024-10-24
Temporary fix for token issue

erikbern:
If `MODAL_TOKEN_ID` and `MODAL_TOKEN_SECRET` is set inside a container, ignore it in `.from_env`

This relies on the existence of a the `MODAL_TASK_SECRET` env var, which is being phased out – the reason it's needed it's it's set in "Modal containers" but not in sandboxes (unlike `MODAL_IS_REMOTE` which is set in both).

I'm going to roll out a worker change that adds a new env var, which will let me remove the use of `MODAL_TASK_SECRET`. But it will take a few days for that to be done.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -31,7 +31,7 @@ from modal_version import __version__
 from ._utils import async_utils
 from ._utils.async_utils import TaskContext, synchronize_api
 from ._utils.grpc_utils import create_channel, retry_transient_errors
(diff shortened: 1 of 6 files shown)


erikbern:
@prbot approve

ekzhang:
Thanks! Looks good to me, playground should still work.

PR #2403, opened 2024-10-25
[wip] Add support for client retries

rculbertson:


modal/functions.py
```diff
@@ -178,6 +210,21 @@ async def pop_function_call_outputs(
                     # return the last response to check for state of num_unfinished_inputs
                     return response
 
+    async def retry_input(self) -> None:
+        if not self._function_call_jwt or not self._input_id or not self._item:
+            raise ValueError("_function_call_jwt, _response_item, and _item must not be empty when retrying an input")
+
+        item = api_pb2.FunctionRetryInputsItem(idx=self._item.idx, input_id=self._input_id, input=self._item.input)
+        request = api_pb2.FunctionRetryInputsRequest(function_call_jwt=self._function_call_jwt, inputs=[item])
```
gongy, line 218: I might be wrong, but I think the `input_id` must be part of the JWT for true validation.

modal/functions.py
```diff
@@ -132,7 +152,16 @@ async def create(
         function_call_id = response.function_call_id
 
         if response.pipelined_inputs:
-            return _Invocation(client.stub, function_call_id, client)
+            input = response.pipelined_inputs[0]
```
freider, line 155: Can there ever be more than one pipelined input, I assume no, but could we add an assertion that the response will not have more than one in this case?

rculbertson: Yea, good point. Will add!

rculbertson: Added.

gongy (approved):
Sorry for taking a while. This is fantastic work @rculbertson and @rohansingh.

PR #2404, opened 2024-10-25
Warn if token id+secret are ignored in a container

erikbern:
Follow up to #2401 – I was thinking more about it and I think this is a bit of a weird case that at least deserves a warning if people try it. Maybe they think the container will inherit the creds or something.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -211,7 +211,10 @@ class _Client:
             token_secret = c["token_secret"]
             if task_secret:
                 if token_id or token_secret:
-                    warnings.warn("Modal token id and secret are ignored inside containers")
+                    warnings.warn(
+                        "Modal tokens provided by MODAL_TOKEN_ID and MODAL_TOKEN_SECRET"
+                        " (or through the config file) are ignored inside containers."
+                    )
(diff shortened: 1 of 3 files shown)


modal/client.py
```diff
@@ -210,6 +210,8 @@ async def from_env(cls, _override_config=None) -> "_Client":
             token_id = c["token_id"]
             token_secret = c["token_secret"]
             if task_secret:
+                if token_id or token_secret:
+                    warnings.warn("Modal token id and secret are ignored inside containers")
```
mwaskom, line 214: Should we clarify that these are the one in environment variables?

mwaskom: We currently don't read `.modal.toml` in the container: https://github.com/modal-labs/modal-client/blob/f804d0278fac7d3399bef8f3e87529ec1301488a/modal/config.py#L106-L124

Although I guess that's a different way that we might confusingly ignore tokens!

erikbern: Nice, maybe I overreacted updating the warning message then! It's probably fine to err on the side of being extra verbose

mwaskom (approved):

PR #2406, opened 2024-10-25
Update tasklist cli command changelog

kramstrom:
Adds changes in https://github.com/modal-labs/modal-client/pull/2395#issuecomment-2437852047 (`modal container list` filtering by environment) to the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ We appreciate your patience while we speedily work towards a stable release of t
 
 ### 0.64.227 (2024-10-25)
 
-- The `modal container list` CLI command now only shows containers for your active environment. You can pass `--env` to list containers in other environments.
+- The `modal container list` CLI command now only shows containers for the active profile's environment if there is one, otherwise it uses the default environment. You can pass `--env` to list containers in other environments.
 
 ### 0.64.223 (2024-10-24)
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ We appreciate your patience while we speedily work towards a stable release of t
 
 <!-- NEW CONTENT GENERATED BELOW. PLEASE PRESERVE THIS COMMENT. -->
 
+### 0.64.227 (2024-10-25)
+
+- The `modal container list` CLI command now only shows containers for your active environment. You can pass `--env` to list containers in other environments.
```
mwaskom, line 15: "Active environment" is maybe slightly inaccurate (though in a very common and confusing way) — we don't technically have this concept. We have an active _profile_, and a profile can have an environment configuration, but it doesn't have to, in which case the server determines the environment to use. But I'm not certain how better to articulate it. We should maybe rethink how this works.

mwaskom (approved):
Thanks for following up here!

CHANGELOG.md
kramstrom: 😅 yeah just switching environments was confusing when trying this out too, definitely something we should rethink!

Updated the changelog to better match how it works though, thanks!

PR #2407, opened 2024-10-25
Add back CLIENT_TYPE_WORKER

erikbern:
This was removed in 827b6ef49cbf9de94a192190fda237b55c5f4ec5 but afaict had been deprecated for years at that point.


mwaskom (commented):
Make sense!

PR #2408, opened 2024-10-25
Use uv to install client dependencies with 2024.10 image builder

mwaskom:
In practice we find that this doesn't achieve huge speedups over pip (mostly because we need to compile the bytecode) but it does save a handful of seconds for every custom base Image build.

Closes CLI-149


mwaskom:
@prbot approve

thundergolfer (commented):
LGTM

Love to see it.

PR #2409, opened 2024-10-25
Redact token_secret in default `modal config show` output

mwaskom:
`modal config show` prints config values by default which almost always include the modal token + secret. It's easy to accidentally leak your credentials this way.

This adds a default-enabled switch that redacts the token secret by printing it as `"..."`. I've keep the token itself in the output. In the event that users need to debug an issue with the secret, they can pass `--no-redact` to see it.

I also switched the printing to use `rich` instead of `pprint` to align better with the rest of our CLI.

### DIFF
--- a/modal/cli/config.py
+++ b/modal/cli/config.py
@@ -1,7 +1,6 @@
 # Copyright Modal Labs 2022
-import pprint
-
 import typer
+from rich.console import Console
 
 from modal.config import _profile, _store_user_config, config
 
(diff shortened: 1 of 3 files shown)


devennavani (approved):
LGTM

modal/cli/config.py
```diff
@@ -17,10 +16,15 @@
...
 
-@config_cli.command(help="Show configuration values for the current profile (debug command).")
-def show():
+@config_cli.command(help="Show current configuration values (debugging command).")
+def show(redact: bool = typer.Option(True, help="Redact the `token_secret` value.")):
     # This is just a test command
-    pprint.pprint(config.to_dict())
+    config_dict = config.to_dict()
+    if redact and config_dict.get("token_secret"):
+        config_dict["token_secret"] = "..."
```
luiscape, line 24: Using `***` seems to be a lot more common than `...`.

ekzhang (approved):

modal/cli/config.py
mwaskom: Oh yeah good idea

PR #2410, opened 2024-10-25
Deprecate modal.Proxy

ekzhang:
I think we should also add a page about Proxy IPs to the docs, so it can be searchable!

### DIFF
--- a/modal/proxy.py
+++ b/modal/proxy.py
@@ -9,7 +9,8 @@ from .object import _get_environment_name, _Object
 
 
 class _Proxy(_Object, type_prefix="pr"):
-    """
+    """This is deprecated. Please use the **Proxy IP** feature, available on the team plan instead.
+
     Proxy objects are used to setup secure tunnel connections to a private remote address, for example
     a database.
 


mwaskom:
My understanding of @luiscape's current plan was that we would be migrating proxies to the new system and continue using the `modal.Proxy` name. Could be wrong about where we landed there, though.

ekzhang:
> My understanding of @luiscape's current plan was that we would be migrating proxies to the new system and continue using the `modal.Proxy` name. Could be wrong about where we landed there, though.

We can continue using the `modal.Proxy` name but that will probably require a complete rewrite? We're currently moving people onto `_experimental_proxy_ip` for the time being.

mwaskom:
> That will probably require a complete rewrite of this class, since it's built for the deprecated Proxy object at the moment?

This is what I wasn't sure about — it doesn't look like the object class in the client actually does anything relevant to the proxy mechanism?

ekzhang:
Yeah, it's just the code in `_proxy_tunnel.py` right now. And the `apt_install("autossh")` hack. I think we could maybe deprecate this object, remove it, then reintroduce it in a future version so people don't get confused by the docs page for the old object?

image

ekzhang:
@luiscape I indexed the "Proxy IPs" guide page you wrote in https://github.com/modal-labs/modal/pull/16743 so this PR will remove the docs for the deprecated modal.Proxy object

erikbern:
My understanding is also that we'll keep the `Proxy` object in the client, but we'll swap out the code underneath.

PR #2412, opened 2024-10-27
Enable cltr-c tests on windows

freider:
We were previously not testing Ctrl-C behavior on windows since SIGINT doesn't work and the python standard library makes it tricky to send CTRL_C_EVENT to a subprocess.

This was worked around in recent #2322 and this patch generalizes that fix and reactivates our CLI tests for windows

### DIFF
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -25,8 +25,8 @@ from typing import (
 )
 
 import synchronicity
-from exceptions import NestedEventLoops
 from synchronicity.async_utils import Runner
+from synchronicity.exceptions import NestedEventLoops
 from typing_extensions import ParamSpec
 
 from ..exception import InvalidError
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -3,9 +3,6 @@ import asyncio
 import concurrent.futures
 import functools
 import inspect
-import signal
-import sys
-import threading
 import time
 import typing
 from contextlib import asynccontextmanager
@@ -28,6 +25,8 @@ from typing import (
 )
 
 import synchronicity
+from exceptions import NestedEventLoops
+from synchronicity.async_utils import Runner
 from typing_extensions import ParamSpec
 
 from ..exception import InvalidError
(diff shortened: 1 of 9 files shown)


freider:
```FAILED test/cli_test.py::test_keyboard_interrupt_during_app_load - subprocess.TimeoutExpired: Command '['C:\\hostedtoolcache\\windows\\Python\\3.10.11\\x64\\python.exe', '-m', 'modal', 'run', 'D:\\a\\modal-client\\modal-client\\test\\supports\\hello.py::hello']' timed out after 5 seconds```
I guess I'll have to bring out the ol' windows laptop again 😿 🪟

**Comment by freider** [comment]
Turns out Ctrl-C doesn't actually work on the windows client in many cases, since synchronicity (though actually the underlying concurrent.futures.Future) doesn't support aborting a long running blocking call. Attempt at fix in https://github.com/modal-labs/synchronicity/pull/179

File: modal/_utils/async_utils.py [hunk 1]
```diff
@@ -454,121 +453,30 @@ async def asyncnullcontext(*args, **kwargs):
...
-                # first sigint is graceful
-                self._loop.call_soon_threadsafe(coro_task.cancel)
-                return
-
-            # this should normally not happen, but the second sigint would "hard kill" the event loop
-            # by raising KeyboardInterrupt inside of it
-            raise KeyboardInterrupt()
-
-        original_sigint_handler = None
+def run_async_gen(
```
freider, line 456: Just pulled this out of the `Runner` here and made it use the identical `Runner` implementation from synchronicity itself instead

setup.cfg
```diff
@@ -27,7 +27,7 @@ install_requires =
     grpclib==0.4.7
     protobuf>=3.19,<5.0,!=4.24.0
     rich>=12.0.0
-    synchronicity~=0.8.3
+    synchronicity~=0.9.1
```
freider, line 30: This version upgrade the actual windows issues

test/helpers.py
```diff
@@ -54,3 +55,25 @@ def deploy_app_externally(
         print(f"Deploying app failed!\n### stdout ###\n{stdout_s}\n### stderr ###\n{stderr_s}")
         raise Exception("Test helper failed to deploy app")
     return stdout_s
+
+
+class PopenWithCtrlC(subprocess.Popen):
```
freider, line 60: We have an identical helper in the tests of synchronicity, but it's not part of the synchronicity *package*, so we can't reuse it here

PR #2413, opened 2024-10-27
Add proto changes for container filesystem

azliu0:
## Describe your changes

proto changes for WRK-414 

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -757,10 +757,6 @@ message ContainerFileFlushRequest {
   string file_descriptor = 1;
 }
 
-message ContainerFileFlushResponse {
-  string exec_id = 1;
-}
-
 message ContainerFileOpenRequest {
   // file descriptor is hydrated when sent from server -> worker
   optional string file_descriptor = 1;
@@ -772,17 +768,9 @@ message ContainerFileReadLineRequest {
   string file_descriptor = 1;
 }
 
-message ContainerFileReadLineResponse {
-  string exec_id = 1;
-}
-
 message ContainerFileReadRequest {
   string file_descriptor = 1; 
   optional uint32 n = 2;
-}
-
-message ContainerFileReadResponse {
-  string exec_id = 1;
 }
 
 message ContainerFileSeekRequest {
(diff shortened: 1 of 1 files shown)


mwaskom:
Is this exposing the *container* filesystem? Or a "sandbox fileststem"?

azliu0:
This is meant to expose some operations on the filesystem inside of the container, so I guess the latter? My intention with naming was that _technically_ these changes aren't tied to Sandboxes, and could be used to implement other things, similar to how Sandbox `exec` is implemented with a container exec.

erikbern:
I think it's nice to keep it more general than for sandboxes. A lot of what we do with sandboxes today could (and probably should) be extended to containers at some point

modal_proto/api.proto
```diff
@@ -720,6 +743,102 @@ message ContainerExecWaitResponse {
...
+  string file_descriptor = 1;
+  optional uint32 start_inclusive = 2;
+  optional uint32 end_exclusive = 3;
+}
+
+message ContainerFileFlushRequest {
+  string file_descriptor = 1;
+}
+
+message ContainerFileFlushResponse {
```
pawalt, line 760: Everything here looks good except looks like there are a few of these responses laying around that aren't used anymore. Remove these and I'll LGTM

modal_proto/api.proto
```diff
@@ -720,6 +743,86 @@ message ContainerExecWaitResponse {
...
+
+message ContainerFilesystemExecResponse {
+  string exec_id = 1;
+  // only set when the request opens a new file, i.e., ContainerFileOpenRequest
+  optional string file_descriptor = 2;
+}
+
+message ContainerFilesystemGetOutputRequest {
+  float timeout = 1;
+  uint64 last_batch_index = 2;
```
pawalt, line 823: It's actually a mistake that the current RPCs have last_batch_index - it doesn't work with `DataChannel`s. Let's remove it. We should keep `batch_index` on the batch responses just for informational reasons though.

pawalt (approved):
I'm super excited about this!

PR #2414, opened 2024-10-27
Add container filesystem for sandboxes

azliu0:
## Describe your changes

Resolves WRK-414 by adding a new container filesystem API. 

This API is designed to closely follow `io.FileIO`, supporting most of the major functions (`open`, `read`, `readline`, `readlines`, `write`, `seek`, `flush`, `close`) along with a few higher-level functions (`delete_bytes`, `write_replace_bytes`) to help make Sandboxes a bit easier to use.

All file modes (e.g. text, bytes, r/w/a/x, etc.) should also be supported, as well as most common errors/exceptions. If an unknown exception is thrown, then a new `modal.exception.FilesystemExecutionError` is thrown instead.

Example usage:
```
app = modal.App.lookup("sandbox-fs", create_if_missing=True)

with sb.open("test.txt", "w") as f:
  f.write("Hello World\n")

f = sb.open("test.txt", "rb")
print(f.read())
```

---

---

</details>

## Changelog

Add filesystem support for Sandboxes.

### DIFF
--- a/modal/exception.py
+++ b/modal/exception.py
@@ -214,3 +214,11 @@ class ModuleNotMountable(Exception):
 
 class ClientClosed(Error):
     pass
+
+
+class UnsupportedOperation(Error):
+    """Raised when an unsupported operation is performed on a file descriptor."""
+
+
+class FilesystemExecutionError(Error):
+    """Raised when an error occurs during a container filesystem operation."""
--- a/modal/exception.py
+++ b/modal/exception.py
@@ -216,9 +216,5 @@ class ClientClosed(Error):
     pass
 
 
-class UnsupportedOperation(Error):
-    """Raised when an unsupported operation is performed on a file descriptor."""
-
-
 class FilesystemExecutionError(Error):
-    """Raised when an error occurs during a container filesystem operation."""
+    """Raised when an unknown error is thrown during a container filesystem operation."""
--- a/modal/file_io.py
+++ b/modal/file_io.py
@@ -63,12 +63,12 @@ class _FileIO:
                 break
             stderr += data
         if errored:
-            raise FilesystemExecutionError(f"Error executing filesystem command: {stderr}")
+            raise FilesystemExecutionError(stderr)
         return stdout
 
     async def _open_file(self, path: str, mode: str) -> None:
         resp = await self._client.stub.ContainerFileOpen(
-            api_pb2.ContainerFileOpenRequest(path=path, mode=mode, exec_id=self._task_id)
+            api_pb2.ContainerFileOpenRequest(path=path, mode=mode, task_id=self._task_id)
         )
         self._file_descriptor = resp.file_descriptor
         await self._wait(resp.exec_id)
@@ -128,6 +128,16 @@ class _FileIO:
         resp = await self._client.stub.ContainerFlushFile(api_pb2.ContainerFlushFileRequest(exec_id=self._task_id))
         await self._wait(resp.exec_id)
 
+    def _get_whence(self, whence: int) -> api_pb2.SeekWhence:
+        if whence == 0:
+            return api_pb2.SeekWhence.SEEK_SET
+        elif whence == 1:
+            return api_pb2.SeekWhence.SEEK_CUR
+        elif whence == 2:
+            return api_pb2.SeekWhence.SEEK_END
+        else:
+            raise ValueError(f"Invalid whence value: {whence}")
+
     # TODO
     async def seek(self, offset: int, whence: int = 0) -> None:
         """Move to a new position in the file.
@@ -137,7 +147,7 @@ class _FileIO:
         """
         self._check_seekable()
         resp = await self._client.stub.ContainerSeekFile(
-            api_pb2.ContainerSeekFileRequest(exec_id=self._task_id, offset=offset, whence=whence)
+            api_pb2.ContainerSeekFileRequest(exec_id=self._task_id, offset=offset, whence=self._get_whence(whence))
         )
         await self._wait(resp.exec_id)
 
--- a/modal/file_io.py
+++ b/modal/file_io.py
@@ -1,6 +1,6 @@
 # Copyright Modal Labs 2024
 import io
-from typing import AsyncIterator, Optional, Union
+from typing import AsyncIterator, Optional, Union, cast
 
 from modal_proto import api_pb2
 
@@ -43,8 +43,6 @@ class _FileIO:
             seen_chars.add(char)
 
     def _handle_error(self, error: api_pb2.SystemErrorMessage) -> None:
-        print("error_code", error.error_code)
-        print("error_message", error.error_message)
         raise FilesystemExecutionError(error.error_message)
 
     async def _consume_output(self, exec_id: str) -> AsyncIterator[Optional[str]]:
@@ -64,8 +62,6 @@ class _FileIO:
 
     async def _wait(self, exec_id: str) -> Union[bytes, str]:
         output = ""
-        # The filesystem API shouldn't involve any long-running processes,
-        # so it should be safe to consume output/err separately here.
         async for data in self._consume_output(exec_id):
             if data is None:
                 break
@@ -102,8 +98,9 @@ class _FileIO:
         self._closed = False
         return self
 
-    async def read(self, n: int | None = None) -> Union[bytes, str]:
+    async def read(self, n: Union[int, None] = None) -> Union[bytes, str]:
         """Read n bytes from the current position, or the entire remaining file if n is None."""
+        self._check_closed()
         self._check_readable()
         resp = await self._client.stub.ContainerFilesystemExec(
             api_pb2.ContainerFilesystemExecRequest(
@@ -115,6 +112,7 @@ class _FileIO:
 
     async def readline(self) -> Union[bytes, str]:
         """Read a single line from the current position."""
+        self._check_closed()
         self._check_readable()
         resp = await self._client.stub.ContainerFilesystemExec(
             api_pb2.ContainerFilesystemExecRequest(
@@ -124,10 +122,14 @@ class _FileIO:
         )
         return await self._wait(resp.exec_id)
 
-    async def readlines(self) -> list[Union[bytes, str]]:
+    async def readlines(self) -> Union[list[bytes], list[str]]:
         """Read all lines from the current position."""
+        self._check_closed()
         self._check_readable()
-        return await self.read().split(b"\n")
+        data = await self.read()
+        if self._binary:
+            return cast(bytes, data).split(b"\n")
+        return cast(str, data).split("\n")
 
     async def write(self, data: Union[bytes, str]) -> None:
         """Write data to the current position.
@@ -136,6 +138,7 @@ class _FileIO:
         can be done manually with `flush()` or automatically when the file is
         closed.
         """
+        self._check_closed()
         self._check_writable()
         self._validate_type(data)
         if isinstance(data, str):
@@ -150,6 +153,7 @@ class _FileIO:
 
     async def flush(self) -> None:
         """Flush the buffer to disk."""
+        self._check_closed()
         self._check_writable()
         resp = await self._client.stub.ContainerFilesystemExec(
             api_pb2.ContainerFilesystemExecRequest(
@@ -159,7 +163,7 @@ class _FileIO:
         )
         await self._wait(resp.exec_id)
 
-    def _get_whence(self, whence: int) -> api_pb2.SeekWhence:
+    def _get_whence(self, whence: int) -> api_pb2.SeekWhence.ValueType:
         if whence == 0:
             return api_pb2.SeekWhence.SEEK_SET
         elif whence == 1:
@@ -175,6 +179,7 @@ class _FileIO:
         `whence` defaults to 0 (absolute file positioning); other values are 1
         (relative to the current position) and 2 (relative to the file's end).
         """
+        self._check_closed()
         resp = await self._client.stub.ContainerFilesystemExec(
             api_pb2.ContainerFilesystemExecRequest(
                 file_seek_request=api_pb2.ContainerFileSeekRequest(
@@ -187,7 +192,9 @@ class _FileIO:
         )
         await self._wait(resp.exec_id)
 
-    async def delete_bytes(self, start_inclusive: int | None = None, end_exclusive: int | None = None) -> None:
+    async def delete_bytes(
+        self, start_inclusive: Union[int, None] = None, end_exclusive: Union[int, None] = None
+    ) -> None:
         """Delete a range of bytes from the file.
 
         `start_inclusive` and `end_exclusive` are byte offsets. If either is
@@ -195,6 +202,8 @@ class _FileIO:
 
         Resets the file pointer to the start of the file.
         """
+        self._check_closed()
+        self._check_writable()
         if start_inclusive is not None and end_exclusive is not None:
             if start_inclusive >= end_exclusive:
                 raise ValueError("start_inclusive must be less than end_exclusive")
@@ -211,7 +220,7 @@ class _FileIO:
         await self._wait(resp.exec_id)
 
     async def write_replace_bytes(
-        self, data: bytes, start_inclusive: int | None = None, end_exclusive: int | None = None
+        self, data: bytes, start_inclusive: Union[int, None] = None, end_exclusive: Union[int, None] = None
     ) -> None:
         """Replace a range of bytes in the file with new data.
 
@@ -220,6 +229,8 @@ class _FileIO:
 
         Resets the file pointer to the start of the file.
         """
+        self._check_closed()
+        self._check_writable()
         if start_inclusive is not None and end_exclusive is not None:
             if start_inclusive >= end_exclusive:
                 raise ValueError("start_inclusive must be less than end_exclusive")
@@ -244,8 +255,8 @@ class _FileIO:
                 task_id=self._task_id,
             )
         )
-        await self._wait(resp.exec_id)
         self._closed = True
+        await self._wait(resp.exec_id)
 
     async def close(self) -> None:
         """Flush the buffer and close the file."""
--- a/modal/file_io.py
+++ b/modal/file_io.py
@@ -8,6 +8,22 @@ from ._utils.async_utils import synchronize_api
 from .client import _Client
 from .exception import FilesystemExecutionError
 
+ERROR_MAPPING = {
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_PERM: PermissionError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_NOENT: FileNotFoundError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_IO: IOError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_NXIO: IOError,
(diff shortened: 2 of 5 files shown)


mwaskom:
Can we improve the changelog a bit here?

modal/file_io.py
```diff
@@ -0,0 +1,309 @@
...
+            raise ValueError("must have exactly one of create/read/write/append mode")
+
+        seen_chars = set()
+        for char in mode:
+            if char in seen_chars:
+                raise ValueError(f"Invalid file mode: {mode}")
+            seen_chars.add(char)
+
+    def _handle_error(self, error: api_pb2.SystemErrorMessage) -> None:
+        if error.error_code == api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_UNSPECIFIED:
```
pawalt, line 66: Can you just make this a case in the error mapping

modal/file_io.py
```diff
@@ -0,0 +1,309 @@
...
+            raise FilesystemExecutionError(error.error_message)
+
+        error_class = ERROR_MAPPING.get(error.error_code, FilesystemExecutionError)
+        raise error_class(error.error_message)
+
+    async def _consume_output(self, exec_id: str) -> AsyncIterator[Optional[bytes]]:
```
pawalt, line 72: Implement retries here similar to how they're implemented in `StreamReader`.

modal/file_io.py
```diff
@@ -0,0 +1,309 @@
...
+        self._check_readable()
+        data = await self.read()
+        if self._binary:
+            lines = cast(bytes, data).split(b"\n")
+            return [line + b"\n" for line in lines[:-1]] + ([lines[-1]] if lines[-1] else [])
+        else:
+            lines = cast(str, data).split("\n")
+            return [line + "\n" for line in lines[:-1]] + ([lines[-1]] if lines[-1] else [])
+
+    async def write(self, data: Union[bytes, str]) -> None:
```
pawalt, line 160: Can we check data size here so we fail before making the req?

modal/file_io.py
```diff
@@ -0,0 +1,309 @@
...
+    async def create(cls, path: str, mode: str, client: _Client, task_id: str) -> "_FileIO":
+        self = cls.__new__(cls)
+        self._validate_mode(mode)
+        self._client = client
+        self._task_id = task_id
+        await self._open_file(path, mode)
+        self._closed = False
+        return self
+
+    async def read(self, n: Union[int, None] = None) -> Union[bytes, str]:
```
pawalt, line 124: Check data size before making the req

modal/sandbox.py
```diff
@@ -425,6 +426,22 @@ async def exec(self, *cmds: str, pty_info: Optional[api_pb2.PTYInfo] = None):
         )
         return _ContainerProcess(resp.exec_id, self._client)
 
+    async def open(self, path: str, mode: str = "r") -> _FileIO:
```
pawalt, line 429: Can we support the context manager api too?

azliu0: It should already be supported

pawalt:
@azliu0 can we convert this back to a draft until we're ready to look to merge?

azliu0:
@pawalt @cweld510 @mwaskom Now that daemon changes are gearing up to be merged, this interface is unlikely to be changed, so client changes are ready for another look. Merging these changes are specifically still blocked on:
- daemon changes merging
- enabling `-host-uds=all` in production for all Sandboxes
- gVisor upgrade

but they are ready for review.

modal/file_io.py
```diff
@@ -0,0 +1,335 @@
...
+                if retries_remaining > 0:
+                    retries_remaining -= 1
+                    if isinstance(exc, GRPCError):
+                        if exc.status in RETRYABLE_GRPC_STATUS_CODES:
+                            await asyncio.sleep(1.0)
+                            continue
+                    elif isinstance(exc, StreamTerminatedError):
+                        continue
+                raise
+        if self._binary:
```
pawalt, line 112: I don't think it makes sense for `_wait` to handle this encoding/decoding. It means the caller has to be aware of the state and act appropriately: see the switching in `readlines`.

modal/file_io.py
```diff
@@ -0,0 +1,335 @@
...
+        return output.decode("utf-8")
+
+    def _validate_type(self, data: Union[bytes, str]) -> None:
+        if self._binary and isinstance(data, str):
+            raise TypeError("Expected bytes when in binary mode")
+        if not self._binary and isinstance(data, bytes):
+            raise TypeError("Expected str when in text mode")
+
+    async def _open_file(self, path: str, mode: str) -> None:
+        resp = await self._client.stub.ContainerFilesystemExec(
```
pawalt, line 123: You should `retry_transient_errors` on all of these calls. Maybe there's a way to write a helper to make this easier.

modal/file_io.py
```diff
@@ -0,0 +1,335 @@
...
+                task_id=self._task_id,
+            )
+        )
+        if not resp.HasField("file_descriptor"):
+            raise FilesystemExecutionError("Failed to open file")
+        self._file_descriptor = resp.file_descriptor
+        await self._wait(resp.exec_id)
+
+    @classmethod
+    async def create(cls, path: str, mode: str, client: _Client, task_id: str) -> "_FileIO":
```
pawalt, line 135: This should have docs since it's public.

modal/file_io.py
```diff
@@ -0,0 +1,335 @@
...
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_MFILE: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_FBIG: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_NOSPC: OSError,
+}
+
+
+# The Sandbox file handling API is designed to mimic Python's io.FileIO
+# See https://github.com/python/cpython/blob/main/Lib/_pyio.py#L1459
+# Unlike io.FileIO, it also implements some higher level APIs, like `delete_bytes` and `overwrite_bytes`.
+class _FileIO:
```
pawalt, line 39: This should have a docstring

modal/file_io.py
```diff
@@ -0,0 +1,335 @@
...
+            )
+        )
+        return await self._wait(resp.exec_id)
+
+    async def readlines(self) -> Union[List[bytes], List[str]]:
+        """Read all lines from the current position."""
+        self._check_closed()
+        self._check_readable()
+        data = await self.read()
+        if self._binary:
```
pawalt, line 175: See my note about why the switching in `_wait` makes this hard.

modal/file_io.py
```diff
@@ -0,0 +1,335 @@
...
+        """Write data to the current position.
+
+        Writes may not appear until the entire buffer is flushed, which
+        can be done manually with `flush()` or automatically when the file is
+        closed.
+        """
+        self._check_closed()
+        self._check_writable()
+        self._validate_type(data)
+        if isinstance(data, str):
```
pawalt, line 192: Is this the behavior of `write` in Python? If it requires you to pass bytes, I think we should too.

modal/sandbox.py
```diff
@@ -495,6 +496,22 @@ async def exec(
         by_line = bufsize == 1
         return _ContainerProcess(resp.exec_id, self._client, stdout=stdout, stderr=stderr, text=text, by_line=by_line)
 
+    async def open(self, path: str, mode: str = "r") -> _FileIO:
+        """Open a file in the Sandbox and return
+        a [`FileIO`](/docs/reference/modal.FileIO#modalfile_io) handle.
```
pawalt, line 501: This link should be a full domain so people using the client library directly can hit the link.

modal/sandbox.py
```diff
@@ -495,6 +496,22 @@ async def exec(
...
+
+        **Usage**
+
+        ```python notest
+        sb = modal.Sandbox.create("sleep", "infinity")
```
pawalt, line 506: Do no args but do `app=app`

test/file_io_test.py
```diff
@@ -0,0 +1,343 @@
+# Copyright Modal Labs 2024
```
pawalt, line 1: I think these tests are pretty good. But I think the real way to catch bugs in this is to write an integration test that fuzzes this API. It should do random writes and reads and check if those results on a regular FS are different from those results in a Sandbox.

modal/file_io.py
azliu0: `self._validate_type` enforces that the correct type is passed in, but we always convert to `bytes` when sending the rpc.

modal/sandbox.py
pawalt: it should yeah

modal/file_io.py
pawalt: Yeah my question is whether python `write` accept string if you're in non-bytes mode

azliu0: python `write` enforces that the correct type is passed in

test/file_io_test.py
azliu0: I'll put a server-side integration test once these changes are merged

modal/file_io.py
```diff
@@ -0,0 +1,403 @@
...
+from ._utils.async_utils import synchronize_api
+from ._utils.grpc_utils import RETRYABLE_GRPC_STATUS_CODES
+from .client import _Client
+from .exception import FilesystemExecutionError
+
+OpenTextModeUpdating = Literal["r+", "+r", "w+", "+w", "a+", "+a", "x+", "+x"]
+OpenTextModeWriting = Literal["w", "a", "x"]
+OpenTextModeReading = Literal["r"]
+OpenTextMode = Union[OpenTextModeUpdating, OpenTextModeWriting, OpenTextModeReading]
+OpenBinaryModeUpdating = Literal[
```
pawalt, line 20: I think there's gotta be a better way to do this than listing out every ordering. Can you not just check that the characters are in the list?

pawalt (commented):
This is looking pretty good to me. Looking for a review from @mwaskom or @freider since this is as pretty big client change.

modal/file_io.py
azliu0: I agree that listing all these out isn't ideal. Type safety is built into Python's native `open`, which was my original incentive to add it here. 

If we want to preserve this, another option is to take these variables from [_typeshed](https://github.com/python/typeshed/tree/main/stdlib/_typeshed). This would also need some changes to synchronicity to get working: https://github.com/modal-labs/synchronicity/pull/193. I don't think these changes are compatible with python 3.8 / I'm not sure if the changes in that draft pr would necessarily be the correct way to approach resolving `_typeshed`.

I'm ok with removing the type safety is this is too much, though. Deferring to @mwaskom / @freider.

pawalt (commented):
No more comments from me. Deferring to @freider and @mwaskom

modal/sandbox.py
```diff
@@ -490,7 +494,7 @@ async def exec(
         _pty_info: Optional[api_pb2.PTYInfo] = None,
     ):
         """Execute a command in the Sandbox and return
-        a [`ContainerProcess`](/docs/reference/modal.ContainerProcess#modalcontainer_process) handle.
+        a [`ContainerProcess`](https://modal.com/docs/reference/modal.ContainerProcess#modalcontainer_process) handle.
```
freider, line 497: I think we are intentionally not using fully qualified domains in docs links, so that we can publish staging docs etc. on other domains for testing purposes and have links refer to that domain.

modal/file_io.py
```diff
@@ -0,0 +1,375 @@
...
+        Writes may not appear until the entire buffer is flushed, which
+        can be done manually with `flush()` or automatically when the file is
+        closed.
+        """
+        self._check_closed()
+        self._check_writable()
+        self._validate_type(data)
+        if isinstance(data, str):
+            data = data.encode("utf-8")
+        if len(data) > LARGE_FILE_SIZE_LIMIT:
```
freider, line 234: Would it make sense to chunk the data for the user in this case instead of erroring? If they pass this "fileio" object to some library where they can't control the write chunk size, it might be problematic otherwise?

modal/file_io.py
```diff
@@ -0,0 +1,375 @@
...
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_FBIG: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_NOSPC: OSError,
+}
+
+T = TypeVar("T", str, bytes)
+
+
+# The Sandbox file handling API is designed to mimic Python's io.FileIO
+# See https://github.com/python/cpython/blob/main/Lib/_pyio.py#L1459
+# Unlike io.FileIO, it also implements some higher level APIs, like `delete_bytes` and `overwrite_bytes`.
```
freider, line 44: Could we document the reasoning for having those specific extensions/higher level API methods. (I assume it's for efficiency reasons, not having to shuffle all of the file data back and forth to rewrite parts of a file).

Also, to make the distinction of custom vs standard clear, maybe we could move these non-standard methods to their own top level helper functions in the module? I can see the case for having them be in the FileIO class too (IDE discoverability/tab completion), but at the same time API discoverability through the web docs might be easier if these special methods are top level and not buried among the standard functions that a user wouldn't explore (since it looks exactly like io.FileIO which they already know otherwise)

freider (commented):
Didn't look at the implementation details, but I like the interfaces overall. Added some smaller suggested changes.

modal/file_io.py
azliu0: Thanks for bringing up the web docs! Moving the implementation in a helper makes sense to me; unfortunately the docs end up looking something like this: 

Screenshot 2024-12-11 at 5 14 55 AM

where the higher-level functions are separated but pushed to the bottom (in addition, the docstrings are also duplicated between methods). Is this what you had in mind?

freider: That looks pretty normal to me - what would be your expected output?

What do you mean by the docstrings being duplicated between methods?

modal/file_io.py
freider: Ah thanks for the context! Makes sense

modal/file_io.py
azliu0: Ah, just that the docstring on the `delete_bytes` in `modal.file_io.FileIO` and `modal.file_io.delete_bytes` is the same

modal/exception.py
```diff
@@ -218,6 +218,10 @@ class ClientClosed(Error):
     pass
 
 
+class FilesystemExecutionError(Error):
```
mwaskom, line 221: Just being annoying but are we consistent about whether we treat "file system" as one or two words when we camelcase? You're also using `snapshot_fs` right?

modal/file_io.py
```diff
@@ -0,0 +1,374 @@
...
+async def delete_bytes(file: "_FileIO", start: Optional[int] = None, end: Optional[int] = None) -> None:
+    """Delete a range of bytes from the file.
+
+    `start` and `end` are byte offsets. `start` is inclusive, `end` is exclusive.
+    If either is None, the start or end of the file is used, respectively.
+    """
+    file._check_closed()
+    if start is not None and end is not None:
+        if start >= end:
+            raise ValueError("start must be less than end")
```
mwaskom, line 51: `modal.exceptions.InvalidError`?

mwaskom: TBH I'm not sure I prefer that we lump a lot of stuff into that instead of using the existing Python exception typology but I guess we should try to be consistent.

modal/file_io.py
```diff
@@ -0,0 +1,374 @@
...
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_INVAL: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_MFILE: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_FBIG: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_NOSPC: OSError,
+}
+
+T = TypeVar("T", str, bytes)
+
+
+async def delete_bytes(file: "_FileIO", start: Optional[int] = None, end: Optional[int] = None) -> None:
```
freider, line 42: Minor thing, but we've more or less standardized using `_` prefixes on the underlying async implementation versions of synchronicity wrapped functions etc.

modal/file_io.py
```diff
@@ -0,0 +1,374 @@
...
+                start_inclusive=start,
+                end_exclusive=end,
+            ),
+            task_id=file._task_id,
+        )
+    )
+    await file._wait(resp.exec_id)
+
+
+async def overwrite_bytes(file: "_FileIO", data: bytes, start: Optional[int] = None, end: Optional[int] = None) -> None:
```
mwaskom: If the new bytes don't need to have the same range I might think `replace_bytes` would be a better name but not a big deal if there's a reference point for this name.

modal/file_io.py
freider: Btw,  just want to loop in @mwaskom for a second opinion here - am I crazy to want these "specialization" methods be separate module functions in this file rather than members on the FileIO object which is otherwise compliant with python's own FileIO.

Makes IDE-based discovery of these methods harder, but documentation-based discovery somehwat easer imo. Not 100% sure I'm right though :D 

I'll leave it to @mwaskom to decide - I'd be fine with either decision

freider (approved):
lgtm, but I'd like @mwaskom (or @erikbern if he has time) to chime in on the best place to put these "non-standard" methods (in class or as separate functions) so I'm not making a crazy decision here on my own :)

modal/file_io.py
```diff
@@ -0,0 +1,376 @@
...
+            ),
+            task_id=file._task_id,
+        )
+    )
+    await file._wait(resp.exec_id)
+
+
+# The Sandbox file handling API is designed to mimic Python's io.FileIO
+# See https://github.com/python/cpython/blob/main/Lib/_pyio.py#L1459
+class _FileIO(Generic[T]):
```
mwaskom, line 96: How specific is this to sandboxes? If very specific, should the name reflect that?

modal/file_io.py
```diff
@@ -0,0 +1,376 @@
...
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_INVAL: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_MFILE: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_FBIG: OSError,
+    api_pb2.SystemErrorCode.SYSTEM_ERROR_CODE_NOSPC: OSError,
+}
+
+T = TypeVar("T", str, bytes)
+
+
+async def _delete_bytes(file: "_FileIO", start: Optional[int] = None, end: Optional[int] = None) -> None:
```
mwaskom, line 42: j/w but why are these standalone functions instead of methods on a FileIO?

mwaskom (approved):

modal/file_io.py
azliu0: I think we have plans to use this logic to eventually implement other methods, like `container cp`. The logic isn't directly tied to Sandboxes

modal/file_io.py
azliu0: Our rpcs also have this name, I agree that `replace_bytes` seems better.

modal/file_io.py
azliu0: some discussion here: https://github.com/modal-labs/modal-client/pull/2414#discussion_r1881014114

modal/exception.py
azliu0: Yep, the existing `snapshot_fs` treats it as one word.

PR #2415, opened 2024-10-28
Improve modal.Cls documentation

thundergolfer:
## Describe your changes

Originally just wanted an example in `modal shell` docs of how to use it with `modal.Cls`, but then pulled on that thread and realized how many more places we had missing or inadequate documentation of `modal.Cls`.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -374,7 +374,7 @@ def shell(
         ),
     ),
 ):
-    """Run an interactive shell inside a Modal image.
+    """Run an interactive shell inside a Modal container.
 
     **Examples:**
 
@@ -384,10 +384,16 @@ def shell(
     modal shell
     ```
 
-    Start a bash shell using the spec for `my_function` in your app:
+    Start a bash shell using the spec for `my_function` in your App:
 
     ```
     modal shell hello_world.py::my_function
+    ```
+
+    Or, if you're using a [modal.Cls](/docs/reference/modal.Cls), you can refer to a `@modal.method` directly:
+
+    ```
+    modal shell hello_world.py::MyClass.my_method
     ```
 
     Start a `python` shell:
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -225,6 +225,14 @@ Obj = synchronize_api(_Obj)
 
 
 class _Cls(_Object, type_prefix="cs"):
+    """
+    Cls adds method pooling and [lifecycle hook](/docs/guide/lifecycle-functions) behavior
+    to [modal.Function](/docs/reference/modal.Function).
+
+    Generally, you will not construct a Cls directly.
+    Instead, use the [`@app.cls()`](/docs/reference/modal.App#cls) decorator on the App object.
+    """
+
     _user_cls: Optional[type]
     _class_service_function: Optional[
         _Function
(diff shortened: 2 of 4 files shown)


modal/cls.py
```diff
@@ -225,6 +225,14 @@ def __getattr__(self, k):
 
 
 class _Cls(_Object, type_prefix="cs"):
+    """
+    Cls adds method pooling and [lifecycle hook](/docs/guide/lifecycle-functions) behavior
```
thundergolfer, line 229: I tried to find documentation on the "method pooling" behavior and could only find it in the changelog? That seems like a documentation gap that needs filling.

modal/cli/run.py
```diff
@@ -384,12 +384,18 @@ def shell(
     modal shell
     ```
 
-    Start a bash shell using the spec for `my_function` in your app:
+    Start a bash shell using the spec for `my_function` in your App:
```
freider, line 387: Unrelated, but we don't really explain what a "spec" is anywhere (don't even know if we have an official name for "the group of options that you can set on a function that affect what the container environment looks like") - cc @mwaskom since this ties back to the decorator brainstorming you've been doing

freider (approved):
nice!

modal/cli/run.py
mwaskom: Yeah agreed — was talking to another Modal eng recently who was confused about the difference between `Function` / `api_pb2.Function` / `FunctionInfo` / `FunctionSpec`. I don't believe that we define the "spec" externally anywhere so it's also not clear to users what parameters will be picked up by `modal shell` (@thundergolfer this is part of why I'm a little resistant to exposing this feature in other CLI tools).

If we embrace additional decorators it would be nice to be able to define the "spec" as "everything in `app.function(...)` and not need reference to a separate concept, but I suspect that won't be possible. First because I think autoscaling parameters probably still belong in `@app.function` rather than a separate decorator (but open to arguments from converts who are even more zealous than me... 😄) and also because we have some random parameters that don't really correspond to the "container spec" but that also don't really seem to merit their own decorators (`serialized`, `is_generator`, `enable_memory_snapshot`, etc.).

thundergolfer: > this is part of why I'm a little resistant to exposing this feature in other CLI tools

Yes I noticed that usage of "spec" as potentially something we're too vague on. I think of it as '`modal.Function` but with the body code replaced with a shell', but that's actually wrong on the details you write out below.

mwaskom: Yeah, in fact it's probably surprising from a certain perspective that you can define a function with `concurrency_limit=1` and then "connect two shells" to it at the same time. Which isn't exactly that we're doing, but it's what it _looks_ like we're doing.

modal/cls.py
mwaskom: This is missing "parameterization" BTW. Agreed that we should document that method pooling behavior somewhere. Somewhat unclear to me if it is in itself a "feature", or just a fact about how classes work.

BTW maybe a pet peeve (depending on whether you consider [PEP257](https://peps.python.org/pep-0257/#multi-line-docstrings) a standard) but I think the one-line summary in a docstring really should fit on one line.

It's unfortunate that we need to spell out the cross-ref to the API docs with a url path, which is most of why your summary spills over. In sphinx you could write it as
```
:class:`Function`
```
and the docs framework would handle the hyperlink in the HTML build.

PR #2416, opened 2024-10-28
Add proto definitions for exec output options

pawalt:
## Describe your changes

When running an exec, we block the process if its stdout or stderr has not been consumed. This can be surprising behavior because if a user blindly execs, their process could hang without them knowing. This commit adds proto definitions enabling subprocess-like selection of output style.

---

---

</details>


freider (approved):
lgtm!

PR #2417, opened 2024-10-28
Improve import stats performance by emitting start and end together

op:
Decrease lock contention on imports by sending start and end events all at once.

### DIFF
--- a/modal/_telemetry.py
+++ b/modal/_telemetry.py
@@ -65,7 +65,7 @@ class InterceptedModuleLoader(importlib.abc.Loader):
 
 
 class ImportInterceptor(importlib.abc.MetaPathFinder):
-    loading: typing.Dict[str, typing.Tuple[str, float]]
+    loading: typing.Dict[str, float]
     tracing_socket: socket.socket
     events: queue.Queue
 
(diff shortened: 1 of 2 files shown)


danielnorberg:
With the current [internal telemetry collection implementation](https://github.com/modal-labs/modal/blob/dca60d098a98431ddfd7dfc0f257feb7d03462e7/crates/modal-runtime/src/task/container_telemetry.rs#L157) emitting start and stop events at the same time would cause the resulting spans to become zero length, although maybe that's something that can be fixed.

dflemstr:
Maybe out of scope for this particular PR but I wonder how we can attack the main bottleneck of this? If the main issue is contending for the queue, this PR reduces the overhead by a constant factor which is great but I guess the contention is still `O(n)` in terms of something (nr of threads)?

danielnorberg:
I would be surprised if contention is the problem given that python is mostly single threaded. It might be good to benchmark and measure to get a sense of where the bottleneck is.

mwaskom:
Doing PR queue cleanup — closing as stale and conflicted. Feel free to revisit!

PR #2418, opened 2024-10-28
MOD-2454: add async_cocat

kramstrom:
## Describe your changes

---

---

</details>

### DIFF
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -5,6 +5,7 @@ import functools
 import inspect
 import time
 import typing
+from collections import defaultdict
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from typing import (
@@ -561,8 +562,10 @@ class StopSentinelType:
 STOP_SENTINEL = StopSentinelType()
 
 
-async def async_merge(*inputs: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGenerator[T, None]:
-    queue: asyncio.Queue[Tuple[int, Union[ValueWrapper[T], ExceptionWrapper, StopSentinelType]]] = asyncio.Queue()
+async def async_merge(*iterables: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGenerator[T, None]:
+    queue: asyncio.Queue[Tuple[int, Union[ValueWrapper[T], ExceptionWrapper, StopSentinelType]]] = asyncio.Queue(
+        maxsize=len(iterables) * 10
+    )
 
     async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[T]]):
         try:
@@ -573,8 +576,8 @@ async def async_merge(*inputs: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGen
         finally:
             await queue.put((producer_id, STOP_SENTINEL))
 
-    tasks = [asyncio.create_task(producer(i, it)) for i, it in enumerate(inputs)]
-    active_producers = set(range(len(inputs)))
+    tasks = [asyncio.create_task(producer(i, it)) for i, it in enumerate(iterables)]
+    active_producers = set(range(len(iterables)))
 
     try:
         while active_producers:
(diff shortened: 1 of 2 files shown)


modal/_utils/async_utils.py
```diff
@@ -561,8 +562,10 @@ class StopSentinelType:
 STOP_SENTINEL = StopSentinelType()
 
 
-async def async_merge(*inputs: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGenerator[T, None]:
-    queue: asyncio.Queue[Tuple[int, Union[ValueWrapper[T], ExceptionWrapper, StopSentinelType]]] = asyncio.Queue()
+async def async_merge(*iterables: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGenerator[T, None]:
+    queue: asyncio.Queue[Tuple[int, Union[ValueWrapper[T], ExceptionWrapper, StopSentinelType]]] = asyncio.Queue(
+        maxsize=len(iterables) * 10
```
kramstrom, line 567: realized we didn't have a max queue size here so added one arbitrarily

modal/_utils/async_utils.py
```diff
@@ -593,3 +596,60 @@ async def producer(producer_id: int, iterable: Union[AsyncIterable[T], Iterable[
...
+
+
+@typing.overload
+def async_concat(*iterables: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGenerator[Tuple[T, ...], None]:
+    ...
+
+
+async def async_concat(*iterables):
+    queue: asyncio.Queue[Tuple[int, Union[ValueWrapper[T], ExceptionWrapper, StopSentinelType]]] = asyncio.Queue(
+        maxsize=len(iterables) * 10
```
kramstrom, line 615: same as for `async_merge` -- a bit arbitrary number

modal/_utils/async_utils.py
```diff
@@ -561,35 +561,82 @@ class StopSentinelType:
...
+                    pass
 
 
 async def callable_to_agen(awaitable: Callable[[], Awaitable[T]]) -> AsyncGenerator[T, None]:
     yield await awaitable()
+
+
+@typing.overload
+def async_concat(
+    i1: Union[AsyncIterable[T], Iterable[T]], i2: Union[AsyncIterable[V], Iterable[V]], /
```
freider, line 629: This type signature is wrong right? This looks like a zip signature... I think async_concat could just have:

```
def async_concat(*iterables: AsyncIterable[T]) -> AsyncGenerator[T, None]:
```
and no overloads.

The reason we had to do overloads for zip is because the Tuple item outputs is variable in length, so would require specialization for each length of the tuple we don't want to have to assert length on in static usage

modal/_utils/async_utils.py
```diff
@@ -561,35 +561,82 @@ class StopSentinelType:
...
+                except asyncio.CancelledError:
+                    pass
 
 
 async def callable_to_agen(awaitable: Callable[[], Awaitable[T]]) -> AsyncGenerator[T, None]:
     yield await awaitable()
+
+
+@typing.overload
+def async_concat(
```
freider, line 628: nit: I'd prefer the itertools naming convention, e.g. `async_chain`, since it more clearly communicates "lazyness" of the operation

modal/_utils/async_utils.py
```diff
@@ -561,35 +561,82 @@ class StopSentinelType:
...
+    ...
+
+
+@typing.overload
+def async_concat(*iterables: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGenerator[Tuple[T, ...], None]:
+    ...
+
+
+async def async_concat(*iterables):
+    for it in iterables:
```
freider, line 640: nit: In some of our other generators we "take ownership" of the input iterables and make sure to cancel/close them if the generator itself is cancelled/closed. Might want to be consistent in how we do this across all of these utils

modal/_utils/async_utils.py
```diff
@@ -561,35 +561,82 @@ class StopSentinelType:
 STOP_SENTINEL = StopSentinelType()
 
 
-async def async_merge(*inputs: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGenerator[T, None]:
-    queue: asyncio.Queue[Tuple[int, Union[ValueWrapper[T], ExceptionWrapper, StopSentinelType]]] = asyncio.Queue()
+async def async_merge(*iterables: Union[AsyncIterable[T], Iterable[T]]) -> AsyncGenerator[T, None]:
```
freider, line 564: A bit of a late realization, but I just realized it might be nice to enforce the signature of these to only take async iterables, and let the user of the function apply `sync_or_async_iter` on the input iterators when there is ambiguity. Makes it a bit more explicit in usage

kramstrom: makes sense, do we need them to be async iterables though? Updated the branch https://github.com/modal-labs/modal-client/pull/2421 this is taken from to just be async generators since that is what we return from `sync_or_async_iter` anyway

freider (approved):
lgtm

PR #2419, opened 2024-10-28
Align sandbox exec arguments with subprocess

pawalt:
## Describe your changes

This commit adds new sandbox exec arguments, allowing the user to specify if they want to consume a stream directly, ignore the result, or consume it via a pipe.

Resolves WRK-422
Resolves WRK-413

---

---

</details>

I've tested this behavior out locally:

```python
import modal
from modal.io_streams import StreamType

app = modal.App("peyton-sb-test")

@app.local_entrypoint()
async def main():
    sandbox = modal.Sandbox.create(
        timeout=1000,
        app=app,
    )

    exc = sandbox.exec("echo", "hello", stdout=StreamType.PIPE)
    print(exc.stdout.read(), end="")

    exc = sandbox.exec("echo", "hi2", stdout=StreamType.DEVNULL)
    try:
        print(exc.stdout.read())
    except modal.exception.InvalidError as e:
        print(e)

    exc = sandbox.exec("echo", "hi3", stdout=StreamType.STDOUT)
    exc.wait()
    try:
        print(exc.stdout.read())
    except modal.exception.InvalidError as e:
        print(e)

    sandbox.terminate()
```

## Changelog

Output from `Sandbox.exec` can now be directed to `/dev/null`, `stdout`, or stored for consumption. This behavior can be controlled via the new `StreamType` arguments.

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -1,15 +1,17 @@
 # Copyright Modal Labs 2022
 import asyncio
-from typing import TYPE_CHECKING, AsyncIterator, Literal, Optional, Tuple, Union
+from typing import TYPE_CHECKING, AsyncGenerator, Literal, Optional, Tuple, Union
 
 from grpclib import Status
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 
+from modal.exception import ClientClosed, InvalidError
 from modal_proto import api_pb2
 
 from ._utils.async_utils import synchronize_api
 from ._utils.grpc_utils import RETRYABLE_GRPC_STATUS_CODES, retry_transient_errors
 from .client import _Client
+from .stream_type import StreamType
 
 if TYPE_CHECKING:
     pass
@@ -17,7 +19,7 @@ if TYPE_CHECKING:
 
 async def _sandbox_logs_iterator(
     sandbox_id: str, file_descriptor: int, last_entry_id: Optional[str], client: _Client
-) -> AsyncIterator[Tuple[Optional[api_pb2.TaskLogs], str]]:
+) -> AsyncGenerator[Tuple[Optional[str], str], None]:
     req = api_pb2.SandboxGetLogsRequest(
         sandbox_id=sandbox_id,
         file_descriptor=file_descriptor,
@@ -35,22 +37,19 @@ async def _sandbox_logs_iterator(
 
 
 async def _container_process_logs_iterator(
-    process_id: str, file_descriptor: int, last_entry_id: Optional[str], client: _Client
-):
+    process_id: str, file_descriptor: int, client: _Client
+) -> AsyncGenerator[Optional[str], None]:
     req = api_pb2.ContainerExecGetOutputRequest(
         exec_id=process_id,
         timeout=55,
-        last_batch_index=last_entry_id or 0,
         file_descriptor=file_descriptor,
     )
     async for batch in client.stub.ContainerExecGetOutput.unary_stream(req):
         if batch.HasField("exit_code"):
-            yield (None, batch.batch_index)
+            yield None
             break
         for item in batch.items:
-            # TODO: do this on the server.
-            if item.file_descriptor == file_descriptor:
-                yield (item.message, batch.batch_index)
+            yield item.message
 
 
 class _StreamReader:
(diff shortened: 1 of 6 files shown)


modal/io_streams.py
```diff
@@ -17,7 +19,7 @@
 
 async def _sandbox_logs_iterator(
     sandbox_id: str, file_descriptor: int, last_entry_id: Optional[str], client: _Client
-) -> AsyncIterator[Tuple[Optional[api_pb2.TaskLogs], str]]:
+) -> AsyncIterator[Tuple[Optional[str], str]]:
```
freider, line 22: The return type here should preferably use the `AsyncGenerator` generic which is more specific (letting the type checking system know that you can call `.aclosing()` etc. on it). I see it was AsyncIterator even before this change though, but could be nice to fix

modal/io_streams.py
```diff
@@ -35,22 +37,28 @@ async def _sandbox_logs_iterator(
 
 
 async def _container_process_logs_iterator(
-    process_id: str, file_descriptor: int, last_entry_id: Optional[str], client: _Client
-):
+    process_id: str, file_descriptor: int, client: _Client
+) -> AsyncIterator[Optional[str]]:
```
freider, line 41: Nit: `AsyncGenerator[Optional[str], None]`

modal/io_streams.py
```diff
@@ -35,22 +37,28 @@ async def _sandbox_logs_iterator(
...
         for item in batch.items:
-            # TODO: do this on the server.
-            if item.file_descriptor == file_descriptor:
-                yield (item.message, batch.batch_index)
+            yield item.message
+
+
+class StreamType(Enum):
+    # Discard all logs from the stream.
+    DEVNULL = 0
```
freider, line 57: To be foot-gun-safe, you could reuse the values of the subprocess equivalents here (they are also int valued afaik)
e.g. `DEVNULL = subprocess.DEVNULL`

Then, even though we expose our own constant, things wouldn't break if a user unknowingly used subprocess.DEVNULL instead etc. Minor but could save us some support issues in the future :)

modal/io_streams.py
```diff
@@ -124,6 +146,63 @@ async def read(self) -> str:
 
         return data
 
+    async def _consume_container_process_stream(self):
+        """mdmd:hidden
```
freider, line 150: don't need mdmd:hidden for underscored methods

modal/io_streams.py
```diff
@@ -89,12 +98,25 @@ def __init__(
...
+        if object_type == "sandbox" and stream_type != StreamType.PIPE:
+            raise ValueError("Sandbox streams must be piped.")
+        self._stream_type = stream_type
+
+        if self._object_type == "container_process":
+            # Container process streams need to be consumed as they are produced,
+            # otherwise the process will block. Use a buffer to store the stream
+            # until the client consumes it.
+            self._container_process_buffer = []
+            asyncio.create_task(self._consume_container_process_stream())
```
freider, line 118: You need to keep a reference to this task, otherwise it could get preemptively garbage collected and just stop. Also, it should preferably be created in an async context manager that can be entered/exited by some "owner" of this stream, so we don't "leak" this task in case of cancellations/early returns (garbage collection of async tasks can't be trusted to execute `finally`-statements etc. of those tasks since the gc happens outside of the event loop)

freider: I just realized the context manager thing might be tricky to pull of w/ how we currently structure sandbox calls to implicitly just start streaming logs in the background when created.

That part could be left as a future improvement where we make ContainerProcess into an (optional) context manager, which could allow graceful termination of the underlying streams (and even the process itself) when exited.

modal/sandbox.py
```diff
@@ -396,7 +397,13 @@ async def _get_task_id(self):
                 await asyncio.sleep(0.5)
         return self._task_id
 
-    async def exec(self, *cmds: str, pty_info: Optional[api_pb2.PTYInfo] = None):
+    async def exec(
+        self,
+        *cmds: str,
+        pty_info: Optional[api_pb2.PTYInfo] = None,
```
freider, line 403: We might have talked about this before and it might be a TODO, but I don't think we should expose this proto type directly to users. I guess it's mostly for internal use with modal shell anyway - maybe we should document it as internal and underscore it or something for the time being?

freider (changes requested):
Some minor comments and major realizations :D

modal/sandbox.py
pawalt: Added this as an underscored option but left the other so we can smooth migrate on our side. I'll follow up with a removal once we migrate our code.

modal/sandbox.py
```diff
@@ -396,7 +397,16 @@ async def _get_task_id(self):
...
+        self,
+        *cmds: str,
+        # Deprecated: internal use only
+        pty_info: Optional[api_pb2.PTYInfo] = None,
+        stdout: StreamType = StreamType.PIPE,
+        stderr: StreamType = StreamType.PIPE,
+        # Internal option to set terminal size and metadata
+        _pty_info: Optional[api_pb2.PTYInfo] = None,
+    ):
         """Execute a command in the Sandbox and return
```
freider, line 410: Might want to put an assertion/nice error message if the user provides a non-StreamType as stdout/stderr, like an actual file object 😬

freider (commented):
lgtm! didn't look to closely at the internals, but the public API makes sense to me! wdyt @mwaskom ?

freider (approved):

modal/sandbox.py
pawalt: good idea

pawalt:
@prbot automerge

PR #2421, opened 2024-10-29
MOD-2454: add async cancellation handling for async_utils

kramstrom:
Adds fix to `async_merge` to make it handle cancellation errors in the input generator, adds max size to the queue and adds tests

PR #2422, opened 2024-10-29
MOD-2454: async_map uses async_merge

kramstrom:
Depends on https://github.com/modal-labs/modal-client/pull/2421

Alternative implementation to https://github.com/modal-labs/modal-client/pull/2349 of `async_map`. 

This implementation creates different mapper workers and merges them together using `async_merge` instead

oh, and it also works

### DIFF
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -561,7 +561,7 @@ class ValueWrapper(typing.Generic[T]):
 
 @dataclass
 class ExceptionWrapper:
-    value: Union[Exception, asyncio.CancelledError]
+    value: Exception
 
 
 class StopSentinelType:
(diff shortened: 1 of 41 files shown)


modal/_utils/async_utils.py
```diff
@@ -560,7 +561,7 @@ class ValueWrapper(typing.Generic[T]):
 
 @dataclass
 class ExceptionWrapper:
-    value: Exception
+    value: Union[Exception, asyncio.CancelledError]
```
freider, line 564: doesn't have to have CancelledError anymore

modal/_utils/async_utils.py
```diff
@@ -631,3 +632,66 @@ async def producer(generator: AsyncGenerator[T, None]):
...
+            try:
+                if isinstance(item, ValueWrapper):
+                    yield await async_mapper_func(item.value)
+                elif isinstance(item, ExceptionWrapper):
+                    raise item.value
+                else:
+                    assert_type(item, StopSentinelType)
+                    break
+            finally:
+                queue.task_done()
```
freider, line 668: You can remove the task_done since we don't use `queue.join()` anymore

PR #2423, opened 2024-10-29
Fix Image.imports not re-raising when using a Cls

mwaskom:
Fixes CLI-210 (at least in the simple case! this code is pretty removed from the defective behavior so hard to say completely.)

The issue was that we were not including the `Image` that is a dependency of the "class service function" when importing a `Cls`, and we do the import error re-raising as part of `Image` metadata hydration.

## Changelog

- Fixed a bug where the `Image.imports` context manager would not correctly propagate ImportError when using a `modal.Cls`.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -630,10 +630,16 @@ def import_class_service(
 
     if isinstance(cls, Cls):
         # The cls decorator is in global scope
-        method_partials = synchronizer._translate_in(cls._get_partial_functions())
+        _cls = synchronizer._translate_in(cls)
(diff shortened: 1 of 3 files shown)


erikbern:
Nice – this I think points to some sketchy part of the code base we might want to rethink a bit

freider:
Yeah esp the container entrypoint has some particular 🍝 parts, because of all the special cases of how functions can be run and imported... we have a ticket for cleaning it up.

freider (approved):
lgtm. Undecorated classes are probably broken in a lot of ways... need to fix that separately

PR #2428, opened 2024-10-30
run_function return type specificity

thundergolfer:
## Describe your changes

**context:** https://modal-com.slack.com/archives/C06UKEPF3HR/p1730314009504369

Just a type-checking change and docs change.

---

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1396,7 +1396,7 @@ class _Image(_Object, type_prefix="im"):
 
     def run_function(
         self,
-        raw_f: Callable[..., Any],
+        raw_f: Callable[..., None],
         secrets: Sequence[_Secret] = (),  # Optional Modal Secret objects with environment variables for the container
         gpu: Union[
             GPU_T, List[GPU_T]
@@ -1415,7 +1415,8 @@ class _Image(_Object, type_prefix="im"):
     ) -> "_Image":
         """Run user-defined function `raw_f` as an image build step. The function runs just like an ordinary Modal
         function, and any kwargs accepted by `@app.function` (such as `Mount`s, `NetworkFileSystem`s,
-        and resource requests) can be supplied to it.
+        and resource requests) can be supplied to it. The function should return `None`.
+
         After it finishes execution, a snapshot of the resulting container file system is saved as an image.
 
         **Note**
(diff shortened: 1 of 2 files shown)


modal/image.py
```diff
@@ -1396,7 +1396,7 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
 
     def run_function(
         self,
-        raw_f: Callable[..., Any],
+        raw_f: Callable[..., None],
```
thundergolfer, line 1399: This is a narrowing but I can't think of a case where returning something is desired. These should just be side-effect functions.

mwaskom:
I could imagine users doing something like this

```python
def load_model_with_caching():
    return transformers.AutoModel.from_pretrained("whatever-bert")


image = modal.Image.debian_slim().run_function(load_model_with_caching)

...

@app.function()
def do_inference(input):
    model = load_model_with_caching()
    model(input)
```

Similar to our advertised "stack `@modal.build` / `@modal.enter` pattern, to take advantage of huggingface's caching during the Image build.

Still it's helpful to clarify that we don't use the return value. Maybe that could just be done with narrative documentation? Not sure.

thundergolfer:
Yeh I think I've even done something similar myself. But the problem is that in the `run_function` case Modal's client will actually try pickle that return value and send it as an output, which is undesirable.

mwaskom:
> But the problem is that in the run_function case Modal's client will actually try pickle that return value and send it as an output, which is undesirable.

I guess it wouldn't be trivial not to do this since the container doesn't know it's running a "build function"?

thundergolfer:
> I guess it wouldn't be trivial not to do this since the container doesn't know it's running a 

It might actually be easy enough. I think the container can access the Function definition which has something like `is_build_function`

freider:
Good point, if it's not too hard structurally in the container entrypoint we should add something that just returns None as the result value when running a builder function! I've ran into this issue where builds take ages because of doing things like `run_function(load_model_with_caching)` like above to reuse the same model loading function without declaring a separate builder function

mwaskom:
Going to close here and track the suggested change to container entrypoint in https://linear.app/modal-labs/issue/CLI-265/dont-try-to-serialize-builder-function-outputs

PR #2429, opened 2024-10-30
Remove old container heartbeat fields and add new ones.

danielshaar:
Prereq for WRK-11.

This cleans up fields that are no longer used on the server or by any client version. We also report whether the client supports returning terminated outputs for canceled inputs, so we can avoid killing the task.

There's also indefinite retries for the container heartbeat since the canceled inputs should be idempotent.

Won't submit till after [e0508fa](https://github.com/modal-labs/modal/pull/16889) is in all workers.


freider (approved):
Love it

PR #2430, opened 2024-10-30
Remove _experimental_boost proto field

irfansharif:
Unused now - grouped functions just use the new gang-scheduler unconditionally.


mwaskom (approved):
Bye!

PR #2431, opened 2024-10-30
Adds new new enum `ProxyType`

luiscape:
The new enum allows us to differentiate between the different proxy systems offered to users. The "legacy" proxy system is being deprecated and all new proxies will use the new system ("vprox").


luiscape:
@prbot approve

mwaskom (approved):
Looks good!

PR #2434, opened 2024-10-31
Adds py.typed to get type safe usage of modal_proto package

freider:
This marker is required by many (all?) type checkers to treat the import as something other than "Any"


danielnorberg:
Nice!

PR #2437, opened 2024-11-01
Move fixture logic from ClientHello to the middleware

erikbern:
Broke this out of a PR I have that removes `ClientHello` just so I can merge it in a few smaller steps.


mwaskom (commented):
✅

PR #2438, opened 2024-11-01
Simplify `from_env` test

erikbern:
Break up big test in 3 tests

1. We reset the env client using an auto-fixture anyway, no need to reset it explicitly
2. With `ClientHello` going away in the future, we won't have the ability to "verify" an env client on creation, resetting it if it doesn't work. I think this is a bit of a weird use case anyway – in theory it's a singleton determined by its environment, so it shouldn't change – the user can always reset it manually if needed. We don't need to reset it automatically on connection failures.

### DIFF
--- a/test/client_test.py
+++ b/test/client_test.py
@@ -138,6 +138,13 @@ def test_client_from_env_failing(servicer, credentials):
         client_from_env("https://foo.invalid", credentials)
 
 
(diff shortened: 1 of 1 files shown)


erikbern:
@prbot approve

mwaskom (commented):
:+1:

PR #2439, opened 2024-11-01
Remove ClientHello on client construction

erikbern:
Will let this sit for a few days – the server side isn't quite ready for it

Should remove 2 server round trips for most uses of Modal – both ephemeral runs and calling deployed apps.

### DIFF
--- a/modal_version/_version_generated.py
+++ b/modal_version/_version_generated.py
@@ -1,4 +1,4 @@
 # Copyright Modal Labs 2024
 
 # Note: Reset this value to -1 whenever you make a minor `0.X` release of the client.
-build_number = 44  # git: cb01050
+build_number = 45  # git: 3e8e42d


test/shutdown_test.py
```diff
@@ -64,6 +64,7 @@ async def _mocked_logs_loop(client: Client, app_id: str):
         await t
 
     with Client(servicer.client_addr, api_pb2.CLIENT_TYPE_CLIENT, credentials) as client:
+        await client.hello.aio()  # TODO(erikbern): why is this needed?
```
freider, line 67: I looked at this the other week I think - if I'm not mistaken this was due to grpclib's channels being lazily connected by the first rpc call. Removing the hello from the client opening sequence made it instead set up the channel in the user space event loop used by `_mocked_logs_loop` here, which invalidates this test (but probably has no real effect on users of the SDK).

This test was mostly just in place for internal usage warnings during development anyway (the warnings etc.).
An alternative fix would be to make sure that `Client._open` still calls `channel.__connect__` to set up the connection (or something like that).

Not related to this PR, but I think we should just prohibit cross-event loop usage of the client altogether instead of trying to be accomodating here.

erikbern: > An alternative fix would be to make sure that `Client._open` still calls `channel.__connect__` to set up the connection (or something like that).

This seems useful in itself, I can take a look at this!

erikbern: Thanks for explanation!

freider (approved):
nice!

erikbern:
Very confused what happened, but Github just decided to close this PR

erikbern:
reopening

PR #2448, opened 2024-11-04
Replace macos-12 CI runner with macos-13

kramstrom:
[maco-12 runner is being deprecated](https://github.com/actions/runner-images/issues/10721) with brownouts scheduled beforehand -- all jobs running on the runner will fail and block our CICD

Replacing it with macos-13 which also runs on `x86-64`

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -28,6 +28,9 @@ jobs:
         os: ["ubuntu-20.04"]
         proto-version: ["latest"]
         include:
+          - os: "macos-13" # x86-64
+            python-version: "3.10"
+            proto-version: "latest"
           - os: "macos-14" # ARM64 (M1)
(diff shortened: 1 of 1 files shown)


howard36 (approved):

kramstrom:
@prbot automerge

freider (approved):

mwaskom:
I wonder if we even need to test on 13/14 or if just the latest supported macos is enough? Not familiar with the backstory for why we previously did both 12 and 14 though.

kramstrom:
> I wonder if we even need to test on 13/14 or if just the latest supported macos is enough? Not familiar with the backstory for why we previously did both 12 and 14 though.

I imagine it's because we want to catch the x86 architecture, 14+ (unless you do large containers) all run on arm

mwaskom:
Oh good point!

PR #2449, opened 2024-11-04
Upgrade CI runner to MacOS 13

howard36:
The `macos-12` github actions runner is now deprecated. According to [this table](https://github.com/actions/runner-images?tab=readme-ov-file#available-images), `macos-13` uses `x86-64`, so we should use that instead.


howard36:
Oops, didn't see #2448

PR #2451, opened 2024-11-04
Implement aclosing without a generator

freider:
Not sure exactly why, but using a generator with `@asynccontextmanager` to implement it leads to a potential issues with `generator didn't stop after athrow()` which I can't reproduce in isolation, but appear when running many tests. 

This seems to fix it  🤷 

I'm thinking the issue is probably something either exiting the context aclosing context manager early, cancellation or gc of the underlying generator.


kramstrom (approved):
🪄

mwaskom:
Nice, this fixes something I was just about to report :D

PR #2454, opened 2024-11-04
Connect socket in create_channel

erikbern:
`grpclib` connects the socket lazily, but we can force it to connect on creation (and not on the first call) by calling an internal method. This solves a weird test failure in https://github.com/modal-labs/modal-client/pull/2439#pullrequestreview-2412843577

Once we remove `ClientHello`, checking connection upfront is I think a bit better since it will fail in the proper context and not later. It means we can still fail fail on client construction in many cases (although some issues like failing credentials will be raised lazily). It means we can remove a few dummy hello calls in tests in #2439

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -70,7 +70,7 @@ RETRYABLE_GRPC_STATUS_CODES = [
 ]
 
 
-def create_channel(
+async def create_channel(
     server_url: str,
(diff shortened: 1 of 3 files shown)


freider (approved):
👍

erikbern:
Turns out we're using `create_channel` in a bunch of internal code, so it's a bit annoying to change the signature (to `async`)

PR #2455, opened 2024-11-04
Improve base image deployment script

mwaskom:
Closes CLI-221. Originally, I just wanted to prevent the build from failing on incompatible python/builder versions. But I noticed that the Debian Slim and Micromamba base images were basically the exact same code, so I consolidated them into one script. And we also now build all of the Python versions in parallel in a single script / app, reducing the number of concurrent jobs in the GitHub CI matrix. Should speed up CI a tiny bit. In theory I could have parallelized across the base image too, but I think this way will make it slightly easier to diagnose any issues (otherwise, with the interleaved build logs, it would be hard to know where to start).

Also enables deployments of the 2024.10 base image builds.


mwaskom:
We don't really need this script nested under `modal_global_objects/images` anymore but I didn't want to get into that :)

mwaskom:
@prbot approve; this is making our CI show up as failing (although deploys are going out)

modal_global_objects/images/base_images.py
```diff
@@ -0,0 +1,28 @@
...
+    _, name = sys.argv
+    constructor = getattr(modal.Image, name)
+
+    builder_version = os.environ.get("MODAL_IMAGE_BUILDER_VERSION")
+    assert builder_version, "Script requires MODAL_IMAGE_BUILDER_VERSION environment variable"
+    python_versions = SUPPORTED_PYTHON_SERIES[cast(ImageBuilderVersion, builder_version)]
+
+    app = modal.App(f"build-{name.replace('_', '-')}-image")
+    for v in python_versions:
+        app.function(image=constructor(python_version=v), name=f"{v}")(dummy)
```
freider, line 24: h/u that `name` is kind of broken when not using `serialized=True` since it is used to look up the actual function in code too at the moment, but should be fine in this case since never actually *run* these functions, which would crash the container otherwise

freider:
lgtm

PR #2456, opened 2024-11-05
Manage output state in rich-safe output module

mwaskom:
This PR addresses an issue with the 2024.10 Image Builder version that one can encounter when running a `Function.map` remotely. It adds some tracking of the output-enabled state in the rich-safe `modal.output` module so that we only try to import the richified `modal._output` module when output has been explicitly enabled.

I am still formulating a plan for improving rich safety — I think with the right refactor we can make rich totally optional and thus have it be a lot more safe to potentially run code that would try to use rich in a container environment. I don't love the global variable approach here. But I think it's simple enough and addresses an acute issue.


mwaskom:
@prbot approve

modal/output.py
```diff
@@ -13,6 +13,9 @@
     from ._output import OutputManager
 
 
+OUTPUT_ENABLED = False
```
erikbern, line 16: confused why this is needed vs checking the singleton?

erikbern (approved):
seems a bit hacky but i'm sure this is indicative of some cleanup we should do later

modal/output.py
mwaskom: Right now we can't check the singleton without importing the `_output` module, which has a ton of rich imports in global scope. We could move those imports into the relevant function bodies (and leave some guarded under `if TYPE_CHECKING` for rich types used in the `OutputManager` interfaces) but that's cumbersome/hacky in its own way.

I'd prefer to turn the `OutputManager` into more of a _manager_ with no display responsibilities itself, and make the code that uses rich directly more modular and easier to replace with plain versions in contexts where rich isn't available.

PR #2457, opened 2024-11-05
Add flag to certain RPCs to indicate no usage of method placeholders

devennavani:
Add flag `only_class_function` to certain RPCs to indicate (if `True`) no usage of method placeholders on the `Cls`. These are defaulting to `False` for now, and the client PR where I flip the switch so that the `Cls` object doesn't need hydrated method placeholder functions on it, this flag will be passed as `True`.


devennavani:
@prbot approve

freider:
✅

PR #2458, opened 2024-11-05
Remove WRK-241 fix and rely on worker-side fix

thundergolfer:
## Describe your changes

- See #17074 on the worker side. It's a much simpler way to fix this problem.
- 🚧 **Don't merge this** until worker-side change merges and rolls out.
   -  **Update:** Worker-side changed merged at 3:20 PM EST 5th Nov.

---

---

</details>


luiscape (approved):
Much better indeed. I'm sorry I didn't think of it beforehand either. :)

PR #2461, opened 2024-11-05
Push terminated outputs before exiting the IO context.

danielshaar:
This logic is moved from inside of the container heartbeat where it incorrectly would inform of cancellation success too early.


modal/_container_io_manager.py
```diff
@@ -388,15 +392,18 @@ async def _heartbeat_handle_cancellations(self) -> bool:
...
-                    current_input_ids_to_cancel = [
-                        input_id for input_id in input_ids_to_cancel if input_id in self.current_inputs
-                    ]
-                    await self._push_terminated_outputs(current_input_ids_to_cancel)
-                    for input_id in current_input_ids_to_cancel:
-                        self.current_inputs[input_id].cancel()
-
-                elif self.current_input_id and self.current_input_id in input_ids_to_cancel:
+                    for input_id in new_input_ids_to_cancel:
+                        if input_id in self.current_inputs:
```
gongy, line 403: What if the container doesn't have the input yet? Does this swallow the cancellation forever

danielshaar: Rehashed the logic to use the existence of the cancel callback which is deregistered when the cancellation is in progress.

modal/_container_io_manager.py
```diff
@@ -389,12 +389,12 @@ async def _heartbeat_handle_cancellations(self) -> bool:
...
-                        input_id for input_id in input_ids_to_cancel if input_id in self.current_inputs
-                    ]
-                    await self._push_terminated_outputs(current_input_ids_to_cancel)
-                    for input_id in current_input_ids_to_cancel:
-                        self.current_inputs[input_id].cancel()
+                    for input_id in input_ids_to_cancel:
+                        if input_id in self.current_inputs:
+                            # Avoid issuing cancellation signals for the same inputs repeatedly.
+                            io_context = self.current_inputs[input_id]
+                            if io_context._cancel_callback:
```
gongy, line 396: not sure about private field read here, maybe we can adjust `IOContext.cancel()` to behave idempotently?

danielshaar: done

gongy (approved):
Looks great

PR #2465, opened 2024-11-05
[Multi Node] Add Clustered Functions

thecodingwizard:
This PR removes the old logic we had for grouped functions and replaces them with clustered functions, where input broadcasting and container broadcasting happens in the server.

See https://github.com/modal-labs/modal/pull/17130 for the server-side implementation.

See #2491 for the protobuf changes.


modal/_clustered_functions.py
```diff
@@ -0,0 +1,56 @@
+from modal._utils.async_utils import synchronize_api
+
+
+async def _initialize_clustered_function(cluster_id: str, cluster_rank: int, cluster_size: int):
+    print(f"Initializing clustered function with cluster ID {cluster_id}, rank {cluster_rank}, size {cluster_size}")
```
thecodingwizard, line 5: Temporary placeholder. I plan to make a blocking `ClusterHello` RPC that will return every container's rank + IP address in the cluster; we can figure out what to do with this information / how to expose it to the user later.

modal/_container_entrypoint.py
```diff
@@ -2,6 +2,8 @@
 # ruff: noqa: E402
 import os
 
+from modal._clustered_functions import initialize_clustered_function
```
mwaskom, line 5: Is there a reason we do this above all the other Modal imports?

modal/_container_entrypoint.py
```diff
@@ -816,6 +815,13 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
                 metadata: Message = container_app.object_handle_metadata[object_id]
                 obj._hydrate(object_id, _client, metadata)
 
+        # Initialize clustered functions.
+        if function_def._experimental_group_size > 1:
+            initialize_clustered_function(
+                client,
+                container_args.task_id,
+            )
```
mwaskom, line 823: j/w but did you consider getting the cluster info lazily the first time user code asks for it rather than as part of the container startup? Not really sure if that's better not; I'm just not totally certain if this is the right place to be doing it.

test/i6pn_clustered_test.py
```diff
@@ -39,17 +39,6 @@ def test_experimental_group(servicer, client):
...
 
 
-def test_spawn_experimental_group(client, servicer, monkeypatch):
-    # We need to set a custom function body here since grouped function's kwargs
-    # aren't compatible with the default servicer function body.
-    @servicer.function_body
-    def grouped_func(*args, **kwargs):
-        return sum(args)
-
+def test_spawn_experimental_cluster(client, servicer, monkeypatch):
```
mwaskom, line 42: Nit: but why "spawn" in the name if we are exercising `.remote`? (I know you're just modifying  an existing test here).

test/i6pn_clustered_test.py
```diff
@@ -39,17 +39,6 @@ def test_experimental_group(servicer, client):
...
     with app.run(client=client):
-        # Needed for type checker to be happy. This isn't the best solution, but
-        # the feature is experimental anyway...
-        assert isinstance(f1, modal.experimental.GroupedFunction)
-        f1.client = client
-
-        assert f1.remote(2, 4) == [6] * 2
+        assert f1.remote(2, 4) == 2**2 + 4**2
```
mwaskom, line 44: I'm confused where this result is coming from. Is it hardcoded into the mock servicer?

modal/_container_entrypoint.py
thecodingwizard: no, thanks for catching this!

test/i6pn_clustered_test.py
thecodingwizard: Good point 😅 I'll rename it to `test_run_experimental_cluster`

test/i6pn_clustered_test.py
thecodingwizard: Yeah. I'll add a comment

mwaskom (approved):

PR #2466, opened 2024-11-05
[CLI-223] Retry UNKNOWN status codes by default

gongy:
## Describe your changes

Closes CLI-223

---

a
---

</details>


rculbertson:
Since we already retry on INTERNAL errors, I think it's consistent to also retry on UNKNOWN. 

Will we also fix [the nginx issue](https://modal-com.slack.com/archives/C031K7RJLQZ/p1730821734881449) @samaross found?

I'll say that in general, I'm a little hesitant to automatically retry on INTERNAL and UNKNOWN errors. If a system is failing in unexpected ways, retrying can increase load and possibly make the problem worse. So in an ideal world, I'd say it would be better to fix the root cause - but that's easier said than done :)

freider:
Sounds good to me. Not sure why the Windows test failed, I haven't seen that specific flake before. Just re-triggered it now

freider (approved):

PR #2467, opened 2024-11-06
Include mounts from image build steps in "watch" mounts - CLI-219

freider:
## Changelog

* Fixes issue in `modal serve` where files used in `Image.copy_*` commands were not watched for changes


modal/app.py
```diff
@@ -478,11 +478,14 @@ def _get_default_image(self):
             return _default_image
 
     def _get_watch_mounts(self):
+        if not self._running_app:
+            raise InvalidError("`_get_watch_mounts` requires a running app.")
```
mwaskom, line 482: Should we be referencing a private method in a user-facing error? Can we catch this earlier?

freider: This is not supposed to be a user facing error, so maybe InvalidError is actually a bit misleading. Should basically never happen unless we mess up in the code calling `_get_watch_mounts` (which is only used by `modal serve`)

freider:
@prbot approve

mwaskom (approved):
LGTM

PR #2468, opened 2024-11-06
Make it possible to use the Resolver without rich installed

mwaskom:
The `Resolver` object uses some tools from `rich` directly, and it imports some utilities from the non-rich-safe `modal._output` module. This doesn't affect minimal apps but it is relevant for apps that use lazily-hydrated objects, which instantiate a `Resolver` when they get hydrated.

This PR adds some more guardrails around the imports so that we can hydrate objects remotely without rich. It's not my preferred solution — I don't think the `Resolver` should need to touch `rich` at all — but it's some more duct tape until we can more thoroughly refactor the output pipeline.

Also added a test that mocks an environment without `rich`. We could potentially use it in more places too.

- Fixes CLI-222

### DIFF
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -19,13 +19,11 @@ if TYPE_CHECKING:
 
 class StatusRow:
     def __init__(self, progress: "typing.Optional[Tree]"):
-        from ._output import (
-            step_progress,
-        )
-
         self._spinner = None
         self._step_node = None
         if progress is not None:
+            from ._output import step_progress
+
             self._spinner = step_progress()
             self._step_node = progress.add(self._spinner)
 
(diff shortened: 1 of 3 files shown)


mwaskom:
@prbot approve

erikbern (approved):
This seems fine but a bit janky to catch the `ImportError` imo.

I think maybe a cleaner solution would be to move all the resolver-related output to the `OutputManager` singleton instead, although that's a bit annoying for other reasons – there's no obvious mechanism to construct the tree. But maybe we could have some sort of context manager that creates temporary trees on the `OutputManager` singleton? idk.

Anyway seems fine to ship this for now, as long as we agree that the `Resolver` class is a big piece of tech debt and we should revisit it later.

PR #2469, opened 2024-11-06
Adds `ProxyGet` to public API

luiscape:
Introduces `ProxyGet` to public API. This precedes an update to `Proxy.from_name("my-proxy")` to replace `ProxyGetOrCreate`.

I intended to have that RPC support both the client get method and internal proxy create methods but came to the conclusion that it is better to separate those methods because it is not desirable to create a Proxy lazily. (That's the case because it takes several seconds to create.)

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1831,6 +1831,15 @@ message ProxyGetOrCreateResponse {
   string proxy_id = 1;
 }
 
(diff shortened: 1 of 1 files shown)


luiscape:
@prbot approve

mwaskom (approved):
Makes sense!

PR #2471, opened 2024-11-06
Make tag parameter non-optional in from_name and lookup for Function and Cls

devennavani:
Make tag parameter non-optional in from_name and lookup for Function and Cls

Neither `FunctionGet` nor `ClassGet` support `request.object_tag=None`.


mwaskom (approved):
LGTM; we should maybe think about how to change this name / make these postional-only arguments; users have gotten confused as this now collides with the "deployment tag" concept, and we don't otherwise use the "tag" terminology for "function name" AFAIK.

PR #2472, opened 2024-11-06
Add user telemetry protos

vishaalram02:
## Adds protos for sending user telemetry data to the worker.

- This PR resolves WRK-483

<details>
---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -112,6 +112,13 @@ enum DeploymentNamespace {
   DEPLOYMENT_NAMESPACE_GLOBAL = 3;
 }
 
+// Encoding type used for user telemetry data.
+enum EncodingType {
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -112,6 +112,13 @@ enum DeploymentNamespace {
   DEPLOYMENT_NAMESPACE_GLOBAL = 3;
 }
 
+// Encoding type used for user telemetry data.
```
pawalt, line 115: Maybe `TelemetryEncodingType` instead so we don't get this confused with other kinds of encodings.

pawalt (approved):

pawalt:
@vishaalram02 make sure to wait for tests before merging! If you want to have things merge automatically, you can use `@prbot automerge`

PR #2473, opened 2024-11-06
Add warnings on app publish response proto

vishaalram02:
## Add warning on app publish proto for deployed limits

- See SVC-88

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -412,6 +412,7 @@ message AppPublishRequest {
 
 message AppPublishResponse {
   string url = 1;
+  repeated string warnings = 2;
 }
 
 message AppRollbackRequest {


modal_proto/api.proto
```diff
@@ -412,6 +412,7 @@ message AppPublishRequest {
 
 message AppPublishResponse {
   string url = 1;
+  repeated string warnings = 2;
```
mwaskom, line 415: Will we want there to be any metadata for each warning (e.g., a warning class, or something like that?) I'm wondering if we want to be future-compatible and use some kind of repeated proto message. Not sure exactly how you plan to use this, though.

mwaskom (approved):
LGTM but see comment about whether string is the right type.

modal_proto/api.proto
vishaalram02: Makes sense! I think in this particular case there's only one type of warning (and at most one) so wanted to keep things simple. If we add more warnings in other rpc responses, I think we should revisit this.

PR #2474, opened 2024-11-06
Add timeout, workdir, secrets to sandbox exec

pawalt:
## Describe your changes

- This change aligns the sandbox exec interface with the create interface, allowing timeout, workdir, and secrets to be provided.

Resolves WRK-424

---

---

</details>

## Changelog

`Sandbox.exec` can now accept `timeout`, `workdir`, and `secrets`. See the `Sandbox.create` function for context on how to use these arguments.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -93,6 +93,9 @@ class _Sandbox(_Object, type_prefix="sb"):
                 "Specify a single GPU configuration, e.g. gpu='a10g'"
             )
 
+        if workdir is not None and not workdir.startswith("/"):
+            raise InvalidError(f"workdir must be an absolute path, got: {workdir}")
+
         # Validate volumes
         validated_volumes = validate_volumes(volumes)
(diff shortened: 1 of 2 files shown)


modal/sandbox.py
```diff
@@ -424,13 +430,24 @@ async def exec(
         ```
         """
 
+        if workdir is not None and not workdir.startswith("/"):
+            raise InvalidError(f"workdir must be an absolute path, got: {workdir}")
+
+        # Force secret resolution so we can pass the secret IDs to the backend.
+        for secret in secrets:
```
pawalt, line 437: I'm looking for advice here on how to hydrate these secrets. Overwriting `_hydrate_lazily` feels very wrong, but I couldn't figure out how to resolve these otherwise.

mwaskom:
Uh oh, that ellipsis thing is gonna get shipped into the changelog I think 😖

modal/sandbox.py
pawalt: Ah cool thanks. Done

pawalt:
@prbot approve

PR #2475, opened 2024-11-07
Add text and line buffering mode to sandbox exec

azliu0:
## Describe your changes

Resolves WRK-424. In `Sandbox.exec`, this change gives users the option to switch between byte and text reads. This change also gives users the option to switch between streaming output and single lines at a time.

## Changelog

`Sandbox.exec` now accepts arguments `text` and `by_line` for streaming output, which controls text output and line buffering.


modal/io_streams.py
```diff
@@ -116,7 +118,7 @@ def __init__(
     def file_descriptor(self):
         return self._file_descriptor
 
-    async def read(self) -> str:
+    async def read(self) -> Union[str, bytes]:
```
freider, line 121: Having a union type for the return type here would make type-checkers complain whenever someones uses this API without checking the return type.

Type safety could be solved by making `_StreamReader` into a `typing.Generic[T]` with T determined by which `text` mode it uses, but considering that our transfer protocol is already a "string" and all we do with `text=False` is re-encode it as hard-coded utf8, I'd lean towards skipping the `text=False` option entirely and let users do that encoding themselves if they really want bytes?

Somewhat unrealted - but it feels somewhat upside down that we use `str` as the basic transfer protocol for these "logs" instead of `bytes` - what happens if the output string of a command isn't a valid utf-8 string for example? Would that cause an error? And if so, `text=False` is a bit misleading for users too...

modal/sandbox.py
```diff
@@ -400,12 +400,12 @@ async def _get_task_id(self):
...
         *cmds: str,
-        # Deprecated: internal use only
-        pty_info: Optional[api_pb2.PTYInfo] = None,
+        pty_info: Optional[api_pb2.PTYInfo] = None,  # Deprecated: internal use only
         stdout: StreamType = StreamType.PIPE,
         stderr: StreamType = StreamType.PIPE,
-        # Internal option to set terminal size and metadata
-        _pty_info: Optional[api_pb2.PTYInfo] = None,
+        text: bool = True,  # Encode output as text
+        by_line: bool = False,  # Line-buffered output
```
freider, line 407: See above comment about potentially skipping the `text` argument here. Users can achieve the same behavior by just `.encode("utf8")`:ing the output instead, with the benefit of better type safety.

I'd also be in favor of keeping this a closely mirrored to the `subprocess.Popen` interface as possible, so perhaps `by_line` could be replaced by `bufsize` with only allowed values being `-1` ("auto") or `1` (line buffered), at least in the public-facing interface

freider: Also, should we add the same argument(s) to Sandbox.create too to keep them mirrored?

freider (changes requested):
Some thoughts. I don't have the background on this change so let me know if something doesn't make sense :)

azliu0:
> Am I missing something or is `by_line` not implemented here?

It gets passed into the `ContainerProcess` handle, which is subsequently passed into the `StreamReaders`.

modal/io_streams.py
azliu0: I agree! I'll look into changing the transfer protocol before merging this change. I think we'd like to keep the interface close to `subprocess.Popen`, which was the original incentive for adding `text`.

azliu0:
@pawalt @freider This is ready for another look. The changes in https://github.com/modal-labs/modal/pull/17149, https://github.com/modal-labs/modal/pull/17148 make it so that raw bytes are streamed back to the client by default (and should be fully rolled out at this point). I've opted not to add these to `Sandbox.create` as doing the same for the Sandbox logs touches more fundamental parts of the system—I've created a ticket to support this eventually, though.

freider (commented):
Some comments/questions

modal/io_streams.py
```diff
@@ -132,13 +141,30 @@ async def read(self) -> str:
         ```
 
         """
+        if self._object_type == "sandbox":
+            return await self._read_sandbox()
+        else:
+            return await self._read_container_process()
+
+    async def _read_sandbox(self) -> str:
```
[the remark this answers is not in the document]
azliu0: Unfortunately this isn't possible, unless I make the change that the sandbox logs iterator should return bytes by encoding the string output (see below comment). I agree that the logic depending on object_type is ugly, but the sandbox `_get_logs` always returns string, while the container process `_get_logs` always returns bytes.

modal/io_streams.py
```diff
@@ -244,21 +270,35 @@ async def _get_logs(self) -> AsyncGenerator[Optional[str], None]:
...
         """mdmd:hidden
         Processes logs from the server and yields complete lines only.
         """
         async for message in self._get_logs():
             if message is None:
-                if self._line_buffer:
-                    yield self._line_buffer
-                    self._line_buffer = ""
-                yield None
+                if self._object_type == "sandbox":
```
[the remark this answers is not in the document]
azliu0: I disabled line buffering for non-text mode to simplify this logic.

modal/io_streams.py
```diff
@@ -220,7 +246,7 @@ async def _get_logs(self) -> AsyncGenerator[Optional[str], None]:
         while not completed:
             try:
                 if self._object_type == "sandbox":
-                    iterator = _sandbox_logs_iterator(
+                    iterator: AsyncGenerator[Tuple[Optional[Union[T, str]], str], None] = _sandbox_logs_iterator(
```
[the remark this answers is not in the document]
azliu0: I've changed `_sandbox_logs_iterator` to encoding to simplify the branching logic

azliu0:
@freider this should be ready for another look. I've verified that static type checkers work correctly with the new spec in `type_assertions.py`. Let me know if you think it'd be appropriate to add a Sandbox `exec` test to that file, or if there's anything else you'd like to see changed

freider:
I think you need to add type annotations to the Sandbox.stdout properties etc. as well - I'm getting `Any` as the type on those, and it would be good to add some actual assertions to the test file as regression tests. Here is what I added at the end of the type_assertions.py to find the issue (could add some more I suppose):

```
# check sandboxes
sandbox = modal.Sandbox.create("dummy")
assert_type(sandbox.stdout.read(), str)

cmd = sandbox.exec("other")
typing_extensions.reveal_type(cmd)
assert_type(cmd.stdout.read(), str)

cmd2 = sandbox.exec("other_bin", text=False)
assert_type(cmd2.stdout.read(), bytes)

```

What about stdin btw? Should/is that also affected by the text flag?

azliu0:
Thanks, added the tests. By default, `stdin` accepts any type. This behavior seems ok to me for now, as we're only adding the `text` flag to `exec` and not `Sandbox.create`. If at some point we unify these two APIs, then it might make more sense to enforce this behavior for `stdin` as well.

freider (approved):
Thanks for the cleanup! Looks great!

I realized we don't actually check the implementation types unless we add the file to the pyright allowlist so I tried that locally and added a new PR on top of this one that fixes the few outstanding type annotations (found no implementation issues). I'll merge that after you merge this 👍

PR #2477, opened 2024-11-07
Typo in error msg

freider:
Method name is `from_local_environ` but error used other name


freider:
@prbot approve

mwaskom (commented):
Nice catch. I would if it would be friendlier to aggregate all missing keys before erroring out rather than doing it one-by-one. Not an issue with this PR obviously, just hadn't looked at the code before.

PR #2478, opened 2024-11-07
Add proto for streaming raw bytes

azliu0:
This change adds a proto field for serving raw bytes from the server rather than text. 

The upstream motivation for this is to eventually allow users more fine-grained control over streamed Sandbox output, e.g., by adding a `text` flag to choose to receive string (text) rather than raw bytes output. It didn't necessarily make sense to have the behavior for `text=False` to decode string back to bytes on the client.

---

---

</details>

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -49,7 +49,12 @@ async def _container_process_logs_iterator(
             yield None
             break
         for item in batch.items:
-            yield item.message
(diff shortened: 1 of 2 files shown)


azliu0:
@pawalt this should be good to look at now

pawalt (approved):
excellent. make sure to clean the changelog

PR #2479, opened 2024-11-07
Make the runner module rich-safe

mwaskom:
This PR makes it so that the `modal.runner` module can be imported in an Image that doesn't have `rich` installed. The main usecase is for our integration tests, though I could see it also being useful for Apps that do programmatic deployments.

To reduce the amount of deferred-import junk I converted the `step_*` helper functions to be staticmethods on the `OutputManager`, so that they can be called directly in contexts where we have just acquired the `OutputManager` singleton. There were a few other cases where they were re-used _without_ an `OutputManager`.

Not 100% sure what the best long-term solution is here. Should we always have an `OutputManager` when we're generating output? Or should we make it possible to import the step helpers directly, e.g. falling back to a static progress symbol when a `Spinner` is not a available.


mwaskom:
I also added a test for whether we can deploy apps without rich installed, which turned up a couple other places that we could potentially try to import rich in a container (if you're programmatically deploying apps). The easiest solution was to move the `FunctionCreationStatus` object, which is itself rich-safe, out of `modal._output` for now.

mwaskom:
@prbot approve, more rich hackery :)

erikbern (approved):

erikbern:
> Not 100% sure what the best long-term solution is here. Should we always have an `OutputManager` when we're generating output? Or should we make it possible to import the step helpers directly, e.g. falling back to a static progress symbol when a `Spinner` is not a available.

My feeling on this (but could be wrong) is that
1. Modal the library should never enable the output manager, so if you're calling Modal stuff directly, it doesn't print stuff unless you enable output.
2. All cli commands enable output
3. Modal running inside containers fall under point 1, which is maybe a bit confusing, but that way we def don't need Rich in containers

PR #2483, opened 2024-11-07
Adapt to typer 0.13 deprecation of `is_flag`

mwaskom:
Typer 0.13 deprecates the `is_flag` parameter of `typer.Option`. We use that in one place. As far as I can tell, this has no effect when the type of the option parameter is `bool`, as is the case where we use it. So it seems like a straightforward fix to remove it.

Closes CLI-234


mwaskom:
@prbot approve, unblocks deployments

mwaskom:
@prbot automerge

thecodingwizard (approved):
thanks!

PR #2486, opened 2024-11-08
Adds ProxyCreate proto definition

luiscape:
This adds `ProxyCreate` for creating proxies. This has the same motivations of https://github.com/modal-labs/modal-client/pull/2469: I am splitting the `ProxyGetOrCreate` RPC into two RPCs because the DX for creating proxies disallow them to be created lazily with `ProxyGetOrCreate`.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1857,9 +1857,8 @@ message Proxy {
   repeated ProxyIp proxy_ips = 4;
 }
 message ProxyCreateRequest {
-  string deployment_name = 1 [ (modal.options.audit_target_attr) = true ];
-  DeploymentNamespace namespace = 2;
-  string environment_name = 3;
+  string name = 1 [ (modal.options.audit_target_attr) = true ];
+  string environment_name = 2;
 }
 
 message ProxyCreateResponse {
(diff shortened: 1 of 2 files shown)


luiscape:
@prbot approve

luiscape:
@prbot automerge

modal_proto/api.proto
```diff
@@ -1856,6 +1856,15 @@ message Proxy {
   string proxy_id = 5;
   repeated ProxyIp proxy_ips = 4;
 }
+message ProxyCreateRequest {
+  string deployment_name = 1 [ (modal.options.audit_target_attr) = true ];
+  DeploymentNamespace namespace = 2;
```
ekzhang, line 1861: Can we get rid of DeploymentNamespace? We're using it for global objects, but I think we should perhaps deprecate that unless we really need it

ekzhang:
@prbot cancel

modal_proto/api.proto
luiscape: I think we should. It's not useful for proxies. Dicts, Mounts, Queues, and Proxies take the `namespace` parameter in `from_name()`. Do you think I should remove it now?

luiscape: Removed it from both the proto and `Proxy.from_name()`. :)

PR #2490, opened 2024-11-09
Re-applies Proxy URL

luiscape:
Path is now merged in docs.


mwaskom (approved):

mwaskom (changes requested):
Looks like https://modal.com/docs/guide/proxy does not exist yet

PR #2491, opened 2024-11-10
[Multi Node] Add `TaskClusterHello` RPC

thecodingwizard:
This PR adds the `TaskClusterHello` RPC needed for multi-node clusters. See #2465 for more details.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2375,9 +2375,8 @@ message TaskClusterHelloRequest {
 
 message TaskClusterHelloResponse {
   string cluster_id = 1;
-  uint32 cluster_size = 2;
-  uint32 cluster_rank = 3;
-  repeated string container_ips = 4;
+  uint32 cluster_rank = 2;
+  repeated string container_ips = 3;
 }
 
 message TaskCurrentInputsResponse {
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2368,6 +2368,18 @@ message SystemErrorMessage {
   string error_message = 2;
 }
 
+message TaskClusterHelloRequest {
+  string task_id = 1;
+  string container_ip = 2;
+}
+
+message TaskClusterHelloResponse {
+  string cluster_id = 1;
+  uint32 cluster_size = 2;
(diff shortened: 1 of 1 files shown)


mwaskom:
One not-super-important comment is that we're in the process of getting rid of `ClientHello` which is the only existing "Hello" verb RPC. Not sure if there's a different category that would make sense to group this with, though.

modal_proto/api.proto
```diff
@@ -2368,6 +2368,17 @@ message SystemErrorMessage {
   string error_message = 2;
 }
 
+message TaskClusterHelloRequest {
+  string task_id = 1;
+  string container_ip = 2;
+}
+
+message TaskClusterHelloResponse {
```
thundergolfer, line 2376: cluster members are started at the same time but they won't execute this RPC at the same time, and it seems plausible that the last hello could arrive >60s after the first hello, causing an RPC timeout because not all `container_ips` are available.

If this is a problem in practice, the server could stop polling redis after `timeout - 5s` and return `FAILED_PRECONDITION` or some other retry-able status when not all members say hello in time, or we could split out "hello" from a new "get cluster IPs" RPC.

modal_proto/api.proto
```diff
@@ -2368,6 +2368,17 @@ message SystemErrorMessage {
...
+  string cluster_id = 1;
+  uint32 cluster_rank = 2;
+  repeated string container_ips = 3;
```
thundergolfer, line 2379:
```suggestion
  All IP addresses in cluster, ordered by cluster rank
  repeated string container_ips = 3;
```

thundergolfer (approved):
LGTM. 

To respond to @mwaskom's comment. I think "Hello" still works best here, and we do have other "Hello" RPCs in other APIs (e.g. internal).

modal_proto/api.proto
thecodingwizard: > If this is a problem in practice

I don't think it should be -- 60 seconds is more than enough.

Currently, the server implementation stops polling Redis after ~55 seconds, then returns `DEADLINE_EXCEEDED` and the client will retry.

PR #2492, opened 2024-11-12
Adds client side ImageMetadata

freider:
Prep for supplying image metadata from successful ImageJoins, which can be useful for the client to take conditional measures in subsequent build/run steps (like warning users of missing requirements, or adjusting commands based on workdir)

This somewhat matches the format of the metadata we already collect on the backend after @thecodingwizard 's recent patch, but unpacks the raw json-formatted info into structured proto, avoiding having to do that parsing in the client and allowing an easier migration if we want to change the format we store data data on the backend.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1706,21 +1706,6 @@ message ImageGetOrCreateResponse {
   string image_id = 1;
 }
 
-message PythonPackageVersions {
-  // wrapper around map to get presence for the field
-  map<string, string> packages = 1;
-}
-
-message ImageMetadata {
-  // The output of `python -VV. None if missing (not recently built, or failed collection)
-  optional string python_version_info = 1;
-  // Installed python packages, as listed by by `pip list`.
-  // package name -> version. Not present if missing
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1926,6 +1938,11 @@ message ProxyListResponse {
   repeated Proxy proxies = 1;
 }
 
+message PythonPackageVersions {
```
mwaskom, line 1941: j/w but why does this need its own message class?

mwaskom (approved):

modal_proto/api.proto
freider: It was to be able to differentiate empty map from missing map, but I just changed that since I realized we store the map as a PGDict anyways on the backend which doesn't differentiate those. Might add a boolean in the future if we want "presence" for the field

mwaskom: Got it. I don't think we should ever have an "true" empty map of third-party Python packages given that we should install the client deps in all base images right? Or maybe I'm missing something.

PR #2493, opened 2024-11-12
Image metadata hydration

freider:
Followup to #2492 

Actually store the image metadata on the Image object after hydration.

~Note: server changes are not live yet to return this data, but it's not required to exist so should be fine to merge this anyway~ Server changes are now live


freider:
@prbot approve

freider:
@prbot auto-merge

mwaskom (commented):
Nice! Excited for what we can do with this.

PR #2494, opened 2024-11-12
Suppress some previously deprecated methods from API docs

mwaskom:
Housekeeping PR to hide some fully-deprecated (i.e., erroring out) methods from our API docs. I also standardized some of the messages and reformatted a few to show up better in the AST-based `inv show-deprecations` helper.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -126,6 +126,38 @@ class _FunctionDecoratorType:
         ...
 
 
+_app_attr_error = """\
+App assignments of the form `app.x` or `app["x"]` are deprecated!
+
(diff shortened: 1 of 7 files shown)


mwaskom:
Could also be convinced that it's time to remove a couple of these fully, but I think it's friendly to keep the errors around for a while in case people try to revive old code (or use an LLM with a stale knowledge base or something).

mwaskom:
@prbot approve (docs changes only)

erikbern:
Nice. Some of these are super old and we can probably remove them at this point!

PR #2495, opened 2024-11-13
Support ImageGetOrCreate returning build results and skipping ImageJoinStreaming

freider:
If an image is already built we let that get returned by ImageGetOrCreate instead of doing another roundtrip with ImageJoinStreaming. 

If a result is not provided, we still go through ImageJoinStreaming as usual, but this lets us shave of one RPC per image build layer at deploy/run time


tasks.py
```diff
@@ -149,6 +149,7 @@ def type_check(ctx):
         "modal/_utils/shell_utils.py",
         "test/cls_test.py",  # see mypy bug above - but this works with pyright, so we run that instead
         "modal/_container_io_manager.py",
+        "modal/image.py",
```
freider, line 152: I ran pyright on image.py while developing this to give some more confidence in not breaking in branching of the code that might not be tested, so I fixed the few type errors in that file while I was at it and added it to the allow list here so we check it going forward

modal/image.py
```diff
@@ -323,17 +361,18 @@ def _from_args(
         if build_function and len(base_images) != 1:
             raise InvalidError("Cannot run a build function with multiple base images!")
 
-        def _deps() -> List[_Object]:
-            deps: List[_Object] = list(base_images.values()) + list(secrets)
+        def _deps() -> Sequence[_Object]:
```
freider, line 364: type fixes - `list` is invariant and doesn't allow subtypes in the container typing-wise, so I changed it to a `Sequence` which is covariant (and used tuples internally for the same reason). a bit annoying

modal/image.py
```diff
@@ -264,6 +263,45 @@ class DockerfileSpec:
     context_files: Dict[str, str]
 
 
+async def _image_await_build_result(image_id: str, client: _Client) -> api_pb2.ImageJoinStreamingResponse:
```
freider, line 266: Mostly moved as is from below, to make code at call site more readable after branching

mwaskom (approved):
Great!

PR #2498, opened 2024-11-13
Improve lookup/from_name docs on Modal objects

mwaskom:
I don't think we explain the distinction between `.lookup` and `.from_name` anywhere in our docs.

We've had various conversations about how to avoid having this dichotomy (which is somewhat easier now that nearly all objects support lazy hydration) but in the meantime I think it's helpful to explain a little bit about the difference.

I don't think the explanation is _super_ helpful because I'm not certain the average user will be able to reason from the explanation that we give to the practical implications for their code. Which is why it's generally preferable to just have One Way To Do It. But hopefully the examples help a little bit in terms of guiding usage.

I also took the opportunity to modernize + standardize some of the other doctstring copy and examples, along with Jonofication of the capitalization.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -288,12 +288,14 @@ class _App:
         environment_name: Optional[str] = None,
         create_if_missing: bool = False,
     ) -> "_App":
-        """Look up an app with a given name. When `create_if_missing` is true,
-        the app will be created if it doesn't exist.
+        """Look up an App with a given name, creating a new App if necessary.
+
+        Note that Apps created through this method will be in a deployed state
+        but will not have any Functions or Classes assocaited with them. This
+        is mainly useful for creating an App to associate with a Sandbox:
 
         ```python
         app = modal.App.lookup("my-app", create_if_missing=True)
-
         modal.Sandbox.create("echo", "hi", app=app)
         ```
         """
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -586,6 +586,8 @@ class _Mount(_Object, type_prefix="mo"):
         namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
         environment_name: Optional[str] = None,
     ) -> "_Mount":
+        """mdmd:hidden"""
+
         async def _load(provider: _Mount, resolver: Resolver, existing_object_id: Optional[str]):
             req = api_pb2.MountGetOrCreateRequest(
                 deployment_name=label,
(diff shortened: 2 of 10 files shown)


modal/mount.py
```diff
@@ -586,6 +586,8 @@ def from_name(
         namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
         environment_name: Optional[str] = None,
     ) -> "_Mount":
+        """mdmd:hidden"""
```
mwaskom, line 589: I don't think this is useful as a public-facing method?

modal/secret.py
```diff
@@ -206,13 +210,7 @@ async def lookup(
...
         required_keys: List[str] = [],
     ) -> "_Secret":
-        """Lookup a secret with a given name
-
-        ```python
-        s = modal.Secret.lookup("my-secret")
-        print(s.object_id)
-        ```
-        """
+        """mdmd:hidden"""
```
mwaskom, line 213: I don't think users have any reason to `.lookup` a Secret? Could be wrong though...

modal/app.py
```diff
@@ -288,12 +288,14 @@ async def lookup(
         environment_name: Optional[str] = None,
         create_if_missing: bool = False,
     ) -> "_App":
-        """Look up an app with a given name. When `create_if_missing` is true,
-        the app will be created if it doesn't exist.
+        """Look up an App with a given name, creating a new App if necessary.
+
+        Note that Apps created through this method will be in a deployed state
+        but will not have any Functions or Classes assocaited with them. This
```
danielshaar, line 294: associated*

danielshaar, line 294: nit, comma before but

mwaskom: thanks, a fun fact about me is that i literally cannot spell

danielshaar: nitheir cna i

mwaskom: I disagree! The text after "but" isn't an independent clause. :D

modal/secret.py
```diff
@@ -169,7 +169,11 @@ def from_name(
             str
         ] = [],  # Optionally, a list of required environment variables (will be asserted server-side)
     ) -> "_Secret":
-        """Create a reference to a persisted Secret
+        """Reference a Secret by its name.
+
+        In contrast to most other Modal objects, named Secrets must be provisioned
+        from the Dashboard. See other methods for alternate ways of creating a new
```
danielshaar, line 175: This sentence took me a sec to parse, maybe better: See other methods for alternative ways to create a new Secret from code.

mwaskom:
@prbot approve; just docs

erikbern:
👍 

thanks for doing this!

PR #2500, opened 2024-11-13
Drop support for Python 3.8

mwaskom:
## Changelog

- Modal no longer supports Python 3.8, which has reached its [official EoL](https://devguide.python.org/versions/).


test/image_test.py
```diff
@@ -1063,7 +1063,7 @@ def get_hash(img: Image) -> str:
         spec_file=test_dir / "supports" / "test-conda-environment.yml",
         channels=["conda-forge", "my-channel"],
     )
-    assert get_hash(img) == "d9d4c9fe24769ce587877b9752a64486e8f7d8520731110bd2fa666de82f43fd"
+    assert get_hash(img) == "f8701ce500d6aa1fecefd9c2869aef4a13c77ab03925333c011d7eca60bbf08a"
```
mwaskom, line 1066: This isn't a real change to the Image recipe, but the support `test-conda-environment.yml` support file we use had `python_version: 3.8`, which I've updated to avoid other issues.

.github/workflows/ci-cd.yml
```diff
@@ -78,10 +78,14 @@ jobs:
...
       matrix:
         image-builder-version: ["2024.04", "2024.10"]
-        python-version: ["3.8", "3.13"]
+        python-version: ["3.9", "3.13"]
     steps:
       - uses: actions/checkout@v3
 
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
```
mwaskom, line 87: Whoops, we were always running this check on 3.8 anyway...

erikbern:
looks good! Looking forward to removing the `from typing import` shortly

ekzhang (commented):
I know you asked people to ✔️ but I'll let @freider 😅

freider:
🪓

.github/workflows/ci-cd.yml
```diff
@@ -91,7 +103,13 @@ jobs:
         run: |
           python -m venv venv
           source venv/bin/activate
-          pip install grpcio-tools==1.59.2 grpclib==0.4.7
+          if [ "${{ matrix.python-version }}" == "3.9" ]; then
```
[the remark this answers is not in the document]
mwaskom: This check is just meant to assert that we haven't added any third-party dependencies to `python -m modal._container_entrypoint` that aren't in the expected container requirements, so the specific way that we compile the protos isn't super important because whatever choice we make wouldn't introduce transitive dependencies. But (as we're discussing on Slack) there definitely are larger and trickier questions about our proto packaging!

mwaskom:
> But I think you need to bump the minor version again (?)

Yeah the workflow where subsequent commits introduce a merge conflict with the "reset the micro version" diff is annoying. I've been meaning to automate that (i.e. just detect a minor version bump and handle the micro version accordingly) in the CD workflow but it doesn't happen quite often enough to make that feel worthwhile...

PR #2505, opened 2024-11-13
Misc cluster interface improvements

thecodingwizard:
- Throw `InvalidError` if `get_cluster_info` is called outside a cluster
- Make clustered functions behave the same if size == 1 and if size > 1


thecodingwizard:
@prbot approve

mwaskom (commented):
👍

PR #2509, opened 2024-11-14
Send method definitions in FunctionPrecreateRequest

devennavani:
## Describe your changes

This PR modifies the client to include the "method definitions" in the `FunctionPrecreateRequest` for a class service function. As a reminder, the method definition of a class method contains its webhook config, function name, and function type. Method definitions were introduced because with the removal of method placeholder functions we won't have a function definition per class method anymore.

This change is more or less a no-op. `FunctionPrecreate`, which was modified server-side in [16997](https://github.com/modal-labs/modal/pull/16997), will now return a `FunctionHandleMetadata` for each method of the class containing the function name, function type, and derived (not claimed) web url, but I haven't modified the client to do anything with this method handle metadata yet.

And if we send the `method_definitions` as a part of the request, it means we're working with a class service, so we don't need to send a `webhook_config`, since class functions don't have a webhook config.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -523,6 +523,9 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         _experimental_custom_scaling_factor: Optional[float] = None,
     ) -> None:
         """mdmd:hidden"""
+        # Needed to avoid circular imports
+        from .partial_function import _find_partial_methods_for_user_cls, _PartialFunctionFlags
+
         tag = info.get_tag()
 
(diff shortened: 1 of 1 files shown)


devennavani:
@prbot approve

mwaskom:
> And if we send the method_definitions as a part of the request, it means we're working with a class service, so we don't need to send a webhook_config, since class functions don't have a webhook config.

Non-constructive comment but it's sort of dissastisfying that we still have this complexity in the object model. Not sure I know what the answer is, though.

PR #2510, opened 2024-11-14
Add `_method_functions` attribute to `_Function`

devennavani:
## Describe your changes

This PR adds a `_method_functions` attribute to `_Function`. It is a mapping between method names (strings) and `_Function` objects. Only class service `_Function`s will have this mapping populated. Th use of this attribute will effectively replace the use of the current `_method_functions` attribute on `_Cls`. For example, `MyClass.foo` will look up the `foo` key in the `_method_functions` mapping on the class service function for `MyClass`.

This mapping will be populated in the `_Function.from_args` / `_Function.lookup` invocations for the class service function. As of this PR each `_Function` value of the mapping will be un-hydrated. Further PRs will add the code to hydrate these objects based on the `FunctionHandleMetadata` returned by the server for each class method. Note that these `_Function` objects will not correspond to unique functions on the backend, e.g. the `object_id` of all the `_Function`s in the mapping will be the function ID of the class service function.

`_bind_method_old` will go away with the client version that stops creating method placeholder functions.


devennavani:
@prbot approve

modal/functions.py
```diff
@@ -954,6 +988,12 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
         obj._is_method = False
         obj._spec = function_spec  # needed for modal shell
 
+        if info.user_cls:
+            obj._method_functions = {}
```
freider, line 992: I feel like the `_method_functions` should be living on the `_Cls` object instead, since _Function is quite large and complicated as it is, and the methods would only ever get used through the class anyway.

modal/functions.py
```diff
@@ -338,6 +339,40 @@ def _bind_method(
     ):
         """mdmd:hidden
 
+        Creates a _Function that is bound to a specific class method name. This _Function is not uniquely tied
+        to any backend function -- its object_id is the function ID of the class service function.
```
freider, line 343: I don't think the new "methods" should even have an object id or be fully fledged `_Function`s (apart from their functionality).

The outline of what I'd like to do is roughly:

* Pull out all the user-callable `_Function` functionality (.remote, .map etc.) into its own base class `_AbstractFunctionActions` with an abstract `_get_invocation_metadata() -> (use_object_id, use_method_name)` method that's used by all places that make function calls etc.
* Make both `_Function` and `_MethodHandle` inherit from `_AbstractFunctionActions` and override `_get_invocation_metadata` in their own way
* Class/instance service functions would be `_Function`s (with object ids and load-hydration etc.) but class methods would be "dumb" `MethodHandle` objects that "hydrate" using data from their service functions.

I think a structure like that would let us separate more of the class-specific from the function logic to make the latter less complicated and fragile.

freider (commented):
I think I'd like to take this a slightly different route. Maybe we should book a call to sync on how to proceed?

PR #2512, opened 2024-11-14
Update FunctionData proto with recently added Function fields

irfansharif:
Picking up the Function -> FunctionData migration work again. Catching up to some recent fields we added to the former, fields that will get "moved up" to the latter.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1328,6 +1328,11 @@ message FunctionData {
   uint32 warm_pool_size = 4;
   uint32 concurrency_limit = 5;
   uint32 task_idle_timeout_secs = 6;
+  // When the function is a "grouped" one, this records the # of tasks we want
+  // to schedule in tandem.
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -1350,13 +1361,13 @@ message FunctionData {
...
 
-  // When the function is a "grouped" one, this records the # of tasks we want
-  // to schedule in tandem.
-  uint32 _experimental_group_size = 19;
-
   Schedule schedule = 20;
 
   reserved 21;
+
+  bool untrusted = 27; // If set, the function will be run in an untrusted environment.
```
mwaskom, line 1368: Did we ever end up using this btw?

mwaskom (approved):

modal_proto/api.proto
irfansharif: Not yet, never trust @pawalt with anything

PR #2514, opened 2024-11-14
Hydrate `_method_functions` attribute on `_Function`

devennavani:
## Describe your changes

This PR adds code to hydrate each of the `_Function` objects in the newly added `_method_functions` mapping on `_Function`.

Note that `_Function._method_functions` is not yet being used anywhere.

This hydration occurs after the results of `FunctionPrecreate`, `FunctionCreate`, and `FunctionGet` are received. This PR is a no-op for the latter two, since the client is not yet sending `method_definitions` in the definition for the class service function, so `FunctionHandleMetadata.method_handle_metadata` from the server's RPC response will be empty.

For `FunctionPrecreate`, `FunctionHandleMetadata.method_handle_metadata` will be populated since we're [now sending the method definitions](https://github.com/modal-labs/modal-client/pull/2509) in `FunctionPrecreateRequest`, so `_Function._method_functions` will at minimum be hydrated with function names+types and web urls.


modal/functions.py
```diff
@@ -517,6 +517,26 @@ def _deps():
         fun._spec = class_bound_method._spec
         return fun
 
+    def _hydrate_function_and_method_functions(
```
devennavani, line 520: changed my mind on this, renamed it to _hydrate so that it overrides `_Object`'s `_hydrate` method, in this PR: https://github.com/modal-labs/modal-client/pull/2525

this is because sometimes we call obj._hydrate and I don't want to have to check `isinstance(obj, Function` and then call `obj._hydrate_function_and_method_functions`. it's cleaner to always call `._hydrate` and just have this special logic for `_Function`.

freider:
Discussed over hangout 👍

PR #2515, opened 2024-11-15
Add log when receiving cancellation signal for a single input.

danielshaar:
Missed this case when adding the log in the IOContext cancel() call. For single inputs, we just issue a SIGUSR1 and the signal handler then does the task cancellation. Everything works as intended, just missing the log.


danielshaar:
@prbot approve

PR #2522, opened 2024-11-17
snapshots: handle file indexes in CheckpointInfo

thundergolfer:
## Describe your changes

Prerequisite of WRK-186

---

    - yes old servers will not deserialize the new field
    - just adds a new bool, which is fine to be false by default

---

</details>


mwaskom (approved):

cweld510 (commented):
LGTM

PR #2523, opened 2024-11-17
Remove dead protos AppDeploySingleObject

erikbern:
While tearing down "dummy apps" I randomly noticed this dead RPC which isn't used anywhere.

It was added in #1009 but I'm not sure why.


freider (approved):
Don't know why we had it 🤷

PR #2525, opened 2024-11-18
Have _Function override _Object _hydrate method

devennavani:
Have _Function override _Object _hydrate method

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -523,25 +523,22 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         fun._spec = class_bound_method._spec
         return fun
 
-    def _hydrate_function_and_method_functions(
-        self, function_id: str, client: _Client, handle_metadata: api_pb2.FunctionHandleMetadata
-    ):
-        self._hydrate(function_id, client, handle_metadata)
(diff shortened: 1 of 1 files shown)


modal/functions.py
```diff
@@ -523,25 +523,22 @@ def _deps():
         fun._spec = class_bound_method._spec
         return fun
 
-    def _hydrate_function_and_method_functions(
-        self, function_id: str, client: _Client, handle_metadata: api_pb2.FunctionHandleMetadata
-    ):
-        self._hydrate(function_id, client, handle_metadata)
+    def _hydrate(self, object_id: str, client: _Client, metadata: Optional[Message]):
```
freider, line 526: iirc `_hydrate_metadata` is the method you should be overriding in object sub classes to customize how they use the metadata message, without affecting the "base logic" for hydration that could potentially change

freider: (but to hydrate an object you still call `._hydrate()` (the object implementation then calls into _hydrate_metadata for the custom logic)

freider: Also note this comment in separate PR: https://github.com/modal-labs/modal-client/pull/2510#discussion_r1846186256

freider (changes requested):

PR #2526, opened 2024-11-18
Fix mock `FunctionPrecreate` to only return web url for class method that has webhook config

devennavani:
Fix mock `FunctionPrecreate` to only return web url for class method that has webhook config

Analogous to the fix [here](https://github.com/modal-labs/modal/pull/17387)


devennavani:
@prbot approve

freider (commented):
👍

PR #2529, opened 2024-11-18
Add `_method_handle_metadata` attr to `_Function`

devennavani:
This is so that the `Cls` object can use `class_service_function._method_handle_metadata` to hydrate its `method_functions` mapping

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -329,6 +329,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
     _parent: Optional["_Function"] = None
 
     _class_parameter_info: Optional["api_pb2.ClassParameterInfo"] = None
+    _method_handle_metadata: Optional[Dict[str, "api_pb2.FunctionHandleMetadata"]] = None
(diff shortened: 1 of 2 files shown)


devennavani:
@prbot approve

modal/functions.py
```diff
@@ -1261,6 +1262,7 @@ def _hydrate_metadata(self, metadata: Optional[Message]):
         self._use_function_id = metadata.use_function_id
         self._use_method_name = metadata.use_method_name
         self._class_parameter_info = metadata.class_parameter_info
+        self._method_handle_metadata = dict(metadata.method_handle_metadata)
```
freider, line 1265: Might be cleaner at this point to just store the full `metadata` object on self - especially if we move out a bunch of the class-specific stuff to a separate non-_Object `MethodHandle` class - since most of these attributes would then never be used by _Function itself.

Could do that when we create that class and move out the `_bind_method()` to a constructor of that class instead. Just a thought

freider (commented):
cool!

PR #2532, opened 2024-11-18
don't require method name for modal shell on class

devennavani:
don't require method name for modal shell on class


mwaskom:
Looks like we fixed this elsewhere; closing here.

PR #2533, opened 2024-11-19
io streams types

freider:
Double checked some type stuff locally using pyright and figured I could just fix the few outstanding issues and add io_streams.py to the ci-checked pyright allowlist


freider:
@prbot approve

PR #2534, opened 2024-11-19
Quickfix for leaking implementation objects when using Modal objects as cls parameters

freider:
Still not something that should be encouraged, given that we'll most likely be deprecating custom constructor patterns in favor of the new structured parameterization...


mwaskom:
> given that we'll most likely be deprecating custom constructor patterns in favor of the new structured parameterization
are we tracking this in the 1.0 project btw? :D

freider:
@prbot approve

mwaskom (commented):
✅

PR #2536, opened 2024-11-20
Separate modules used within container into its own sub-package (modal.runtime)

freider:
Wanted/needed to refactor some stuff into new modules and the spreading of a bunch of underscore-modules at the top level package just itched too much.

Would be nice to also move _container_entrypoint but that's more complicated since it's referenced directly by the backend runtime.


freider:
> Could maybe try to make the underscoring of modules in `modal.runtime` consistent (should it be `modal._runtime` with internal modules not underscored? Or the reverse?) but that's obviously just fussy. Thanks for doing this :)

Makes sense! I'd opt to have the modal._runtime package underscore-prefixed in that case and the modules without th underscore (although I kind of want to get away from underscore module/packages at some point since they mess with my editor :P)

mwaskom:
> Makes sense! I'd opt to have the modal._runtime package underscore-prefixed in that case and the modules without th underscore (although I kind of want to get away from underscore module/packages at some point since they mess with my editor :P)

That sounds good to me and matches what we do with `_utis`. What's the issue with your editor though?

freider:
Running integration tests on this now to make sure I haven't accidentally broken anything else...

PR #2538, opened 2024-11-20
Container user code import tests

freider:
Starting to add some tests for various user code import cases - there is a lot of branching here depending on how functions are defined and we don't have great test coverage of that, since our previous tests have been more complicated container tests.

This is a step towards getting better test coverage.

Note the TODO with the service.app not being defined for class imports, which is an actual bug we would have caught with better test coverage here.


freider:
@prbot approve

devennavani (commented):
lgtm

PR #2539, opened 2024-11-20
Fix class imports not setting service.app

freider:
This bug leads the container entrypoint to the fallback app lookup logic instead (only intended for undecorated classes/functions), which can fail in certain cases - preventing hydration of sibling functions/classes.

### DIFF
--- a/modal/_runtime/user_code_imports.py
+++ b/modal/_runtime/user_code_imports.py
@@ -310,8 +310,8 @@ def import_class_service(
 
     See import_function.
     """
-    active_app: Optional["modal.app._App"] = None
-    code_deps: Optional[List["modal.object._Object"]] = None
+    active_app: Optional["modal.app._App"]
+    code_deps: Optional[List["modal.object._Object"]]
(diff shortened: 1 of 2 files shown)


devennavani (approved):
should we raise an error if `active_app` is still `None` after invoking `get_active_app_fallback`?

freider:
> should we raise an error if `active_app` is still `None` after invoking `get_active_app_fallback`?

I think there are cases where that's expected but things would still work, although very particular edge cases. Things like creating both the app and decorating the function in a private scope (or using serialized=True).

Having the `active_app` is only required in order to hydrate sibling methods (and only when serialized=False). And maybe Images - not 100% sure actually. Not all functions make use of sibling methods or image hydration, so they wouldn't break unless they used those specific features.

We should add some more tests that the attribute is set and not in the expected cases though, and if possible I'd like to move the fallback code and hydration logic into `user_code_imports` too add a clear separation of concerns between this module and container_entrypoint (also good for testing)

freider:
@prbot automerge

PR #2540, opened 2024-11-20
Remove extra period from interpolated string

mwaskom:
Tiny fix that avoids an error message ending with two periods.


mwaskom:
@prbot approve; the smallest possible change

PR #2541, opened 2024-11-20
Remove aiostream from library dependencies

mwaskom:
Closes CLI-241

We have to keep the library in the requirements files for legacy Image Builder versions but it's not part of `2024.10`. Even though it will be installed in the container with older Image Builders, up-to-date clients won't actually import it.

## Changelog

- Removed the `aiostream` package from the modal client library dependencies.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,7 +20,6 @@ packages = find:
 python_requires = >=3.9
 install_requires =
     aiohttp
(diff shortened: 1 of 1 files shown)


freider:
lgtm!

mwaskom:
@prbot lgtm

freider (approved):

PR #2542, opened 2024-11-20
Remove a few references to 'Stub'

mwaskom:
Hides the `Stub` class from the API docs and removes a couple mentions where we glossed App with the old name.


modal/cli/import_refs.py
```diff
@@ -107,35 +107,6 @@ def get_by_object_path(obj: Any, obj_path: Optional[str]) -> Optional[Any]:
...
-
-    if obj_path:
-        return get_by_object_path(obj, obj_path)
-    else:
-        app = get_by_object_path(obj, DEFAULT_APP_NAME)
-        stub = get_by_object_path(obj, "stub")
-        if isinstance(app, App):
-            return app
-        elif app is not None and isinstance(stub, App):
-            deprecation_warning(
```
erikbern, line 121: in hindsight maybe we should have escalated this into a deprecation error but it's been so long that i think it's fine to remove it

erikbern (approved):

erikbern:
exciting!

PR #2546, opened 2024-11-20
Remove telemetry protos

vishaalram02:
## Describe your changes

Remove the previously added user telemetry protos, as we're no longer writing client exporters. 

---

---

</details>


pawalt (approved):
down with protos

PR #2547, opened 2024-11-20
Increase fetch depth in deployment workflow

mwaskom:
I think this is the answer to why the changelog step fails when we have concurrent deployments: the default behavior for the checkout action is to fetch the most recent commit. That's why we crash when we try to get the log for the commit that triggered the build: if there's been a subsequent merge to main while CI ran on our commit, that's the commit we fetch, and we otherwise have no record of the commit that `$GIT_SHA` points at.


devennavani (commented):
lgtm

PR #2551, opened 2024-11-21
Adds Proxy support to Sandboxes

luiscape:
Allows for Sandboxes to use Proxies.


pawalt (approved):
LGTM but we should probably wait for worker changes to fully roll before merging

PR #2548, opened 2024-11-20
Move to next phase of `show_progress` deprecation

erikbern:
See outline of the migration in https://github.com/modal-labs/modal-client/pull/2030#issuecomment-2237266186

Basically we're deprecating `show_progress` in favor of `enable_output` now

PR #2552, opened 2024-11-21
remove way to suppress show_progress deprecation warning

erikbern:
The whole point of this was to not break internal integration tests for a short period but they broke anyway so I fixed it in a different way.

Either way we can remove this now.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -1,6 +1,5 @@
 # Copyright Modal Labs 2022
 import inspect
-import os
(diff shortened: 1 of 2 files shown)


howard36 (approved):
We should remove the fixture from conftest too

erikbern:
> We should remove `disable_app_run_warning` from the client's conftest too

I removed it in #2548 actually!

PR #2553, opened 2024-11-21
make Mount._deploy implicitly Environment-scoped

thundergolfer:
## Describe your changes

Currently we have a benchmark failure (large_mount_data) because `Mount.lookup` is implicitly Environment-scoped by `Mount._deploy` is not. In one Environment, the named Mount is deleted and so the lookup fails but the `_deploy` looks at our "main" Environment which does have the named Mount.

I expect that Modal's workspace is the only consumer of `._deploy`.

---

---

</details>


erikbern (approved):
lgtm!

PR #2554, opened 2024-11-21
Add custom_domain_info field to MethodDefinition

devennavani:
This is so we can display custom domains created in the terminal when a user does `modal deploy` on a new-style class that doesn't have method placeholders. `FunctionDefinition` already has this field.

See corresponding server side change [here](https://github.com/modal-labs/modal/pull/17571)


freider (commented):
goodie

PR #2556, opened 2024-11-21
Retire modal-client PyPI package: Step one

mwaskom:
Step one of retiring the `modal-client` alias package. This does one final release (version 1.0!!) with a module that just raises an exception on import directing users to the `modal` PyPI package.

I'll follow up with a subsequent PR to stop publishing the alias package.


mwaskom:
@prbot approve; worst thing that happens here is this fails and blocks its own deployment

erikbern:
👍

PR #2559, opened 2024-11-22
Add prbot token step back to versioning job

mwaskom:
Fixes deployments after https://github.com/modal-labs/modal-client/pull/2558

I took out a step that looked vestigial, but I guess it wasn't 😁

PR #2560, opened 2024-11-22
Fix deployments better

mwaskom:
OK I think this is what I was missing in https://github.com/modal-labs/modal-client/pull/2559

We need to use the token to check out the repo — then we can use the fact that prbot is allowed to push directly to main in the subsequent step. That is set up in the GitHub settings UI (not the config) and I am not 100% sure how the token propagates from one step to the other, but I reckon this is the solution.


mwaskom:
No code changes here, so gonna merge before tests finish.

ekzhang:
yep!

PR #2562, opened 2024-11-22
Add docstring for async_merge

rohansingh:


kramstrom (approved):
Great!

modal/_utils/async_utils.py
```diff
@@ -573,6 +573,34 @@ class StopSentinelType:
...
+        *generators: One or more async generators to be merged.
+
+    Yields:
+        The values produced by the input async generators.
+
+    Raises:
+        Exception: If any of the input generators raises an exception, it is propagated.
+
+    Example:
+        async def gen1():
```
freider, line 592: if you add a markdown codefence with ```` ```py````around the example and optionally add assertions at the end, the inlined example will be automatically tested in CI to prevent example-drift if we change something

freider: (see https://pypi.org/project/pytest-markdown-docs/)

freider (approved):

modal/_utils/async_utils.py
rohansingh: Tried that but looks like all docstring tests are being skipped in CI. Might take a look later.

rohansingh: OK hopefully should be fixed in #2576.

freider:
Interesting - the reactivated docstring testing seems to trigger an error on Python 3.9 which also triggers a bug in pytest-markdown-docs that prevents it from showing the traceback of the error 🤦 I'll take a look

freider:
Fix for actual bug in Volumes that triggered the failing test: https://github.com/modal-labs/modal-client/pull/2581

Also added an issue on pytest-markdown-docs for the borked traceback, but won't prioritize fixing that right now: https://github.com/modal-labs/pytest-markdown-docs/issues/36

modal/_utils/async_utils.py
```diff
@@ -573,6 +573,46 @@ class StopSentinelType:
...
+        yield 4
+
+    async def example():
+        values = set()
+        async for value in async_merge(gen1(), gen2()):
+            values.add(value)
+
+        return values
+
+    # Output could be: {1, 'a', 2, 'b'} (order may vary)
```
freider, line 611: This would be {1, 2, 3, 4} right? (I assume the letter were from a previous version?) (that set should be usable in assertion too without sorting?)

PR #2563, opened 2024-11-22
Remove heartbeat constants

erikbern:
Randomly saw this dead code


erikbern:
actually it's not dead, it's just used in a different place. a bit confusing.

PR #2564, opened 2024-11-22
Add _experimental_fs_api

azliu0:
## Describe your changes

This pr adds a flag enabling testing of modal-daemon (i.e., filesystem api) in production.

This flag is necessary as enabling the daemon requires enabling `-host-uds=all` in gVisor, which is currently unstable (see [here](https://github.com/google/gvisor/issues/11202)). Adding the flag here temporarily unblocks us from merging and testing the daemon changes.


pawalt (commented):
Do we need this? I thought we were only going to control this internally for now?

PR #2567, opened 2024-11-25
[CLI-229] Fix for `self.method.remote` calling FunctionBindParams in containers

freider:
**Note** This only fixes the special case of un-parameterized classes calling stuff on `self` when they could simply use the class servicer function as is done when using the class locally.

To fix the general case, we'd need to add a way to hydrate parameter-bound instances of classes ("instance service functions") using the container args or similar.

The root cause of the problem was that when used in side containers, the class isn't hydrated at import time when the "self" is created (unlike when it's used in an active app locally). This means that when the Cls is instantiated (in `get_user_class_instance`) it doesn't trigger the short-circuit logic in `_bind_parameters` which prevents FunctionBindParams for parameter-less functions. This simple patch moves that short-circuit logic into the _load function instead.


modal/functions.py
```diff
@@ -1064,10 +1070,9 @@ async def _load(param_bound_func: _Function, resolver: Resolver, existing_object
 
         fun: _Function = _Function._from_loader(_load, "Function(parametrized)", hydrate_lazily=True)
 
-        if can_use_parent and self.is_hydrated:
-            # Edge case that lets us hydrate all objects right away
-            # if the instance didn't use explicit constructor arguments
-            fun._hydrate_from_other(self)
+        if can_use_parent and parent.is_hydrated:
```
freider, line 1073: Keeping this for now since it breaks a test due to how our Resolver works (the _load methods requires setting up a resolver + client, so it can't short circuit that part from within the function)

test/supports/user_code_import_samples/cls.py
```diff
@@ -14,5 +14,11 @@ def f(self, arg):
...
 
+    @modal.method()
+    def calls_f_remote(self, arg):
+        return self.f.remote(arg)
+
+
+UndecoratedC = C  # keep a reference to original class before overwriting
 
-C = app.cls()(UndecoratedC)
+C = app.cls()(C)  # type: ignore[misc]   # "decorator" of C
```
freider, line 24: Changing this around since the function won't ever import the decorated entity if it has a different name than the underlying one at decoration time

freider:
@prbot approve

devennavani (commented):
nice thanks!

PR #2569, opened 2024-11-25
More docs and types for io_streams

ekzhang:
This might seem totally random but I wanted to take another look at this before `1.0`. Didn't find anything that needed to be changed though, just some minor documentation improvements.

https://modal.com/docs/reference/modal.io_streams

A lot of what I wanted to do was already handled by @pawalt and @azliu0's recent work on making it generic on bytes | str.

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -146,9 +146,7 @@ class _StreamReader(Generic[T]):
         return self._file_descriptor
 
     async def read(self) -> T:
-        """Fetch and return contents of the entire stream.
-
-        If EOF was received, return an empty string.
+        """Fetch the entire contents of the stream until EOF.
 
         **Usage**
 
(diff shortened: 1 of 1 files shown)


modal/io_streams.py
```diff
@@ -140,12 +141,14 @@ def __init__(
...
+    def file_descriptor(self) -> int:
+        """Possible values are `1` for stdout and `2` for stderr."""
         return self._file_descriptor
 
     async def read(self) -> T:
-        """Fetch and return contents of the entire stream. If EOF was received,
-        return an empty string.
+        """Fetch and return contents of the entire stream.
+
+        If EOF was received, return an empty string.
```
pawalt, line 151: This was here before, but this text is a bot confusing. Technically EOF is always received, in which case it returns the full buffer. Maybe something about it waits for the stream to close would be more appropriate.

pawalt (approved):
Thanks for the cleanup!

PR #2571, opened 2024-11-25
Add client retry support to .map

rohansingh:
Implements similar logic for `.map` as #2403 did for `.remote`.

To do this, we create a `TimedPriorityQueue`, which holds all pending inputs and pending retries, along with the timestamp at which they should be sent to the server.

When inputs are first read from the generator, they are added to the queue with a timestamp of "now". And if a failed output is received from the server, it's added back to the queue with a configured retry delay.


rohansingh:
Continued in #2734.

PR #2572, opened 2024-11-25
Separate "indexed objects" by type

erikbern:
We have a few `Dict[str, _Object]` where the object is either a function or a class. It makes sense to separate these by type. Starting with `App.indexed_objects`. I'll do `Running_app.tag_to_object_id` later.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -517,7 +517,10 @@ class _App:
 
     @property
     def indexed_objects(self) -> Dict[str, _Object]:
-        # TODO(erikbern): deprecate
+        deprecation_warning(
+            (2024, 11, 25),
+            "`app.indexed_objects` is deprecated! Use `app.registered_functions` or `app.registered_classes` instead.",
+        )
         return dict(**self._functions, **self._classes)
 
     @property
--- a/modal/app.py
+++ b/modal/app.py
@@ -177,7 +177,8 @@ class _App:
 
     _name: Optional[str]
     _description: Optional[str]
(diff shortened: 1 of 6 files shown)


modal/app.py
```diff
@@ -507,7 +517,11 @@ def registered_entrypoints(self) -> Dict[str, _LocalEntrypoint]:
 
     @property
     def indexed_objects(self) -> Dict[str, _Object]:
-        return self._indexed_objects
+        deprecation_warning(
+            (2024, 11, 25),
+            "`app.indexed_objects` is deprecated! Use `app.registered_functions` or `app.registered_classes` instead.",
```
mwaskom, line 522: I think we might be planning on deprecating those too? 😁

mwaskom (approved):
:+1: we have some very annoying code to unmix the `tag_to_object_ids` into the separate `function_ids` / `class_ids` for the `AppPublishRequest`, will be nice to clean that up!

modal/app.py
```diff
@@ -177,7 +177,8 @@ def foo():
 
     _name: Optional[str]
     _description: Optional[str]
-    _indexed_objects: Dict[str, _Object]
+    _functions: Dict[str, _Function]
+    _classes: Dict[str, _Cls]
```
mwaskom, line 181: `classes`? Or ... `clse`? 🤔 😭

erikbern: yeah idk, no strong feelings

PR #2576, opened 2024-11-26
ci:  Run docstring tests again

rohansingh:
This comparison expression has always been broken.

However in #2334 it changed from `!= 'true'` to `== 'false'`. So the tests went from always running to never running.


rohansingh:
@prbot approve

mwaskom:
Thanks, gonna blame the robot for this one.

PR #2577, opened 2024-11-26
 Modify FunctionGetOutputs proto for client retries

rculbertson:
~This is based on https://github.com/modal-labs/modal-client/pull/2403. That should me reviewed and merged first. ~
Update: That PR has been merged, and this one has been rebased on it.

This modifies the request and response of FunctionGetOutputs so the client can retry lost inputs. This applies to sync inputs only.

The request has a new field `expected_input_ids`, so the client can optionally pass the server a list of input ids that it expects the server to be processing.

The server will check if it has a record of all these IDs. If not, it will immediately return, populating the response's new `lost_input_ids` field with the IDs that missing. The client can the resend those inputs.

Merging the proto changes first so we can implement and test this in the server.

Part of SVC-171

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1426,6 +1426,8 @@ message FunctionGetOutputsRequest {
   string last_entry_id = 6;
   bool clear_on_success = 7; // expires *any* remaining outputs soon after this call, not just the returned ones
   double requested_at = 8; // Used for waypoints.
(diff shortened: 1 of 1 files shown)


gongy (commented):
I think we may need `entry_id`s here.

rculbertson:
Closing in favor of #2600

PR #2578, opened 2024-11-26
Check that environment exists when setting in config

### LINKED ISSUE #2086: No check if environment exists when setting it by default
When I wanted to change the environment I confused the name (development) with the web suffix (dev), but it seemed to be fine

image

mwaskom:
Super simple, fixes #2086

image

## Changelog

- `modal config set-environment` will now raise if the requested environment does not exist.

### DIFF
--- a/modal/config.py
+++ b/modal/config.py
@@ -268,7 +268,7 @@ class Config:
         return repr(self.to_dict())
 
     def to_dict(self):
-        return {key: self.get(key) for key in _SETTINGS.keys()}
+        return {key: self.get(key) for key in sorted(_SETTINGS)}
 
 
 config = Config()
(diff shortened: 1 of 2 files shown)


modal/config.py
```diff
@@ -268,7 +268,7 @@ def __repr__(self):
         return repr(self.to_dict())
 
     def to_dict(self):
-        return {key: self.get(key) for key in _SETTINGS.keys()}
+        return {key: self.get(key) for key in sorted(_SETTINGS)}
```
mwaskom, line 271: Not directly related but when confirming that we can still set a valid environment I noticed that the `modal config show` output was hard to read because it was unordered.

mwaskom:
@prbot approve

devennavani (commented):
lgtm

PR #2579, opened 2024-11-26
Remove autoapprove workflow

mwaskom:
We don't use this in the client development workflow; we never really need to do a "hotfix" release because users affected by client bugs can always downgrade their client version until we get a fix out.


parjohansson (approved):
Great, thanks for following up!

PR #2580, opened 2024-11-27
Only check if returned object ID is different from existing_object_id for functions

devennavani:
Only check if returned object ID is different from existing_object_id for functions.

Currently only functions and classes use `existing_object_id` in function/class creation RPCs. 

After the server-side fix [here](https://github.com/modal-labs/modal/pull/17677), ID of an existing class might change if the set of methods of the class change.

existing function IDs should not change because the way we modify functions is by assigning them a new definition id

### DIFF
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -130,9 +130,14 @@ class Resolver:
                         raise NotFoundError(exc.message)
                     raise
 
-                # Check that the id of functions and classes didn't change
+                # Check that the id of functions didn't change
                 # Persisted refs are ignored because their life cycle is managed independently.
-                if not obj._is_another_app and existing_object_id is not None and obj.object_id != existing_object_id:
+                if (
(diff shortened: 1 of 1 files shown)


devennavani:
@prbot approve

freider:
👍

modal/_resolver.py
```diff
@@ -130,9 +130,14 @@ async def loader():
...
                     raise
 
-                # Check that the id of functions and classes didn't change
+                # Check that the id of functions didn't change
                 # Persisted refs are ignored because their life cycle is managed independently.
-                if not obj._is_another_app and existing_object_id is not None and obj.object_id != existing_object_id:
+                if (
+                    not obj._is_another_app
+                    and existing_object_id is not None
+                    and existing_object_id.startswith("fu-")
```
erikbern, line 138: There's a helper function for this, see the _object base class!

PR #2581, opened 2024-11-27
Fixes issue where Volume.from_name eagerly assigns event loop

freider:
Lock was created in sync function as part of client-side object creation, causing it to bind the lock to the main thread event loop instead of the synchronizer loop on Python 3.9.

This surfaced in some tests where the event loop was un-set by some prior test and then triggered an error when a subsequent test called `Volume.from_name()` in global scope.

I *think* in most cases the locks still function across event loops on Python 3.9, as long as both the awaiter and holder run on the same loop, so this probably created no issue in practice (unless users unset the main thread event loop before initializing a volume)

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1714,7 +1714,7 @@ class _Image(_Object, type_prefix="im"):
         """
 
         def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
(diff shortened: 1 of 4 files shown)


freider:
@prbot approve

erikbern:
ohhhh interesting

PR #2582, opened 2024-11-27
[WIP] CLI-227: add implicit `context_mount` to `Image` docker methods

kramstrom:
Adds deprecation warning when using `context_mount` parameter on `Image` docker methods as part of the effort to remove mounts. This parameter which used to be necessary for `COPY` commands to work.

This PR instead creates an implicit context mount from the dockerfile/docker commands by parsing the `COPY` source patterns and use them as mount filters for the current directory.

Also adds support for finding a dockerignore file that matches the given dockerfile or if there's a generic one if using docker commands


modal/image.py
```diff
@@ -1436,9 +1541,6 @@ def from_aws_ecr(
     def from_dockerfile(
         # Filepath to Dockerfile.
         path: Union[str, Path],
-        # modal.Mount with local files to supply as build context for COPY commands.
-        # NOTE: The remote_path of the Mount should match the Dockerfile's WORKDIR.
-        context_mount: Optional[_Mount] = None,
```
freider, line 1441: We can't remove this argument immediately or we'll break user code.

Instead, make the None value here use the new functionality and add a pending deprecation warning if users set it to non-None, in which the error message tells them to stop providing it since we'll automatically detect files for them based on their COPY commands, and contact us if that doesn't work for them (good for us to be able to patch before we remove it).

freider (commented):
Haven't looked at the details too much, but looks good on the surface!

I just realized we might also want to (separately) add more fine grained control mechanisms/filter support to give people a way to exclude particular files they don't want to include (dockerignore style) before we can forcibly push people to use this.

As a start, maybe we could just start exposing `condition` in a more user-friendly way. Will discuss separately in slack

modal/_utils/match.py
```diff
@@ -0,0 +1,161 @@
+# Copyright Modal Labs 2024
```
freider, line 1: Maybe call this `docker_copy_match.py` or something less generic than `match` :)

modal/image.py
```diff
@@ -236,6 +239,23 @@ def _get_image_builder_version(server_version: ImageBuilderVersion) -> ImageBuil
     return version
 
 
+def _create_context_mount(docker_commands: list[str], ignore: Callable[[Path], bool]) -> _Mount:
```
freider, line 242: Nit: I like to use `Sequence` instead of list since it will statically prevent accidental mutation of the argument

modal/image.py
```diff
@@ -1162,6 +1181,24 @@ def dockerfile_commands(
         if not cmds:
             return self
 
+        if context_mount is not None:
```
freider, line 1184: Might want to add the `ignore` argument here as well if people want manual inline ignore patterns without external files, as a replacement for context_mount. And the pending deprecation could point people towards that argument

freider: (or dockerignore obviously)

freider: Wait, does dockerfile_commands support dockerignore with this? We only call find_dockerignore_file in from_dockerfile right? 🤔 
No biggie if it doesn't - as long as we support `ignore` for now

kramstrom: added to `dockerfile_commands` as well -- currently both methods will try to find a dockerignore file unless a custom ignore argument is passed

modal/_utils/docker_utils.py
```diff
@@ -88,3 +88,24 @@ def valid_dockerignore_file(fp):
...
+        if self._custom_repr:
+            return self._custom_repr
+
+        return super().__repr__()
+
+    def __call__(self, _: Path) -> bool:
+        return True
+
+
+AUTO_DOCKERIGNORE = _AutoDockerIgnore().with_repr(f"{__name__}.AUTO_DOCKERIGNORE")
```
erikbern, line 111: what's all this repr magic used for?

kramstrom: Moved the repr inside the class, but it's just a placeholder value for type hints & docs to make it simpler to understand what will happen if you don't pass in a custom ignore pattern-list or callable --> then we default to trying to find a matching dockerignore to your dockerfile (or just a generic one if you're using `docker_commands`) -- same as we have for [add_local_python_source](https://modal.com/docs/reference/modal.Image#add_local_python_source)

Screenshot 2024-12-20 at 13 49 59

freider: TLDR for why it's needed (this was my suggestion)

It's so that we can use these as default values in signatures and still emit valid python signatures in our type stubs, since things like `inspect.signature(...)` will just use the repr of all default values, and can't figure out stuff like import path of the specific constant we use as a default. 

It's definitely a bit of a hack, but not something users should see. An alternative would be to do a bunch of `ast` magic to extract the signatures.

PR #2364, opened 2024-10-20
Remove usage of method placeholder functions

devennavani:
This PR introduces a new minor client version `0.67.x` that does not create method placeholder functions for modal classes.

I've tested all forward lookup scenarios (`.lookup()` of a `0.67` class from a pre `0.67` client) as well as backwards lookup scenarios (`.lookup()` of a pre `0.67` class from a `0.67` client) and all work, except for a `0.62` client looking up a `0.67` class, which will be explicitly not supported (this maintains our current restriction of not being able to lookup a `0.63+` class from a `0.62` client).

## Changelog

New minor client version `0.67.x` comes with an internal data model change for how Modal creates functions for Modal classes. There are no breaking or backwards-incompatible changes with this new minor version. All forward lookup scenarios (`.lookup()` of a `0.67` class from a pre `0.67` client) as well as backwards lookup scenarios (`.lookup()` of a pre `0.67` class from a `0.67` client) work, except for a `0.62` client looking up a `0.67` class (this maintains our current restriction of not being able to lookup a `0.63+` class from a `0.62` client).

PR #2583, opened 2024-11-28
hydrate functions then classes in container entrypoint

devennavani:
## Describe your changes

https://github.com/modal-labs/modal-client/pull/2364 introduced a subtle requirement that hydrating a class requires that the class service function for the class is hydrated first.

in `_run_app` we make sure to `resolver.load` an object's `deps` before loading the object itself, but in container entrypoint we are hydrating a class before we hydrate the class service function, so the `_method_functions` of the `Cls` was empty because the class service function was unhydrated (its `method_handle_metadata` attribute was not populated).

this PR changes the entrypoint to load the app's functions, then its classes

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -493,12 +493,17 @@ class _App:
         _App._container_app = running_app
 
         # Hydrate objects on app
-        indexed_objects = dict(**self._functions, **self._classes)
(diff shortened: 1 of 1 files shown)


devennavani:
@prbot approve

mwaskom (commented):
✅

PR #2584, opened 2024-11-28
Add space in _assert_no_mount_layers error message

kramstrom:
Adds a space to the error message in `_assert_no_mount_layers` as it currently will concat the words _files_ and _directly_ into _filesdirectly_

Screenshot 2024-11-28 at 14 27 52

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -375,7 +375,7 @@ class _Image(_Object, type_prefix="im"):
                 "\n"
                 "Run `image.add_local_*` commands last in your image build to avoid rebuilding images with every local "
                 "file change. Modal will then add these files to containers on startup instead, saving build time.\n"
-                "If you need to run other build steps after adding local files, set `copy=True` to copy the files"
+                "If you need to run other build steps after adding local files, set `copy=True` to copy the files "
                 "directly into the image, at the expense of some added build time.\n"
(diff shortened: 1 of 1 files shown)


kramstrom:
@prbot automerge

modal/image.py
```diff
@@ -375,7 +375,7 @@ def _assert_no_mount_layers(self):
                 "\n"
                 "Run `image.add_local_*` commands last in your image build to avoid rebuilding images with every local "
                 "file change. Modal will then add these files to containers on startup instead, saving build time.\n"
-                "If you need to run other build steps after adding local files, set `copy=True` to copy the files"
+                "If you need to run other build steps after adding local files, set `copy=True` to copy the files "
                 "directly into the image, at the expense of some added build time.\n"
```
mwaskom, line 379: Nice thanks for catching this. I've adopted habit of putting spaces at the beginning of the next line when doing implicit string concatenation because they'll line up and it's easier to see when you're missing one, e.g. 

```
                "If you need to run other build steps after adding local files, set `copy=True` to copy the files"
                " directly into the image, at the expense of some added build time.\n"
```

We're definitely not consistent across the codebase and I don't think you need to revisit this, just FYI in case it feels like a better approach going forward.

PR #2585, opened 2024-11-28
Upgrades pytest-markdown-docs to version with better code traversal

freider:
All tests weren't previously detected, since the code traversal only looked at the docstrings that were transferred by synchronicity to the wrapper objects/functions. Now we use the implementation classes/functions instead as the test objects, which finds a bunch more tests (many of which were failing)

* Disables and splits up a bunch of tests that got detected and failed
* Fixes a bunch of docstring tests that had missing args etc.

**Note** Would be nice to find a way to test async code and inject async fixtures like our MockServicer into the markdown tests (some kind of pytest-asyncio compatibility). Until then, we'll have to disable a bunch of tests that assume prior existence of resources or event loop usage.

### DIFF
--- a/modal/_tunnel.py
+++ b/modal/_tunnel.py
@@ -61,7 +61,7 @@ async def _forward(port: int, *, unencrypted: bool = False, client: Optional[_Cl
 
     **Usage:**
 
-    ```python
+    ```python notest
     import modal
     from flask import Flask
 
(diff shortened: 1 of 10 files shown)


freider:
@prbot approve

mwaskom (commented):
Agree it would be good to test these! But better to have them explicitly `notest` than to silently skip them of course.

PR #2587, opened 2024-11-28
CLI-271: Add X-Forwarded-For header to @modal.web_server http requests

kramstrom:
Adds the request IP as `X-Forwarded-For` header to `@modal.web_server` requests

### DIFF
--- a/modal/_runtime/asgi.py
+++ b/modal/_runtime/asgi.py
@@ -280,11 +280,47 @@ def wait_for_web_server(host: str, port: int, *, timeout: float) -> None:
                 ) from ex
 
 
+def _add_forwarded_for_header(scope):
+    # we strip X-Forwarded-For headers from the scope
+    # but we can add it back from the ASGI scope
+    # https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope
+
+    # X-Forwarded-For headers is a comma separated list of IP addresses
+    # there may be multiple X-Forwarded-For headers
+    # we want to prepend the client IP to the first one
+    # but only if it doesn't exist in one of the headers already
+
+    first_x_forwarded_for_idx = None
+    if "headers" in scope and "client" in scope:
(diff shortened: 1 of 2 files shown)


ekzhang (approved):

modal/_runtime/asgi.py
```diff
@@ -283,6 +283,9 @@ def wait_for_web_server(host: str, port: int, *, timeout: float) -> None:
 async def _proxy_http_request(session: aiohttp.ClientSession, scope, receive, send) -> None:
     proxy_response: aiohttp.ClientResponse
 
+    if "client" in scope and "headers" in scope:
+        scope["headers"].append((b"X-Forwarded-For", scope["client"][0].encode()))
```
erikbern, line 287: why does `scope["client"][0]` have this header? maybe add a comment referring to the documentation or something, since this is highly nonobvious

ekzhang: It's the asgi spec https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope

It doesn't have the header, it just has the source IP which this PR places in the header

freider: @ekzhang do you remember why we are not sending this as `X-Forwarded-For` from Caddy? 

Couldn't it make sense to have caddy set the `X-Forwarded-For` headers using `http.request.remote.host` too? (We shouldn't pass along a user-supplied such value, but we could make caddy set the header using its "trusted" info from the nlb?).
Not sure what the usage % of the scope["client"] value vs the standard header is among asgi apps in general, but it wouldn't hurt to fill both, right?

freider: @kramstrom regardless of caddy changes, it could make sense to check if the X-Forwarded-For header is already present in headers as well and just updating the existing header value in that case (if the scope value is already present in the X-Forwarded-For value, no change would be needed).

PR #2588, opened 2024-11-29
_Obj private properties

freider:
Trying to minimize how much we clutter the "public namespace" on _Obj, to avoid issues where users declare these names and it clashes with internal implementation details.

Turns out we ourselves had one of these clashes inside one of our unit tests - hiding the fact that async `@enter` methods don't actually work with `.local()` at the moment 🤦 This patch doesn't have a fix for that but I'll follow up with that...

Side note: I think we should change the API for how .local() (and containers) interacts with class lifecycles. Would be nice if the `.local()` usage of lifecycles was explicit using a context manager or something that also includes an exit trigger. Maybe something like:

```py
with myobj.lifecycle():
    ...
```
or just (could be a bit messy in case users still use the `__enter__` overloading syntax):
```py
with myobj:
    ...
```

Inside containers the lifecycle of `self` should be a no-op since that's handled by the container entrypoint instead and we don't want double triggers.

### DIFF
--- a/modal/_runtime/user_code_imports.py
+++ b/modal/_runtime/user_code_imports.py
@@ -195,7 +195,7 @@ def get_user_class_instance(
     if isinstance(cls, modal.cls.Cls):
         # globally @app.cls-decorated class
         modal_obj: modal.cls.Obj = cls(*args, **kwargs)
-        modal_obj.entered = True  # ugly but prevents .local() from triggering additional enter-logic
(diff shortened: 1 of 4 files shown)


mwaskom:
> could be a bit messy in case users still use the __enter__ overloading syntax
This has been a hard error since June so should be fine to change at this point if we want to go that route.

erikbern (approved):
I would love to reduce the complexity of `.local` on classes. The auto-enter thing was probably a bad idea in hindsight. I like your idea of using a context manager!

PR #2589, opened 2024-11-29
Fix modal run on class method defined on base of wrapped class

mwaskom:
Fixes CLI-22

Hard to explain exactly what the issue was here: it's fairly complex code. But there was a gap in the conditionals for resolving the name of the method to execute given (1) a Python function object (i.e. an unbound method) and (2) a Python class that has been registered to an App.

We have a lot of undefined behavior and edge cases around class inheritance.

## Changelog

- Fixed a bug where pointing `modal run` at a method on a Modal Cls would fail if the method was inherited from a parent.

### DIFF
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -133,6 +133,9 @@ class FunctionInfo:
         elif f is None and user_cls:
             # "service function" for running all methods of a class
             self.function_name = f"{user_cls.__name__}.*"
+        elif f and user_cls:
+            # Method may be defined on superclass of the wrapped class
+            self.function_name = f"{user_cls.__name__}.{f.__name__}"
         else:
             self.function_name = f.__qualname__
 
--- a/test/supports/class_hierarchy.py
+++ b/test/supports/class_hierarchy.py
@@ -1,0 +1,21 @@
+# Copyright Modal Labs 2024
+import modal
+
+app = modal.App("class-hierarchy")
+
+
+class Base:
+    @modal.method()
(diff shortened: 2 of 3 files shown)


test/supports/class_hierarchy.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Base:
+    @modal.method()
+    def defined_on_base(self):
+        print("base")
+
+    @modal.method()
+    def overridden_on_wrapped(self):
+        raise NotImplementedError()
+
```
mwaskom, line 14: Was hoping this would also fix the bug where `modal run my_app.py::Wrapped.overridden_on_wrapped` executes the base version of the method, but apparently that is a separate defect.

mwaskom: Resolved here: https://github.com/modal-labs/modal-client/pull/2590

mwaskom:
@prbot approve; unblocks a user

modal/_utils/function_utils.py
```diff
@@ -133,6 +133,9 @@ def __init__(
         elif f is None and user_cls:
             # "service function" for running all methods of a class
             self.function_name = f"{user_cls.__name__}.*"
+        elif f and user_cls:
```
freider, line 136: I don't think we should even be constructing FunctionInfo for methods (except @build ones, and we might want to deprecate those), but that's a separate matter.

freider (commented):
nice catch! should get rid of Methods even using FunctionInfo (which should really be ServiceInfo)

PR #2590, opened 2024-11-29
Fix modal.method overriding in subclass

mwaskom:
Fixes CLI-47

Previously if you had a `@modal.method` defined on a class and on a parent of that class:

```python
class AbstractService:
    @modal.method()
    def foo(self):
        raise NotImplementedError()

@app.cls()
class Service:
    @modal.method()
    def foo():
        print("Doing a foo")
```

Then running `Service().foo.remote()` would run the method on the base class.

I think this is a simple bug: we're iterating through the `user_cls.mro()` in the wrong order, so we always use the most ancestral version of the method.

## Changelog

- Fixed a bug that executes the wrong method when a Modal Cls overrides a `@modal.method` inherited from a parent.

### DIFF
--- a/test/cli_test.py
+++ b/test/cli_test.py
@@ -201,6 +201,7 @@ def test_run_quiet(servicer, set_env_client, test_dir):
 def test_run_class_hierarchy(servicer, set_env_client, test_dir):
     app_file = test_dir / "supports" / "class_hierarchy.py"
     _run(["run", app_file.as_posix() + "::Wrapped.defined_on_base"])
+    _run(["run", app_file.as_posix() + "::Wrapped.overridden_on_wrapped"])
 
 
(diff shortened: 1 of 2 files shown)


test/cli_test.py
```diff
@@ -202,6 +202,7 @@ def test_run_quiet(servicer, set_env_client, test_dir):
 def test_run_class_hierarchy(servicer, set_env_client, test_dir):
     app_file = test_dir / "supports" / "class_hierarchy.py"
     _run(["run", app_file.as_posix() + "::Wrapped.defined_on_base"])
+    _run(["run", app_file.as_posix() + "::Wrapped.overridden_on_wrapped"])
```
mwaskom, line 205: Added the test support in #2589 but deferred the fix to a separate PR

freider (approved):
lgtm

PR #2591, opened 2024-11-29
Run pyupgrade for Py3.9+

mwaskom:
We dropped support for Python 3.8 on 15 November and haven't heard of that causing any problems, so I'd say we can start taking advantage to remove vestigial code.

This is a ~fully~ mostly automated PR, resulting from 

```
pyupgrade --py39-plus `find . -name "*.py"`
```

and then some manual fixes for a couple edge cases:

- Change a `from modal import Dict` import to `import modal; modal.Dict`, as pyupgrade overzealously turned the first one into `dict`
- Rename CLI `list` functions to `list_` to avoid shadowing the builtin type (no external effect as we provide a command name in the typer decorator)

### DIFF
--- a/modal/_clustered_functions.py
+++ b/modal/_clustered_functions.py
@@ -2,7 +2,7 @@
 import os
 import socket
 from dataclasses import dataclass
-from typing import List, Optional
+from typing import Optional
 
 from modal._utils.async_utils import synchronize_api
 from modal._utils.grpc_utils import retry_transient_errors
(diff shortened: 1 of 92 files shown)


mwaskom:
Oh that's really annoying, pyupgrade gets confused about `modal.Dict`.

mwaskom:
I should read up on exactly what `pyupgrade` does; it seems to be leaving some unnecessary constructs like `typing.Union[...]` where the `|` operator should work. Going to leave those for now but maybe follow up in smaller PRs.

mwaskom:
@prbot approve; huge diff, but very straightforward

erikbern (approved):
Nice!

ekzhang:
Just saw this, thanks for this upgrade! 🙏

PR #2595, opened 2024-12-02
Avoid sending empty messages on process start

pawalt:
## Describe your changes

Previously, we added an empty message on container start so that the client could tell the process had booted. This turned out to be a breaking change since clients would now receive an extra message on initialization. This commit refactors output consumption to only return the empty message in `modal shell` usage.

---

---

</details>


azliu0 (commented):
This looks reasonable to me, deferring to @freider who may have more opinions.
More context on the regression here: https://modal-com.slack.com/archives/C044YUDC1QU/p1732752148985199

pawalt:
@prbot approve

freider:
👍

PR #2600, opened 2024-12-03
Have client retry lost inputs

rculbertson:
Closes SVC-224

Update client to retry inputs if FunctionGetOutputs reports that an input is lost. If an input is lost, FunctionGetOutputs returns a failed output with status `GENERIC_STATUS_INTERNAL_FAILURE`. The client then retries immediately.

---

---

</details>

### DIFF
--- a/modal/exception.py
+++ b/modal/exception.py
@@ -110,9 +110,7 @@ class ServerWarning(UserWarning):
 
 class InternalFailure(Error):
     """
-    Raised when the server returns GENERIC_STATUS_INTERNAL_FAILURE. This is a retiable error which can be
-    caused by events like 1) redis crashing and the server losing track of inputs, or 2) a worker being
-    preempted, which terminates the input.
+    Retriable internal error.
     """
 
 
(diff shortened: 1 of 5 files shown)


rculbertson:
@gongy This is now has all the changes we discussed.

modal/exception.py
```diff
@@ -108,6 +108,14 @@ class ServerWarning(UserWarning):
     """Warning originating from the Modal server and re-issued in client code."""
 
 
+class InternalFailure(Error):
+    """
+    Raised when the server returns GENERIC_STATUS_INTERNAL_FAILURE. This is a retiable error which can be
+    caused by events like 1) redis crashing and the server losing track of inputs, or 2) a worker being
+    preempted, which terminates the input.
```
gongy, line 115:
```suggestion
    Retriable internal error.
```

rculbertson:
@prbot autoapprove

rculbertson:
@gongy I think you already review this last week and gave me a +1 over slack.  Mind marking it as approved?

rohansingh (approved):

PR #2601, opened 2024-12-03
fix sync_or_async_iter: check if async iterable has `aclose` before aclosing it

kramstrom:
Changing `sync_or_async_iter` to accept the more generic `AsyncIterable` instead of `AsyncGenerator` and also then checking if the object has an `aclose` method (which doesn't necessarily exist on `AsyncIteratable`s ) before calling it

### DIFF
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -6,7 +6,7 @@ import inspect
 import itertools
 import time
 import typing
-from collections.abc import AsyncGenerator, Awaitable, Iterable, Iterator
+from collections.abc import AsyncGenerator, AsyncIterator, Awaitable, Iterable, Iterator
 from contextlib import asynccontextmanager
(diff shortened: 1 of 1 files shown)


modal/_utils/async_utils.py
```diff
@@ -484,14 +484,17 @@ async def __aexit__(self, exc, exc_type, tb):
         await self.agen.aclose()
 
 
-async def sync_or_async_iter(iter: Union[Iterable[T], AsyncGenerator[T, None]]) -> AsyncGenerator[T, None]:
+async def sync_or_async_iter(iter: Union[Iterable[T], AsyncIterator[T]]) -> AsyncGenerator[T, None]:
```
freider, line 487: I think this should be `AsyncIterable`?

freider (approved):
lgtm!

PR #2603, opened 2024-12-03
Include method definitions in FunctionData message

mwaskom:
Fixes CLI-275.

I'm not really sure how to write a unit test for this -- open to suggestions. See the ticket for a simple repro.

---

---

</details>

## Changelog

- Fixed a bug introduced in v0.67.0 where it was impossible to call `modal.Cls` methods when passing a list of requested GPUs.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -886,6 +886,8 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
                         is_method=function_definition.is_method,
                         use_function_id=function_definition.use_function_id,
                         use_method_name=function_definition.use_method_name,
+                        method_definitions=function_definition.method_definitions,
+                        method_definitions_set=function_definition.method_definitions_set,
                         _experimental_group_size=function_definition._experimental_group_size,
                         _experimental_buffer_containers=function_definition._experimental_buffer_containers,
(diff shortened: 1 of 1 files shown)


modal/functions.py
```diff
@@ -886,6 +886,8 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
                         is_method=function_definition.is_method,
                         use_function_id=function_definition.use_function_id,
                         use_method_name=function_definition.use_method_name,
+                        method_definitions=method_definitions,
```
irfansharif, line 889: [nit] Can we make this look like `method_definitions=function_definition.method_definitions`, ditto for the other field. If it helps, after the FunctionData migration, we'll be able to have just one form this in the client - i.e. filling out the FunctionData proto directly. 

In the interm - we have this code in the server. We could pull it into the client and use that instead:

https://github.com/modal-labs/modal/blob/159c468655850cd8f53d6506bc3138b9e29c4e58/server/modal_server/function_definition.py#L89.

irfansharif (approved):
🤦🏽

modal/functions.py
mwaskom: Dumb on my part for not noticing this.

mwaskom:
@prbot automerge

PR #2604, opened 2024-12-03
Unwrap functions when extracting global variable references

mwaskom:
Fixes CLI-276

## Changelog

- Image build functions that use a `functools.wraps` decorator will now have their global variables included in the cache key. Previously, the cache would use global variables referenced within the wrapper itself. This will force a rebuild for Image layers defined using wrapped functions.

### DIFF
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -248,7 +248,13 @@ class FunctionInfo:
     def get_globals(self) -> dict[str, Any]:
         from .._vendor.cloudpickle import _extract_code_globals
 
+        if self.raw_f is None:
+            return {}
+
         func = self.raw_f
+        while hasattr(func, "__wrapped__") and func is not func.__wrapped__:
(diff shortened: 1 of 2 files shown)


modal/_utils/function_utils.py
```diff
@@ -249,6 +249,12 @@ def get_globals(self) -> dict[str, Any]:
         from .._vendor.cloudpickle import _extract_code_globals
 
         func = self.raw_f
+        while True:
+            # Unwrap functions decorated using functools.wrapped (potentially multiple times)
+            try:
```
aksh-at, line 254: This is a really contrived example and I hope no one will do this:

```python
import functools


def f():
    print("hi")


functools.wraps(f)(f)

print(f.__wrapped__ is f) # True
print(f.__wrapped__.__wrapped__ is f) # True
```

But just in case, we could guard against `func != func.__wrapped__` so we don't go into an infinite loop?

aksh-at (approved):
Nice, thanks for tracking this down! Left an evil comment.

PR #2606, opened 2024-12-04
Move _bind_instance_method out of _Function

freider:
Trying to make small changes toward a simpler Cls/Obj/Function code base.

This is a first tiny step towards decoupling "Function - the interface" from "Function - the autoscaling container pool", and having methods only be an instance of the former.

This PR just moves the `_bind_instance_method` method out of `_Function` as a simple separation.

---

---

</details>


devennavani (commented):
LGTM

PR #2607, opened 2024-12-04
Small comment and type correctness for Obj constructor

freider:
Just improves type + comment (got confused when I tried to refactor and this wasn't set)

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -90,7 +90,7 @@ class _Obj:
 
     def __init__(
         self,
-        user_cls: type,
+        user_cls: Optional[type],  # this would be None in case of lookups
         class_service_function: Optional[_Function],  # only None for <v0.63 classes
         classbound_methods: dict[str, _Function],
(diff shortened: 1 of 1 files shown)


freider:
@prbot approve

freider:
@prbot automerge

mwaskom (approved):

modal/cls.py
```diff
@@ -90,7 +90,7 @@ def _uses_common_service_function(self):
 
     def __init__(
         self,
-        user_cls: type,
+        user_cls: Optional[type],  # this would be None in case of lookups
```
mwaskom, line 93: Just a random comment about consistency; I think the analogous attribute is `raw_f` for `Function`?

freider: Yup. There is a bunch of code here that needs cleanup/clarification. `FunctionInfo` is quite annoying how it bundles a bunch of different use cases but is hard to provide in some cases.

PR #2608, opened 2024-12-04
[FS-149] New blobs upload API

dflemstr:
A work-in-progress PR to start using the new Blobs service for client-side file uploads.  Both the API and implementation details are subject to change.

- [ ] Client+Server: this change is compatible with old servers (NOTE: *not backwards compatible with old servers*)

</details>


protoc_plugin/plugin.py
```diff
@@ -106,7 +106,7 @@ def render(
                         raise TypeError(cardinality)
 
                     original_method = f"grpclib_stub.{name}"
-                    buf.add(f"self.{name} = {wrapper_cls}({original_method}, client)")
+                    buf.add(f"self.{name} = {wrapper_cls}({original_method}, client, grpclib_stub_attr)")
```
freider, line 109: Couldn't we just pass the `grpclib_stub` directly to the wrapper_cls intead of the name of the attribute?

dflemstr: Because of this stuff:

https://github.com/modal-labs/modal-client/blob/main/modal/client.py#L307-L325

ie. if the process is forked, we might want to re-establish connections and create new stubs, and retaining a reference to the old stub would likely break in many different ways as it would still be using the gRPC channel from the parent process.

modal/_utils/blob_utils.py
```diff
@@ -122,79 +122,78 @@ def remaining_bytes(self):
...
+    blob_size: int,
     max_part_size: int,
-    part_urls: list[str],
-    completion_url: str,
     upload_chunk_size: int = DEFAULT_SEGMENT_CHUNK_SIZE,
     progress_report_cb: Optional[Callable] = None,
-) -> None:
+):
+    def ceildiv(a, b):
+        return -(a // -b)
```
erikbern, line 179: 🤯

PR #2610, opened 2024-12-04
Add a way to tell the client to blob upload large async inputs when the runnable queue is full.

danielshaar:
One thing I had to think through when writing this simple PR is the nuance around how we plan to handle multiple inputs passed into FunctionMap. While we currently only pass in one input for experimental_spawn, we should have a reasonable behavior for the future. Let's say the client passes in 3 inputs, and the second one is too large. We would then send back a response containing 3 items - the first and third would just contain the index and nothing else (we wouldn't create those inputs). The second would contain the index and requires_blob_upload = true. This would inform the client of which inputs they should upload before trying again.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1552,6 +1552,7 @@ message FunctionPutInputsResponseItem {
   int32 idx = 1;
   string input_id = 2;
   string input_jwt = 3;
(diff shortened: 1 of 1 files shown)


danielshaar:
Deleting in favor of creating a new RPC.

PR #2611, opened 2024-12-04
Format modal client deprecation warning better

mwaskom:
We have some code that does rich formatting of deprecation warnings that was malfunctioning for client deprecations:

image

The main problem is that we assume the `modal.exception.DeprecationError` warnings are issued from `modal.exception.deprecation_warning`, which prefixes the message with the date of deprecation. But when we issue the client deprecation warning we don't include a date.

We also have a flag in `deprecation_warning` to suppress the source of the error in cases where the surrounding context is not relevant for the warning; I copied that approach here to clean up the message. Result:

image

I don't love how the alarm emoji look here, but they're helpful in the app logs. Could maybe omit them when `modal.is_local`.

### DIFF
--- a/modal/cli/app.py
+++ b/modal/cli/app.py
@@ -115,7 +115,7 @@ def logs(
     ```
 
     """
-    app_identifier = warn_on_name_option("stop", app_identifier, name)
+    app_identifier = warn_on_name_option("logs", app_identifier, name)
     app_id = get_app_id(app_identifier, env)
     stream_app_logs(app_id)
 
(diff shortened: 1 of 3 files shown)


modal/cli/app.py
```diff
@@ -115,7 +115,7 @@ def logs(
     ```
 
     """
-    app_identifier = warn_on_name_option("stop", app_identifier, name)
+    app_identifier = warn_on_name_option("logs", app_identifier, name)
```
mwaskom, line 118: Unrelated but happened to run this to verify that normal deprecation warnings are still formatted as we expect and noticed this bug!

mwaskom:
cc @erikbern is there somewhere else we're handling the client deprecation warnings now that we're trying to get rid of ClientHello?

erikbern (approved):

PR #2612, opened 2024-12-05
Fix 0.62 class with build method updating to 0.67

devennavani:
Fix 0.62 class with build method updating to 0.67. This was breaking because the build function runs before the app is updated to have a class service function for the class.

When the build function was running, the `AppGetObjects` request in container entrypoint wouldn't return any class service function (as one wasn't created). But `Cls.from_local` adds a `_class_service_function` attribute to the `Cls` object that points to the un-hydrated class service function `Function` object and is expected to by hydrated with the results of `AppGetObjects`.

The fix is to have `_hydrate_metadata` of `Cls`, in the branch where it is hydrating a 0.63+ class, check if the class service function object is actually hydrated (it has an object_id set).

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -354,7 +354,7 @@ class _Cls(_Object, type_prefix="cs"):
                     self._method_functions[method_name] = _Function._new_hydrated(
                         self._class_service_function.object_id, self._client, method_handle_metadata
                     )
-        elif self._class_service_function:
+        elif self._class_service_function and self._class_service_function.object_id:
             # A class with a class service function and method placeholder functions
             self._method_functions = {}
             for method in metadata.methods:


devennavani:
@prbot approve

modal/cls.py
```diff
@@ -354,7 +354,7 @@ def _hydrate_metadata(self, metadata: Message):
                     self._method_functions[method_name] = _Function._new_hydrated(
                         self._class_service_function.object_id, self._client, method_handle_metadata
                     )
-        elif self._class_service_function:
+        elif self._class_service_function and self._class_service_function.object_id:
```
freider, line 357: I think you can use `.is_hydrated`.

freider: Can we add a comment of when we are expected to hit this branch?

Could there still be cases for 0.63+ where the `class_service_function` is defined, but not yet hydrated, when we hit this code path?

freider (approved):
Ugh, I'd love to get rid of 0.62 support asap :'( A bit of a catch 22 though, if the upgrade is painful...

modal/cls.py
devennavani: > Could there still be cases for 0.63+ where the class_service_function is defined, but not yet hydrated, when we hit this code path?

Nope, that won't happen, which is why I decided not to include a comment, since we'll be deprecating 0.62 soon.

PR #2613, opened 2024-12-05
disable test_pdm_cache_automount_exclude due to ci failure

kramstrom:
`test_pdm_cache_automount_exclude` currently fails on ubuntu-20 `python3.13` probably due to some unpinned underlying library, disabling for not to unblock publishing new client versions


kramstrom:
@prbot approve

freider (commented):
THanks!

PR #2614, opened 2024-12-05
Enable `Cls.from_name`

freider:
## Support for `Cls.from_name`

This adds support for `Cls.from_name` by pushing method vs attribute resolution into the `__getattr__` method instead of the `Obj` constructor.

My first attempt separated `RemoteObj` from `LocalObj`, but when ironing out the details in that branch (#2305 ) it seemed to me like a lot of code would still have to be shared between the two and this might be a simpler path to something working. We could still split it out later if we feel it would simplify things.

### Complexity
The complexity of the classes code base stems from the huge matrix of possible ways they can currently be used:

* hydrated vs unhydrated class  (+ possibly hydration of each Obj service Function and the related method's Function)
* local vs non-local definition of the class available
* .local() vs .remote() calls
* decorated vs undecorated class at the import path in the container
* 0.62 vs 0.63 functionality difference (not having service_functions pre 0.63, but those classes are still lookup-able)
* accessing methods vs accessing attributes on the obj

## Current state

There are still a bunch of unrelated inconsistencies in how classes work that need to be fixed separately:
* .local() lifecyclehandling is semi-broken (only works with blocking enter functions at the moment, doesn't call exit lifecycles)
* `self` within containers isn't an `Obj` and *only* supports `self.method...` (not stuff like self.keep_warm())
* `self` for serialized or undecorated classes has members that evaluate to vanilla python functions, not modal functions
* hydration of classes and methods inside of containers is inconsistent - any parameterized class still needs ad-hoc rpc calls to hydrate itself (isn't hydrated at container startup)

---

---

</details>

## Changelog

* `Cls.from_name(...)` now works as a lazy alternative to `Cls.lookup()` that doesn't perform any IO until a method on the class is used for a .remote() call or similar


modal/cls.py
```diff
@@ -161,45 +161,60 @@ def __init__(
             check_valid_cls_constructor_arg(i + 1, arg)
         for key, kwarg in kwargs.items():
             check_valid_cls_constructor_arg(key, kwarg)
-
-        self._method_functions = {}
```
freider, line 165: Instead of eagerly constructing these lazy _Functions on class instantiation, we now conditionally run the same logic for each method in the `__getattr__` of Obj below, when users attempt ot access attribute or method of an instance.
This means we can now construct an `_Obj` without having method definitions of the class through local code or hydration metadata, as long as we have a `_Cls` object which itself is lazily loadable later on, when a user calls .remote()

modal/cls.py
```diff
@@ -161,45 +161,60 @@ def __init__(
...
-        user_cls_instance._modal_functions = self._method_functions  # Needed for PartialFunction.__get__
+        # TODO: OR (if simpler for now) replace all the PartialFunctions on the user cls
+        #   with getattr(self, method_name)
+
+        # user cls instances are only created locally, so we have all partial functions available
+        instance_methods = {}
+        for method_name in _find_partial_methods_for_user_cls(self._user_cls, _PartialFunctionFlags.FUNCTION):
+            instance_methods[method_name] = getattr(self, method_name)
+
+        user_cls_instance._modal_functions = instance_methods
```
freider, line 217: Not a fan of this, but leaving it as is in this PR

modal/cls.py
```diff
@@ -266,23 +283,74 @@ async def _aenter(self):
...
+            await resolver.load(method_function)  # get the appropriate method handle (lazy)
+            fun._hydrate_from_other(method_function)
+
+        # The reason we don't *always* use this lazy loader is because it precludes attribute access
+        # on local classes.
+        return _Function._from_loader(
+            method_loader,
+            repr,
+            deps=lambda: [],  # TODO: use cls as dep instead of loading inside method_loader?
+            hydrate_lazily=True,
```
erikbern, line 352: we could get rid of this argument in a subsequent PR since i think every object is hydrated lazily now

erikbern (approved):
Won't pretend I follow all the logic, but this seems fairly straightforward and less complex than I expected (taking into account the fact that we can remove a lot of complexity once 0.63 is gone). So that's great!

Still feels like there's maybe room to simplify the class code, but that's obviously a separate conversation!

freider:
> Still feels like there's maybe room to simplify the class code, but that's obviously a separate conversation!

Yeah. There is a ton of stuff I'd like to do to make it simpler and more consistent across usage patterns. Might push a few small PRs over the coming days to fix some of my main peeves

PR #2615, opened 2024-12-05
change `publish-base-images` cd step to require `client-versioning` and `publish-client`

kramstrom:
Updates `publish-base-images` to depend on `client-versioning` and `publish-client` steps instead of running in parallel to them, which causes a race condition where it can depend on the previous version of the published client mount, rather than the current one.

This PR ensures it will depend on the current one always

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -163,45 +163,6 @@ jobs:
           python -c 'import modal._container_entrypoint'
           if [ "${{ matrix.image-builder-version }}" == "2024.04" ]; then python -c 'import fastapi'; fi
 
-  publish-base-images:
-    name: |
-      Publish base images for ${{ matrix.image-name }} ${{ matrix.image-builder-version }}
-    if: github.ref == 'refs/heads/main'
-    needs: [client-test]
-    runs-on: ubuntu-20.04
-    timeout-minutes: 5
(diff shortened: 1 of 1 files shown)


mwaskom:
I wonder if the right solution here was to move the publishing of the client mount into the `client-versioning` workflow actually.

kramstrom:
> I wonder if the right solution here was to move the publishing of the client mount into the `client-versioning` workflow actually.

It would essentially result in the same thing right? Except that we separate it from the pypi upload step (which is quite fast) + then we would need to install all dependencies twice, in the client versioning job and in the client publish job. We could cache them I suppose but it would really just be very similar?

mwaskom:
We already needed to add the full `setup-cached-python` to `client-versioning`, IIRC because the `tasks.py` that we use for `inv` has various third-party dependencies.

> It would essentially result in the same thing right?

Maybe but not sure exactly what you're thinking of. Currently the problem is that, between doing `client-versioning` and the `publish-client-mount` step of `publish-client`, the `main` branch of the client repo is unusable because it specifies a client version that we haven't published a mount for. This affects CI but also development workflows (although you can work around it by doing `MODAL_SYNC_ENTRYPOINT=1`.

That said moving this step might surface other issues, has has been the case with every other part of this reorganization 😁

PR #2617, opened 2024-12-05
Include original task ID in checkpoint info response

pawalt:
## Describe your changes

This commit adds the task ID which creategd the checkpoint to CheckpointInfo. This allows us to restore a container with the exact same arguments which were used to create it.

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -589,6 +589,7 @@ message CheckpointInfo {
   string runtime_fingerprint = 4;
   int64 size = 5;
   bool checksum_is_file_index = 6;
+  string original_task_id = 7;
 }
 
 message ClassCreateRequest {


modal_proto/api.proto
```diff
@@ -588,7 +588,8 @@ message CheckpointInfo {
   string checkpoint_id = 3;
   string runtime_fingerprint = 4;
   int64 size = 5;
-  bool checksum_is_file_index = 6;
+  bool checksum_is_file_index = 6;
```
pawalt, line 591: I'm not sure why this is a diff

modal_proto/api.proto
```diff
@@ -589,6 +589,7 @@ message CheckpointInfo {
   string runtime_fingerprint = 4;
   int64 size = 5;
   bool checksum_is_file_index = 6;
+  string task_id = 7;
```
cweld510, line 592: Can we call this `original_task_id` or something else that makes it clear which task it refers to?

cweld510 (approved):

modal_proto/api.proto
pawalt: ya

pawalt:
@prbot automerge

PR #2618, opened 2024-12-05
Exclusively start using the FunctionData form ...

irfansharif:
.. when creating functions. We were already using the FunctionData form when creating functions that use GPU fallbacks. On the server, when trying to create functions using the Function form, we automatically translate it to the FunctionData form. By starting to do this in the client, we can start actually moving some of the configuration from the lower api_pb2.Function proto to the upper api_pb2.FunctionData proto for good. We'll still have a period where we need to support the duplication, because of older clients. But the fields can now be treated as being deprecated.

### DIFF
--- a/modal/functions.py
+++ b/modal/functions.py
@@ -746,95 +746,66 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
                         raise Exception(f"Dependency {dep} isn't hydrated")
                     object_dependencies.append(api_pb2.ObjectDependency(object_id=dep.object_id))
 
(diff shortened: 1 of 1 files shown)


irfansharif:
Hmm this is massively broken. I'll have to come back to it. We use the `api_pb2.Function` in the container entrypoint and then all through out in the client code, uses that use these fields present in `api_pb2.Function`. We'd have to start plumbing in the `api_pb2.FunctionData` equivalents from the worker, perhaps through the FunctionExtended union type? I'll try to sync up with folks first before proceeding.

PR #2619, opened 2024-12-05
refactor: make obvious that checkpoint_id is required

thundergolfer:
## Describe your changes

We've currently got old deprecated code paths and things will be simpler if we clear them up. As far as I can see the worker is _always_ sending a checkpoint ID to a snapshotting container.

---

    -  Old _workers_ are already sending the checkpoint ID and have been for a while.

---


thundergolfer:
Oh nice this breaks tests. Will fix.

cweld510 (approved):

PR #2621, opened 2024-12-06
Log specific deps and IDs when there's a mismatch

aksh-at:
It's painful to debug these when users run into them because we don't know what's changing, only the numbers. We can probably give a more specific warning, but quick fix for now.

Now prints:
image


aksh-at:
@prbot approve

erikbern (commented):
👍

PR #2622, opened 2024-12-06
Add ContainerHello endpoint

erikbern:
I've been working on removing the need for `ClientHello` which adds 10-100ms on every container start, as well as every local client creation. @gongy pointed out that it would be useful to have something container-side that at least hits the worker. It wouldn't be used for anything right now, but that way we have the optionality later on to add any logic we want. Since the worker would respond to this request, it would not add more than maybe a few milliseconds.

Adding it to the proto for now. Will add it on the worker at some point later. After that's rolled out, I'll wire it together so that we call this method from all containers on startup.


erikbern:
@prbot approve

gongy (commented):
:+1:

PR #2623, opened 2024-12-06
Add example of persisted queue to reference

erikbern:
Realized this wasn't part of the docstring.

Also put the code sample up top. The docs currently start with a bunch of stuff about partitions which seems a bit distracting for the average user.

image


erikbern:
@prbot approve

gongy (commented):
thanks for cleaning up

PR #2624, opened 2024-12-08
Decode messages for `StreamType.STDOUT`

aksh-at:
## Describe your changes

#2478 started returning raw bytes from stdout/stderr stream, which broke `StreamType.STDOUT` added in #2419. There were no tests for it, adding one now.

---

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -184,7 +184,7 @@ class _StreamReader(Generic[T]):
 
                 async for message in iterator:
                     if self._stream_type == StreamType.STDOUT and message:
-                        print(message, end="")
+                        print(message.decode("utf-8"), end="")
                     elif self._stream_type == StreamType.PIPE:
                         self._container_process_buffer.append(message)
                     if message is None:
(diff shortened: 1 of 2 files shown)


modal/io_streams.py
```diff
@@ -184,7 +184,7 @@ async def _consume_container_process_stream(self):
 
                 async for message in iterator:
                     if self._stream_type == StreamType.STDOUT and message:
-                        print(message, end="")
+                        print(message.decode("utf-8"), end="")
```
azliu0, line 187: should we condition this on `self._text`? it probably makes sense for the behavior to differ passing in `text=True` vs. `text=False` to `exec`

azliu0 (commented):
thanks for catching this!

modal/io_streams.py
aksh-at: Hm I'm not sure. When would you want to write raw bytes to stdout?

azliu0: good point—`subprocess.Popen` seems to only use the `text` arg when stdout/stderr are piped, so always sending text here seems more consistent

azliu0 (approved):

PR #2626, opened 2024-12-09
Handle non-PTY case for `modal shell` and `modal container exec`

aksh-at:
Resolves CLI-239

`modal container exec` now checks `stdout.isatty()` to determine whether to run in PTY mode or not. If not, it uses `StreamType.STDOUT` to write output directly to stdout. This means that you can now do this:

```python
modal shell -c 'uv pip list' > env.txt
```

It didn't make sense for `ContainerProcess.attach()` to support `pty=False`, so I removed that option and made it a `deprecation_error` directly. 

Added a test for `modal container exec` since none existed, but it's a bit tricky to test PTY behavior directly because of how we run this. Tested some basic cases manually:

image

Depends on #2624.

---

---

</details>

## Changelog
- `modal container exec` and `modal shell` now work correctly even when a pseudoterminal (PTY) is not present. This means, for example, that you can pipe the output of these commands to a file:

    ```python
    modal shell -c 'uv pip list' > env.txt
    ```

### DIFF
--- a/modal/cli/container.py
+++ b/modal/cli/container.py
@@ -1,5 +1,4 @@
 # Copyright Modal Labs 2022
-import sys
 from typing import Optional, Union
 
 import typer
@@ -8,7 +7,7 @@ from rich.text import Text
 from modal._pty import get_pty_info
 from modal._utils.async_utils import synchronizer
 from modal._utils.grpc_utils import retry_transient_errors
-from modal.cli.utils import ENV_OPTION, display_table, stream_app_logs, timestamp_to_local
+from modal.cli.utils import ENV_OPTION, display_table, is_tty, stream_app_logs, timestamp_to_local
 from modal.client import _Client
 from modal.config import config
 from modal.container_process import _ContainerProcess
@@ -67,7 +66,7 @@ async def exec(
     """Execute a command in a container."""
 
     if pty is None:
-        pty = sys.stdout.isatty()
+        pty = is_tty()
 
     client = await _Client.from_env()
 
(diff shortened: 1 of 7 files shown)


modal/cli/run.py
```diff
@@ -423,9 +423,7 @@ def shell(
     """
     env = ensure_env(env)
 
-    console = Console()
-    if not console.is_terminal:
-        raise click.UsageError("`modal shell` can only be run from a terminal.")
+    pty = sys.stdout.isatty()
```
pawalt, line 426: If we want to support this from a non-interactive terminal, we should change the doc comment to reflect this - right now, it only talks about using `shell` for interactive terminals.

modal/cli/container.py
```diff
@@ -71,7 +79,11 @@ async def exec(
     )
     res: api_pb2.ContainerExecResponse = await client.stub.ContainerExec(req)
 
-    await _ContainerProcess(res.exec_id, client).attach(pty=pty)
+    if pty:
+        await _ContainerProcess(res.exec_id, client).attach()
+    else:
+        # TODO: redirect stderr to its own stream?
```
pawalt, line 85: Yeah maybe it should go in its own stream. The API is implemented like this to match `subprocess`, but we could easily add `StreamType.STDERR`.

pawalt (approved):
Nice!

modal/cli/run.py
aksh-at: image
wow this doesn't show up well right now

PR #2627, opened 2024-12-09
Experimental spawn now uses FunctionAsyncInvoke

danielshaar:
## Describe your changes

SVC-220

Context in the linear ticket's linked notion doc.

---

  - **This is NOT compatible with servers that don't implement the new RPC!**

---


danielshaar:
I'll fix the tests once I have the server logic in.

danielshaar:
I'm not sure we're going with this route...

PR #2631, opened 2024-12-09
Retry transient sandbox exec failures

pawalt:
The lack of `retry_transient_errors` was an oversight.

## Describe your changes

Added transient error retry to Sandbox exec.

---

---

</details>

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -504,17 +504,16 @@ class _Sandbox(_Object, type_prefix="sb"):
             await secret.resolve(client=self._client)
 
         task_id = await self._get_task_id()
-        resp = await self._client.stub.ContainerExec(
-            api_pb2.ContainerExecRequest(
-                task_id=task_id,
-                command=cmds,
-                pty_info=_pty_info or pty_info,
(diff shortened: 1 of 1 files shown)


pawalt:
@prbot approve

pawalt:
@prbot automerge

azliu0 (commented):
thanks. since it seems like we're expecting changes to this soon, maybe it might be good to add a test? like [this](https://github.com/modal-labs/modal-client/blob/dff5ae94e46fb391d37da930e696bde2c4400537/test/file_io_test.py#L321)

azliu0 (approved):
thanks. since it seems like we're expecting changes to this soon, maybe it might be good to add a test? like [this](https://github.com/modal-labs/modal-client/blob/dff5ae94e46fb391d37da930e696bde2c4400537/test/file_io_test.py#L321)

pawalt:
@azliu0 I think we should test this thing more holistically - there's a change coming down the pipe to make all client stub calls retried automatically, so I'd rather test as part of that change if it's cool with you

PR #2632, opened 2024-12-09
Let FunctionGet return warnings

erikbern:
Since we're removing `ClientHello`, @mwaskom pointed out there won't be a great way to warn that a client is deprecated. However we can move the logic to `AppPublish` and `FunctionGet` and cover most use cases. `AppPublish` already returns a list of warnings. Adding warnings support to `FunctionGet`.

Will make the server change separately.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -27,7 +27,7 @@ from ._utils import async_utils
 from ._utils.async_utils import TaskContext, synchronize_api
 from ._utils.grpc_utils import connect_channel, create_channel, retry_transient_errors
 from .config import _check_config, _is_remote, config, logger
-from .exception import AuthError, ClientClosed, ConnectionError, DeprecationError, VersionError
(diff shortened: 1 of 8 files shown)


erikbern:
Actually let me add a test for this as well.

mwaskom:
Do we need to do `ClassGet` too? (Or is it `ClsGet`?)

mwaskom:
I don't love that we're propagating the semi-magical "getting a warning must mean a deprecation" assumption from `ClientHello`; I think that has constrained us in the past. Maybe better to start adopting a more structured message for these warnings to so we can distinguish them?

In fact, for `AppPublish`, the existing `warnings` are not assumed to be deprecation-related (we're currently using them to warn about deployments reaching a plan limit). So we might need to do that anyway. Not sure how complicated the proto migration will be :/

erikbern:
Will make a proto-only change for this first

PR #2633, opened 2024-12-09
Michael/2024 12 09 parameter renaming

mwaskom:
A little bit too much in this PR:

- Renamed two parameters that are potentially confusing in object constructors,`label` and `tag`, to use `name` consistently
- Introduced a `renamed_parameter` decorator to do this in a minimally-invasive way while providing backwards compatibility for people spelling their function calls with full kwargs
- Moved the `deprecation_warning` / `deprecation_error` helpers out of public API into a new utilities module with `renamed_parameter`
- Updated the `inv show-deprecations` task to understand deprecations handled through `renamed_parameter`

Closes CLI-284

## Changelog

- The `label` and `tag` parameters in Modal's `.from_name` and `.lookup` object constructors have been changed to `name`, avoiding confusion with other concepts.
- The internal `deprecation_warning` and `deprecation_error` utility functions have been moved out of public API.


mwaskom:
Argh typing is going to be complicated.

modal/_utils/deprecation.py
```diff
@@ -0,0 +1,75 @@
...
+    date: tuple[int, int, int],
+    old_name: str,
+    new_name: str,
+) -> Callable[[Callable[P, R]], Callable[P, R]]:
+    def decorator(func: Callable[P, R]) -> Callable[P, R]:
+        @functools.wraps(func)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+            if old_name in kwargs:
+                kwargs[new_name] = kwargs.pop(old_name)
+                func_name = func.__qualname__.removeprefix("_")  # Avoid confusion when synchronicity-wrapped
```
freider, line 65: This would be "properly" done using `synchronizer._translate_out(func).__qualname__`, but it's possible that would have its own snags so if it doesn't work this *should* be ok given that we have it standardized to use `_` prefixes on our implementation functions

modal/_utils/deprecation.py
```diff
@@ -0,0 +1,75 @@
...
+def renamed_parameter(
+    date: tuple[int, int, int],
+    old_name: str,
+    new_name: str,
+) -> Callable[[Callable[P, R]], Callable[P, R]]:
+    def decorator(func: Callable[P, R]) -> Callable[P, R]:
+        @functools.wraps(func)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+            if old_name in kwargs:
+                kwargs[new_name] = kwargs.pop(old_name)
```
freider, line 64: These two lines's usage of `kwargs` trip up `sigtools.signature`'s "smart" wrapper parser that we use to emit type stubs, which is why the type checker fails :(

modal/_utils/deprecation.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # This is a lower-level function that warnings.warn uses
+    warnings.warn_explicit(f"{date(*deprecated_on)}: {msg}", warning_cls, filename, lineno)
+
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+def renamed_parameter(
```
erikbern, line 55: would it be simpler if this isn't a decorator and we just have code like this
```python
def f(
    new_parameter: int = 42,
    old_parameter: Optional[int] = None
):
    if old_parameter is not None:
        deprecation_warning(...)
        new_parameter = old_parameter
```

mwaskom: Simpler in the sense of avoiding having to fight with whatever synchronicity is doing but it will be a lot more clutter in the code and produce signatures that are less clear about the right way going forward. I would call that more complex?

Also some of the parameters that we intent to rename have `None` as a valid argument so we'd need to introduce a "default" sentinel ... again doable just more headache. The problem is that synchronicity is too clever.

erikbern: Probably more lines in total but I think the maintenance cost of those lines is very low – the behavior is self-explanatory and they don't interfere with anything else.

Two problems with decorator is (a) typing (b) longer tracebacks. Those are issues even without synchronicity. I do agree that synchronicity blows things up though!

Not a hill I would die on but I would have personally just used "dumb" if-statements for these deprecations. Speculative but it also helps if we want to do any sort of value transformation.

mwaskom:
Can't get people on board with decorators so will pursue the wordier implementation.

PR #2636, opened 2024-12-10
CLI-287: add `ignore` parameter to`Image.add_local_dir` and `Image.copy_local_dir` methods

kramstrom:
- Adds an `ignore` parameter to our `Image` `add_local_dir` and `copy_local_dir` methods. It is similar to the `condition` method on `Mount` methods but instead operates on a `Path` object. It takes either a list of string patterns to ignore which follows the `dockerignore` syntax implemented in our `FilePatternMatcher` class, or you can pass in a callable which allows for more flexible selection of files.

Also renamed `PatternMatcher` to `FilePatternMatcher` to better match what it's doing and moved it to the public API.

Usage:

```python
img.add_local_dir(
  "./local-dir", 
  remote_path="/remote-path", 
  ignore=FilePatternMatcher("**/*", "!*.txt") # ignore everything except files ending with .txt
)

img.add_local_dir(
  ...,
  ignore=~FilePatternMatcher("**/*.py") # can be inverted for when inclusion filters are simpler to write
)

img.add_local_dir(
  ...,
  ignore=["**/*.py", "!module/*.py"] # ignore all .py files except those in the module directory
)

img.add_local_dir(
  ...,
  ignore=lambda fp: fp.is_relative_to("somewhere") # use a custom callable
)
```

which will add the `./local-dir` directory to the image but ignore all files except `.txt` files


modal/mount.py
```diff
@@ -324,25 +324,39 @@ def add_local_dir(
...
         # add files from subdirectories as well
         recursive: bool = True,
+        include_files: Optional[Callable[[Path], bool]] = None,
     ) -> "_Mount":
         """
         Add a local directory to the `Mount` object.
         """
+
+        if condition is not None:
+            deprecation_warning((2024, 12, 10), "`condition` is deprecated, please use `include_files` instead")
```
mwaskom, line 334: Since we plan to deprecate `Mount` as public-facing API I'm not sure it's worth the churn to substitute the new name here.

modal/_utils/local_file_filter.py
```diff
@@ -0,0 +1,25 @@
...
+
+from .pattern_matcher import PatternMatcher
+
+
+class LocalFileFilter:
+    """
+    git-/docker-ignore-like pattern matcher for local files.
+    """
+
+    def __init__(self, pattern: str, *more_patterns: str) -> None:
```
mwaskom, line 11: Why not just `self, *patterns: str`?

modal/image.py
```diff
@@ -638,7 +638,14 @@ def add_local_file(self, local_path: Union[str, Path], remote_path: str, *, copy
...
         return self._add_mount_layer_or_copy(mount, copy=copy)
 
-    def add_local_dir(self, local_path: Union[str, Path], remote_path: str, *, copy: bool = False) -> "_Image":
+    def add_local_dir(
+        self,
+        local_path: Union[str, Path],
+        remote_path: str,
+        *,
+        copy: bool = False,
+        include_files: Optional[Callable[[Path], bool]] = None,
```
mwaskom, line 647: I was sort of imagining that we would support a string-based API for pattern based filtering, seems easier than asking users to construct their own callable. Also less flexible though! Would be fun to think through the different things users are going to want to do with this.

mwaskom (commented):
Just a drive-by; saw it's still WIP so apologies if you're thinking about these already.

modal/mount.py
kramstrom: Yeah makes sense, think we should just let condition be as is & set a deprecation warning on mounts as a whole later instead?

modal/image.py
kramstrom: Added this signature now: `ignore: Union[list[str], Callable[[Path], bool]] = []` to allow for both use cases!

modal/mount.py
```diff
@@ -324,6 +324,9 @@ def add_local_dir(
         condition: Optional[Callable[[str], bool]] = None,
         # add files from subdirectories as well
         recursive: bool = True,
+        # Predicate filter function for file selection, which should accept a filepath and return `True` for inclusion.
+        # Defaults to including all files.
+        ignore: Union[list[str], Callable[[Path], bool]] = [],
```
mwaskom, line 329: Note that we are 3.9+ now
```suggestion
        ignore: list[str] | Callable[[Path], bool] = [],
```
will probably add pyupgrade as a pre-commit step at some point so you don't need to think about this (although it had a few annoyin edge cases, e.g. around `modal.Dict`.

modal/mount.py
```diff
@@ -324,6 +324,9 @@ def add_local_dir(
         condition: Optional[Callable[[str], bool]] = None,
         # add files from subdirectories as well
         recursive: bool = True,
+        # Predicate filter function for file selection, which should accept a filepath and return `True` for inclusion.
```
mwaskom, line 327: Is the explanation backwards now that we consider this as exclusive rather than inclusive?

mwaskom: (Also I guess it's missing an explanation for the list of strings spelling).

modal/__init__.py
```diff
@@ -79,4 +80,5 @@
     "web_endpoint",
     "web_server",
     "wsgi_app",
+    "LocalFileFilter",
```
mwaskom, line 83: I'm not understanding why we need this new public object.

modal/image.py
```diff
@@ -638,7 +639,14 @@ def add_local_file(self, local_path: Union[str, Path], remote_path: str, *, copy
...
         return self._add_mount_layer_or_copy(mount, copy=copy)
 
-    def add_local_dir(self, local_path: Union[str, Path], remote_path: str, *, copy: bool = False) -> "_Image":
+    def add_local_dir(
+        self,
+        local_path: Union[str, Path],
+        remote_path: str,
+        *,
+        copy: bool = False,
+        ignore: Union[list[str], Callable[[Path], bool]] = [],
```
mwaskom, line 648: Juxtaposing "copy" and "ignore" like this makes me read `ignore` as wanting a boolean. I guess the type signature is clear though. But we do need to document how the different arguments are interpreted.

modal/mount.py
```diff
@@ -23,11 +23,12 @@
 from ._utils.async_utils import aclosing, async_map, synchronize_api
```
mwaskom, line 23: Do we need to make any changes to `Mount` if we're planning to deprecate it as public API? I could see the argument that it would be confusing to have an inversion between the public interface / internal implementation I guess.

modal/__init__.py
kramstrom: Removed!

modal/mount.py
kramstrom: Yeah old explanation

modal/_utils/local_file_filter.py
kramstrom: yeah fixed now

modal/mount.py
kramstrom: Yeah I think the inversion is annoying but also since the `ignore` callable works on `Path` objects and the `condition` we have for mount methods only allows `str` you would basically have to wrap every `ignore` to work on strings instead

modal/_utils/local_file_filter.py
```diff
@@ -0,0 +1,58 @@
+# Copyright Modal Labs 2024
+from pathlib import Path
+
+from .pattern_matcher import PatternMatcher
+
+
+class LocalFileFilter:
+    """Allows checking paths against one or more patterns."""
+
+    def __init__(self, *pattern: str, invert: bool = False) -> None:
```
freider, line 10: I'd prefer to not make invert an argument to the constructor, it can just be a lambda in the `__invert__` method since all we need is a callable :)

modal/_utils/local_file_filter.py
```diff
@@ -0,0 +1,58 @@
...
+        ```python
+
+        filter = LocalFileFilter("*.py")
+
+        assert filter(Path("foo.py"))
+
+        assert not ~filter(Path("foo.py"))
+        ```
+        """
+        return LocalFileFilter(*self.patterns, invert=True)
```
freider, line 58: I.e. this could be implemented like:
```
return lambda path: not self(path)
```

modal/mount.py
kramstrom: I think this was introduced in 3.10 actually, we use this syntax in some places but only when we use `from __future__ import annotations` which I would like to avoid

modal/mount.py
```diff
@@ -324,6 +324,9 @@ def add_local_dir(
         condition: Optional[Callable[[str], bool]] = None,
         # add files from subdirectories as well
         recursive: bool = True,
+        # Predicate filter function for file selection, which should accept a filepath and return `True` for inclusion.
+        # Defaults to including all files.
+        ignore: Union[Sequence[str], Callable[[Path], bool]] = [],
     ) -> "_Mount":
```
freider, line 330: Thinking about it, it might actually make sense to basically copy the Mount add_local* methods into new "private" variants (underscore prefixed) that only have the new ignore arg and not the old condition (and leave the old methods untouched)

Prevents users from starting to use it with Mount that we intend to deprecate, and gives them a reason to switch to the image alternatives.

modal/_utils/local_file_filter.py
kramstrom: makes sense!

modal/mount.py
kramstrom: yeah I like it!

modal/mount.py
kramstrom: As per @freider's comment below I opted to add a private method to the `Mount` object for the image methods to use instead and leave the signature of the `Mount` copy & add methods as is

modal/mount.py
```diff
@@ -155,31 +156,30 @@ def top_level_paths(self) -> list[tuple[Path, PurePosixPath]]:
         return [(self.local_dir, self.remote_path)]
 
 
-def module_mount_condition(module_base: Path):
+def module_mount_ignore_condition(module_base: Path) -> Callable[[Path], bool]:
```
freider, line 159: Just in case someone is using this (it's public after all) it might be better to create a new one than modifying the existing one

modal/_utils/local_file_filter.py
```diff
@@ -0,0 +1,54 @@
+# Copyright Modal Labs 2024
+from pathlib import Path
+from typing import Callable
+
+from .pattern_matcher import PatternMatcher
+
+
+class LocalFileFilter:
```
freider, line 8: Maybe this should be called "IgnorePatterns" or something, and make it clear in that it *excludes* on `True` values.  The word "Filter" and some of these docstrings make it sound like its the other way around

kramstrom: Hm, this class in itself doesn't exclude or include specifically though, I guess that's more how it's used in the `Image` methods, but I'm open to renaming it, but maybe something like `FilePathPatternMatcher` or something to show that it's related to our `PatternMatcher`?

modal/mount.py
mwaskom: oh waow you're right

modal/_utils/local_file_filter.py
mwaskom: I'm still a little confused about we need to be wrapping our own `PatternMatcher` object here

freider: Sorry, you're right, not sure what I was thinking in the original comment. This returns Truthy for matching items and then it makes sense that it's called a Matcher or Filter or something like this

kramstrom: Went for using the `PatternMatcher` object directly but modifying it to take in `Path` objects and add an invert operator

modal/_utils/pattern_matcher.py
```diff
@@ -151,44 +152,43 @@ def match(self, path: str) -> bool:
...
                 self.exclusions = True
             # In Python, we can proceed without explicit syntax checking
-            new_pattern.cleaned_pattern = pattern
-            new_pattern.dirs = pattern.split(os.path.sep)
+            new_pattern.cleaned_pattern = p
+            new_pattern.dirs = p.split(os.path.sep)
             self.patterns.append(new_pattern)
 
-    def matches(self, file_path: str) -> bool:
+    def matches(self, file_path: Path) -> bool:
```
freider, line 183: Perhaps name it `_matches` to indicate this is not part of the public interface of a pattern matcher in Modal.

modal/_utils/pattern_matcher.py
```diff
@@ -151,44 +152,43 @@ def match(self, path: str) -> bool:
...
         """Check if the file path or any of its parent directories match the patterns.
 
         This is equivalent to `MatchesOrParentMatches()` in the original Go
         library. The reason is that `Matches()` in the original library is
         deprecated due to buggy behavior.
         """
         matched = False
-        file_path = os.path.normpath(file_path)
-        if file_path == ".":
+        if str(file_path) == ".":
```
freider, line 191: Is the `normpath` not needed for some cases?

freider, line 191: If we're treating file_path here as a string everywhere, we might as well make the `_matches` method take a string like before and do the conversion to str when we call it from `__call__`, instead of having to re-cast it in multiple places in this method

modal/image.py
```diff
@@ -650,12 +661,44 @@ def add_local_dir(self, local_path: Union[str, Path], remote_path: str, *, copy:
         copy=True can slow down iteration since it requires a rebuild of the Image and any subsequent
         build steps whenever the included files change, but it is required if you want to run additional
         build steps after this one.
+
+        **Usage:**
+
+        ```python
+        from modal._utils.pattern_matcher import PatternMatcher
```
freider, line 668: Lets put PatternMatcher (or whatever we want to call it) in a publicly importable module (no underscores), exported through `modal.__init__`. 

Otherwise we'd start breaking user's code when we refactor these internals

modal/_utils/pattern_matcher.py
```diff
@@ -151,44 +152,43 @@ def match(self, path: str) -> bool:
 class PatternMatcher:
```
freider, line 152: Since we might want this as a top level `modal.X` entity, we might have to think of a name that is less general and conveys that it's matching paths specifically. Naming is hard though... What about `MatchingFiles`?

modal/image.py
freider: See comment above about naming this class too (if we're moving it to a user importable, we need to make the name more specific)

freider (commented):
good stuff! Some smaller final suggestions in comments

modal/_utils/pattern_matcher.py
kramstrom: How about the `noun` `FilePatternMatcher` ? 😄

modal/image.py
kramstrom: Moved it!

modal/_utils/pattern_matcher.py
kramstrom: Makes sense!

modal/image.py
```diff
@@ -697,19 +740,56 @@ def add_local_python_source(self, *modules: str, copy: bool = False) -> "_Image"
...
 
         This works in a similar way to [`COPY`](https://docs.docker.com/engine/reference/builder/#copy)
         works in a `Dockerfile`.
+
+        **Usage:**
+
+        ```python
+        from modal import FilePatternMatcher
+
+        image = modal.Image.debian_slim().copy_local_dir(
```
freider, line 765: Use `.add_local_dir()` instead in the examples - we're getting rid of the `copy_` equivalents soon! 😬

freider (approved):
lgtm!

PR #2637, opened 2024-12-10
Proto definition for server warnings

erikbern:
Just proto definitions or now. I'll populate them on the server next. Once that's done, I'll go back to the client and add code that reads these warnings. In that PR I'll also remove the current check in `ClientHello` so that we don't catch multiple warnings.

This is the only remaining thing before we can remove `ClientHello`.


mwaskom (approved):
Thanks for bearing with me!

PR #2638, opened 2024-12-11
Add protos for directory level sandbox api

azliu0:
## Describe your changes

This pr adds protos for the directory-level api in WRK-521.

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -811,6 +811,20 @@ message ContainerFilesystemExecGetOutputRequest {
   float timeout = 2;
 }
 
+message ContainerFileLsRequest {
+  string path = 1;
+}
+
+message ContainerFileMkdirRequest {
+  string path = 1;
+  bool parents = 2;
+}
+
+message ContainerFileRmRequest {
+  string path = 1;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -772,6 +772,15 @@ message ContainerFileFlushRequest {
   string file_descriptor = 1;
 }
 
+message ContainerFileLsRequest {
+  string path = 1;
+}
+
+message ContainerFileMkdirRequest {
+  string path = 1;
+  bool parents = 2;
```
pawalt, line 781: maybe `make_parents` would be more obvious

pawalt (approved):
LGTM

modal_proto/api.proto
azliu0: The naming mostly comes from here: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir. 
Does that work? Don't feel too strongly about this, as it's a proto anyways

pawalt: I don't feel strongly either - this is fine; it's just internal terminology that isn't obvious to the reader IMO

PR #2639, opened 2024-12-11
Use AppPublishResponse.server_warnings instead

erikbern:
This gets rid of the `warnings` field and replaces it with `server_warnings` which will be populated server-side very soon.


modal/runner.py
```diff
@@ -455,7 +455,7 @@ class DeployResult:
     app_id: str
     app_page_url: str
     app_logs_url: str
-    warnings: list[str]
+    warnings: list[api_pb2.Warning]
```
mwaskom, line 458: Meh since the `DeployResult` is public API we should probably coerce the proto message into a dataclass or some other representation that users are meant to interact with.

mwaskom (approved):

PR #2641, opened 2024-12-11
Remove ClientHello (attempt 2)

erikbern:
Not sure what happened but I did something weird with Git and Github decided to close https://github.com/modal-labs/modal-client/pull/2439 – I'm not able to reopen it for whatever reason so I'm just creating a new PR instead. This is rebased on latest main as well! Otherwise no changes.

Anyway, this PR removes `ClientHello` finally. The point of this is to eliminate a server roundtrip for every container startup, which can be 100-300ms in remote regions. It eliminates two server roundtrips for ephemeral apps since one is happening locally as well. There are still other server roundtrips (`AppGetObjects`) that we should remove separately.

Bumped the minor version to 0.68 so that there's a clear deprecation point for `ClientHello` support in the server.

Note that we still have `client.hello()` as a dummy method to verify connections though – we mostly use it in some tests, and we use it to verify the credentials during the token flow. But we can remove all logic later so it's a pure dummy method.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -155,11 +155,6 @@ class _Client:
 
     async def __aenter__(self):
         await self._open()
-        try:
-            await self.hello()
-        except BaseException:
-            await self._close()
-            raise
         return self
 
     async def __aexit__(self, exc_type, exc, tb):
@@ -175,7 +170,6 @@ class _Client:
         client = cls(server_url, api_pb2.CLIENT_TYPE_CLIENT, credentials=None)
         try:
             await client._open()
-            # Skip client.hello
             yield client
         finally:
             await client._close()
@@ -226,7 +220,6 @@ class _Client:
             client = _Client(server_url, client_type, credentials)
             await client._open()
             async_utils.on_shutdown(client._close())
-            await client.hello()
             cls._client_from_env = client
             return client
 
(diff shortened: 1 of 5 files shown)


erikbern:
eCDF of the `ClientHello` latency, from before:

image

modal/client.py
```diff
@@ -262,8 +250,8 @@ async def verify(cls, server_url: str, credentials: tuple[str, str]) -> None:
         """mdmd:hidden
         Check whether can the client can connect to this server with these credentials; raise if not.
         """
-        async with cls(server_url, api_pb2.CLIENT_TYPE_CLIENT, credentials):
-            pass  # Will call ClientHello RPC and possibly raise AuthError or ConnectionError
+        async with cls(server_url, api_pb2.CLIENT_TYPE_CLIENT, credentials) as client:
+            client.hello()  # Will call ClientHello RPC and possibly raise AuthError or ConnectionError
```
mwaskom, line 254: I guess we could also have just gotten rid of the `hello` method and moved the logic in here if it's the only place we still use it.

mwaskom (approved):

modal/client.py
erikbern: I was thinking about it and I think it's actually useful to keep the `hello` method indefinitely as a simple ping method – it's useful for tests and verifying tokens. I might revert some of the recent changes and add back logic for the warnings etc, just in a way where it's not a part of client construction.

erikbern:
Finally – this only took like 2 months (not more than a couple of days total work though, just lots of dependencies)

PR #2643, opened 2024-12-11
Improve presentation of warnings originating from modal server

mwaskom:
- Adds a `ServerWarning` class that we now also use to decide which warnings get a special Rich panel in CLI output
- Moves `print_server_warnings` into `_traceback` to suppress it from public API
- Calls `print_server_warnings` from `publish_app` so we warn about deprecated client versions for both ephemeral and deployed apps
- Removes the logic to explicitly print deployment warnings from the `deploy` CLI command, since it now happens automatically with the warnings interception in the CLI

Warnings now look like this when using the CLI:
image

When not using the CLI, they appear like regular Python warnings, but with the origin more clearly attributed to the Modal server:
image

### DIFF
--- a/modal/cli/_traceback.py
+++ b/modal/cli/_traceback.py
@@ -1,5 +1,6 @@
 # Copyright Modal Labs 2024
 """Helper functions related to displaying tracebacks in the CLI."""
+
 import functools
 import re
 import warnings
@@ -11,7 +12,7 @@ from rich.syntax import Syntax
 from rich.text import Text
 from rich.traceback import PathHighlighter, Stack, Traceback, install
 
-from ..exception import DeprecationError, PendingDeprecationError
+from ..exception import DeprecationError, PendingDeprecationError, ServerWarning
 
 
 @group()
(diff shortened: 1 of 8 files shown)


modal/cli/_traceback.py
```diff
@@ -180,10 +181,16 @@ def showwarning(warning, category, filename, lineno, file=None, line=None):
             except OSError:
                 # e.g., when filename is "<unknown>"; raises FileNotFoundError on posix but OSError on windows
                 pass
+            if issubclass(category, ServerWarning):
+                title = "Modal Warning"
```
erikbern, line 185: could make this a property of the class (or an abstract method!) so you don't have to check the type

erikbern (approved):

mwaskom:
@prbot automerge

modal/cli/_traceback.py
mwaskom: Yeah not a bad idea. We've talked a little bit about other useful metadata that we could add to our exceptions (e.g. surfacing the ID of the modal object that an exception is from / about).

PR #2644, opened 2024-12-11
Spawn should just use the async function call invocation

danielshaar:
## Describe your changes

SVC-220

Since the server treats it the same as async legacy, this is a NOOP.


danielshaar:
@prbot approve

PR #2645, opened 2024-12-11
Add server_warnings to ClientHello

erikbern:
Thinking more about it, I think it's actually useful to keep the `ClientHello` method indefinitely, but in a way where it's mostly used for tests and some other fringe things (the token flow uses it to verify credentials).

It simplifies testing a bit to have the same behavior with `ClientHello` as with `AppPublish` etc and return structured server warnings, so let's add this and populate it server-side. This will let me bring back this test and some other stuff: https://github.com/modal-labs/modal-client/pull/2640/files#diff-c3f9238c73a319c67ee49a92bcd534772c62b66a7a3c3b61da1ad630d178488bL102

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -132,10 +132,10 @@ class _Client:
         # Remove cached client.
         self.set_env_client(None)
 
-    async def hello(self, total_timeout: Optional[float] = None):
+    async def hello(self):
         """Connect to server and retrieve version information; raise appropriate error for various failures."""
         logger.debug(f"Client ({id(self)}): Starting")
-        resp = await retry_transient_errors(self.stub.ClientHello, empty_pb2.Empty(), total_timeout=total_timeout)
+        resp = await retry_transient_errors(self.stub.ClientHello, empty_pb2.Empty())
         print_server_warnings(resp.server_warnings)
 
     async def __aenter__(self):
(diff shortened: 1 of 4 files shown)


modal/client.py
```diff
@@ -134,24 +132,11 @@ async def _close(self, prep_for_restore: bool = False):
         # Remove cached client.
         self.set_env_client(None)
 
-    async def hello(self):
+    async def hello(self, total_timeout: float | None = None):
```
mwaskom, line 135: What makes a "total_timeout" different from a "timeout"?

mwaskom: Is the distinction important for the external interface?

erikbern: no, I just added this argument for testing purposes. I can prefix it with underscore.

Not even sure if that test is needed tbh. We're already covering the same functionality in `grpc_utils_test.py`. I'm open to just removing this tests (along with a few other tests for connection issues). wdyt?

erikbern: I removed 2 tests in the last commit, don't think they make sense anymore.

mwaskom (approved):
LGTM

erikbern:
Thanks @mwaskom 

will follow up with server changes for this right after

PR #2649, opened 2024-12-12
blobs: omit redundant sha256 digest

danielnorberg:
Currently when uploading large blobs for volume put we hash the file twice:

1. First in `_get_file_upload_spec`
2. Then again in `blob_upload_file`

This PR passes in the initial digest from the file spec in to `blob_upload_file` so that it can skip the second digest. Note that `blob_upload_file` still computes an md5 digest for use in the presigned upload url. We could possibly also remove that, but that requires server side changes as well. Future work.

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -305,9 +305,12 @@ async def blob_upload(payload: bytes, stub: ModalClientModal) -> str:
 
 
 async def blob_upload_file(
-    file_obj: BinaryIO, stub: ModalClientModal, progress_report_cb: Optional[Callable] = None
+    file_obj: BinaryIO,
+    stub: ModalClientModal,
+    progress_report_cb: Optional[Callable] = None,
+    sha256_hex: Optional[str] = None,
 ) -> str:
-    upload_hashes = get_upload_hashes(file_obj)
+    upload_hashes = get_upload_hashes(file_obj, sha256_hex=sha256_hex)
     return await _blob_upload(upload_hashes, file_obj, stub, progress_report_cb)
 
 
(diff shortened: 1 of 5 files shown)


modal/_utils/blob_utils.py
```diff
@@ -385,6 +385,7 @@ def _get_file_upload_spec(
         if size >= LARGE_FILE_LIMIT:
             use_blob = True
             content = None
+            sha256_hex = hashlib.file_digest(fp, "sha256")
```
danielnorberg, line 388: derp

modal/_utils/hash_utils.py
```diff
@@ -49,11 +49,18 @@ class UploadHashes:
     sha256_base64: str
 
 
-def get_upload_hashes(data: Union[bytes, BinaryIO]) -> UploadHashes:
+def get_upload_hashes(data: Union[bytes, BinaryIO], sha256_hex: Optional[str] = None) -> UploadHashes:
     md5 = hashlib.md5()
```
danielnorberg, line 53: We could also compute the md5 digest in parallel with the first sha256 digest, on a separate thread. Future work.

freider (approved):
Nice find!

gimaker (approved):
:rocket:

danielnorberg:
### Before
```
[ThreadPoolExecutor-0_0] 2024-12-13T10:44:22+0100 get_sha256_hex took 1.972s
[Thread-1 (thread_inner)] 2024-12-13T10:44:30+0100 get_upload_hashes took 8.001s
```

### After
```
[ThreadPoolExecutor-0_0] 2024-12-13T10:44:57+0100 get_sha256_hex took 2.057s
[Thread-1 (thread_inner)] 2024-12-13T10:45:03+0100 get_upload_hashes took 6.289s (['md5'])
```

It seems that unfortunately the remaining md5 hash is very expensive. We should probably omit that and only use the sha256 hash.

PR #2650, opened 2024-12-12
Add directory api for sandboxes

azliu0:
## Describe your changes

Resolves WRK-521.

### Blockers
- [x] https://github.com/modal-labs/modal/pull/18168 roll out to all workers.

---

---

</details>

## Changelog

- Sandboxes now support additional filesystem commands `mkdir`, `rm`, and `ls`.
```python
app = modal.App.lookup("sandbox-fs", create_if_missing=True)
sb.mkdir("/foo")
with sb.open("/foo/bar.txt", "w") as f:
    f.write("baz")
print(sb.ls("/foo"))
```


azliu0:
@ellipsis-dev quiet

azliu0:
@ellipsis quiet

modal/file_io.py
```diff
@@ -188,6 +190,12 @@ async def _wait(self, exec_id: str) -> bytes:
                 raise
         return output
 
+    async def _parse_list_output(self, output: bytes) -> list[str]:
```
pawalt, line 193: Don't think this needs to be `async`.

modal/file_io.py
```diff
@@ -337,6 +345,49 @@ async def seek(self, offset: int, whence: int = 0) -> None:
         )
         await self._wait(resp.exec_id)
 
+    @classmethod
+    async def ls(cls, path: str, client: _Client, task_id: str) -> list[str]:
+        """List the contents of the provided directory."""
+        self = cls.__new__(cls)
```
pawalt, line 351: Is there any cleanup that needs to be done of these beyond simple garbage collect?

pawalt (commented):
Looks mostly reasonable

modal/file_io.py
azliu0: don't think so, as these operations don't affect any state anywhere else

modal/file_io.py
```diff
@@ -188,6 +190,12 @@ async def _wait(self, exec_id: str) -> bytes:
                 raise
         return output
 
+    def _parse_list_output(self, output: bytes) -> list[str]:
```
pawalt, line 193: On further thought, I don't think it makes sense for this to be pulled out into its own function. It's only used in `ls`, so I think you should keep it there.

pawalt (approved):
I think it's a bit undesirable how we're tying the client to the `daemon` serialization, but for something as simple as `paths` I think it's OK given how much it simplifies the implementation.

Wait for all changes to roll on workers before merging.

modal/sandbox.py
```diff
@@ -567,6 +567,21 @@ async def open(
         task_id = await self._get_task_id()
         return await _FileIO.create(path, mode, self._client, task_id)
 
+    async def ls(self, path: str) -> list[str]:
```
mwaskom, line 570: j/w but did you consider using `Path` objects here instead of `str`?

azliu0: I hadn't, but it seems like it would be a good addition to support it. There are a few other instances in `FileIO` that take in `str` paths only

PR #2653, opened 2024-12-13
Adds more extensive testing of classes

freider:
Particularly which operations incur rpc requests

Done in order to prevent regressions on larger class refactors


freider:
@prbot approve

PR #2655, opened 2024-12-13
blobs: omit md5 for multipart uploads

danielnorberg:
Turns out that the server does not use the md5 of files when performing multi-part uploads. And the creation of the unused md5 digest is [very expensive](https://github.com/modal-labs/modal-client/pull/2649#issuecomment-2541013862).

This PR omits sending a real md5 for files larger than 1GiB, which we know** will always be multipart uploaded.

** We'll likely remove md5 entirely and only use the sha256 hash. Until then we'll make sure not to raise the multipart upload threshold higher. If anything we're [lowering it](https://github.com/modal-labs/modal/pull/18208). 

TODO:
- [x] Test volume put 2GiB file
- [x] Test volume put 16MiB file
- [x] Test mount 2GiB file
- [x] Test mount 16MiB file

## Benchmark

TL;DR: We save ~2s per GiB by not doing the md5 hash. For a 64GiB file that's ~2 minutes before we even start uploading. For a 256GiB file that's ~8 minutes. 

The below benchmark uploads a 16GiB file.

```
MODAL_LOGLEVEL=DEBUG modal volume put dano-test data-16g  --force >/dev/null
```

### Before
```
[ThreadPoolExecutor-0_0] 2024-12-13T15:30:34+0100 get_sha256_hex took 9.462s
[Thread-1 (thread_inner)] 2024-12-13T15:31:01+0100 get_upload_hashes took 27.037s (get_md5_base64)
```

### After
```
[ThreadPoolExecutor-0_0] 2024-12-13T15:29:25+0100 get_upload_hashes took 9.033s (dict_keys(['sha256']))
[Thread-1 (thread_inner)] 2024-12-13T15:29:25+0100 get_upload_hashes took 0.000s (dict_keys([]))
```

---

---

</details>

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -422,7 +422,12 @@ def get_file_upload_spec_from_path(
     # Python appears to give files 0o666 bits on Windows (equal for user, group, and global),
     # so we mask those out to 0o755 for compatibility with POSIX-based permissions.
     mode = mode or os.stat(filename).st_mode & (0o7777 if platform.system() != "Windows" else 0o7755)
-    return _get_file_upload_spec(lambda: open(filename, "rb"), filename, mount_filename, mode)
+    return _get_file_upload_spec(
+        lambda: open(filename, "rb"),
+        filename,
+        mount_filename,
+        mode,
+    )
 
 
 def get_file_upload_spec_from_fileobj(fp: BinaryIO, mount_filename: PurePosixPath, mode: int) -> FileUploadSpec:
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -385,7 +385,6 @@ def _get_file_upload_spec(
     source_description: Any,
     mount_filename: PurePosixPath,
     mode: int,
-    md5_hex: Optional[str] = None,
 ) -> FileUploadSpec:
     with source() as fp:
         # Current position is ignored - we always upload from position 0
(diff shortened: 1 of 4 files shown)


modal/_utils/blob_utils.py
```diff
@@ -412,12 +421,13 @@ def get_file_upload_spec_from_path(
     # Python appears to give files 0o666 bits on Windows (equal for user, group, and global),
     # so we mask those out to 0o755 for compatibility with POSIX-based permissions.
     mode = mode or os.stat(filename).st_mode & (0o7777 if platform.system() != "Windows" else 0o7755)
-    return _get_file_upload_spec(
-        lambda: open(filename, "rb"),
-        filename,
-        mount_filename,
-        mode,
-    )
+    if os.path.getsize(filename) > MULTIPART_UPLOAD_THRESHOLD:
```
freider, line 424: Can't we do this size check in the same place that we check for LARGE_FILE_LIMIT, in `_get_file_upload_spec`?
`_get_file_upload_spec` supports BinaryIO objects as import as well, and it feels like this fix could be beneficial for those too?

freider (approved):
general concept sounds good, but added some comments about structure since we now get multiple levels of size checks. But I don't have this code fresh in my head so it's possible things are complicated in ways that prevent us from moving stuff around as I suggested 🤔

modal/_utils/blob_utils.py
danielnorberg: how can we check the length of a BinaryIO object?

modal/_utils/blob_utils.py
```diff
@@ -412,12 +421,13 @@ def get_file_upload_spec_from_path(
...
+        # TODO(dano): remove this once we stop requiring md5 for blobs
+        md5_hex = "baadbaadbaadbaadbaadbaadbaadbaad"  # placeholder - will not be used by server
```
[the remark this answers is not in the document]
danielnorberg: I wanted it to be possible to quickly find the source of this value if we attempt to use it somewhere and end up debugging the resulting failures.

modal/_utils/blob_utils.py
danielnorberg: Let me take a look.

danielnorberg: Done!

PR #2656, opened 2024-12-13
Don't import ipykernel if it's not already in sys.modules

erikbern:
I was looking at import times using `python -X importtime -c 'import modal' 2>&1 | sort -n -k 5` and I noticed we waste like 25ms (on my laptop, probably more in other environments) importing ipykernel. This seems kind of wasteful. Optimized the check a bit by
1. Using `sys.modules` to see if the `ipython.iostream` is even present (if not, then it can't be inside a notebook)
2. Do the import in local scope

Actually reduces total import times on my laptop fairly consistently from ~240ms to ~200ms which seems good! The import modal benchmark has gotten significantly slower in the last month or two so I hope this plus a few other quick hacks could make a meaningful difference.

image

### DIFF
--- a/modal/_ipython.py
+++ b/modal/_ipython.py
@@ -1,21 +1,11 @@
 # Copyright Modal Labs 2022
 import sys
-import warnings
-
-ipy_outstream = None
-try:
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore")
-        import ipykernel.iostream
-
-    ipy_outstream = ipykernel.iostream.OutStream
-except ImportError:
-    pass
 
 
 def is_notebook(stdout=None):
-    if ipy_outstream is None:
+    ipykernel_iostream = sys.modules.get("ipykernel.iostream")
+    if ipykernel_iostream is None:
         return False
     if stdout is None:
         stdout = sys.stdout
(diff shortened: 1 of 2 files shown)


erikbern:
not sure why tests are failing, will look at it later

mwaskom:
> not sure why tests are failing, will look at it later

I think it's because it has the effect of deferring an import of some notebook stuff to `test/notebook_test.py` where we aren't catching a DeprecationWarning.

Pretty sure the warning itself is irrelevant to us, and pretty annoying:

> ERROR test/notebook_test.py - DeprecationWarning: Jupyter is migrating its paths to use standard platformdirs
given by the platformdirs library.  To remove this warning and
see the appropriate new directories, set the environment variable
`JUPYTER_PLATFORM_DIRS=1` and then run `jupyter --paths`.

erikbern:
Dumb test:

image

erikbern:
Btw I'm not sure how meaningful of a delta this will be inside containers, because most likely they won't have ipykernel installed anyway. But it helps me optimize package imports further since `ipykernel` is present in my local environment.

mwaskom (approved):
Nice, a lot cleaner

PR #2657, opened 2024-12-13
Don't import aiohttp in global scope

erikbern:
This cuts down the time it takes to run `import modal` from about 205ms to about 135ms on my devbox, i.e. a 70ms improvement. Pretty good!

Actually it might even be more. With `python -X importtime -c 'import modal'` I see a speed up of about 70ms. When I run `time python -c 'import modal'`, I see a speedup of about 80-90ms.

Note that the impact might be 2x in many cases since it removes it both locally and in the container.

I don't _love_ this code though, it's a bit janky. The reason it's complex is that we were subclassing an `aiohttp` internal class in global scope so we had to break it out into a submodule that's optionally imported. Maybe there's a more elegant way to do it?

Note that we couldn't have done this without #2656 since `ipykernel` actually imports `aiohttp` too (among other things).

### DIFF
--- a/modal/_runtime/user_code_imports.py
+++ b/modal/_runtime/user_code_imports.py
@@ -55,42 +55,37 @@ def construct_webhook_callable(
     webhook_config: api_pb2.WebhookConfig,
     container_io_manager: "modal._runtime.container_io_manager.ContainerIOManager",
 ):
+    # Note: aiohttp is a significant dependency of the `asgi` module, so we import it locally
+    from modal._runtime import asgi
+
     # For webhooks, the user function is used to construct an asgi app:
     if webhook_config.type == api_pb2.WEBHOOK_TYPE_ASGI_APP:
         # Function returns an asgi_app, which we can use as a callable.
-        from modal._runtime.asgi import asgi_app_wrapper
-
-        return asgi_app_wrapper(user_defined_callable(), container_io_manager)
+        return asgi.asgi_app_wrapper(user_defined_callable(), container_io_manager)
 
     elif webhook_config.type == api_pb2.WEBHOOK_TYPE_WSGI_APP:
         # Function returns an wsgi_app, which we can use as a callable
-        from modal._runtime.asgi import wsgi_app_wrapper
-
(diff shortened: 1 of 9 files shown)


erikbern:
@mwaskom pointed out we're still importing `aiohttp` from containers – it can be seen by running `python -X importtime -c 'import modal._container_entrypoint'`

I'll fix

erikbern:
Ok, now when running `python -X importtime -c 'import modal._container_entrypoint' 2>&1 | tail -n 1`, it's consistently 200-210ms on `main`, vs consistently 127-130ms on this branch. So that's potentially a 70-80ms improvement for container cold start.

modal/_runtime/user_code_imports.py
```diff
@@ -66,20 +58,28 @@ def construct_webhook_callable(
     # For webhooks, the user function is used to construct an asgi app:
     if webhook_config.type == api_pb2.WEBHOOK_TYPE_ASGI_APP:
         # Function returns an asgi_app, which we can use as a callable.
+        from modal._runtime.asgi import asgi_app_wrapper
```
mwaskom, line 61: Nit, but it looks like we use this module in every branch (or raise an error) so could import it as `asgi` once at the top of the function?

erikbern: yeah good point, will fix

erikbern:
Added a test too. I verified that this test is failing on `main`.

test/supports/slow_dependencies_container.py
```diff
@@ -0,0 +1,10 @@
+# Copyright Modal Labs 2024
+import sys
+
+import modal  # noqa
```
mwaskom, line 4: Should this do `import modal._container_entrypoint`?

erikbern: 🤦

mwaskom (approved):
🐎

erikbern:
Thanks for spotting dumb mistakes in my code @mwaskom! Hoping this will be significant enough to show up in our benchmarks

PR #2658, opened 2024-12-13
Add `requires_proxy_auth` arg to webhooks

2022tgoel:
## Describe your changes

---

---

</details>

### DIFF
--- a/modal/partial_function.py
+++ b/modal/partial_function.py
@@ -334,6 +334,7 @@ def _asgi_app(
     *,
     label: Optional[str] = None,  # Label for created endpoint. Final subdomain will be <workspace>--<label>.modal.run.
     custom_domains: Optional[Iterable[str]] = None,  # Deploy this endpoint on a custom domain.
+    requires_proxy_auth: bool = False,  # Require Proxy-Authenticate HTTP Headers on requests to the endpoint
     wait_for_response: bool = True,  # DEPRECATED: this must always be True now
 ) -> Callable[[Callable[..., Any]], _PartialFunction]:
     """Decorator for registering an ASGI app with a Modal function.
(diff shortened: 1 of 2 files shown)


luiscape (changes requested):
We are missing a few places: `_wsgi_app`, `_web_server`, etc. Can you find instances of `api_pb2.WebhookConfig` and add this parameter everywhere?

modal/partial_function.py
```diff
@@ -332,6 +334,7 @@ def _asgi_app(
     *,
     label: Optional[str] = None,  # Label for created endpoint. Final subdomain will be <workspace>--<label>.modal.run.
     custom_domains: Optional[Iterable[str]] = None,  # Deploy this endpoint on a custom domain.
+    requires_proxy_auth: bool = False,  # Require Proxy-Authenticate HTTP Headers on requests to the endpoint
```
luiscape, line 337: `Proxy-Authenticate` is a response header. This requires the `Proxy-Authorization` request header.

luiscape (approved):
Please update comments as per review. Good to merge after. 

@mwaskom is this alright with you?

modal/partial_function.py
2022tgoel: 👍

modal/partial_function.py
```diff
@@ -332,6 +334,7 @@ def _asgi_app(
     *,
     label: Optional[str] = None,  # Label for created endpoint. Final subdomain will be <workspace>--<label>.modal.run.
     custom_domains: Optional[Iterable[str]] = None,  # Deploy this endpoint on a custom domain.
+    requires_proxy_auth: bool = False,  # Require Proxy-Authorization HTTP Headers on requests to the endpoint
```
mwaskom, line 337: Nit, but is "the endpoint" ambiguous for the asgi_app/wsgi_app/web_server decorators?

mwaskom (commented):
Couple questions:

- Would it make sense to cover this in the unit tests for these decorators?
- Do we need to do this in two steps (i.e., a proto PR, then do the client interface PR after we've enabled server support?)
- Can we add a changelog entry?

mwaskom:
Oops ... guess I'm a little later there.

PR #2659, opened 2024-12-13
Add `--write-result` flag to `modal run` to save function return values locally

mwaskom:
## Changelog

- The `modal run` CLI now has a `--write-result` option. When you pass a filename, Modal will write the return value of the entrypoint function to that location on your local filesystem. The return value of the function must be either `str` or `bytes` to use this option; otherwise, an error will be raised. It can be useful for exercising a remote function that returns text, image data, etc.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -133,15 +133,15 @@ def _get_clean_app_description(func_ref: str) -> str:
         return " ".join(sys.argv)
 
 
-def _save_local_result(output_path: str, res: Any):
+def _write_local_result(result_path: str, res: Any):
     if isinstance(res, str):
         mode = "wt"
     elif isinstance(res, bytes):
         mode = "wb"
     else:
         res_type = type(res).__name__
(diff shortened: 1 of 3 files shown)


mwaskom:
😠 this passes locally!

mwaskom:
rip that's because i forgot to git add the new supports file 🤦

erikbern:
Tbh not in love with `--output` as the flag here, since it's an overloaded term. Maybe `--write-result` or something? Idk.

mwaskom:
OK yeah — unfortunate since `--output` is the canonical name for this CLI option but on balance better to reduce the chance for confusion with the existing terminology.

mwaskom:
OK renamed to `--write-result`. Thought about adding a parallel `--print-result` but wasn't certain; might add that later.

mwaskom:
@prbot approve

erikbern (approved):
Great! Agree that `--write-result` is a bit less discoverable, but it feels like the reduced ambiguity makes it a better one

PR #2660, opened 2024-12-13
Add file watching client changes

azliu0:
## Describe your changes

WRK-484

### Blockers
- [x] https://github.com/modal-labs/modal/pull/18215 roll out to all workers

---

---

</details>

## Changelog

- Sandboxes now support fsnotify-like file watching:
```python
from modal.file_io import FileWatchEventType

app = modal.App.lookup("file-watch", create_if_missing=True)
events = sb.watch("/foo")
for event in events:
    if event.type == FileWatchEventType.Modify:
        print(event.paths)
```

### DIFF
--- a/modal/file_io.py
+++ b/modal/file_io.py
@@ -410,49 +410,6 @@ class _FileIO(Generic[T]):
         await self._wait(resp.exec_id)
 
     @classmethod
-    async def ls(cls, path: str, client: _Client, task_id: str) -> list[str]:
-        """List the contents of the provided directory."""
-        self = cls.__new__(cls)
-        self._client = client
-        self._task_id = task_id
-        resp = await self._make_request(
-            api_pb2.ContainerFilesystemExecRequest(
-                file_ls_request=api_pb2.ContainerFileLsRequest(path=path),
-                task_id=task_id,
-            )
-        )
-        output = await self._wait(resp.exec_id)
-        return await self._parse_list_output(output)
-
-    @classmethod
-    async def mkdir(cls, path: str, client: _Client, task_id: str, parents: bool = False) -> None:
-        """Create a new directory."""
-        self = cls.__new__(cls)
-        self._client = client
-        self._task_id = task_id
-        resp = await self._make_request(
-            api_pb2.ContainerFilesystemExecRequest(
-                file_mkdir_request=api_pb2.ContainerFileMkdirRequest(path=path, make_parents=parents),
-                task_id=self._task_id,
-            )
-        )
-        await self._wait(resp.exec_id)
-
-    @classmethod
-    async def rm(cls, path: str, client: _Client, task_id: str, recursive: bool = False) -> None:
-        """Remove a file or directory in the Sandbox."""
-        self = cls.__new__(cls)
-        self._client = client
-        self._task_id = task_id
-        resp = await self._make_request(
-            api_pb2.ContainerFilesystemExecRequest(
-                file_rm_request=api_pb2.ContainerFileRmRequest(path=path, recursive=recursive),
-                task_id=self._task_id,
-            )
-        )
-        await self._wait(resp.exec_id)
-
-    @classmethod
     async def watch(
         cls, path: str, client: _Client, task_id: str, timeout: Optional[int] = None, recursive: bool = False
     ) -> AsyncIterator[FileWatchEvent]:
--- a/modal/file_io.py
+++ b/modal/file_io.py
@@ -6,6 +6,8 @@ from typing import TYPE_CHECKING, AsyncIterator, Generic, Optional, Sequence, Ty
 if TYPE_CHECKING:
     import _typeshed
 
+from collections.abc import AsyncGenerator
+
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 
 from modal._utils.grpc_utils import retry_transient_errors
@@ -336,6 +338,71 @@ class _FileIO(Generic[T]):
             )
         )
         await self._wait(resp.exec_id)
+
+    @classmethod
+    async def ls(cls, path: str, client: _Client, task_id: str) -> list[str]:
+        """List the contents of the provided directory."""
+        self = cls.__new__(cls)
+        self._client = client
+        self._task_id = task_id
+        resp = await self._make_request(
+            api_pb2.ContainerFilesystemExecRequest(
+                file_ls_request=api_pb2.ContainerFileLsRequest(path=path),
+                task_id=task_id,
+            )
+        )
+        output = await self._wait(resp.exec_id)
+        return await self._parse_list_output(output)
+
+    @classmethod
+    async def mkdir(cls, path: str, client: _Client, task_id: str, parents: bool = False) -> None:
+        """Create a new directory."""
+        self = cls.__new__(cls)
+        self._client = client
+        self._task_id = task_id
+        resp = await self._make_request(
+            api_pb2.ContainerFilesystemExecRequest(
+                file_mkdir_request=api_pb2.ContainerFileMkdirRequest(path=path, make_parents=parents),
+                task_id=self._task_id,
+            )
+        )
+        await self._wait(resp.exec_id)
+
+    @classmethod
+    async def rm(cls, path: str, client: _Client, task_id: str, recursive: bool = False) -> None:
+        """Remove a file or directory in the Sandbox."""
+        self = cls.__new__(cls)
+        self._client = client
+        self._task_id = task_id
+        resp = await self._make_request(
+            api_pb2.ContainerFilesystemExecRequest(
+                file_rm_request=api_pb2.ContainerFileRmRequest(path=path, recursive=recursive),
+                task_id=self._task_id,
+            )
+        )
+        await self._wait(resp.exec_id)
+
+    @classmethod
+    async def watch(
+        cls, path: str, client: _Client, task_id: str, timeout: Optional[int] = None, recursive: bool = False
+    ) -> AsyncGenerator[str, None]:
+        self = cls.__new__(cls)
+        self._client = client
+        self._task_id = task_id
+        resp = await self._make_request(
+            api_pb2.ContainerFilesystemExecRequest(
+                file_watch_request=api_pb2.ContainerFileWatchRequest(
+                    path=path,
+                    recursive=recursive,
+                    timeout_secs=timeout,
+                ),
+                task_id=self._task_id,
+            )
+        )
+        print("resp", resp)
+        while True:
+            await asyncio.sleep(1.0)
+            yield "dummy event"
 
     async def _close(self) -> None:
         # Buffer is flushed by the runner on close
--- a/modal/file_io.py
+++ b/modal/file_io.py
@@ -9,6 +9,8 @@ if TYPE_CHECKING:
     import _typeshed
 
 
+import json
+
 from grpclib.exceptions import GRPCError, StreamTerminatedError
 
 from modal._utils.grpc_utils import retry_transient_errors
@@ -96,10 +98,11 @@ async def _replace_bytes(file: "_FileIO", data: bytes, start: Optional[int] = No
 
 
 class WatchEvent(Enum):
-    Access = "ACCESS"
-    Create = "CREATE"
-    Modify = "MODIFY"
-    Remove = "REMOVE"
+    Unknown = "Unknown"
+    Access = "Access"
+    Create = "Create"
+    Modify = "Modify"
+    Remove = "Remove"
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


azliu0:
@ellipsis-dev quiet

modal/file_io.py
```diff
@@ -166,6 +184,53 @@ async def _consume_output(self, exec_id: str) -> AsyncIterator[Optional[bytes]]:
             for message in batch.output:
                 yield message
 
+    async def _consume_watch_output(self, exec_id: str) -> None:
```
pawalt, line 187: We've got identical consuming code to this in a few places. Cant it be genericized into an iterator where the only thing that changes is the work you're doing in the body?

modal/file_io.py
```diff
@@ -385,6 +454,27 @@ async def rm(cls, path: str, client: _Client, task_id: str, recursive: bool = Fa
         )
         await self._wait(resp.exec_id)
 
+    @classmethod
+    async def watch(
```
pawalt, line 458: maybe we should give an option to only watch for certain even types?

modal/file_io.py
```diff
@@ -385,6 +454,27 @@ async def rm(cls, path: str, client: _Client, task_id: str, recursive: bool = Fa
...
+            api_pb2.ContainerFilesystemExecRequest(
+                file_watch_request=api_pb2.ContainerFileWatchRequest(
+                    path=path,
+                    recursive=recursive,
+                    timeout_secs=timeout,
+                ),
+                task_id=self._task_id,
+            )
+        )
+        asyncio.create_task(self._consume_watch_output(resp.exec_id))
```
pawalt, line 474: Can you hold onto this reference and cancel it when this exits?

modal/file_io.py
```diff
@@ -190,6 +255,10 @@ async def _wait(self, exec_id: str) -> bytes:
                 raise
         return output
 
+    async def _wait_watch(self, exec_id: str) -> None:
```
pawalt, line 258: Why is this needed?

modal/file_io.py
```diff
@@ -166,6 +184,53 @@ async def _consume_output(self, exec_id: str) -> AsyncIterator[Optional[bytes]]:
...
+                        continue
+                    elif isinstance(exc, ClientClosed):
+                        break
+                raise
+
+    async def _parse_watch_output(self, event: bytes) -> Optional[FileWatchEvent]:
+        try:
+            event_json = json.loads(event.decode())
+            return FileWatchEvent(type=FileWatchEventType(event_json["event_type"]), paths=event_json["paths"])
+        except Exception:
```
pawalt, line 214: Be specific about what kind of exceptions this could throw

modal/file_io.py
```diff
@@ -166,6 +184,53 @@ async def _consume_output(self, exec_id: str) -> AsyncIterator[Optional[bytes]]:
...
+
+    async def _stream_watch_output(self) -> AsyncIterator[FileWatchEvent]:
+        buffer = b""
+        while True:
+            if len(self._watch_output_buffer) > 0:
+                item = self._watch_output_buffer.pop(0)
+                if item is None:
+                    break
+                buffer += item
+                if buffer.endswith(b"\n\n"):
```
pawalt, line 226: When is this not the case?

modal/file_io.py
azliu0: accident thanks

modal/file_io.py
azliu0: when a single event is split across multiple items

pawalt: Can you add a comment to this effect

pawalt (approved):
Looking good

modal/io_streams_helper.py
```diff
@@ -0,0 +1,53 @@
+# Copyright Modal Labs 2024
```
aksh-at, line 1: Filename should be prefixed with `_` if it's meant to be private.

modal/io_streams_helper.py
```diff
@@ -0,0 +1,53 @@
...
+    item_handler: Callable[[T], None],
+    completion_check: Callable[[T], bool],
+    max_retries: int = 10,
+    retry_delay: float = 1.0,
+) -> None:
+    """mdmd:hidden
+    Helper function to consume a stream with retry logic for transient errors.
+
+    Args:
+        stream_generator: Function that returns an AsyncIterator to consume
```
aksh-at, line 25: nit: docstring doesn't match

modal/file_io.py
```diff
@@ -167,6 +185,44 @@ async def _consume_output(self, exec_id: str) -> AsyncIterator[Optional[bytes]]:
...
                 yield message
 
+    async def _consume_watch_output(self, exec_id: str) -> None:
+        def item_handler(item: Optional[bytes]):
+            self._watch_output_buffer.append(item)
+
+        def completion_check(item: Optional[bytes]):
+            return item is None
+
+        await consume_stream_with_retries(
```
aksh-at, line 195: Is there any reason this uses this, but `_wait` in the same file exists and has very similar logic?

modal/io_streams.py
```diff
@@ -176,34 +177,21 @@ async def _consume_container_process_stream(self):
...
-                        if exc.status in RETRYABLE_GRPC_STATUS_CODES:
-                            await asyncio.sleep(1.0)
-                            continue
-                    elif isinstance(exc, StreamTerminatedError):
-                        continue
-                    elif isinstance(exc, ClientClosed):
-                        # If the client was closed, the user has triggered a cleanup.
-                        break
-                raise exc
+        iterator = _container_process_logs_iterator(self._object_id, self._file_descriptor, self._client)
```
aksh-at, line 189: Mentioned this on Slack, but this will only make one RPC ever, which is not what we want. We want to make a new call in every iteration of the loop.

modal/io_streams_helper.py
```diff
@@ -0,0 +1,53 @@
...
+from grpclib.exceptions import GRPCError, StreamTerminatedError
+
+from modal.exception import ClientClosed
+
+from ._utils.grpc_utils import RETRYABLE_GRPC_STATUS_CODES
+
+T = TypeVar("T")
+
+
+async def consume_stream_with_retries(
```
aksh-at, line 14: I see why you're doing this, but I think it's not worth the additional complexity to break this out right now. If you fix the function, it will take 3 callables, which is quite complicated to follow. Maybe let's just restore `io_streams.py` to work how it did before?

modal/file_io.py
```diff
@@ -167,6 +185,44 @@ async def _consume_output(self, exec_id: str) -> AsyncIterator[Optional[bytes]]:
...
+
+    async def _parse_watch_output(self, event: bytes) -> Optional[FileWatchEvent]:
+        try:
+            event_json = json.loads(event.decode())
+            return FileWatchEvent(type=FileWatchEventType(event_json["event_type"]), paths=event_json["paths"])
+        except (json.JSONDecodeError, KeyError, ValueError):
+            # skip invalid events
+            return None
+
+    async def _stream_watch_output(self) -> AsyncIterator[FileWatchEvent]:
```
aksh-at, line 209: Does this need to be a separate function?

modal/file_io.py
```diff
@@ -391,6 +447,36 @@ async def rm(cls, path: str, client: _Client, task_id: str, recursive: bool = Fa
...
+        client: _Client,
+        task_id: str,
+        filter: Optional[list[FileWatchEventType]] = None,
+        recursive: bool = False,
+        timeout: Optional[int] = None,
+    ) -> AsyncIterator[FileWatchEvent]:
+        self = cls.__new__(cls)
+        self._client = client
+        self._task_id = task_id
+        resp = await self._make_request(
```
aksh-at, line 463: I think this is from a separate PR, but it's weird to have `_make_request` be a separate function here.

modal/file_io.py
```diff
@@ -391,6 +447,36 @@ async def rm(cls, path: str, client: _Client, task_id: str, recursive: bool = Fa
...
+            api_pb2.ContainerFilesystemExecRequest(
+                file_watch_request=api_pb2.ContainerFileWatchRequest(
+                    path=path,
+                    recursive=recursive,
+                    timeout_secs=timeout,
+                ),
+                task_id=self._task_id,
+            )
+        )
+        task = asyncio.create_task(self._consume_watch_output(resp.exec_id))
```
aksh-at, line 473: We should use `modal._utils.TaskContext` here.

modal/file_io.py
```diff
@@ -167,6 +185,44 @@ async def _consume_output(self, exec_id: str) -> AsyncIterator[Optional[bytes]]:
...
+        def completion_check(item: Optional[bytes]):
+            return item is None
+
+        await consume_stream_with_retries(
+            self._consume_output(exec_id),
+            item_handler,
+            completion_check,
+        )
+
+    async def _parse_watch_output(self, event: bytes) -> Optional[FileWatchEvent]:
```
aksh-at, line 201: Also think a separate function here is not necessary.

modal/file_io.py
```diff
@@ -391,6 +447,36 @@ async def rm(cls, path: str, client: _Client, task_id: str, recursive: bool = Fa
...
+    async def watch(
+        cls,
+        path: str,
+        client: _Client,
+        task_id: str,
+        filter: Optional[list[FileWatchEventType]] = None,
+        recursive: bool = False,
+        timeout: Optional[int] = None,
+    ) -> AsyncIterator[FileWatchEvent]:
+        self = cls.__new__(cls)
```
aksh-at, line 460: Why not use a constructor here?

modal/file_io.py
azliu0: the main difference is `_wait` expects to return relatively quickly, whereas file watching handles could in theory stay alive for the duration of the `Sandbox`, which requires handling the buffers differently

modal/io_streams_helper.py
azliu0: Hmm, I guess I felt a bit uneasy about duplicating the retry logic, especially for e.g. testing purposes, or in this case being able to isolate the user issue to single place. Do you think it's worth duplicating? I don't feel too strongly about this

modal/file_io.py
azliu0: I think the incentive here was to deduplicate `assert self._client is not None`, since pyright fails without this check

helloworld:
# Code Review: Add file watching client changes

## Overview

This PR implements file watching functionality for Modal sandboxes, enabling users to monitor and react to file system events in real-time. Key additions include:

1. New `FileWatchEventType` enum and `FileWatchEvent` dataclass to represent file system events
2. `watch()` method on the `Sandbox` class to initiate file watching with filtering options
3. Stream handling utilities extracted into a new `io_streams_helper.py` module for code reuse
4. Comprehensive test coverage for the new functionality

## Suggestions

### Improve error handling for file watch stream parsing

In `modal/file_io.py:201`, the method for parsing watch output catches exceptions but doesn't provide any logging:

```python
async def _parse_watch_output(self, event: bytes) -> Optional[FileWatchEvent]:
    try:
        event_json = json.loads(event.decode())
        return FileWatchEvent(type=FileWatchEventType(event_json["event_type"]), paths=event_json["paths"])
    except (json.JSONDecodeError, KeyError, ValueError):
        # skip invalid events
        return None
```

Consider adding some logging to help diagnose when events are skipped due to parsing errors, which would make troubleshooting easier:

```python
async def _parse_watch_output(self, event: bytes) -> Optional[FileWatchEvent]:
    try:
        event_json = json.loads(event.decode())
        return FileWatchEvent(type=FileWatchEventType(event_json["event_type"]), paths=event_json["paths"])
    except (json.JSONDecodeError, KeyError, ValueError) as e:
        logger.debug(f"Skipping invalid file watch event: {e} (event data: {event[:100]}...)")
        return None
```

### Potential busy loop in _stream_watch_output

In `modal/file_io.py:208-217`, the `_stream_watch_output` method has a sleep when the buffer is empty:

```python
else:
    await asyncio.sleep(0.1)
```

This creates a busy loop that executes every 100ms even when there's no activity. Consider using a more efficient event-based approach or at least increasing the sleep time to reduce CPU usage.

## Questions

1. In the `watch` method in `modal/sandbox.py:583-590`, the `recursive` parameter has a default value of `None`, but in `file_io.py` it defaults to `False`. Is this intentional? It seems inconsistent and could cause confusion.

<details>
<summary><h3>Minor Comments</h3></summary>
<br>
<ul>
<li>In `modal/file_io.py:208`, the busy loop implementation in `_stream_watch_output` could benefit from some explanatory comments about why this approach was chosen over alternatives.</li>
<li>The docstring for the new `consume_stream_with_retries` function in `io_streams_helper.py` has a `mdmd:hidden` tag - is this intentional or a leftover from documentation generation?</li>
<li>Consider adding a public example or documentation for filtering by multiple event types, as this is a useful feature that's implemented but not demonstrated in the PR description.</li>
</ul>
</details>

[View in Exponent](https://exponent.run/workflow/9abb54b2-6c53-4e6c-8f79-39182508065c)

PR #2661, opened 2024-12-14
await `client.hello()`

aksh-at:
Just noticed this

image

---

---

</details>


aksh-at:
@prbot approve

erikbern (commented):
Oh wow thanks! We should add a test for this also

PR #2662, opened 2024-12-14
Write all data in `write_to_fd`

aksh-at:
Originally from a user report that `top` in `modal shell` was behaving weirdly sometimes. Was coming from our `write_to_fd` function just giving up after a partial write to the buffer (originally from #1188, @thecodingwizard any idea if this is for a reason?). 

After the fix:

https://github.com/user-attachments/assets/d5721bb4-2108-4382-b9d0-300dc5f5a6e4

---

---

</details>

### DIFF
--- a/modal/_utils/shell_utils.py
+++ b/modal/_utils/shell_utils.py
@@ -19,14 +19,20 @@ def write_to_fd(fd: int, data: bytes):
     future = loop.create_future()
 
     def try_write():
+        nonlocal data
         try:
             nbytes = os.write(fd, data)
-            loop.remove_writer(fd)
-            future.set_result(nbytes)
+            data = data[nbytes:]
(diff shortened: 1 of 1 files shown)


modal/_utils/shell_utils.py
```diff
@@ -17,16 +17,23 @@
 def write_to_fd(fd: int, data: bytes):
```
aksh-at, line 17: We could probably simplify this to just register a writer with `add_writer` once, but that requires a bigger refactor

aksh-at:
@prbot approve

thecodingwizard (commented):
lgtm, thanks! I think when I originally implemented this I forgot to handle the case where `os.write` did not write the entire buffer, oops 😅

PR #2663, opened 2024-12-15
Test for Client.verify

erikbern:
Add test for fix in #2661 

This made me realized the token flow isn't super well tested in general. I'll take a look later.


erikbern:
@prbot approve

aksh-at (commented):
Thanks for adding a test!

PR #2664, opened 2024-12-15
Truncate dict items if too long

aksh-at:
Currently fails when your Dict has large values.

---

- [x\ Client forward compatibility: this change ensures client can accept data intended for later versions of itself

---

</details>

### DIFF
--- a/modal/cli/dict.py
+++ b/modal/cli/dict.py
@@ -89,6 +89,10 @@ async def get(name: str, key: str, *, env: Optional[str] = ENV_OPTION):
     console.print(val)
 
 
+def _truncate(val: str) -> str:
+    return val[:80] + "..." if len(val) > 80 else val
+
+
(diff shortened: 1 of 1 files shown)


modal/cli/dict.py
```diff
@@ -118,7 +122,7 @@ async def items(
                 display_item = key, val
             else:
                 cast = repr if use_repr else str
-                display_item = cast(key), cast(val)  # type: ignore  # mypy/issue/12056
+                display_item = _truncate(cast(key)), _truncate(cast(val))  # type: ignore  # mypy/issue/12056
```
mwaskom, line 125: Could probably combine the `cast` and `_truncate` into a single call but nbd

mwaskom (approved):
Makes sense!

modal/cli/dict.py
aksh-at: Good point, done!

PR #2665, opened 2024-12-15
Proto prep for deprecating AppGetObjects

erikbern:
I'm planning to get rid of the need to call `AppGetObjects` by having the worker pass the data to the container through `ContainerArguments`.

This removes a server round trip which should cut down 10-200ms cold start latency (200ms+ will be for remote regions). It also reduces complexity.

As a part of this change, I'm using a slightly different message structure. We currently have

```protobuf
message AppGetObjectsResponse {
  repeated AppGetObjectsItem items = 2;
}

message AppGetObjectsItem {
  string tag = 1;
  Object object = 6;
}
```

But only functions and classes have tags since a long time so we can make it a bit more direct. So we do this instead:

```protobuf
message AppLayout {
  repeated Object object = 1;
  map<string, string> function_ids = 2;  // tag -> function id
  map<string, string> class_ids = 3;  // tag -> class id
}
```

Note that the `object_id` is a part of the `Object` definition.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -360,6 +360,12 @@ message AppHeartbeatRequest {
   string app_id = 1;
 }
 
+message AppLayout {
+  repeated Object object = 1;
+  map<string, string> function_ids = 2;  // tag -> function id
+  map<string, string> class_ids = 3;  // tag -> class id
+}
+
 message AppListRequest {
   string environment_name = 1;
 }
@@ -710,6 +716,7 @@ message ContainerArguments {  // This is used to pass data from the worker to th
   string runtime = 11;
   string environment_name = 13;
   optional string checkpoint_id = 14;
+  AppLayout app_layout = 15;
 }
 
 message ContainerCheckpointRequest {


modal_proto/api.proto
```diff
@@ -360,6 +360,12 @@ message AppHeartbeatRequest {
   string app_id = 1;
 }
 
+message AppLayout {
+  repeated Object object = 1;
+  map<string, string> function_ids = 2;  // tag -> function id
```
mwaskom, line 365: I know it's just a comment but should we use different terminology for the keys since we're trying to deconflict "tag" with the deployment tag concept?

mwaskom (approved):
Makes sense; I think we have some TODOs in various places about migrating `indexed_object_ids` to `function_ids` / `class_ids`.

I'm mildly sad about having `class` in some places and `cls` in others, but may be nothing we can do there.

erikbern:
My thought is we should call it class in the proto since it's not a reserved keyword, and cls in Python since it is. But not a strong preference.

modal_proto/api.proto
erikbern: Yeah makes sense. Any thoughts on the best term?

mwaskom:
> My thought is we should call it class in the proto since it's not a reserved keyword, and cls in Python since it is. But not a strong preference.

Yeah I think it's fine (although that would also suggest `modal.Class` is preferred); I guess since the internal table is named `class` there's going to be a discontinuity somewhere and best to have it at the public/private boundary.

PR #2667, opened 2024-12-16
Adds warnings if automounts seem to be required

freider:
Adds a warning when packages from global scope imports haven't been added to an image using `.add_local_python_source` if using the not setting an explicit include_source mode, and not using MODAL_AUTOMOUNT=0.

This will be used in a longer transition period towards deprecation of "auto mounting", getting us to a future with the following standard:

* Auto-mounting of first-party packages is disabled by default, except for the *function's own package*
* Function's own package is still auto-mounted by default with the change that it will only include .py-files
* Automounting behavior can be specified on a function and/or app level using the `include_source=False` flag
* Auto-mounting can be configured to one of two values: True (new default, similar but not exactly the same as today's automount=0) and False (not even adding the function's own source package/module)

--------

---

</details>

## Changelog

* Warns users if the `modal.Image` of a Function/Cls doesn't include all the globally imported "local" modules (using `.add_local_python_source()`), and the user hasn't explicitly set an `include_source` value of True/False. This is in preparation for an upcoming deprecation of the current "auto mount" logic.


mwaskom:
FWIW I think it's fine to keep "automount" as the terminology. As a standalone verb I think it's a little more clear; "add" could mean so many things.

freider:
> FWIW I think it's fine to keep "automount" as the terminology. As a standalone verb I think it's a little more clear; "add" could mean so many things.

What about `include_source`? :)

modal/functions.py
```diff
@@ -435,7 +435,7 @@ def _bind_method(
         return fun
 
     @staticmethod
-    def from_args(
+    def from_local(
```
freider, line 438: Semi-unrelated, but I've been wanting to do this rename, and since I replaced `_from_args` usage in Image I changed this while I was at it...

modal/image.py
```diff
@@ -843,7 +853,9 @@ def add_local_python_source(
         ```
         """
         mount = _Mount._from_local_python_packages(*modules, ignore=ignore)
-        return self._add_mount_layer_or_copy(mount, copy=copy)
+        img = self._add_mount_layer_or_copy(mount, copy=copy)
+        img._added_python_source_set |= set(modules)
```
freider, line 857: This is the key part of this change - track all `add_local_python_source` modules on the Image objects so we can inspect it later when the automount logic executes to see which modules are missing. It's not perfect, but people can get rid of false positive warnings by setting `include_source` to any boolean value (with the expectation that they know what they are doing)

modal/_functions.py
```diff
@@ -501,18 +501,43 @@ def from_args(
...
+                    auto_mounts.pop(entrypoint_module, None)
+
+                warn_missing_modules = set(auto_mounts.keys()) - image._added_python_source_set
+
+                if warn_missing_modules:
+                    python_stringified_modules = ", ".join(f'"{mod}"' for mod in sorted(warn_missing_modules))
+                    deprecation_warning(
+                        (2024, 1, 31),
+                        (
+                            "Automatic adding of local python source will be deprecated in the future.\n"
```
freider, line 525: @mwaskom let me know if you have ideas for other ways of phrasing this, and if we should mention the `include_source` option or not (I'm leaning towards no since it's not typically needed)

modal/_functions.py
```diff
@@ -501,18 +501,43 @@ def from_args(
...
+                            f"Make sure you have explicitly added the source for the following modules to the Image "
+                            f"used by `{info.function_name}`:\n"
+                        )
+                        + ", ".join(sorted(warn_missing_modules))
```
mwaskom, line 529: Nit: can we make this a bulleted list? It looks visually strange for a single module imo:
```
╭─ Modal Deprecation Warning (2024-01-31) ───────────────────────────────────────────────────────────────╮
│ Automatic adding of local python source will be deprecated in the future.                              │
│ Make sure you have explicitly added the source for the following modules to the Image used by `f`:     │
│ helper                                                                                                 │
│                                                                                                        │
│ e.g.:                                                                                                  │
```

modal/_functions.py
```diff
@@ -501,18 +501,43 @@ def from_args(
...
+                        + "\n\n"
+                        + (
+                            "e.g.:\n"
+                            f"image = Image.debian_slim().add_local_python_source"
```
mwaskom, line 533: I'm a little worried about this tripping up an overly-literal user who has an Image based on something other than `.debian_slim`(). I don't know that we have a great way to show a more app-tuned suggestion though since we don't track the Image constructor that was used in an explicit way.

modal/_functions.py
mwaskom: Yeah I think that `include_source` is not relevant at this stage of the deprecation process; advising users on how to update their Image definitions

modal/_functions.py
```diff
@@ -501,18 +501,43 @@ def from_args(
...
+                            f"({python_stringified_modules})\n"
+                            f"@app.function(image=image)\n"
```
mwaskom, line 535: Likewise I'm a little worried about a literal-minded user thinking this change requires the Image to be passed directly to the function (and not the App constructor)

I think it's fine for the suggestion just to demonstrate the Image construction syntax and now how you'd use it.

modal/_functions.py
```diff
@@ -501,18 +501,43 @@ def from_args(
...
+                            f"..."
+                        ),
```
mwaskom, line 537: Can we add a line like "For more information, see https://modal.com/docs/guide/modal-1-0-migration"? When users start seeing all the new deprecation warnings I think it will be helpful to call their attention to the idea that this is "one last hurrah" to manage their annoyance :D.

We can also add a more verbose explanation of what's going on and the various options for ameliorating it there. I can work on that.

modal/_functions.py
mwaskom: Suggested rephrasing of the prose:

> Modal will stop implicitly adding local Python modules to the Image ("automounting") in a future update. The following modules need to be explicitly added for future compatibility:

modal/image.py
```diff
@@ -657,9 +660,18 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
...
 
         rep = f"Image({dockerfile_function})"
-        obj = _Image._from_loader(_load, rep, deps=_deps)
-        obj.force_build = force_build
-        return obj
+        img = _Image._from_loader(_load, rep, deps=_deps)
+        img.force_build = force_build
+        return img
+
+    def _extend(self, **kwargs) -> "_Image":
```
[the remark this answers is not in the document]
mwaskom: Is there some reason `Image._from_args` can't automatically transfer the added source set from the provided base Image?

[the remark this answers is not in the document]
freider: I can't seem to come up with one now, but I have a feeling there was something that made me go this route back in december 😬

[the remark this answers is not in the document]
freider: Reverted it and managed to get it working now. Also found a bug and added a test for that in the process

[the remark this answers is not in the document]
mwaskom: Hm, seems not ideal to have the logic required to extend an Image distributed over two methods instead of just one.

mwaskom (approved):
:shipit:

PR #2668, opened 2024-12-16
Support file_io async context manager

azliu0:
## Describe your changes

Implements async context manager for `modal.FileIO`:

```python
async with await sb.open(file) as f:
    f.write(b"foo")
```

---

---

</details>


modal/file_io.py
```diff
@@ -374,6 +374,12 @@ def __enter__(self) -> "_FileIO":
     async def __exit__(self, exc_type, exc_value, traceback) -> None:
```
pawalt, line 374: Will synchronicity not generate the `__enter__` and `__exit__`?

azliu0: Nice catch, yeah it will

pawalt (approved):
cool

PR #2672, opened 2024-12-17
CLI-294: Add support to ASGI apps that don't support the lifespan protocol

kramstrom:
Adds support for ASGI apps that don't support the lifespan protocol, such as django apps by detecting if there are any exceptions raised before lifespan startup and if so disabling lifespan on the app.


kramstrom:
@prbot approve

kramstrom:
@prbot automerge

freider:
lgtm!

PR #2673, opened 2024-12-17
[CLI-290] Fix relative mount dedupe

freider:
Ensure deduplication of mounts when using modal run on a relative parent directory (having `..` in path segments)

Mostly a superficial change as this didn't have real world effects previously, but this will help preventing false positives in warnings when we disable auto mounting.

---

---

</details>

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -131,6 +131,8 @@ class _MountDir(_MountEntry):
         return str(self.local_dir.expanduser().absolute())
 
     def get_files_to_upload(self):
+        # we can't use .resolve() eagerly here since that could end up "renaming" symlinked files
+        # see test_mount_directory_with_symlinked_file
         local_dir = self.local_dir.expanduser().absolute()
 
         if not local_dir.exists():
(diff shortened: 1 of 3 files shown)


modal/mount.py
```diff
@@ -279,8 +280,11 @@ def _new(entries: list[_MountEntry] = []) -> "_Mount":
         rep = f"Mount({entries})"
 
         async def mount_content_deduplication_key():
+            def normalized_selected_files():
```
freider, line 283: We're most likely getting rid of this way of deduplicating mounts soon. It won't be needed when we don't have auto mounts.

Instead I'll try to make something simpler that possibly auto-detects if the function's definition file itself is added to the image already and use the path of that instead of adding a separate entrypoint mount

freider:
@prbot approve

freider:
@prbot automerge

mwaskom (approved):

freider:
@mwaskom (couldn't reply to comment for some reason)
> We're not actually planning to get rid of auto mounts soon though right? I thought the target for 1.0 was to switch the default but preserve the functionality.
Yeah we're not getting rid of it completely for a good while

PR #2674, opened 2024-12-17
Enforce deprecation on App.spawn_sandbox and Sandbox.create without an App

mwaskom:
## Changelog

- Two previously-introduced deprecations are now enforced and raise an error:
    - The `App.spawn_sandbox` method has been removed in favor of `Sandbox.create`
    - `Sandbox.create` now requires an `App` object to be passed


erikbern (approved):
great!

PR #2676, opened 2024-12-17
Move deprecation utilities to private namespace

mwaskom:
Splitting this off from https://github.com/modal-labs/modal-client/pull/2633. We're not very clear about whether functions in nested public modules: we should clean that up.

In particular the exception classes in `modal.exception` _are_ public, but it makes sense not to pull them up into the the top-level namespace, since there's a bunch and referencing them is niche. So it seems particularly important to keep this namespace clean. That said, can't imagine any users are calling these utilities directly.

## Changelog

- The internal `deprecation_error` and `deprecation_warning` utilities have been moved to a private namespace

### DIFF
--- a/modal/_utils/deprecation.py
+++ b/modal/_utils/deprecation.py
@@ -1,0 +1,44 @@
+# Copyright Modal Labs 2024
+import sys
+import warnings
+from datetime import date
+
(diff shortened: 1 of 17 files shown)


mwaskom:
@prbot approve; big diff and want to avoid merge conflicts

PR #2677, opened 2024-12-17
Use worker-provided app layout

erikbern:
Instead of hitting the `AppGetLayout` endpoint, read the app layout provided through the container arguments.

As it turns out, this actually gets rid of a few ugly hacks with test fixtures that we had, which is nice.

I'm planning to bump the client version to 0.69 as a part of this.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -34,7 +34,7 @@ from grpclib import GRPCError, Status
 from grpclib.events import RecvRequest, listen
 
 import modal._serialization
-from modal import __version__, config
+from modal import Cls, Function, __version__, config
 from modal._runtime.container_io_manager import _ContainerIOManager
 from modal._serialization import serialize_data_format
 from modal._utils.async_utils import asyncify, synchronize_api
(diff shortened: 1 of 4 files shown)


test/conftest.py
```diff
@@ -443,31 +463,6 @@ async def AppGetLogs(self, stream):
...
-            for object_id in object_ids.values():
-                if object_id.startswith("fu-"):
-                    definition = self.app_functions[object_id]
-                    if isinstance(definition, api_pb2.FunctionData):
-                        for ranked_fn in definition.ranked_functions:
-                            unindexed_object_ids |= {obj.object_id for obj in ranked_fn.function.object_dependencies}
-                    else:
-                        unindexed_object_ids |= {obj.object_id for obj in definition.object_dependencies}
-            objects += [(None, object_id) for object_id in unindexed_object_ids]
-            # TODO(michael) This perpetuates a hack! The container_test tests rely on hardcoded unindexed_object_ids
```
erikbern, line 461: glad we were able to get rid of this

erikbern:
ugh ran into a frustrating thing – we also call `AppGetObjects` in `_init_local_app_existing`. so we might need to preserve that, or maybe create an external endpoint with the same format.

erikbern:
While implementing this, I ran into the fact that `AppGetObjects` is still needed by the client. I'm taking a slightly different route and I'm planning to rewrite it so we replace `AppGetObjects` with `AppGetLayout` first. That's a more incremental approach towards reducing the container side roundtrip.

PR #2680, opened 2024-12-18
Clarify purpose of FunctionInfo class

freider:
Will get rid of everything in this class that isn't strictly used to serialize/deserialize functions (actually services)

This is the first of many small patches.

Without refactoring this, the new mount code would get really messy

---

</details>


mwaskom (approved):
Makes sense — would be great to simplify this too!

PR #2687, opened 2024-12-18
Update tests using Mount

freider:
Test needed updating since it was using explicit Mounts anyway

### DIFF
--- a/test/function_test.py
+++ b/test/function_test.py
@@ -878,12 +878,15 @@ def test_calls_should_not_unwrap_modal_objects_gen(servicer, client):
     assert len(servicer.client_calls) == 1
 
 
-def test_mount_deps_have_ids(client, servicer, monkeypatch, test_dir):
-    # This test can possibly break if a function's deps diverge between
-    # local and remote environments
+def test_function_deps_have_ids(client, servicer, monkeypatch, test_dir, set_env_client):
     monkeypatch.syspath_prepend(test_dir / "supports")
     app = App()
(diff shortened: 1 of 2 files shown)


test/function_test.py
```diff
@@ -878,38 +878,54 @@ def test_calls_should_not_unwrap_modal_objects_gen(servicer, client):
...
-def test_no_state_reuse(client, servicer, supports_dir):
+def test_no_state_reuse(client, servicer, supports_dir, disable_auto_mount):
     # two separate instances of the same mount content - triggers deduplication logic
-    mount_instance_1 = Mount.from_local_file(supports_dir / "pyproject.toml")
-    mount_instance_2 = Mount.from_local_file(supports_dir / "pyproject.toml")
 
+    img = (
+        Image.debian_slim()
+        .add_local_file(supports_dir / "pyproject.toml", "/root/")
+        .add_local_file(supports_dir / "pyproject.toml", "/root/")
```
mwaskom, line 910: naive question but does it make sense to continue testing this in the `function_test` context or is the deduplication of the "soft" layers now something that should be handed by the Image code?

mwaskom (approved):

test/function_test.py
freider: Yeah that might make sense. But I don't think this test is intended to check deduplication logic itself, it was just some side-effect of deduplication that at some point caused errors in object_id resolution in general for app runs/function loads, which might be why it's being done here.

Will probably rework a lot of this soon when changing automounting, since that will mostly make deduplication unnecessary

PR #2688, opened 2024-12-18
Add helper function for indicating when parameters have been renamed

mwaskom:
Revives the `renamed_parameter` decorator from #2633.

This decorator can be used to simplify the process of renaming a function parameter

```python
@renamed_parameter((2024, 12, 18), "use_ml", "use_ai")
def do_data_science(input: np.ndarray, use_ai: bool) -> float:
    ...
```

At runtime, any invocations that use the old name of the parameter will continue to work, issuing a deprecation warning with a helpful migration error.

Note that this doesn't support graceful migrations for static analysis: type checking code that uses `old_name` as a keyword will fail. In some cases this can be supported by adding the `old_name` as an additional parameter, ideally with an annotated type hint indicating the deprecation.

Note that we exclude this wrapper from tracebacks that originate from within Modal.

### DIFF
--- a/modal/_traceback.py
+++ b/modal/_traceback.py
@@ -74,11 +74,15 @@ def append_modal_tb(exc: BaseException, tb_dict: TBDictType, line_cache: LineCac
 
 def reduce_traceback_to_user_code(tb: Optional[TracebackType], user_source: str) -> TracebackType:
     """Return a traceback that does not contain modal entrypoint or synchronicity frames."""
-    # Step forward all the way through the traceback and drop any synchronicity frames
+
+    # Step forward all the way through the traceback and drop any "Modal support" frames
+    def skip_frame(filename: str) -> bool:
+        return "/site-packages/synchronicity/" in filename or "modal/_utils/deprecation" in filename
+
     tb_root = tb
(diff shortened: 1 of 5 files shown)


modal/_traceback.py
```diff
@@ -74,11 +74,15 @@ def append_modal_tb(exc: BaseException, tb_dict: TBDictType, line_cache: LineCac
 
 def reduce_traceback_to_user_code(tb: Optional[TracebackType], user_source: str) -> TracebackType:
```
erikbern, line 75: side note but this code seems a bit hacky and i wonder if it would make sense to not serialize traceback objects but rather have utilities for translating back and forth to an internal protobuf-based format. we would need that to support other languages in the future anyway.

completely unrelated to this PR, just something i've thought about in the past

mwaskom: I don't think this function has much to do with the need to serialize tracebacks. We have it because execution running through a lot of modal framework (the cli, all the internal stuff with nested with closures getting passed around etc., and then also synchronicity) means that raw tracebacks have a _lot_ of frames that aren't relevant to a user who wants to debug their code.

PR #2689, opened 2024-12-18
Fix sandbox proxy & add to shell

aksh-at:
Was propagating `proxy` info from the function spec to `modal shell`, and then realized proxies never worked with sandboxes. Fixing that here.

(edit: they worked sometimes if the `Proxy` object was resolved separately via an `app.run()`)

---

---

</details>


aksh-at:
Seeing if I can add a test.

luiscape (approved):

PR #2690, opened 2024-12-18
Rename 'label' and 'tag' to 'name' in from_name/lookup/delete methods

mwaskom:
Closes CLI-284

The external interface uses some mixed terminology for the "name of things":

- The `.from_name`/`.lookup`/`.name` methods on most Modal objects uses "label"
- These methods on `Function` and `Cls` use "app_name" and "tag"

It's confusing to have so many synonyms, and in the case of "tag" it's conflicted with the deployment tag concept (this has led a few users astray).

This PR standardizes everything on "name", aligning with the `.from_name` terminology.

I expect that nearly all user code will invoke these parameters with positional argument syntax. I've used the new `renamed_parameter` to issue runtime deprecation warnings in case anyone was using keyword arguments but don't expect we need anything more than that.

## Changelog

- Standardized terminology in `.from_name`/`.lookup`/`.delete` methods to use `name` consistently where `label` and `tag` were used interchangeably before. Code that invokes these methods using `label=` as an explicit keyword argument will issue a deprecation warning and will break in a future release.

### DIFF
--- a/modal/_utils/deprecation.py
+++ b/modal/_utils/deprecation.py
@@ -56,6 +56,7 @@ def renamed_parameter(
     date: tuple[int, int, int],
     old_name: str,
     new_name: str,
+    show_source: bool = True,
 ) -> Callable[[Callable[P, R]], Callable[P, R]]:
     """Decorator for semi-gracefully changing a parameter name.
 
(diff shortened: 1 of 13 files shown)


mwaskom:
Note that we still use `tag` internally in a bunch of places; I'm leaving that alone for now.

erikbern (approved):
🚢

PR #2691, opened 2024-12-18
Remove deprecation errors for legacy lifecycle methods

erikbern:
These were deprecated on 2024-02-21 and turned into errors on 2024-06-03 (#1872). Our minimum client version 0.62 was released in late March. So this is already a deprecation for 100% of users, and it's an error for something like 99.99% of users. Seems fine to clean up.


mwaskom (approved):
See ya!

PR #2692, opened 2024-12-18
Remove deprecated .new methods for storage objects

erikbern:
These were deprecated on 2024-03-19 and turned into errors on 2024-05-29 (#1854)

Seems fine to remove. Happy to wait until 0.63 is the minimum version though.


mwaskom (approved):
Looks like we started issuing a warning before releasing v0.62.0 so I'm fine removing these.

PR #2697, opened 2024-12-19
Require syncronicity 0.9.8 to fix decorated async exception unwrapping

mwaskom:
Bug was fixed in synchronicity; this just updates the required version

## Changelog

- Fixed a bug introduced in v0.68.39 that changed the exception type raise when the target object for `.from_name`/`.lookup` methods was not found.


mwaskom:
@prbot approve

PR #2698, opened 2024-12-20
Make sure background task for asgi lifespans is cancelled before container shuts down

freider:
Will prevent at least one class of "Task was destroyed but is pending" errors on container shutdown

Side note: I want to refactor the modal lifecycle + asgi lifespan execution to be a part of the same context manager, preferably the same that we'll let users use with `.local()`

---

---

</details>


kramstrom (approved):
nice! lgtm

PR #2701, opened 2024-12-20
Auto context-mount without dockerignore support

freider:
Broke out the most useful functionality, fixed the tests and cleaned up some stuff from the ongoing larger PR by @kramstrom 

---

---

</details>

## Changelog

* `Image.from_dockerfile()` and `image.dockerfile_commands()` now auto-infer which files need to be uploaded based on COPY commands in the source if `context_mount` is omitted. The `ignore=` argument to these methods can be used to selectively omit files using a set of glob patterns.


freider:
@prbot approve

test/conftest.py
```diff
@@ -2133,3 +2133,10 @@ def decode_function_call_jwt(function_call_jwt: str) -> tuple[str, str]:
     parts = function_call_jwt.split(":")
     assert len(parts) == 2
     return parts[0], parts[1]
+
+
+@pytest.fixture
```
kramstrom, line 2138: nice! 💡

kramstrom (commented):
Thanks for helping bringing this across the finish line! ⭐

PR #2705, opened 2024-12-20
Replace `AppGetObjects` with `AppGetLayout`

erikbern:
This uses the new `AppGetLayout` endpoint instead of `AppGetObjects` endpoint.

Note that it's used in two places:
1. containers on cold start
2. runner.py when redeploying to an existing app

We will be able to provide the app layout through the container argument soon, obviating the need for 1.

### DIFF
--- a/modal/running_app.py
+++ b/modal/running_app.py
@@ -4,16 +4,42 @@ from typing import Optional
 
 from google.protobuf.message import Message
 
+from modal._utils.grpc_utils import get_proto_oneof
+from modal_proto import api_pb2
+
 from .client import _Client
 
 
 @dataclass
 class RunningApp:
     app_id: str
+    client: _Client
     environment_name: Optional[str] = None
     app_page_url: Optional[str] = None
     app_logs_url: Optional[str] = None
     tag_to_object_id: dict[str, str] = field(default_factory=dict)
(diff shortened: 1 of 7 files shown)


modal/running_app.py
```diff
@@ -4,16 +4,42 @@
...
     environment_name: Optional[str] = None
     app_page_url: Optional[str] = None
     app_logs_url: Optional[str] = None
     tag_to_object_id: dict[str, str] = field(default_factory=dict)
     object_handle_metadata: dict[str, Optional[Message]] = field(default_factory=dict)
     interactive: bool = False
-    client: Optional[_Client] = None
+
+
+def running_app_from_layout(
```
erikbern, line 25: this helper function is probably a slightly unnecessary layer of indirection. We should be able to simplify things later by making `RunningApp` look more like `api_pb2.AppLayout`

mwaskom (approved):
Agreed that it would be nice to use a consistent representation of the App everywhere!

erikbern:
Running integration tests against this before I merge, it's a somewhat risky change.

erikbern:
Ran a ton of integration tests and they all seem good, but I won't have time to babysit this one until probably Monday! Will merge it then.

erikbern:
Turns out I have a few hours to babysit this today. Will revert if it causes issues (like breaking main)

PR #2706, opened 2024-12-20
Add some utilities back to vscode image

mwaskom:
Wrapping up a loose end I forgot to include with #2704; apt-installing a few utilities that were included in the old image and might be useful in the context of development (especially `git`). I'm being more selective than the true `code-server` because I'd like to minimize the risk of issues with esoteric images, and not all of the utilities included in the `code-server` image seemed relevant (e.g. it also installed `vim` and `nano` which seems unnecessary if the whole point is to provide `vscode`).


mwaskom:
@prbot approve

mwaskom:
@prbot automerge

advay-modal (approved):

PR #2708, opened 2024-12-22
Remove `_App._container_app`

erikbern:
Afaict it's not used for anything


erikbern:
oops it's being used somewhat surprisingly by the sandbox code

PR #2709, opened 2024-12-22
feat: add all-extras flag to poetry install from file

shreyashankar:
## Describe your changes

I was running into an issue where I wanted to install dependencies via `poetry install --all-extras`, but the poetry install from file util doesn't support it. My workaround without using the util (below) is kind of clunky:

```python
image=modal.Image.debian_slim().add_local_file("pyproject.toml", "/pyproject.toml", copy=True)
    .pip_install("poetry")
    .run_commands(["poetry config virtualenvs.create false && poetry install --all-extras"])
```

It's not so bad for me to use my workaround; I just spent quite a bit of time debugging and someone may have a similar issue. Feel free to delete this PR if it's not good to have in the codebase; no worries.

---

---

</details>

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1113,6 +1113,8 @@ class _Image(_Object, type_prefix="im"):
         without: list[str] = [],
         # Only install dependency groups specifed in this list.
         only: list[str] = [],
+        # Install all extras (See https://python-poetry.org/docs/cli/#options-2)
+        all_extras: bool = False,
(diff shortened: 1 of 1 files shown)


mwaskom:
Should we also support `--extras` if we're going to support this one?

shreyashankar:
I would appreciate it but I am not sure if people commonly use poetry this way

PR #2711, opened 2024-12-23
`RunningApp` cleanup

erikbern:
I think we can get rid of this class. Starting with the easy stuff.

### DIFF
--- a/modal/_runtime/container_io_manager.py
+++ b/modal/_runtime/container_io_manager.py
@@ -460,7 +460,6 @@ class _ContainerIOManager:
         return running_app_from_layout(
             self.app_id,
             app_layout,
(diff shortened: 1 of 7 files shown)


modal/sandbox.py
```diff
@@ -276,9 +274,6 @@ async def create(
 
             app_id = app.app_id
             app_client = app._client
-        elif _App._container_app is not None:
-            app_id = _App._container_app.app_id
-            app_client = _App._container_app.client
```
erikbern, line 281: afaict these properties are never set when `app._client` etc isn't set

erikbern:
will figure out what's going on with sandboxes separately

PR #2712, opened 2024-12-23
Container app singleton

erikbern:
This changes `_App._container_app` to be an `_App` rather than a `RunningApp`.

It also makes sure this property is _always_ set, regardless of whether we were able to identify the correct app. If we don't identify it, we just create one from scratch.

I also added a `App.container_app()` class method which returns the current app inside a container (which is now always set, whereas previously we wouldn't necessary have an app).

The point of this is to get rid of `RunningApp`. Some properties can be moved to the `_App` class instead. Some properties can be deleted – I removed `client` and `environment_name` as an easy thing. Merging this PR will let me continue this cleanup process.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -1044,6 +1044,9 @@ class _App:
 
     @classmethod
     def container_app(cls) -> Optional["_App"]:
+        """Returns the `App` running inside a container.
+
+        This will return `None` outside of a Modal container."""
         return cls._container_app
 
     @classmethod
--- a/modal/app.py
+++ b/modal/app.py
@@ -294,12 +294,7 @@ class _App:
         app = _App(name)
         app._app_id = response.app_id
         app._client = client
-        app._running_app = RunningApp(
-            response.app_id,
-            client=client,
-            environment_name=environment_name,
-            interactive=False,
-        )
+        app._running_app = RunningApp(response.app_id, interactive=False)
         return app
 
     def set_description(self, description: str):
--- a/modal/app.py
+++ b/modal/app.py
@@ -168,7 +168,7 @@ class _App:
     """
 
     _all_apps: ClassVar[dict[Optional[str], list["_App"]]] = {}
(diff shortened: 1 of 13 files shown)


modal/app.py
```diff
@@ -1047,6 +1042,13 @@ async def _logs(self, client: Optional[_Client] = None) -> AsyncGenerator[str, N
                     if log.data:
                         yield log.data
 
+    @classmethod
+    def container_app(cls) -> Optional["_App"]:
```
mwaskom, line 1046: Is this something we want to be public API? Is it relevant to users in any way?

mwaskom (approved):
Only question is whether we want to be expanding public API on `modal.App` here. Personally I find the overloading of the "App" concept fairly confusing.

modal/app.py
erikbern: Yeah that was my main question. I think it's nice to have a way to get the current app inside containers but I can't think of a super critical use case. Happy to prefix it by underscore if you prefer?

mwaskom: I think there are a few different object types where we'd want to support "get the current instance of this object" and we might to use a consistent method name for doing that.

erikbern: In this case it's specific to containers, but yeah I agree with you. Lmk if you have strong feelings about this naming. I'll merge the PR now, but it's easy to fix as a follow up. Either way it's pretty easy to rename it later and just have a deprecation path.

mwaskom: Generally speaking I guess I would default to keeping things private as it's much less pain to change that later than it is the other way.

> In this case it's specific to containers

i think that' probably true for all the modal objects? e.g. i am thinking it could apply to:

- `App`
- `Function`
- `Container`
- `FunctionCall`

all of which you might want a handle to an object representing the "current" instance. But in all cases that would only be relevant when you're running in a Modal container. But extending the pattern to `Container.container_container` will be a little sad.

PR #2713, opened 2024-12-25
Replace `RunningApp.tag_to_object_id` with `function_ids` and `class_ids`

erikbern:
Slowly peeling apart all the old `indexed_objects` and `tag_to_object_id` nonsense

### DIFF
--- a/modal/runner.py
+++ b/modal/runner.py
@@ -135,8 +135,9 @@ async def _create_all_objects(
     )
     with resolver.display():
         # Get current objects, and reset all objects
-        tag_to_object_id = running_app.tag_to_object_id
-        running_app.tag_to_object_id = {}
+        tag_to_object_id = {**running_app.function_ids, **running_app.class_ids}
+        running_app.function_ids = {}
+        running_app.class_ids = {}
 
         # Assign all objects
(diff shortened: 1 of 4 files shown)


modal/runner.py
```diff
@@ -163,7 +164,12 @@ async def _preload(tag, obj):
         async def _load(tag, obj):
             existing_object_id = tag_to_object_id.get(tag)
             await resolver.load(obj, existing_object_id)
-            running_app.tag_to_object_id[tag] = obj.object_id
+            if _Function._is_id_type(obj.object_id):
```
erikbern, line 167: this is a bit janky but i think we can revisit it later

modal/runner.py
```diff
@@ -183,19 +189,16 @@ async def _publish_app(
     def filter_values(full_dict: dict[str, V], condition: Callable[[V], bool]) -> dict[str, V]:
```
erikbern, line 189: we can probably remove this helper since it's only used in one place now

erikbern:
@prbot approve

mwaskom (commented):
Yay! 🧹

PR #2715, opened 2024-12-30
Fix `COPY --from=` in Dockerfiles

aksh-at:
Noticed we accidentally broke support for `COPY --from=` with the automatic context mount changes.

---

---

</details>


aksh-at:
@prbot approve

freider:
lgtm! I made a mental note about not forgetting this when we first discussed implicit context mounts, but managed to gc that note

PR #2716, opened 2024-12-30
Upgrade vendored cloudpickle to 3.1

aksh-at:

---

---

</details>


mwaskom (approved):
Can we update the docstring, which encodes the cloudpickle version? (GitHub won't let me comment directly on the relevant line, I guess to be paternalistic about review scope 😁).

mwaskom:
Turns out we don't need to do this; avoiding it for now to err on the side of stability.

PR #2718, opened 2025-01-02
Make the container_app classmethod internal

erikbern:
See convo in https://github.com/modal-labs/modal-client/pull/2712#discussion_r1901070185


mwaskom (approved):
✅ can always make it public in the future when we align on a name!

PR #2720, opened 2025-01-03
Read container arguments from file if provided

erikbern:
The worker will provide a `MODAL_CONTAINER_ARGUMENTS_PATH` env var soon, which points to a path on disk containing all the `ContainerArguments`. This will replace the current method where the container gets this data from the command line arguments.

Because workers take a few days to roll out, we fall back to the previous method. This can be removed fairly soon.


pawalt (approved):
Looking good

erikbern:
Won't merge this until I've verified this in prod!

erikbern:
I ran this a lot of times and the container arguments are occasionally being provided by a file. Merging this.

I'll remove the fallback later. In theory we should be able to remove it in ~48h, but I'll probably keep it much longer (a month?) to be on the safe side. I'm really happy I kept a fallback for the app layout since we ran into various unexpected issues after that change that forced us to remove that field in many cases.

PR #2721, opened 2025-01-03
Remove fallbacks related to container arguments

erikbern:
This removes a bunch of code that should in theory be dead.

1. Assume we can always read the container arguments from the `MODAL_CONTAINER_ARGUMENTS_PATH` file
2. Assume that the app layout is always passed to the container as a part of that

We need fallback 1 for another couple of days just waiting for all workers to roll over.

Having fallback 2 has been incredibly useful over the last few weeks because we had to revert a bunch of server changes. 

So I'm a bit nervous about merging this. Planning to let this PR sit for a few weeks and then merge it assuming things are good.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -38,7 +38,7 @@ from modal.partial_function import (
     _find_callables_for_obj,
     _PartialFunctionFlags,
 )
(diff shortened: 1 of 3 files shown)


mwaskom:
Sounds good.

PR #2725, opened 2025-01-04
Pass `pty` correctly in `modal shell <task-id>`

aksh-at:
Since `exec` is decorated with `@click.command` and we're calling it directly here, `pty` was getting a weird default value (`typer.OptionInfo`).

---

---

</details>


mwaskom (approved):
Probably could refactor this a little better at some point.

PR #2728, opened 2025-01-06
Make Sandbox result None if SandboxWait does not return a status

pawalt:
## Describe your changes

Previously, we would always set self._result to resp.result on re-hydrating a Sandbox. This kept some non-None data around, making `poll` return 0 on successive invocations. This commit reuses the exit code logic from `poll` to fix this.

---

---

</details>


pawalt:
@prbot approve

pawalt:
@prbot automerge

azliu0:
lgtm

azliu0 (approved):

PR #2729, opened 2025-01-07
CLI-298: auto detect dockerignore for `Image.from_dockerfile()` and `Image.dockerfile_commands()` methods

kramstrom:
## Describe your changes

- Breakout PR from https://github.com/modal-labs/modal-client/pull/2582 which was split into two parts one (https://github.com/modal-labs/modal-client/pull/2701) already merged which adds auto inferring context mounts for image docker methods and this PR which adds automatic detection of dockerignore for `Image.from_dockerfile()` and `image.dockerfile_commands()`

We already support auto-inferring context mounts for `Image.from_dockerfile()` and `Image.dockerfile_commands()` methods and an `ignore` argument which accepts a sequence of patterns to ignore or a callable to determine which files to ignore.

This PR changes the default behavior when using auto context mounts to automatically attempt to find a suitable `dockerignore` file based on the same ruleset that `docker` uses for finding a `dockerignore` file to use.

---

---

</details>

## Changelog

* The default behavior`Image.from_dockerfile()` and `image.dockerfile_commands()` if no parameter is passed to `ignore` will be to automatically detect if there is a valid dockerignore file in the current working directory or next to the dockerfile following the same rules as `dockerignore` does using `docker` commands. Previously no patterns were ignored.


modal/image.py
```diff
@@ -818,6 +880,13 @@ def copy_local_dir(
             remote_path="/assets",
             ignore=~FilePatternMatcher("**/*.py"),
         )
+
+        # You can also read ignore patterns from a file.
+        image = modal.Image.debian_slim().copy_local_dir(
```
freider, line 885: Preferably use `add_local_dir`, which replaces copy_local_dir

modal/image.py
```diff
@@ -709,6 +763,13 @@ def add_local_dir(
...
```
freider, line 768: This should be `add_local_dir` I think

modal/file_pattern_matcher.py
```diff
@@ -97,6 +112,9 @@ def _matches(self, file_path: str) -> bool:
         library. The reason is that `Matches()` in the original library is
         deprecated due to buggy behavior.
         """
+        if self._delayed_init:
```
freider, line 115: This is quite minor, but after thinking about it agian, I have a slight preference for putting this in from_file instead, making it clear that it's the "constructor's" responsibility to call init. Also, I'd put the `uninitialized._delayed_init = None` inside of this condition rather than in the delayed constructor function itself.

modal/image.py
```diff
@@ -266,6 +278,47 @@ def ignore_with_include(source: Path) -> bool:
...
+            )
+
+            cmds = dockerfile_path.read_text("utf8").splitlines() if dockerfile_path else dockerfile_cmds
+            return _create_context_mount(cmds, ignore_fn=ignore_fn, context_dir=context_dir)
+
+        return auto_created_context_mount_fn
+
+    def auto_created_context_mount_fn() -> Optional[_Mount]:
+        # use COPY commands and ignore patterns to construct implicit context mount
+        cmds = dockerfile_path.read_text("utf8").splitlines() if dockerfile_path else dockerfile_cmds
```
freider, line 316: This is a bit more general than this PR and could be adressed in a followup - but can't `COPY` commands also be split over multiple lines using `\` ?

freider (approved):
nice! some minor inline comments

modal/image.py
kramstrom: Yeah we actually added multiline support for `COPY` commands specifically: https://github.com/modal-labs/modal-client/blob/main/modal/_utils/docker_utils.py#L25C1-L33C34 but for other commands (like `RUN`) we don't support multiline commands, we should add that!

kramstrom: Maybe we can refactor this cmd reading to combine multilines directly instead of relying on the underlying `extract_copy_command_patterns` function to do it for `COPY` commands

modal/file_pattern_matcher.py
kramstrom: discussed offline but we can't move it to `from_file` as that will read the file eagerly and our Image methods require it to be read lazily. Instead moved it to the `__call__` method

PR #2730, opened 2025-01-07
Accelerate 4 years into the future

ekzhang:
This updates all CI to use Ubuntu 24.04. Previously, it was using 20.04.

The main thing to watch out for when upgrading OS version is that the glibc version will change. If there's a binary built in CI, it will be dynamically linked to a newer version of glibc and won't be compatible with older operating systems.

I checked and don't think anything in client CI is building binaries that link to glibc versions, they're manylinux wheels, so this should work!

### DIFF
--- a/.github/actions/setup-cached-python/action.yml
+++ b/.github/actions/setup-cached-python/action.yml
@@ -4,7 +4,7 @@ inputs:
   version:
     description: Which Python version to install
     required: true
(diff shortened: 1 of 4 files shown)


ekzhang:
Ok there are some SSL issues, dropping this for now :/

mwaskom:
I knew we were never getting out of 2020

PR #2731, opened 2025-01-07
Add Image.from_id validation proto

azliu0:
## Describe your changes

WRK-629 and https://modal-com.slack.com/archives/C07JNJMJPQC/p1736275888178649?thread_ts=1736200259.187379&cid=C07JNJMJPQC

Without validation, tasks scheduled with an invalid id get scheduled which is undesirable.

---

---

</details>

## Changelog


modal_proto/api.proto
```diff
@@ -1730,6 +1730,14 @@ message ImageContextFile {
   bytes data = 2;
 }
 
+message ImageFromIdRequest {
+  string image_id = 1;
+}
+
+message ImageFromIdResponse {
```
pawalt, line 1737: This should return the image metadata too

pawalt (approved):
lgtm

PR #2732, opened 2025-01-07
Image.from_id

azliu0:
## Describe your changes

WRK-629 and https://modal-com.slack.com/archives/C07JNJMJPQC/p1736275888178649?thread_ts=1736200259.187379&cid=C07JNJMJPQC

Blocks on server changes: https://github.com/modal-labs/modal/pull/18627

---

---

</details>

## Changelog
- Adds `Image.from_id`, which returns an `Image` object from an existing image id.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -832,6 +832,24 @@ class _Image(_Object, type_prefix="im"):
             ),
         )
 
+    @staticmethod
+    async def from_id(image_id: str, client: Optional[_Client] = None) -> "_Image":
+        """Construct an Image from an id and look up the Image result.
+
+        The ID of an Image object can be accessed using `.object_id`.
+        """
+        if client is None:
+            client = await _Client.from_env()
+
+        async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[str]):
+            resp = await retry_transient_errors(client.stub.ImageFromId, api_pb2.ImageFromIdRequest(image_id=image_id))
+            self._hydrate(resp.image_id, resolver.client, resp.metadata)
+
+        rep = "Image()"
+        obj = _Image._from_loader(_load, rep)
+
+        return obj
+
     def pip_install(
         self,
         *packages: Union[str, list[str]],  # A list of Python packages, eg. ["numpy", "matplotlib>=3.5.0"]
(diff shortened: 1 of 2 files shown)


pawalt (approved):
LGTM pending server changes landing.

mwaskom:
btw i edited the changelog to strike "rich"; I'm not aware of that being a real concept in our ontology, i think we just invented it in the conversation about this feature 😁

mwaskom:
BTW what happens if you do

```python
image = modal.Image.from_id("id-123").pip_install("whatever")
```

and then try to pass that Image to a Function?

azliu0:
Should work as expected! I think this is equivalent to adding a new layer on top of a cached image. Sanity check:
```python
import modal

app = modal.App.lookup("test-from-id", create_if_missing=True)
im_1 = modal.Image.debian_slim().pip_install("numpy")
sb_1 = modal.Sandbox.create(app=app, image=im_1)
image_id = im_1.object_id
print(image_id)

im_2 = modal.Image.from_id(image_id)
sb_2 = modal.Sandbox.create(app=app, image=im_2.pip_install("pandas"))
p = sb_2.exec("python3", "-c", "import pandas; import numpy; print(pandas.__version__, numpy.__version__)")
p.wait()
print(p.stdout.read())
```

mwaskom:
Sounds good; and there's nothing `Sandbox` is doing with the `Image` that's fundamentally different from `Function`?

azliu0:
yeah the behavior should be the same

modal/image.py
```diff
@@ -832,6 +832,27 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+    @staticmethod
+    async def from_id(image_id: str, client: Optional[_Client] = None) -> "_Image":
+        """Construct an Image from an id and look up the Image result.
+
+        The ID of an Image object can be accessed using `.object_id`.
+        """
+        if client is None:
+            client = await _Client.from_env()
+
+        resp = await retry_transient_errors(client.stub.ImageFromId, api_pb2.ImageFromIdRequest(image_id=image_id))
```
mwaskom, line 844: Wait should the RPC call go inside the `_load` closure?

PR #2734, opened 2025-01-07
Client now uses SYNC for .map()

rculbertson:
## Describe your changes

Part of SVC-180. 

This PR has the client start using SYNC for .map calls, and gives it the ability to do client retries. We already did this to .remote() in this PR: https://github.com/modal-labs/modal-client/pull/2403. 

For now, client retries are disabled by default. They can be enabled per-workspace by adding the workspace id to this list in the function_fast.py: https://github.com/modal-labs/modal/blob/bce886b722fad0c6ce5f5aa007707679c93348ec/server/modal_server/objects/function_fast.py#L147

---

---

</details>


rohansingh:
There were some unit tests on the priority queue that could be restored:
https://github.com/modal-labs/modal-client/pull/2734/commits/2c1f62cb7e1bbce6670f2e06bfa5d01418ae79d1#diff-80ca24dfa2bbe913c31c13fd90c8c0a2cef7cb4ba98e66ca0c5eca38d94c7732

modal/parallel_map.py
```diff
@@ -432,3 +562,33 @@ def main():
...
+
+    async def get(self) -> Union[int, None]:
+        while True:
+            (timestamp_seconds, idx) = await self._queue.get()
+            if timestamp_seconds < int(time.time()):
+                return idx
+            if timestamp_seconds == self._MAX_PRIORITY:
+                return None
+            await self._queue.put((timestamp_seconds, idx))
+            await asyncio.sleep(1)
```
rohansingh, line 591: I think using an `asyncio.Condition` like the previous implementation is better than `asyncio.sleep(1)` since the latter becomes a busy wait.

https://github.com/modal-labs/modal-client/blob/afdc7f8e19758352dda69458cfd7c310fd5eef31/modal/_utils/async_utils.py#L806

rculbertson: ah good point. I'll bring back that original implementation.

rohansingh (approved):
:shipit:

rohansingh:
:shipit: 

image

PR #2736, opened 2025-01-08
Protos for VolumeRename RPC

mwaskom:
Adds protos to support `modal volume rename` and `Volume.rename`


modal_proto/api.proto
```diff
@@ -2680,6 +2680,11 @@ message VolumeRemoveFileRequest {
   bool recursive = 3;
 }
 
+message VolumeRenameRequest {
+  string volume_id = 1 [ (modal.options.audit_target_attr) = true ];
+  string name = 2;
```
mwaskom, line 2685: We don't have any other `Rename` RPCs that I can see but this aligns with the pattern we use for deletion; first you do a lookup RPC based on the provided name, then you do a second RPC to mutate the object. This adds a bit of latency, but this isn't a latency-sensitive action IMO, and it simplifies some other things (e.g. our existing server policies are based around the ID).

mwaskom:
@prbot approve

mwaskom:
@prbot approve

modal_proto/api.proto
erikbern: yeah seems reasonable!

erikbern (approved):

PR #2738, opened 2025-01-09
Add web_server decorator container test

kramstrom:
Split out the non-blocking `web_server` container test from https://github.com/modal-labs/modal-client/pull/2255 to a separate PR as it's still useful to add a test for it (we currently don't have any!) even if we don't go forward with running the user code in a background thread


kramstrom:
@prbot approve

freider:
nice

PR #2741, opened 2025-01-09
Add Python standalone mount deployment back to CD pipeline

mwaskom:
Whoops, [this PR](https://github.com/modal-labs/modal-client/pull/1487) from back in March 2024 refactored the "global mounts" and inadvertently (I think?) dropped the "Python standalone" mount from the CD workflow. That didn't actually cause any problems until we added support for Python 3.13, so it went unnoticed for a while. This should re-enable it. (Like all main-only-CD-PRs, this might require an iteration or two to address unforeseen issues).

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -212,6 +212,36 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
         run: twine upload dist/* --non-interactive
 
+  publish-python-standalone:
+    name: Publish Python standalone mounts
+    if: github.ref == 'refs/heads/main'
+    needs: [client-versioning, client-test, publish-client]
+    runs-on: ubuntu-20.04
+    timeout-minutes: 5
+    env:
+      MODAL_LOGLEVEL: DEBUG
+      MODAL_TOKEN_ID: ${{ secrets.MODAL_TOKEN_ID }}
+      MODAL_TOKEN_SECRET: ${{ secrets.MODAL_TOKEN_SECRET }}
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


mwaskom:
@prbot approve; won't break anything if it doesn't work

.github/workflows/ci-cd.yml
```diff
@@ -212,6 +212,36 @@ jobs:
...
+      MODAL_TOKEN_SECRET: ${{ secrets.MODAL_TOKEN_SECRET }}
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: v${{ needs.client-versioning.outputs.client-version}}
+
+      - uses: ./.github/actions/setup-cached-python
+        with:
+          version: "3.11"
```
kramstrom, line 233: I see we use `3.11` for the other cached python steps but is there a reason we don't use the matrix version? And interestingly the matrix does not include `3.11`?

mwaskom: This job isn't matrixed because we don't need to run it on different versions of Python. (While we do create a mount for each Python version, we can just do so in a single Python process).

Where are you seeing a matrix that doesn't include 3.11 btw?

kramstrom: Ah I think I got confused looking at this
https://github.com/modal-labs/modal-client/blob/main/.github/workflows/ci-cd.yml#L132-L146 and then this https://github.com/modal-labs/modal-client/blob/main/.github/workflows/ci-cd.yml#L79-L101

but not sure I understand why we have these specific configurations

mwaskom: Oh yeah maybe there's a more standard way to write that ... basically I just wanted to test on the minimum and maximum Python versions across all image builder versions (instead of the cartesian product) but then the maximum supportable python version started depending on the image builder version

PR #2744, opened 2025-01-09
Add `modal launch hf-download` utility

mwaskom:
Adds a new `modal launch` utility intended to make it simple to cache an asset from the Hugging Face hub onto a modal.Volume.

---

---

</details>

## Changelog

- Added a `modal launch hf-download` utility. This provides a simple CLI for caching an asset from the Hugging Face Hub onto a Modal Volume:

    ```
    modal launch hf-download stabilityai/stable-diffusion-xl-base-1.0 hf-hub-cache --secret=hugging-face
    ```

   To use the assets, mount the Volume to a location that corresponds to the Hugging Face cache directory, as specified by the `HF_HUB_CACHE` environment variable or passed as the `cache_dir` in relevant functions.

### DIFF
--- a/modal/cli/launch.py
+++ b/modal/cli/launch.py
@@ -6,7 +6,7 @@ import os
 from pathlib import Path
 from typing import Any, Optional
 
-from typer import Typer
+from typer import Argument, Option, Typer
 
 from ..app import App
 from ..exception import _CliUserExecutionError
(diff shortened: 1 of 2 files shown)


erikbern:
I'm not entirely sold on this as a utility vs just making sure our examples show how to download stuff. Feels like this as a tool hides the more general principle of how to run data prep in Modal. Also with HF, you can rely on lazy caching in most cases.

mwaskom:
I agree that we should make it clear that Modal can be used throughout the project lifecycle! But this step doesn't seem like a very interesting "preprocessing" step to demonstrate; it's something you'd need to put in every App and it's pure boilerplate.

Relying on implicit caching can also work, but there's a couple advantages here:

- We're enabling the rust-based `hf-transfer` download, which is a lot faster, but requires a few extra steps that aren't otherwise relevant for Apps that use the models
- Writing a Model class so that you can trigger a download as a side effect of spinning up a container isn't always trivial; you need to write some nullary endpoint (which might not be the natural way to write your inference interface), and the download will happen in a container that requests resources you need to actually serve the model, which might be overkill (e.g. it's wasteful to have idling GPUs while the download is happening)

mwaskom:
Closing; didn't seem like there was alignment here.

PR #2745, opened 2025-01-10
Make sure we don't have an event loop when the function is not async

erikbern:
I realized it would be easy to introduce regressions for this – we should have a test.


freider:
👍 

Just curious - did you run into something where this was a critical assertion? I suppose things like trying to run `asyncio.run()` from a sync input would fail if there was already a running loop in the current thread

freider (approved):

PR #2746, opened 2025-01-10
More tests for concurrency info

erikbern:
Following up on #2745. I realized we actually have integration tests for this in the proprietary repo! But those are slow and are much better as isolated tests like this. So I quickly ported them over to the client instead.


erikbern:
@prbot approve

freider (commented):
👍

PR #2747, opened 2025-01-10
Make FilePatternMatcher.with_repr private and update docs

kramstrom:
## Describe your changes

- The current docs unfortunately show `with_repr` as the first method which should be private and is not useful for users. - Moved the documentation from the `__call__` method to the class directly as it doesn't show up in the docs and isn't used directly.
- Added a `from_file` example

Screenshot 2025-01-10 at 16 08 08

---

---

</details>


freider (approved):
thanks for fixing!

PR #2751, opened 2025-01-10
Fix bug with add_python='3.9'

mwaskom:
Fixes a dumb bug that I introduced in https://github.com/modal-labs/modal-client/pull/2743 by doing an ordinal comparison of version number strings 🤦.

Also adds a test for the base behavior of the `from_registry(..., add_python)` functionality (it was untested) and the edge case.

---

</details>

## Changelog

- Fixes a bug introduced in v0.72.2 when specifying `add_python="3.9"` in `Image.from_registry`.


ekzhang (approved):
Ah I see, so 3.9 > 3.13

mwaskom:
Yes exactly. Wouldn't have this problem if we just asked an LLM instead of executing the Python code, of course.

Issue #2748, opened 2025-01-10
Support for setuptools.dynamic when using pip_install_from_pyproject 


peterroelants:
I have a local package configured by a `pyproject.toml` file that looks similar to:
```toml 

[project]
name = "<name>"
description = "<description>"
version = "0.0.1"
readme = "README.md"
license = {text = "Proprietary/Confidential"}
classifiers = [
    "Programming Language :: Python :: 3",
]
requires-python = ">=3.9"
dynamic = [
  "dependencies",
  "optional-dependencies",
]

[build-system]
requires = ["setuptools>=67", "wheel", "pip"]
build-backend = "setuptools.build_meta"

[tool.setuptools.dynamic]
dependencies = {file = ["env/requirements.txt"]}
optional-dependencies.dev = {file = ["env/dev_requirements.txt"]}
optional-dependencies.strict_versions = {file = ["env/local_requirements.txt"]}

[tool.setuptools.packages.find]
where = ["./src"]
```

Note that I'm not using a [`[project.dependencies]`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements) section. However my dependencies are defined as [`dynamic`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#static-vs-dynamic-metadata), and I'm using [`[tool.setuptools.dynamic]`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) to define where to get the dependencies dynamically (from a requirements.txt file)

However, when creating an image and installing this package using [`pip_install_from_pyproject` ](https://modal.com/docs/reference/modal.Image#pip_install_from_pyproject) modal complains that there is `No [project.dependencies] section in pyproject.toml file`:
```
❯ modal serve server.py
✓ Initialized. View run at ...
Stopping app - uncaught exception raised locally: ValueError('No [project.dependencies] section in pyproject.toml file. If your pyproject.toml instead declares [tool.poetry.dependencies], use `Image.poetry_install_from_file()`. See https://packaging.python.org/en/latest/guides/writing-pyproject-toml for further file format guidelines.').
╭─ Error ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ No [project.dependencies] section in pyproject.toml file. If your pyproject.toml instead declares [tool.poetry.dependencies], use `Image.poetry_install_from_file()`. See          │
[...]

PR #2753, opened 2025-01-12
fix[2748]: Use setuptools to parse pyproject.toml

peterroelants:
## Describe your changes

[<details> <summary>Backward/forward compatibility checks</summary>](https://github.com/modal-labs/modal-client/issues/2748)

---

---

</details>


peterroelants:
Not sure what your contributor guidelines are, and how to test this properly. Any help is appreciated.

mwaskom:
Thanks for raising this issue — we're gonna track the feature request internally but I think the scope here is pretty broad and needs some more thought. Basically to claim support here we'd need to parse the toml to discover the build tool the project expects to use; just using `setuptools` will only use for setuptools projects and may be more harmful for other ones.

PR #2755, opened 2025-01-13
Adds deprecation warning when trying to access method directly on class [CLI-302]

freider:
E.g.

`MyClass.my_method.remote()` should not work
but 
`MyClass().my_method.remote()` should

This functionality is unintended and expected to be removed in 1.0


freider:
@prbot approve

freider:
@prbot automerge

PR #2756, opened 2025-01-13
CLI-227: Add deprecation warnings for mount methods on images

kramstrom:
## Describe your changes

---

---

</details>

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -82,6 +82,11 @@ class _AutoDockerIgnoreSentinel:
 
 AUTO_DOCKERIGNORE = _AutoDockerIgnoreSentinel()
 
+COPY_DEPRECATION_MESSAGE_PATTERN = """modal.Image.copy_* methods will soon be deprecated.
+
+Use {replacement} instead, which is functionally and performance-wise equivalent.
+"""
+
 
 def _validate_python_version(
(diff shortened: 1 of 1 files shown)


kramstrom:
@prbot approve

freider (commented):
lgtm

mwaskom:
Would be nice to get these to link to a more verbose migration guide as soon as we can: https://modal-com.slack.com/archives/C056CGAANRM/p1736784312057149

mwaskom:
Also why are we using `pending=True` here?

kramstrom:
> Also why are we using `pending=True` here?

So that we can still run the tests for these usages, without `pending=True` the deprecation warnings will raise in the CI: https://github.com/modal-labs/modal-client/blob/main/modal/_utils/deprecation.py#L45

erikbern:
Let's remove `pending=True` very soon though – the intent is for them to be used very intermittently. Once we add `with pytest.warns(DeprecationWarning)` for all those tests, we can remove `pending=True`

mwaskom:
I see, so you're doing

1. introduce pending deprecations
2. assert that tests for old syntax issue deprecation warnings
3. un-pend the deprecation

that makes sense.

PR #2757, opened 2025-01-13
Fix formatting on `modal shell` docs

yirenlu92:
Remove `\b` characters that were messing up the formatting on the reference docs page for `modal shell`

### DIFF
--- a/modal/cli/launch.py
+++ b/modal/cli/launch.py
@@ -19,6 +19,8 @@ launch_cli = Typer(
     no_args_is_help=True,
     help="""
     Open a serverless app instance on Modal.
+
+    This command is in preview and may change in the future.
     """,
 )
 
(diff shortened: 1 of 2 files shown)


modal/cli/launch.py
```diff
@@ -18,9 +18,7 @@
     name="launch",
     no_args_is_help=True,
     help="""
-    [Preview] Open a serverless app instance on Modal.
-
-    This command is in preview and may change in the future.
+    Open a serverless app instance on Modal.
```
mwaskom, line 21: I think we still consider these interfaces fairly unstable fwiw (but can remove the `[Preview]` from the top-level description.

mwaskom (approved):
Oh annoying. Thanks for fixing. cc @aksh-at this fixes the terminal help but breaks the web docs :(

PR #2758, opened 2025-01-14
Accept string path to FilePatternMatcher.from_file

kramstrom:
## Describe your changes

Allow passing a string to `FilePatternMatcher.from_file` instead of forcing users to pass in a `Path` when reading an ignore file. Consistent with other Python methods such as `open` and our own `Image.from_dockerfile` etc

---

---

</details>

### DIFF
--- a/test/file_pattern_matcher_test.py
+++ b/test/file_pattern_matcher_test.py
@@ -337,18 +337,13 @@ def test_invert_patterns(tmp_path_with_content):
 
 
 @pytest.mark.usefixtures("tmp_cwd")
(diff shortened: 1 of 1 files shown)


mwaskom (approved):
Glad you found the Path of least resistance.

If we have tests for this, would be good to cover string inputs there too.

kramstrom:
> Glad you found the Path of least resistance.
> 
> If we have tests for this, would be good to cover string inputs there too.

💀

kramstrom:
@prbot automerge

PR #2760, opened 2025-01-14
Fix state for detached (disconnected) Apps in `modal app list`

mwaskom:
Previously these had an "Unknown" status in the list.

I thought about distinguishing between "detachable" and "detached", but that doesn't really work with `--detach` being the flag. I guess users don't need to think about the difference.


mwaskom:
@prbot approve

mwaskom:
@prbot thanks bud

kramstrom:
lgtm!

PR #2761, opened 2025-01-15
Fixes issue where `modal run` of a method on a class triggered deprecation warning [CLI-304]

freider:
The internal code used to intermediately make use of the class-bound function rather than instantiating the class first.
 
This required some rather messy refactoring of CLI code which should preferably be simplified further in the future.

---

---

</details>

## Changelog

* Fixes bug introduced in v0.72.9 where `modal run SomeClass.some_method` would incorrectly print a deprecation warning.

### DIFF
--- a/modal/cli/import_refs.py
+++ b/modal/cli/import_refs.py
@@ -93,6 +93,8 @@ def import_file_or_module(file_or_module: str):
 
 @dataclass
 class MethodReference:
+    """This helps with deferring method reference until after the class gets instantiated by the CLI"""
+
     cls: Cls
     method_name: str
 
@@ -260,7 +262,7 @@ You would run foo as [bold green]{base_cmd} app.py::foo[/bold green]"""
 
 def import_object(
     func_ref: str, base_cmd: str, accept_local_entrypoint=True, accept_webhook=False
-) -> Union[Function, LocalEntrypoint, MethodReference]:
+) -> tuple[App, Union[Function, LocalEntrypoint, MethodReference]]:
     """Takes a function ref string and returns something "runnable"
 
     The function ref can leave out partial information (apart from the file name) as
@@ -286,14 +288,19 @@ def import_object(
         # infer function or display help for how to select one
         app = app_function_or_method_ref
         function_handle = _infer_function_or_help(app, module, accept_local_entrypoint, accept_webhook)
-        return function_handle
-    elif isinstance(app_function_or_method_ref, (Function, MethodReference)):
-        return app_function_or_method_ref
+        return app, function_handle
+    elif isinstance(app_function_or_method_ref, Function):
+        return app_function_or_method_ref.app, app_function_or_method_ref
+    elif isinstance(app_function_or_method_ref, MethodReference):
+        return app_function_or_method_ref.cls._get_app(), app_function_or_method_ref
     elif isinstance(app_function_or_method_ref, LocalEntrypoint):
         if not accept_local_entrypoint:
             raise click.UsageError(
                 f"{func_ref} is not a Modal Function (a Modal local_entrypoint can't be used in this context)"
             )
-        return app_function_or_method_ref
+        return app_function_or_method_ref.app, app_function_or_method_ref
     else:
-        raise click.UsageError(f"{app_function_or_method_ref} is not a Modal entity (should be an App or Function)")
+        raise click.UsageError(
+            f"{app_function_or_method_ref} is not a Modal entity (should be an App, Local entrypoint, "
+            "Function or Class/Method)"
+        )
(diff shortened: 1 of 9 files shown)


modal/cli/import_refs.py
```diff
@@ -79,23 +91,37 @@ def import_file_or_module(file_or_module: str):
     return module
 
 
-def get_by_object_path(obj: Any, obj_path: str) -> Optional[Any]:
+@dataclass
+class MethodReference:
```
freider, line 95: This helps with deferring method reference until after the class gets instantiated by the CLI

modal/cli/run.py
```diff
@@ -145,39 +143,7 @@ def _write_local_result(result_path: str, res: Any):
...
-                raise click.UsageError(
-                    f"Please specify a specific method of {class_name} to run, e.g. `modal run foo.py::MyClass.bar`"  # noqa: E501
-                )
-        fun_signature = _get_signature(getattr(cls, method_name).info.raw_f, is_method=True)
-        signature = dict(**cls_signature, **fun_signature)  # Pool all arguments
-        # TODO(erikbern): assert there's no overlap?
-    else:
-        signature = _get_signature(function.info.raw_f)
-
+def _make_click_function(app, inner: Callable[[dict[str, Any]], Any]):
```
freider, line 146: I split this up into separate functions for function and class commands instead of branching inside a single function making things complicated

modal/cls.py
```diff
@@ -397,6 +397,22 @@ def _get_partial_functions(self) -> dict[str, _PartialFunction]:
             raise AttributeError("You can only get the partial functions of a local Cls instance")
         return _find_partial_methods_for_user_cls(self._user_cls, _PartialFunctionFlags.all())
 
+    def _get_app(self) -> "modal.app._App":
```
mwaskom, line 400: Why do we need these new private getters?

mwaskom (commented):
Code here is definitely pretty complicated; seems like we were missing some test coverage, not sure how hard it would be to add? (Input/output behavior feels relatively striaghtforward to define?)

test/cli_test.py
```diff
@@ -314,26 +314,30 @@ def test_run_parse_args_entrypoint(servicer, set_env_client, test_dir):
         assert "Unable to generate command line interface for app entrypoint." in str(res.exception)
 
 
-def test_run_parse_args_function(servicer, set_env_client, test_dir):
+def test_run_parse_args_function(servicer, set_env_client, test_dir, recwarn):
     app_file = test_dir / "supports" / "app_run_tests" / "cli_args.py"
-    res = _run(["run", app_file.as_posix()], expected_exit_code=2, expected_stderr=None)
-    assert "You need to specify a Modal function or local entrypoint to run" in res.stderr
+    # res = _run(["run", app_file.as_posix()], expected_exit_code=2, expected_stderr=None)
+    # assert "You need to specify a Modal function or local entrypoint to run" in res.stderr
```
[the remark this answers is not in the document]
freider: Ooops, unintentional leftover from debugging

modal/cli/import_refs.py
```diff
@@ -62,11 +71,14 @@ def import_file_or_module(file_or_module: str):
         sys.path.insert(0, str(full_path.parent))
 
         module_name = inspect.getmodulename(file_or_module)
+        assert module_name is not None
```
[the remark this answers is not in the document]
freider: Don't think so, or if they would they would also have caused errors before these assertions too (They were added as a result of stricter type checking)

modal/cls.py
freider: To allow access to the variable from the synchronicity wrapper without ugly `synchronizer._translate_in(...)` calls 😬 
Maybe synchronicity should simply getattr/setattr everything to the underlying instance (not sure why we aren't doing that), but that feels like a larger change

freider:
> Code here is definitely pretty complicated; seems like we were missing some test coverage, not sure how hard it would be to add? (Input/output behavior feels relatively striaghtforward to define?)

I feel like the splitup of class vs function CLI command-building logic here makes that part a bit clearer. But the "class path" is still a bit messy as it requires "deferred" instantiation of the class (the `MethodReference` dataclass here) in order to not have to getattr the method directly on the class.

What is still quite complicated is all of the "runnable" inference logic when a non-qualified reference is provided (e.g. just a file). We do have some isolated tests for that spread out across `cli_test` and `cli_import_test`, but they definitely lack some coverage as I did discover some bugs (added to linear).

freider:
Did some additional refactoring and added a bunch of tests for the app -> function inferrence logic

freider:
@prbot approve

freider:
@prbot approve

mwaskom (commented):
✅

PR #2762, opened 2025-01-15
Add environment_name to 'App.run'

mwaskom:
## Describe your changes

`App.run` is just a thin wrapper around `modal.runner.run_app`. The latter allowed you to specify the `environment_name` at runtime, but the former did not.

---

---

</details>

## Changelog

- Added an `environment_name` parameter to the `App.run` context manager.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -365,6 +365,7 @@ class _App:
         show_progress: Optional[bool] = None,
         detach: bool = False,
         interactive: bool = False,
+        environment_name: Optional[str] = None,
     ) -> AsyncGenerator["_App", None]:
         """Context manager that runs an app on Modal.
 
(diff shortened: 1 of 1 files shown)


modal/app.py
```diff
@@ -365,6 +365,7 @@ async def run(
         show_progress: Optional[bool] = None,
         detach: bool = False,
         interactive: bool = False,
+        environment_name: Optional[str] = None,
```
freider, line 368: Can we make it keyword only while we're at it?

freider (approved):

modal/app.py
mwaskom: I think we should make the whole signature keyword only (with a warning) so we can drop `show_progress` without any risk of issues (all very low risk anyway given what the arguments are, but still probably best that way anyway; also the order of parameters differs between `App.run` and `run_app` which is annoying).

freider: I agree, it's just a bit more painful to do it for existing arguments since it could break user code for people upgrading with positinal usage in their code

mwaskom: Right, that's what I mean by "with a warning" — there's a way to make it a gradual transition. (It involves a decorator 🤓)

PR #2764, opened 2025-01-15
[Notebooks] Types for modal-kernelshim to publish outputs

ekzhang:
This allows the server to get outputs from running cells in modal-kernelshim.

Reference: https://jupyter-client.readthedocs.io/en/stable/messaging.html

## Describe your changes

**Why is this public protobuf?** It's in here because kernelshim exists within the sandbox running user code. It can't access public information.

**Why is this re-encoded in this funny way?** It's part of the technical design that will allow us to implement collaborative editing and cloud-based kernels. (Also, on the server side, we'll need to figure out how to store potentially large outputs, possibly separate from other outputs.)

**How will this be authenticated?** Using an authentication passed into the sandbox. We could use task_id / task_secret, except sandboxes don't have that, so probably something else.

## Notebook progress

Map of current progress. This PR is the yellow part.

Progress

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1881,16 +1881,6 @@ message NetworkAccess {
   repeated string allowed_cidrs = 2;
 }
 
-message Object {
-  string object_id = 1;
-  oneof handle_metadata_oneof {
-    FunctionHandleMetadata function_handle_metadata = 3;
-    MountHandleMetadata mount_handle_metadata = 4;
-    ClassHandleMetadata class_handle_metadata = 5;
-    SandboxHandleMetadata sandbox_handle_metadata = 6;
-  }
-}
-
 // A single output from a notebook. When you execute a cell, it produces an
 // array of these outputs as the code runs.
 //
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1880,6 +1881,73 @@ message NetworkAccess {
...
+    ExecuteResult execute_result = 1;
+    DisplayData display_data = 2;
+    Stream stream = 3;
+    Error error = 4;
+  }
+}
+
+message NotebookKernelPublishResultsRequest {
+  // See kernelshim.py for the differences between this and `ExecuteResult`.
+  // https://jupyter-client.readthedocs.io/en/stable/messaging.html#execution-results
```
ekzhang, line 1925: This isn't committed yet, here's an explanation

```python
            # The reply is not returned until after the cell finishes running. Although the
            # `execute_result` output message contains the execution count and status, this is only
            # returned by ipykernel if the last line of the cell evaluates to a non-null value.
            #
            # So, we still need to send this message to the server to get the true execution count
            # for cells that do not produce an `execute_result` message.
            reply = await pending
```

ekzhang:
@prbot approve

Merging for now, still looking forward to feedback though! We discussed a bit in the office already. And after this we'll probably need to iterate further still.

ekzhang:
@prbot approve

Merging for now, still looking forward to feedback though! We discussed a bit in the office already. And after this we'll probably need to iterate further still.

mwaskom (commented):
Exciting!

PR #2765, opened 2025-01-15
Ensure that the sandbox has started before taking a filesystem snapshot

thecodingwizard:
Previously, if you tried to take a filesystem snapshot of a sandbox that hasn't been started yet, it would fail with an error:

```py
import modal

with modal.enable_output():
    sb = modal.Sandbox.create(app=app, gpu="H100")
    sb.terminate()

    img = sb.snapshot_filesystem()
    print(img)
```

This PR fixes this issue by blocking until the sandbox has started (or at least until it has been picked up by a worker) before executing the snapshot request.


pawalt (approved):
I thought you were planning on moving this logic into the RPC?

PR #2768, opened 2025-01-16
SandboxGetTaskId RPC Proto Changes

thecodingwizard:
Protobuf changes for https://github.com/modal-labs/modal/pull/18916

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2242,10 +2242,13 @@ message SandboxGetLogsRequest {
 
 message SandboxGetTaskIdRequest {
   string sandbox_id = 1;
+  float timeout = 2;
+  bool wait_until_ready = 3; // If true, waits until the container's postStart hook has been run before returning. Useful for detecting init failures.
 }
 
 message SandboxGetTaskIdResponse {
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -2242,10 +2242,13 @@ message SandboxGetLogsRequest {
 
 message SandboxGetTaskIdRequest {
   string sandbox_id = 1;
+  float timeout = 2;
+  bool wait_until_ready = 3; // If true, waits until the container's postStart hook has been run before returning. Useful for detecting init failures.
```
pawalt, line 2246: Can this just be represented with `timeout = 0`?

thecodingwizard: I don't think so. `wait_until_ready` does something different if it is true: it will wait until the task fully starts. If `wait_until_ready` is false, it will only wait until the task is scheduled, but it will not wait for the task to fully initialize.

Also, timeout = 0 represents "use the default timeout" for backwards-compatibility with old clients.

pawalt (approved):
ok bossman

PR #2769, opened 2025-01-16
Expose Object.hydrate as public API / deprecate Object.resolve

mwaskom:
## Describe your changes

We are deprecating the `Object.lookup` method, which eagerly hydrates the object upon construction, in favor of the following pattern:

- Most applications can use the lazy `Object.from_name` and rely on operations triggering on-demand hydration
- In rare cases where users need to force hydration without performing some other operation, they can call the new `Object.hydrate()` method.

The `.hydrate` method replaces the existing `.resolve` method. The new name was introduced because we have more commonly used "hydration" terminology (i.e. in error messages) and it is more accurate given the current Modal object model. The new method is marginally more user-friendly, e.g. it returns self so it can be called fluently.

Part of CLI-96

---

---

</details>

## Changelog

- Introduced a new public method, `.hydrate`, for on-demand hydration of Modal objects. This method replaces the existing semi-public `.resolve` method, which is now deprecated.

### DIFF
--- a/modal/_object.py
+++ b/modal/_object.py
@@ -252,6 +252,7 @@ class _Object:
 
     async def hydrate(self, client: Optional[_Client] = None) -> Self:
         """Synchronize the local object with its identity on the Modal server.
+
         It is rarely necessary to call this method explicitly, as most operations
         will lazily hydrate when needed. The main use case is when you need to
         access object metadata, such as its ID.
@@ -271,7 +272,6 @@ class _Object:
             # TODO(michael) can remove _hydrate lazily? I think all objects support it now?
             self._validate_is_hydrated()
         else:
-            # TODO: this client and/or resolver can't be changed by a caller to X.from_name()
             c = client if client is not None else await _Client.from_env()
             resolver = Resolver(c)
             await resolver.load(self)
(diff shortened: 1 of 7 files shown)


mwaskom:
Hm I guess we still need this to avoid remote calls to functions that are unhydrated because their App is not yet running. Will think about it.

mwaskom:
Whoops meant to abandon https://github.com/modal-labs/modal-client/pull/2770, not this one!

erikbern (commented):
👌

mwaskom:
@prbot approve

modal/sandbox.py
```diff
@@ -517,8 +517,9 @@ async def exec(
             raise InvalidError(f"workdir must be an absolute path, got: {workdir}")
 
         # Force secret resolution so we can pass the secret IDs to the backend.
+        # TODO should we parallelize this?
```
freider, line 520: Yes!

modal/_object.py
```diff
@@ -238,31 +238,50 @@ def default_deps(*args, **kwargs) -> Sequence["_Object"]:
...
                 logger.debug(f"rehydrating {self} after snapshot")
                 self._is_hydrated = False  # un-hydrate and re-resolve
                 c = client if client is not None else await _Client.from_env()
                 resolver = Resolver(c)
                 await resolver.load(typing.cast(_Object, self))
                 self._is_rehydrated = True
                 logger.debug(f"rehydrated {self} with client {id(c)}")
-            return
         elif not self._hydrate_lazily:
+            # TODO(michael) can remove _hydrate lazily? I think all objects support it now?
```
freider, line 272: Everything except **locally defined** Functions and Classes (as defined by the function and cls decorators)

freider (approved):
lgtm

PR #2770, opened 2025-01-16
PoC: get rid of hydrate_lazily concept

mwaskom:
In theory, all Modal objects now support lazy / just-in-time hydration. So we can simplify the codebase a bit by getting rid of the `hydrate_lazily` parameter in the Object initialization process, along with some validation logic.

It looks like were weren't passing `hydrate_lazily=True` _everywhere_ we could, but I can't follow the logic to understand whether that's because those routes truly don't support lazy hydration.

As a good empiricist, my first step is just to delete it and see if the tests pass.

### DIFF
--- a/modal/object.py
+++ b/modal/object.py
@@ -39,7 +39,6 @@ class _Object:
     _preload: Optional[Callable[[O, Resolver, Optional[str]], Awaitable[None]]]
     _rep: str
     _is_another_app: bool
-    _hydrate_lazily: bool
     _deps: Optional[Callable[..., list["_Object"]]]
     _deduplication_key: Optional[Callable[[], Awaitable[Hashable]]] = None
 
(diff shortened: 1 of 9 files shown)


modal/object.py
```diff
@@ -123,24 +120,10 @@ def _get_metadata(self) -> Optional[Message]:
         # the object_id is already provided by other means
         return
 
-    def _validate_is_hydrated(self: O):
```
erikbern, line 126: good riddance, this code is a jank parade

freider:
I think the one exception to everything being lazy is "local" Functions and Classes, since those require an App to be created to be hydrated. With `modal run` this is usually not an issue (unless you call functions on *another* local app), but if you have your own script's and call function.remote() outside of a `with app.run()` scope there needs to be some kind of error? (unless we want to auto-run the app in that case, i.e. lazily load it 😵 )

mwaskom:
Yeah the tests helpfully surfaced some additional complexity here. Need to think about it a little more but will follow up on Slack.

PR #2772, opened 2025-01-17
Don't print warning if adding the same Function object again

freider:
This can happen if you use app.include on an app that has a function which is already included by means of including 

---

---

</details>

## Changelog

* No longer prints a warning if app.include re-includes an already included function


freider:
@prbot approve

mwaskom (commented):
Nice catch

PR #2773, opened 2025-01-17
Update sandbox get task id protos

thecodingwizard:
We need to differentiate between old clients and new clients on the server. One way to do this is to make timeout optional and force new clients to always specify this field explicitly.


pawalt (approved):
thank you brother

PR #2774, opened 2025-01-17
Add protos for sandbox snapshots

thecodingwizard:
This PR adds the protobufs needed to support sandbox memory snapshots.

I kept `SandboxCreate` and `SandboxRestore` as separate RPCs just in case their implementations / parameters diverge in the future, though admittedly I don't know why they would.

Corresponding server PR: https://github.com/modal-labs/modal/pull/18923

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2221,6 +2221,30 @@ message Sandbox {
   NetworkAccess network_access = 22;
 
   optional string proxy_id = 23;
+
+  // Enable memory snapshots.
+  bool enable_memory_snapshot = 24;
+
+  // Used to pin gVisor version for memory-snapshottable sandboxes.
+  // This field is set by the server, not the client.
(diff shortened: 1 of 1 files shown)


azliu0 (approved):

azliu0:
Whoops I think we decided to use `enable_snapshot` instead of `enable_memory_snapshot` in the doc—it also seems like @mwaskom left some comments on the doc though so not totally sure what the end state of this will look like 

We can probably make another proto pr once client design is completely finalized but this should be useful enough for testing

mwaskom:
Would generally prefer to use consistent terminology throughout the codebase. In the past we've written a lot of the internals before deciding on the external terminology but then new devs who weren't around for that process find it harder to understand the code.

thecodingwizard:
Makes sense, I'll rename to `enable_snapshot`

PR #2778, opened 2025-01-17
[Notebooks] Add CLIENT_TYPE_NOTEBOOK_KERNEL

ekzhang:
Need a new client type to register the gRPC requests to publish notebook results. Technically we could reuse an existing client type, but I think it's good to err on the side of too much information / observability rather than too little.


luiscape (approved):
No strong opinions against a new client type.

PR #2786, opened 2025-01-21
Send cloud provider as string

erikbern:
Minor annoyance that we have this exposed as an enum in the client.

No strong feelings about casing here – upper case makes I think a tiny bit more sense for string enums, but I'm open minded.


mwaskom:
Assuming we're going to do something similar with GPU types, we currently normalize those to lowercase internally. But I don't think it would break anything to switch that? No strong opinion otherwise.

mwaskom (approved):

mwaskom:
BTW is the plan to eventually rename `cloud_provider_str` -> `cloud_provider`?

erikbern:
> BTW is the plan to eventually rename `cloud_provider_str` -> `cloud_provider`?

Yeah we can make that switch once the old field has been deprecated and removed (will take a long time).

PR #2787, opened 2025-01-21
Make CLI import logic consistent between listings and runnability

freider:
Changes the CLI import logic to use an exhaustive list of commands + filters instead of `getattr()` navigation.

This ensures that the listed modal "runnables" in the CLI error output will always be runnable by the command that you've run, which wasn't previously the case.

While this was a lot of code, I think it's mostly a simplification of the previous code.
The tests are still a bit of a mess since I mostly ported the old tests, and they now overlap quite a lot, but they are quick so it doesn't matter too much.

Note that this does remove a few of the helpful commands for special cases that we had before in favor of a more general error message that always lists the runnable commands if the user passes something that can't resolve to a single "runnable" (function, entrypoint, method)

Also fixes CLI-293

---

---

</details>

## Changelog

* Fixes a CLI bug where you couldn't reference functions via a qualified app, e.g. `mymodule::{app_variable}.{function_name}`.
* The `modal run`, `modal serve` and `modal shell` commands get more consistent error messages in cases where the passed app or function reference isn't resolvable to something that the current command expects.
* Removes the deprecated `__getattr__`, `__setattr__`, `__getitem__` and `__setitem__` methods from `modal.App`

### DIFF
--- a/modal/cli/import_refs.py
+++ b/modal/cli/import_refs.py
@@ -15,9 +15,10 @@ import sys
 import types
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Generator, Union, cast
+from typing import Union, cast
 
 import click
+from click import ClickException
 from rich.console import Console
 from rich.markdown import Markdown
 
@@ -171,7 +172,7 @@ def filter_cli_commands(
     name_prefix: str,
     accept_local_entrypoints: bool = True,
     accept_web_endpoints: bool = True,
-) -> Generator[CLICommand, None, None]:
+) -> list[CLICommand]:
     """Filters by name and type of runnable
 
     Returns generator of (matching names list, CLICommand)
@@ -184,24 +185,26 @@ def filter_cli_commands(
             return False
         return True
 
+    res = []
     for cli_command in cli_commands:
         if not _is_accepted_type(cli_command):
             continue
 
         if name_prefix in cli_command.names:
             # exact name match
-            yield ([name_prefix], cli_command)
+            res.append(cli_command)
             continue
 
         if not name_prefix:
             # no name specified, return all reachable runnables
-            yield (cli_command.names, cli_command)
+            res.append(cli_command)
             continue
 
         # partial matches e.g. app or class name - should we even allow this?
         prefix_matches = [x for x in cli_command.names if x.startswith(f"{name_prefix}.")]
         if prefix_matches:
-            yield (prefix_matches, cli_command)
+            res.append(cli_command)
+    return res
 
 
 def import_app(app_ref: str) -> App:
@@ -271,11 +274,6 @@ You would run foo as [bold green]{base_cmd} app.py::foo[/bold green]"""
     error_console.print(guidance_msg)
 
 
-def _import_cli_commands(file_or_module: str) -> list[CLICommand]:
-    module = import_file_or_module(file_or_module)
-    return list_cli_commands(module)
-
-
 def _get_runnable_app(runnable: Runnable) -> App:
     if isinstance(runnable, Function):
         return runnable.app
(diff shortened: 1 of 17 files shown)


mwaskom:
Is there a typo in the first changelog bullet? I am having trouble parsing it.

freider:
> Is there a typo in the first changelog bullet? I am having trouble parsing it.

Yup, fixed

modal/cli/import_refs.py
```diff
@@ -92,136 +93,149 @@ def import_file_or_module(file_or_module: str):
...
+    module: types.ModuleType,
+) -> list[CLICommand]:
+    """
+    Extracts all runnables found either directly in the input module, or in any of the Apps listed in that module
 
-def _infer_function_or_help(
-    app: App, module, accept_local_entrypoint: bool, accept_webhook: bool
-) -> Union[Function, LocalEntrypoint, MethodReference]:
-    """Using only an app - automatically infer a single "runnable" for a `modal run` invocation
+    Runnables includes all Functions, (class) Methods and Local Entrypoints, including web endpoints.
```
mwaskom, line 120: Why are web endpoints considered "runnable"?

modal/cli/import_refs.py
```diff
@@ -92,136 +93,149 @@ def import_file_or_module(file_or_module: str):
...
+    Where the first name is always the module level name if such a name exists
     """
-    filtered_local_entrypoints = [
-        entrypoint
-        for entrypoint in app.registered_entrypoints.values()
-        if entrypoint.info.module_name == module.__name__
-    ]
+    apps = cast(list[tuple[str, App]], inspect.getmembers(module, lambda x: isinstance(x, App)))
+
+    all_runnables: dict[Runnable, list[str]] = {}
```
mwaskom, line 129: Would using `defaultdict` make some of the updates marginally simpler?

modal/cli/import_refs.py
```diff
@@ -92,136 +93,149 @@ def import_file_or_module(file_or_module: str):
...
 
     if app is None:
-        _show_no_auto_detectable_app(import_ref)
+        error_console = Console(stderr=True)
+        error_console.print(f"[bold red]Could not find Modal app '{object_path}' in {import_path}.[/bold red]")
+
+        if object_path is None:
+            guidance_msg = Markdown(
+                f"Expected to find an app variable named **`{DEFAULT_APP_NAME}`** (the default app name). "
+                "If your `modal.App` is named differently, "
```
mwaskom, line 232: Something I'm a little confused by: is this guiding users to to the _App name_ (e.g. `App("my-app")`) or to the _variable name_ that points at the App?

mwaskom (approved):
Definitely a lot here ... had a few questions but overall lgtm; thanks for cleaning this up.

Does this PR also fix https://linear.app/modal-labs/issue/CLI-293/regression-modal-run-against-class-w-lifecycle-methods-doesnt-auto-run?

modal/cli/import_refs.py
freider: `modal shell` can "run" them - happy to consider other names (but I'm only using this internally at the moment)

mwaskom: Ah interesting, yeah I guess that makes sense.

modal/cli/import_refs.py
freider: 👍

modal/cli/import_refs.py
mwaskom: Maybe could rephrase (e.g. "If your `modal.App` is assigned to a different variable name") to reduce the chance of confusion 🤷‍♂️

freider:
> > Definitely a lot here ... had a few questions but overall lgtm; thanks for cleaning this up.
> > Does this PR also fix https://linear.app/modal-labs/issue/CLI-293/regression-modal-run-against-class-w-lifecycle-methods-doesnt-auto-run?
> 
> Oooh, good question. Not intentionally, but seems very relevant so I should make it so if I haven't! (It would be a bug if it listed/could run the lifecycle methods)

Seems like this patch fixed it, since it extracts the method names using the method metadata on the class. I made sure to add a lifecycle method now to one of the tests to keep this from regressing!

PR #2788, opened 2025-01-21
Add an experimental function for listing deployed Apps

mwaskom:
Currently we don't have a Python API for retrieving the information provided by the `modal app list` CLI. We previously had a private Python function that the CLI called into, but this got dropped in a refactor at some point. A few users were nevertheless using that function, and its removal is blocking them from upgrading past v0.62.

This PR adds back a minimal `list_deployed_apps` function to the `modal.experimental` namespace as a stopgap measure until we can prioritize developing a complete Python API for enumerating and managing Apps. It's very likely that the final API will look different from what's here; this function is intended to minimally support some existing workflows.

### DIFF
--- a/modal/experimental.py
+++ b/modal/experimental.py
@@ -1,16 +1,16 @@
 # Copyright Modal Labs 2022
-from typing import (
-    Any,
-    Callable,
-)
+from dataclasses import dataclass
+from typing import Any, Callable, Optional
 
-import modal._clustered_functions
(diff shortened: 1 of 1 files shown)


mwaskom:
@prbot approve to unblock user

kramstrom:
lgtm!

Removed in this PR it looks like: https://github.com/modal-labs/modal-client/pull/2073

PR #2789, opened 2025-01-21
Add a comment to commands that copy mounts

ekzhang:
## Describe your changes

This changes the build output format of `add_local_dir(copy=True)`, `add_local_file(copy=True)`, and `add_local_python_source(copy=True)` so that they print out a comment in the build output, to recognize which command is being run.

image

Before, it would just show `=> Step {n}: COPY . /`

image

The implementation is a little bit janky, using a newline and a comment to get the command to be printed out as part of the parsed dockerfile span. But I think it's still a little bit better? Not sure.

Also not sure if we need to gate this to a future image builder version.

## Changelog

- The build output format of `add_local_dir(copy=True)`, `add_local_file(copy=True)`, and `add_local_python_source(copy=True)` has been updated to include the names of the files and directories being copied in a comment.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -429,9 +429,15 @@ class _Image(_Object, type_prefix="im"):
             assert isinstance(metadata, api_pb2.ImageMetadata)
             self._metadata = metadata
 
-    def _add_mount_layer_or_copy(self, mount: _Mount, copy: bool = False):
+    def _add_mount_layer_or_copy(
+        self,
+        mount: _Mount,
+        *,
+        copy: bool = False,
+        description: str | None = None,
+    ) -> "_Image":
         if copy:
-            return self.copy_mount(mount, remote_path="/")
+            return self.copy_mount(mount, remote_path="/", description=description)
 
         base_image = self
 
@@ -661,7 +667,13 @@ class _Image(_Object, type_prefix="im"):
         obj.force_build = force_build
         return obj
 
-    def copy_mount(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Image":
+    def copy_mount(
+        self,
+        mount: _Mount,
+        remote_path: Union[str, Path] = ".",
+        *,
+        description: str | None = None,
+    ) -> "_Image":
         """
         **Deprecated**: Use image.add_local_dir(..., copy=True) or similar instead.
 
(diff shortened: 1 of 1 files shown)


ekzhang:
internal Slack reference: https://modal-com.slack.com/archives/C056CGAANRM/p1737155558589629

modal/image.py
```diff
@@ -429,9 +429,15 @@ def _hydrate_metadata(self, metadata: Optional[Message]):
             assert isinstance(metadata, api_pb2.ImageMetadata)
             self._metadata = metadata
 
-    def _add_mount_layer_or_copy(self, mount: _Mount, copy: bool = False):
+    def _add_mount_layer_or_copy(
+        self,
+        mount: _Mount,
+        *,
+        copy: bool = False,
+        description: str | None = None,
```
kramstrom, line 437: I think we still need to use `typing.Union` as we're still supporting python 3.9

kramstrom (approved):
I don't love having it as a comment for the `COPY` statement but this PR is definitely an improvement to the current state and we can always improve upon it in the future! I hadn't realized this regressed when we added the docker context auto mounting 😬

freider (changes requested):
If we want to go this route, it would probably be good to gate it behind a new image builder version as you say, otherwise the changed content of the commands would invalidate existing images on a client upgrade (for existing users of `.add_local*(..., copy=True)`, which may not be tat many though)

Another approach would be to add a new field to the Image proto for "step descriptions", which wouldn't be used for the image hash but could be used by the worker to represent the build step. This would be pretty useful for other things too (like if we want the worker output to more closely mirror the Modal syntax rather than docker syntax, or at least show it in parallel)

modal/image.py
```diff
@@ -683,7 +695,11 @@ def copy_mount(self, mount: _Mount, remote_path: Union[str, Path] = ".") -> "_Im
             raise InvalidError("The mount argument to copy has to be a Modal Mount object")
 
         def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
-            commands = ["FROM base", f"COPY . {remote_path}"]  # copy everything from the supplied mount
+            if description:
+                copy_command = f"COPY \\\n   # {description}\n   . {remote_path}"
```
freider, line 699: Minor, but could the comment be on a separate line before the copy command?

ekzhang: Unfortunately then the dockerfile parser skips over the comment and it wouldn't get parsed :(

ekzhang:
Thank you for the review @kramstrom @freider! I understand, maybe it makes more sense to proceed with the step descriptions as you said, since then we can avoid changing the hash.

mwaskom:
Hm also we should probably add some tests that exercise the new `Image.add_local_*` methods in the unit tests that assert stability of Image definitions within Image Builder Version! https://github.com/modal-labs/modal-client/blob/9b588622aa9dfc7f4f8cbd9345939d70e177f9fc/test/image_test.py#L1388

mwaskom:
Made a ticket to track; I think improving on the worker side makes a lot of sense!

PR #2792, opened 2025-01-22
Remove pending from Object.resolve deprecation warning

mwaskom:
Depends on https://github.com/modal-labs/modal/pull/19124

Part of CLI-96


freider (commented):
👍

PR #2793, opened 2025-01-22
Add experimental support for sandbox memory snapshotting

azliu0:
## Describe your changes

Adds experimental support for sandbox memory snapshots.

### Blockers 

- [x] protos: https://github.com/modal-labs/modal-client/pull/2794
- [x] server changes: https://github.com/modal-labs/modal/pull/18923
- [x] "from_id" server changes: https://github.com/modal-labs/modal/pull/19130
- [x] no TTL worker changes: https://github.com/modal-labs/modal/pull/19076

---

---

</details>


modal/sandbox.py
```diff
@@ -534,6 +542,53 @@ async def exec(
...
+    async def _experimental_from_snapshot(snapshot: _SandboxSnapshot, client: Optional[_Client] = None):
+        client = client or await _Client.from_env()
+
+        restore_req = api_pb2.SandboxRestoreRequest(snapshot_id=snapshot.object_id)
+        restore_resp: api_pb2.SandboxRestoreResponse = await retry_transient_errors(
+            client.stub.SandboxRestore, restore_req
+        )
+        sandbox = await _Sandbox.from_id(restore_resp.sandbox_id, client)
+
+        wait_req = api_pb2.SandboxWaitRequest(sandbox_id=restore_resp.sandbox_id, timeout=0)
```
thecodingwizard, line 583: Is timeout=0 here deliberate? I think this will return immediately?

Also, I think the current implementation of `SandboxWait` does not throw a retryable DEADLINE_EXCEEDED RPC. It just returns an empty SandboxWaitResponse.

Oops also I think `SandboxWait` only returns when the sandbox completely exits, not when the sandbox starts. Maybe we want `SandboxGetTaskId` here (which confusingly supports waiting for a sandbox to start)? But this RPC is called by the `Sandbox` object, so if we call it here, we will end up calling that RPC twice.

modal/sandbox.py
```diff
@@ -81,6 +83,7 @@ def _new(
         unencrypted_ports: Sequence[int] = [],
         proxy: Optional[_Proxy] = None,
         _experimental_scheduler_placement: Optional[SchedulerPlacement] = None,
+        enable_snapshot: bool = False,
```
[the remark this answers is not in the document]
azliu0: yeah

modal/sandbox.py
azliu0: yeah whoops these changes were not final—I probably won't undraft this pr until all of the above prs get in. this should be `SandboxGetTaskId`

cweld510 (approved):
This looks reasonable to me, but please get @mwaskom's review before merge, since this is not my area

modal/sandbox.py
```diff
@@ -534,6 +542,53 @@ async def exec(
         by_line = bufsize == 1
         return _ContainerProcess(resp.exec_id, self._client, stdout=stdout, stderr=stderr, text=text, by_line=by_line)
 
+    async def _experimental_snapshot(self) -> _SandboxSnapshot:
+        if not self._enable_snapshot:
```
cweld510, line 546: This needs to block for the task id, since the server RPC will fail if the task isn't up yet

mwaskom:
This LGTM. Slight preference for making the new module `snapshot.py`; we could add other snapshot classes there if we add them to avoid too much fragmentation.

Still fairly concerned about confusion between "filesystem snapshot" and "sandbox snapshot" but we can continue that conversation.

PR #2795, opened 2025-01-22
Add function_call_id to AppGetLogsRequest

annacarey:
Required to update `/logs-stream` RPC [here](https://github.com/modal-labs/modal/pull/19138)

@mwaskom we are adding the ability to filter logs by `functionCallId` to need this as a new filter on this RPC. The protobuf types live in the client, so wanted to run by you first. LMK if you have ?'s.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -340,6 +340,7 @@ message AppGetLogsRequest {
   string function_id = 5;
   string input_id = 6;
   string task_id = 7;
+  string function_call_id = 9;
   FileDescriptor file_descriptor = 8;
 }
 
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -2598,6 +2599,7 @@ message TaskLogsBatch {
   string image_id = 13;  // Used for image logs
   bool eof = 14;
   string pty_exec_id = 15;  // Used for interactive functions
+  string function_call_id = 16;
```
RohanArepally, line 2602: I don't think we need this here since we already have a have the `function_call_id` on TaskLogs? Might be mistaken

annacarey: We changed [this line](https://github.com/modal-labs/modal/pull/19138/files#diff-491819fb1dbd5a144592d4a13dc948d23c0e126a1f2cfbc0d62899553456ddd8R1548-R1550) to filter on the item instead of batch, so removed `function_call_id` from `TaskLogsBatch`.

annacarey:
@prbot approve

mwaskom:
Nice, I think users have asked about filtering `modal app logs` by function; this should enable that.

PR #2798, opened 2025-01-23
Add App.from_name for construction of a lazily-hydratable App

mwaskom:
## Describe your changes

This PR adds a new `App.from_name` method with similar lazy construction semantics to other `.from_name` methods on Object subclasses, along with an `App.hydrate` method for on-demand hydration. The intention is to make it possible to deprecate `App.lookup` when we deprecate other `Object.lookup` methods, streamlining object construction onto a single lazy path.

As `App` is _not_ a subclass of `Object` we face a decision about whether we should make it one (and leverage the existing hydration plumbing) or add some App-specific plumbing. This PR takes the latter approach, which feels more straightforward now without foreclosing on adapting the other approach later.

There's definitely some confusing things here. I've made it so that you can't call `App.hydrate` if you instantiated an App by calling its main constructor (`App(name)`). Potentially we will want to leverage lazy-hydration of the App layout when we do App redeployments (i.e. the current `_init_local_app_existing`). I think we can punt on that as there are not currently any user-facing workflows where you would want to hydrate the App and then access its layout, but we've tlaked about it, e.g. for Function discovery after looking up a deployed App.

Part of CLI-96

---

---

</details>

## Changelog

- Added a new `App.from_name` methods for lazily constructing an App based on a lookup and an `App.hydrate` method for hydrating it with server metadata on demand. These methods should be used instead of the eager `App.lookup` method, which will be deprecated soon (along with other `.lookup` methods). Note that `App.from_name` is primarily useful when you need an App to associate with a `modal.Sandbox`; other use-cases should continue to create Apps via the main constructor (`app = App(name)`).

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -263,10 +263,26 @@ class _App:
         return self._description
 
     @staticmethod
-    def from_name(name: str, environment_name: Optional[str] = None, create_if_missing: bool = False) -> "_App":
-        """TK"""
+    def from_name(name: str, environment_name: str = "", create_if_missing: bool = False) -> "_App":
+        """Look up an App with a given name, creating a new App if necessary.
 
-        # TODO what are the rules on App names? Can we use check_object_name?
+        Note that Apps created through this method will be in a deployed state,
+        but they will not have any associated Functions or Classes.
+
+        This method is mainly useful for creating an App to associate with a Sandbox:
+
+        ```python
+        app = modal.App.from_name("my-app", create_if_missing=True)
+        app.hydrate()  # Pre-hydrate to accelerate Sandbox creation
+        modal.Sandbox.create("echo", "hi", app=app)
+        ```
+
+        When building an App that will be populated with Functions and Classes, you should
+        call the main `App()` constructor directly.
+        """
+
+        # TODO We need to enforce rules about App names, but we previously weren't in App.lookup
+        # So we probably need to add that gradually (i.e. as a warning first) to avoid breaking people?
         async def _load(self: _App, client: _Client):
             request = api_pb2.AppGetOrCreateRequest(
                 app_name=name,
--- a/modal/app.py
+++ b/modal/app.py
@@ -2,7 +2,7 @@
 import inspect
 import typing
 import warnings
-from collections.abc import AsyncGenerator, Coroutine, Sequence
+from collections.abc import AsyncGenerator, Awaitable, Coroutine, Sequence
 from pathlib import PurePosixPath
 from textwrap import dedent
 from typing import (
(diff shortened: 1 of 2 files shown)


modal/app.py
```diff
@@ -258,6 +262,44 @@ def description(self) -> Optional[str]:
...
+                object_creation_type=(api_pb2.OBJECT_CREATION_TYPE_CREATE_IF_MISSING if create_if_missing else None),
+            )
+            response = await retry_transient_errors(client.stub.AppGetOrCreate, request)
+            self._app_id = response.app_id
+            self._client = client
+            # TODO should we also be calling AppGetLayout and populating the running_app with it?
+            self._running_app = RunningApp(response.app_id, interactive=False)
+
+        app = _App(name)
+        app._load = _load
```
erikbern, line 299: super minor but this could maybe just be a constructor argument

mwaskom: I’m fairly averse to “private parameters” in public interfaces: best case, you confuse users or at least present them with information they don’t need to think about, worse case they try to use it and end up in trouble somehow.

mwaskom: I’d really prefer to just merge `__init__` and .from_name` but couldn’t see a simple way to do that.

freider:
What would be the behavior with this if a user does the seemingly allowed (from a typing perspective):

```
app = modal.App.from_name("blah", create_if_missing=True)

@app.function()
def foo():
    pass

app.remote()
```
?

Maybe `from_name` shouldn't actually return an App, but a more limited smaller subset of its protocol? (Or we should make sure that @app.function and @app.cls raise good exceptions for lazy-created apps)

mwaskom:
> What would be the behavior with this if a user does the seemingly allowed

This is currently (i.e. with `App.lookup`) poorly defined and a little surprising: https://linear.app/modal-labs/issue/CLI-79/define-and-enforce-rules-about-what-you-can-do-with-looked-up-apps

> Am I correct that the only reason we even have the lookup/from_name methods on App is for the sandbox usecase, since those currently require apps?

Yes I think that's currently correct. In practice, Sandbox.create just needs an app_id. If we expand the App API to support more programmatic discovery of a deployed App layout (which seems like a good idea) then you could also imagine a lookup operation being useful for that in the future.

> Maybe from_name shouldn't actually return an App, but a more limited smaller subset of its protocol

That's an interesting idea. Feels like it potentially introduces some complications? E.g. currently with `Sandbox.create` you can `.lookup` a (deployed) App and pass it to `app=`, but you can also pass an ephemeral App that's in a running state. So we'd have something like `app: App | AppView`. We'd achieve a distinction from a strict typing perspective, but there still might be a lot of conceptual confusion.

Personally I think the situation with `App()` vs. `App.lookup()` is fairly unfortunate, but not sure what the best thing to do would be. I think the big-picture question here are orthogonal to the specific need to introduce a lazy complement to `App.lookup` though.

erikbern:
Open to revisiting app-less sandboxes if this is a big can of worms!

PR #2799, opened 2025-01-23
Strike mention of vestigial class from error message

mwaskom:
I didn't even know `AioImage` was a thing! 😁

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -211,12 +211,12 @@ class _App:
         self._name = name
         self._description = name
 
-        check_sequence(mounts, _Mount, "`mounts=` has to be a list or tuple of Mount objects")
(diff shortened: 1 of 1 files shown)


mwaskom:
@prbot approve

freider (commented):
Hehe yeah, synchronicity used to generate distinct types for sync vs async interfaces, so when using async you'd always have to use AioImage etc.

It had its advantages (not being able to accidentally use blocking methods if some async method returned an async type). Don't actually remember all of the reasoning for merging the objects, but I think one reason was that de-serialization of serialized Modal objects was hard to get consistently working like users wanted (i.e. should a serialized *Image be decoded as an Image or an AioImage?)

PR #2801, opened 2025-01-23
Hydrate secrets in parallel during Sandbox.exec

mwaskom:
## Describe your changes

Noticed this while migrating `.resolve` -> `.hydrate`.

If passing a large number of secrets, this could potentially reduce the latency of `Sandbox.exec`.

---

---

</details>


pawalt (approved):
Awesome thanks!

PR #2803, opened 2025-01-24
Handle resizing in pty-shell

azliu0:
## Describe your changes

Very minor QoL improvement: support resizing the terminal window in `pty-shell`.

## Implementation detail

It'd be cleaner to create the `WindowSizeHandler` inside of the `ContainerProcess` handle, but signal modifications must happen in the main thread / event loop. Synchronicity forces all apis wrapped in `synchronize_api` to execute in a different thread, hence the need to create `WindowSizeHandler` externally and pass it in as an arg to `attach`.

## Server pr

- [ ] Pairs with server changes https://github.com/modal-labs/modal/pull/19216

---

---

</details>

### DIFF
--- a/modal/_utils/shell_utils.py
+++ b/modal/_utils/shell_utils.py
@@ -15,7 +15,7 @@ from typing import Callable, Optional
 
 from modal._pty import raw_terminal, set_nonblocking
 
-from .async_utils import asyncify
+from .async_utils import TaskContext, asyncify
 
 
 def write_to_fd(fd: int, data: bytes):
@@ -94,20 +94,16 @@ class WindowSizeHandler:
         """
         self._is_main_thread = threading.current_thread() is threading.main_thread()
 
-        self._current_size: tuple[int, int] = self._get_terminal_size()
+        self._current_size: tuple[int, int] = shutil.get_terminal_size()
         self._size_changed = asyncio.Event()
 
         if self._is_main_thread and hasattr(signal, "SIGWINCH"):
             signal.signal(signal.SIGWINCH, self._signal_resize_event)
 
-    def _get_terminal_size(self) -> tuple[int, int]:
-        c, r = shutil.get_terminal_size()
-        return r, c
-
     def _signal_resize_event(self, signum: Optional[int] = None, frame: Optional[FrameType] = None) -> None:
         """Custom signal handler for SIGWINCH."""
         try:
-            self._current_size = self._get_terminal_size()
+            self._current_size = shutil.get_terminal_size()
             self._size_changed.set()
         except Exception:
             # ignore failed window size reads
@@ -127,11 +123,9 @@ class WindowSizeHandler:
         async def process_events():
             while True:
                 await self._size_changed.wait()
-                await handler(*self._current_size)
                 self._size_changed.clear()
+                await handler(*self._current_size)
 
-        event_task = asyncio.create_task(process_events())
-        try:
+        async with TaskContext() as tc:
+            tc.create_task(process_events())
             yield
-        finally:
-            event_task.cancel()
--- a/modal/_utils/shell_utils.py
+++ b/modal/_utils/shell_utils.py
@@ -3,10 +3,17 @@
 import asyncio
 import contextlib
 import errno
+import fcntl
 import os
 import select
+import signal
+import struct
 import sys
+import termios
+import threading
 from collections.abc import Coroutine
+from queue import Empty, Queue
+from types import FrameType
 from typing import Callable, Optional
 
 from modal._pty import raw_terminal, set_nonblocking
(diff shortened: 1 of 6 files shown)


modal/_utils/shell_utils.py
```diff
@@ -3,10 +3,17 @@
 import asyncio
 import contextlib
 import errno
+import fcntl
```
thecodingwizard, line 6: Do these packages import correctly in Windows? We don't support `container exec` on windows, but just wanted to make sure these imports won't break modal-client for Windows entirely.

modal/_utils/shell_utils.py
```diff
@@ -77,3 +84,51 @@ async def _write():
         yield
     os.write(quit_pipe_write, b"\n")
     write_task.cancel()
+
+
+class WindowSizeHandler:
+    """Handles terminal window resize events."""
+
+    def __init__(self):
+        """Initialize window size handler. Must be called from the main thread to set signals properly.
```
thecodingwizard, line 93: > Must be called from the main thread to set signals properly.

Out of curiosity, is this a system limitation? i.e. Python window size handler doesn't work unless it's run on the main thread?

modal/_utils/shell_utils.py
```diff
@@ -77,3 +84,51 @@ async def _write():
...
+        self._is_main_thread = threading.current_thread() is threading.main_thread()
+        self._event_queue: Queue[tuple[int, int]] = Queue()
+
+        if self._is_main_thread and hasattr(signal, "SIGWINCH"):
+            signal.signal(signal.SIGWINCH, self._queue_resize_event)
+
+    def _queue_resize_event(self, signum: Optional[int] = None, frame: Optional[FrameType] = None) -> None:
+        """Signal handler for SIGWINCH that queues events."""
+        try:
+            hw = struct.unpack("hh", fcntl.ioctl(sys.stdout.fileno(), termios.TIOCGWINSZ, b"1234"))
```
thecodingwizard, line 105: what does b"1234" do here?

modal/_utils/shell_utils.py
thecodingwizard: If it isn't a system limitation, would it be more clear if `enabled: bool` is passed in as an explicit argument (so this can be disabled in tests)? I'm not super opinionated though

modal/_utils/shell_utils.py
```diff
@@ -77,3 +84,51 @@ async def _write():
...
+            yield
+            return
+
+        async def process_events():
+            while True:
+                try:
+                    rows, cols = self._event_queue.get_nowait()
+                    await handler(rows, cols)
+                except Empty:
+                    await asyncio.sleep(0.1)
```
thecodingwizard, line 128: hmm... at first glance I don't really love this whole Queue thing:

- We don't care about having a queue of resize events. We only care about the very last resize event. So I'm not sure that a queue is the right data structure
- having to use asyncio.sleep(0.1) seems a bit unfortunate
- almost feels like watch_window_size is similar to `signal.signal(signal.SIGWINCH, ...)` (i.e. run a function whenever the window is resized). except we have to add a queue around it presumably because we want the handler to be async.

I don't have a well-thought-out alternative at the moment though

modal/container_process.py
```diff
@@ -151,6 +152,14 @@ async def _handle_input(data: bytes, message_index: int):
             self.stdin.write(data)
             await self.stdin.drain()
 
+        async def _send_window_resize(rows: int, cols: int):
+            # resize sequence:
+            # - 2 bytes for the number of rows (big-endian)
+            # - 2 bytes for the number of columns (big-endian)
+            dims = struct.pack(">HH", rows, cols)
+            self.stdin.write(dims)
+            await self.stdin.drain(_terminal_size=(rows, cols))
```
thecodingwizard, line 161: do we still need `self.stdin.write(dims)` here? Isn't it sufficient to just `.drain(_terminal_size=(...))`?

modal/io_streams.py
```diff
@@ -384,7 +384,10 @@ def write_eof(self) -> None:
         """
         self._is_closed = True
 
-    async def drain(self) -> None:
+    async def drain(
+        self,
+        _terminal_size: Optional[tuple[int, int]] = None,  # internal option to send terminal window resize events
```
thecodingwizard, line 389:
```suggestion
        _terminal_size: Optional[tuple[int, int]] = None,  # internal option to send terminal resize events to container exec
```

modal/_utils/shell_utils.py
azliu0: It is a system limitation; from the [docs](https://docs.python.org/3/library/signal.html)

> Besides, only the main thread of the main interpreter is allowed to set a new signal handler.

modal/_utils/shell_utils.py
azliu0: moved to `shutil`, which should import properly everywhere

modal/_utils/shell_utils.py
azliu0: modified to use event-based signalling instead

modal/_utils/shell_utils.py
```diff
@@ -77,3 +81,51 @@ async def _write():
...
+        """
+        if not self._is_main_thread:
+            yield
+            return
+
+        async def process_events():
+            while True:
+                await self._size_changed.wait()
+                await handler(*self._current_size)
+                self._size_changed.clear()
```
thecodingwizard, line 127:
```suggestion
                self._size_changed.clear()
```
```suggestion
                await self._size_changed.wait()
                self._size_changed.clear()
                await handler(*self._current_size)
```

Does the ordering of this need to be flipped to avoid race conditions? where the size changes after the handler runs but before we clear size_changed?

modal/_utils/shell_utils.py
thecodingwizard: Nice, I think this implementation makes sense to me considering `signal.signal` can only be called from the main thread

mwaskom:
Closing as stale; linked to this from a ticket so we can pick it back up at some point in the future.

PR #2806, opened 2025-01-24
CLI-319: Add support for custom pickle type to parameterized functions

kramstrom:
## Describe your changes

Adds pickle typ to parametrized functions but no functionality to use the type yet.

---

---

</details>


modal/_utils/function_utils.py
```diff
@@ -41,6 +41,7 @@ class FunctionInfoType(Enum):
 CLASS_PARAM_TYPE_MAP: dict[type, tuple["api_pb2.ParameterType.ValueType", str]] = {
     str: (api_pb2.PARAM_TYPE_STRING, "string_default"),
     int: (api_pb2.PARAM_TYPE_INT, "int_default"),
+    Any: (api_pb2.PARAM_TYPE_PICKLE, "pickle_default"),
```
kramstrom, line 44: Perhaps this should be a custom catch-all type instead of `Any`

modal/cls.py
```diff
@@ -475,11 +475,10 @@ def validate_construction_mechanism(user_cls):
...
             if t not in CLASS_PARAM_TYPE_MAP:
-                t_name = getattr(t, "__name__", repr(t))
-                supported = ", ".join(t.__name__ for t in CLASS_PARAM_TYPE_MAP.keys())
-                raise InvalidError(
-                    f"{user_cls.__name__}.{k}: {t_name} is not a supported parameter type. Use one of: {supported}"
-                )
+                pass
+                # TODO:
+                # raise if cls has webhooks
+                # and no default value for pickle parameter
```
kramstrom, line 481: I guess the previous check is no longer necessary since the new type will be "catch-all" -- however we could add that too if we add an explicit type that a user must use instead

test/cls_test.py
```diff
@@ -878,6 +878,7 @@ class UsingAnnotationParameters:
     a: int = modal.parameter()
     b: str = modal.parameter(default="hello")
     c: float = modal.parameter(init=False)
+    d: Any = modal.parameter(default={"foo": "bar"})
```
freider, line 881: Thinking some more about this I think it might be better to **not** type these as `Any`, but instead `typing.Annotated[{ActualType}, modal.PickleSerialization]` or similar.

Otherwise users can't both get good typing support in their editor and support for non-supported types.

freider: I.e. then users can do:

```
arr: Annotated[pandas.DataFrame, modal.PickleSerialization] = modal.parameter()
```

kramstrom: Added!

modal/cls.py
kramstrom: nvm we can keep this and require to use the correct annotation

freider:
Decided against this for now

PR #2807, opened 2025-01-27
Accept variable-length arguments for custom argument parsing

aksh-at:
Adds support for accepting variable-length arguments to entrypoints. See changelog for more.

## Describe your changes

---

---

</details>

## Changelog

* Modal functions, methods and entrypoints can now accept variable-length arguments to skip Modal's default CLI parsing. This is useful if you want to use Modal with custom argument parsing via `argparse` or `HfArgumentParser`. For example, the following function can be invoked with `modal run my_file.py --foo=42 --bar="baz"`:

    ```python
    import argparse

    @app.function()
    def train(*arglist):
        parser = argparse.ArgumentParser()
        parser.add_argument("--foo", type=int)
        parser.add_argument("--bar", type=str)
        args = parser.parse_args(args = arglist)
```

### DIFF
--- a/modal/cli/__init__.py
+++ b/modal/cli/__init__.py
@@ -1,1 +1,5 @@
 # Copyright Modal Labs 2022
+
+from .run import get_extra_args
+
+__all__ = ["get_extra_args"]
--- a/modal/cli/__init__.py
+++ b/modal/cli/__init__.py
@@ -1,5 +1,1 @@
 # Copyright Modal Labs 2022
-
-from .run import get_extra_args
-
-__all__ = ["get_extra_args"]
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by aksh-at** [comment]
Could also check that the type of `*args` (if it's passed in) is `tuple[str, …]` or similar but that seems a bit unnecessary?

**Review by mwaskom — APPROVED** [review]
LGTM; I suppose in retrospect it might have been ok to support variadic args and _additional keyword arguments_

```python
@app.function()
def f(*arglist: str, gpu: str = "h100"):
    ...    
```

although I'm not sure how much more complicated that would have been with click, and I like that this is pretty straightforward. Adding it later would not have any compatibility cost though so we can just see if a need for it arises.

aksh-at:
> Adding it later would not have any compatibility cost though so we can just see if a need for it arises.

Yeah exactly; I think it's doable and we can add later if needed.

PR #2809, opened 2025-01-27
app layout refactoring

erikbern:
Something I mostly did while out of wifi on a plane. Idk if this makes sense, but will think about it for a sec. I think if I pull this thread a bit more, it's possible to remove some complexity.


mwaskom:
I'm in favor of this in general but looks pretty conflicted at this point; gonna close for now and we can revisit in the future.

PR #2811, opened 2025-01-27
Add attempt number input item proto

rculbertson:
Part of SVC-180.

This adds attempt number to the input item proto. Adding just the proto definitions so it can be used by the server ([see this PR](https://github.com/modal-labs/modal/pull/19299). Will make follow up PRs to actually use this field in the client.

---

</details>


rculbertson:
@prbot approve

gongy (approved):
Perhaps a name so `0` values feel more natural? Maybe `retry_count`

rculbertson:
@prbot automerge

PR #2813, opened 2025-01-28
Fix and test for Function.from_name in module scope preventing modal run [CLI-323]

freider:
## Changelog

* Fixes bug introduced in v0.72.48 where `modal run` didn't work with files having global `Function.from_name()`/`Function.lookup()`/`Cls.from_name()`/`Cls.lookup()` calls.


freider:
@prbot approve

mwaskom:
✅  Thanks for the quick fix.

PR #2816, opened 2025-01-28
Add `wait_until_ready=True` to sandbox snapshot task id retrieval

azliu0:
## Describe your changes

I was playing around with trying to break snapshots ([see here](https://github.com/modal-labs/modal/pull/19338)) and realized I forgot to include this earlier.

For context, this flag will more gracefully propagate snapshot failures to the client. Without this flag, it may be possible to `exec` something on the broken snapshot and receive a `container exited` failure instead.

---

---

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -581,7 +581,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         )
         sandbox = await _Sandbox.from_id(restore_resp.sandbox_id, client)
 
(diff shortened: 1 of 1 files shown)


azliu0:
@prbot approve

thecodingwizard (commented):
lgtm!

thecodingwizard:
Oops actually @azliu0 I think we need to explicitly specify the timeout field here. The server code uses the presence of the timeout field to determine whether it is a legacy client that expects `task_id=""` or a new client that expects a DEADLINE_EXCEEDED error.

PR #2817, opened 2025-01-28
Chore: Split functions into separate modules for implementation and wrapper

freider:
Similar to what we did for modal.object recently.

It surfaced a few minor typing issues. Mostly just cleans up the type stub by not having to emit all of already typed private implementation details. This also allows mypy to type check implementation since it's no longer "shadowed" by a type stub

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -32,7 +32,6 @@ from ._utils.async_utils import (
     aclosing,
     async_merge,
     callable_to_agen,
-    synchronize_api,
     synchronizer,
     warn_if_generator_is_not_consumed,
 )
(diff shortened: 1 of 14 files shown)


freider:
> Can we hold off until I get the lookup deprecation pr in plz. It has a large surface area.

wdym - this is a tiny change! :trollface: 
image

mwaskom:
Also should we take the opportunity to singular-ize `functions` to match the other object-oriented (heh) modules?

freider:
> Also should we take the opportunity to singular-ize `functions` to match the other object-oriented (heh) modules?

That's a bit messier and would require some deprecation warning and fallback when importing the old one, since a bunch of our examples still use `modal.functions.FunctionCall.from_id()` etc :(

freider:
@prbot approve

mwaskom:
> since a bunch of our examples still use modal.functions.FunctionCall.from_id() etc

ugh, true. we should clean that up.

PR #2812, opened 2025-01-27
Deprecate `.lookup` methods on Modal objects

mwaskom:
## Describe your changes

We're deprecating `.lookup` methods on Modal objects in favor of `.from_name`. This PR adds deprecation warnings to all `.lookup` methods on `modal.Object` subclasses. The one exception is `modal.App.lookup`, which needs to be handled separately.

Part of CLI-96

<details> <summary>Backward/forward compatibility checks</summary>

---

Check these boxes or delete any item (or this section) if not relevant for this PR.

- [x] Client+Server: this change is compatible with old servers
- [x] Client forward compatibility: this change ensures client can accept data intended for later versions of itself

Note on protobuf: protobuf message changes in one place may have impact to
multiple entities (client, server, worker, database). See points above.

---

</details>

## Changelog

- Deprecated `.lookup` methods on Modal objects. Users are encouraged to use `.from_name` instead. In most cases this will be a simple name substitution. See [the 1.0 migration guide](https://modal.com/docs/guide/modal-1-0-migration#deprecating-the-lookup-method-on-modal-objects) for more information.

PR #2818, opened 2025-01-28
Allow lazy hydration after Mount.from_name

mwaskom:
Fixes an issue that popped up when the deployment pipeline ran on #2812

`Mount.from_name` does not instantiate the object with `hydrate_lazily=True`, so it's not possible to chain `modal.Mount.from_name(...).hydrate()` as we need to do for strict replacement of `modal.Mount.lookup`.

I might be missing something but I don't think there's any reason we can't enable lazy hydration here. Risk should be relatively low in any case as `Mount.from_name` is not really user-facing (and `Mount` itself will soon be deprecated from public API).


mwaskom:
@prbot approve

freider (commented):
👍

PR #2820, opened 2025-01-29
Add proto changes for pty-shell resizing

azliu0:
## Describe your changes

Adds proto changes for handling resize in `pty-shell`.
- Full client changes: https://github.com/modal-labs/modal-client/pull/2803
- Worker changes: https://github.com/modal-labs/modal/pull/19216

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2180,6 +2180,9 @@ message RuntimeInputMessage {
   bytes message = 1;
   uint64 message_index = 2;
   bool eof = 3;
+  // only used to handle `container exec` terminal resize events
+  optional uint32 window_rows = 4;
+  optional uint32 window_cols = 5;
 }
 
 message RuntimeOutputBatch {


thecodingwizard (commented):
Would it be better to have `optional uint32 window_rows` and `optional uint32 window_cols` be part of `RuntimeInputMessage`, and have the worker / runtime construct the magical resize sequence based off this information, rather than the client?

That way, in case we decide to change the magical resize sequence in the future, we don't have to worry about backwards-compatibility with older clients.

Please let me know if you disagree though.

modal_proto/api.proto
```diff
@@ -2180,6 +2180,8 @@ message RuntimeInputMessage {
   bytes message = 1;
   uint64 message_index = 2;
   bool eof = 3;
+  optional uint32 window_rows = 4;
+  optional uint32 window_cols = 5;
```
thecodingwizard, line 2184: - maybe we should comment that window_rows and window_cols are only used for `container exec`?
- do you think `window_rows` or `terminal_rows` or `pty_rows` is clearer? (feel free to keep as `window_rows` if you think that makes more sense to you)

thecodingwizard (approved):
lgtm! @thundergolfer should anyone else review this PR (and the other pty-shell resizing PR's)?

modal_proto/api.proto
azliu0: no strong preference on the naming! willing to defer to anyone else who has a strong opinion on what this field should look like

mwaskom:
Closing as stale but made an issue to track and revisit: https://linear.app/modal-labs/issue/CLI-379/improve-handling-of-terminal-resizing-in-modal-shell

PR #2822, opened 2025-01-29
Rename parameterized to parametrized everywhere

kramstrom:
## Describe your changes

SVC-305

Went for parametrized without the e over paramet`e`rized since that's what `pytest` uses 🤷. Both are apparently valid and widespread.

---

---

</details>


mwaskom (approved):
LGTM, but don't merge until the monorepo change goes in since we're now pointing at the new URL for the docs

PR #2823, opened 2025-01-30
Remove 0.62 lookup compatibility code from classes

freider:
Removing some of the more obvious 0.62 compatibility layers from the Cls/Obj code, now that 0.62 is EOL

I think there are more opportunities for refactoring/simplification of the class code that requires some more careful consideration, and once 0.67 is the minimum version we can probably remove a whole lot more (we wouldn't even need backend "class" objects at that point.)

---

---

</details>


freider:
@prbot approve

erikbern:
could maybe just rename it `_function.py` now. I think we called it `functions.py` because the name clashed with something

freider:
> could maybe just rename it `_function.py` now. I think we called it `functions.py` because the name clashed with something

Yeah the internal one can easily be renamed. The "public" functions.py need some deprecation first since we use `modal.functions.FunctionCall` references not going via the top level modal package sometimes in our examples :(

PR #2824, opened 2025-01-30
Update proto so FunctionRetry returns input_jwt

rculbertson:
The FunctionRetry endpoint needs to return the input_jwt of the retried input. On retry, the input will get a new entry_id. We need to return an updated input_jwt to the caller with the correct entry_id, because we use the entry_id to track lost inputs.

---

</details>


rculbertson:
@prbot approve

rculbertson:
@prbot automerge

PR #2826, opened 2025-01-31
Use parameter info to validate parameters on instance creation

freider:
This uses the parameter info from the class service function metadata to validate parameters on instance creation (when implicit/typed constructors are used).

This also fixes the incorrect assumption that the base/parent service for a class can be used whenever the calling arguments are empty, by instead checking if the arguments can be serialized and if they match the defaults

Fixes CLI-328
Fixes CLI-327

--- 

Pretty sure this can have subtle effects on weird code paths and break in unexpected ways, so I'll make sure to run integration tests before merging this (if the client tests pass...)

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -952,8 +952,6 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         Binds a class-function to a specific instance of (init params, options) or a new workspace
         """
 
-        # In some cases, reuse the base function, i.e. not create new clones of each method or the "service function"
-        can_use_parent = len(args) + len(kwargs) == 0 and options is None
         parent = self
 
         async def _load(param_bound_func: _Function, resolver: Resolver, existing_object_id: Optional[str]):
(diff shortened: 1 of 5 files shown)


modal/_functions.py
```diff
@@ -1006,10 +1011,6 @@ async def _load(param_bound_func: _Function, resolver: Resolver, existing_object
 
         fun: _Function = _Function._from_loader(_load, "Function(parametrized)", hydrate_lazily=True)
 
-        if can_use_parent and parent.is_hydrated:
```
freider, line 1009: I don't remember if there were cases where this was a crucial code path - I hope not!

modal/_functions.py
```diff
@@ -952,8 +952,6 @@ def _bind_parameters(
         Binds a class-function to a specific instance of (init params, options) or a new workspace
         """
 
-        # In some cases, reuse the base function, i.e. not create new clones of each method or the "service function"
-        can_use_parent = len(args) + len(kwargs) == 0 and options is None
```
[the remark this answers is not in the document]
mwaskom: For my own understanding: this is apparently the load-bearing change in this diff.

mwaskom (approved):

modal/_functions.py
mwaskom: github reactions need to add the "see no evil" monkey

PR #2828, opened 2025-01-31
Add root_function_id to TaskLogsBatch

RohanArepally:
https://linear.app/modal-labs/issue/PRD-555/migrate-logs-table-primary-key-from-app-id-timestamp-task-id-to-app-id

Need to add a new field to `TaskLogsBatch` so that we can push the `root_function_id` to our new ClickHouse table (once that's up)


devennavani (approved):
LGTM

PR #2829, opened 2025-02-01
Adds experimental support for GPU memory snapshots

luiscape:
Adds support for snapshotting GPU memory using `cuda-checkpoint`. The interface is reserved for internal/experimental use.

### DIFF
--- a/modal/_runtime/container_io_manager.py
+++ b/modal/_runtime/container_io_manager.py
@@ -881,8 +881,13 @@ class _ContainerIOManager:
         # Restore GPU memory.
         if self.function_def._experimental_enable_gpu_snapshot and self.function_def.resources.gpu_config.gpu_type:
             logger.debug("GPU memory snapshot enabled. Attempting to restore GPU memory.")
-            if gpu_memory_snapshot.get_state() == gpu_memory_snapshot.CudaCheckpointState.CHECKPOINTED:
-                gpu_memory_snapshot.toggle()
+            gpu_process_state = gpu_memory_snapshot.get_state()
+            if gpu_process_state != gpu_memory_snapshot.CudaCheckpointState.CHECKPOINTED:
+                raise ValueError(
+                    "Cannot restore GPU state if GPU isn't in a 'checkpointed' state. "
+                    f"Current GPU state: {gpu_process_state}"
+                )
+            gpu_memory_snapshot.toggle()
 
         # Restore input to default state.
         self.current_input_id = None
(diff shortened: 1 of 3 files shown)


modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -0,0 +1,104 @@
+# Copyright Modal Labs 2022
```
erikbern, line 1: feels a little inelegant to do this from within the client code. is it possible to do it from the outside?

not important right now – what you're doing here seems like the easiest approach to get it running. just curious about the long term architecture.

erikbern:
extremely cool!!

modal/_runtime/gpu_memory_snapshot.py
cweld510: presumably we could exec into the container to do this?

erikbern: yeah or even cleaner would be for gvisor to trigger the underlying ioctl stuff directly through the nvproxy – wouldn't even need a binary then

luiscape: Let's think about getting this working outside the container somehow. I couldn't get the PID matching--pid inside the container vs pid outside--to work.

luiscape:
@cweld510 just added `_experimental_enable_gpu_snapshot` in place of the config option. Let me know what you think.

modal/_runtime/container_io_manager.py
```diff
@@ -877,6 +878,12 @@ async def memory_restore(self) -> None:
             if value != "":
                 config.override_locally(key, value)
 
+        # Restore GPU memory.
+        if self.function_def._experimental_enable_gpu_snapshot and self.function_def.resources.gpu_config.gpu_type:
+            logger.debug("GPU memory snapshot enabled. Attempting to restore GPU memory.")
+            if gpu_memory_snapshot.get_state() == gpu_memory_snapshot.CudaCheckpointState.CHECKPOINTED:
```
cweld510, line 884: don't we always expect the state to be `CHECKPOINTED`? Can we raise or log if it isn't?

modal/_runtime/container_io_manager.py
```diff
@@ -892,6 +899,13 @@ async def memory_snapshot(self) -> None:
 
         # Pause heartbeats since they keep the client connection open which causes the snapshotter to crash
         async with self.heartbeat_condition:
+            # Snapshot GPU memory.
+            if self.function_def._experimental_enable_gpu_snapshot and self.function_def.resources.gpu_config.gpu_type:
+                logger.debug("GPU memory snapshot enabled. Attempting to snapshot GPU memory.")
+                if gpu_memory_snapshot.get_state() == gpu_memory_snapshot.CudaCheckpointState.RUNNING:
```
cweld510, line 905: when would the state here not be `RUNNING`? Similarly, can we log if it isn't in the expected state?

modal/_runtime/container_io_manager.py
luiscape: Users may not be running any GPU workloads, even if they have the `_experimental_enable_gpu_snapshot` enabled. Is that a reasonable use-case to account for?

thundergolfer (commented):
I think this can be done entirely from the worker side with no changes to the client. We can exec processes in the container, we control the pre-checkpoint activity after the client has sent the 'checkpoint me' RPC, and we control the pre-restore activity because the container is busy looping on the restore state JSON file. Can discuss details IRL. 

But seeing as this already does the job OK to merge it and use it and then remove it later.

cweld510:
We've discussed the worker-only alternative on the client PR - I think it's pretty reasonable

julien-blanchon:
Any documentation about how to use ?

mwaskom:
Hi @julien-blanchon, as mentioned:

> The interface is reserved for internal/experimental use.

Features exposed through `_experimental` flags are often under active development and not ready for end user adoption.

mwaskom:
Keep your eyes peeled for a stable release though!

julien-blanchon:
Gotcha, I hope this will come soon.

I don't know why I have this error in my case:

```
Error toggling CUDA in process ID 2: "OS call failed or operation not supported on this OS"
```

With `nvidia/cuda:12.4.0-devel-ubuntu22.04`.

But look like cuda-checkpoint is available

luiscape:
Hi @julien-blanchon, GPU memory snapshots are very experimental and have many edge cases. The most important issue is related to GPU passthrough (you can find more details [here](https://github.com/NVIDIA/cuda-checkpoint/issues/18#issuecomment-2638716830)). We're waiting for a new driver release that fixes those issues.

PR #2830, opened 2025-02-03
Adds experimental GPU snapshotting flag

luiscape:
Creates new experimental flag to be used with GPU snapshotting.

PR #2831, opened 2025-02-03
Adds experimental GPU snapshotting flag to Sanboxes

luiscape:
Adds interface for experimental GPU support for memory snapshots, similar to https://github.com/modal-labs/modal-client/pull/2830


cweld510:
We don't really need this yet - there are currently no plans to support snapshotting GPU sandboxes.

PR #2835, opened 2025-02-04
Fix the task count status text

mwaskom:
## Describe your changes

A while back a change was made server-side to stop differentiating between TASK_STATE_IDLE and TASK_STATE_ACTIVE. For arcane reasons, TASK_STATE_IDLE was chosen to represent all tasks that had finished cold starting and not yet been terminated. That broke the status message that we show for ephemeral apps, which was counting "active" tasks as the numerator. I'm changing that to show the sum of "active" and "idle" tasks. As mentioned, "active" is currently always 0, but this approach is more future-proof in case we decide to switch the server to use the more intuitive value or reinstate the distinction between active and idle in the future.

(This does change the meaning of the status message relative to the original intention of just counting the tasks that are currently handling inputs, but it's been broken for a while so I don't think that's a problem).

- Fixes SVC-204
 
---

---

</details>

## Changelog

- Fixed the status message shown in terminal logs for ephemeral Apps to accurately report the number of active containers.

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -291,7 +291,9 @@ class OutputManager:
 
         # The most advanced state that's present informs the message.
         if api_pb2.TASK_STATE_ACTIVE in states_set or api_pb2.TASK_STATE_IDLE in states_set:
-            tasks_running = tasks_at_state(api_pb2.TASK_STATE_ACTIVE)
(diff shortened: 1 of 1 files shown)


thecodingwizard (approved):
thank you!

PR #2836, opened 2025-02-04
Remove GPUConfig.memory

erikbern:
Legacy field that isn't used anywhere (including the server)


mwaskom (approved):
Looks like the tests surfaced an issue, but otherwise I'm on board with this.

erikbern:
thanks – fixed some broken tests

PR #2837, opened 2025-02-05
Add native bytes type to class parameters

kramstrom:
Adds support for new strict `bytes` type for  `modal.parameter`

Usage:

```py
@app.cls()
class C:
    a: bytes = modal.parameter(default=b"hello")
```

---

---

</details>

## Changelog

* Adds support for new strict `bytes` type for  `modal.parameter`

Usage:
```py
import typing
import modal

app = modal.App()


@app.cls()
class Foo:
    a: bytes = modal.parameter(default=b"hello")

    @modal.method()
    def bar(self):
        return f"hello {self.a}"


@app.local_entrypoint()
def main():
    foo = Foo(a=b"world")
    foo.bar.remote()
```

**Note**: For parameterized web endoints you must base64 encode the bytes before passing them in as a query parameter.


kramstrom:
Need to merge the protos before merging this & merge the server-side code before this

modal/_utils/function_utils.py
```diff
@@ -296,12 +289,13 @@ def class_parameter_info(self) -> api_pb2.ClassParameterInfo:
         signature = _get_class_constructor_signature(self.user_cls)
         for param in signature.parameters.values():
             has_default = param.default is not param.empty
-            if param.annotation not in CLASS_PARAM_TYPE_MAP:
-                raise InvalidError("modal.parameter() currently only support str or int types")
-            param_type, default_field = CLASS_PARAM_TYPE_MAP[param.annotation]
-            class_param_spec = api_pb2.ClassParameterSpec(name=param.name, has_default=has_default, type=param_type)
+            if param.annotation not in PYTHON_TO_PROTO_TYPE:
+                raise InvalidError("modal.parameter() currently only support str, int, or bytes types")
```
mwaskom, line 293: nit
```suggestion
                raise InvalidError("modal.parameter() currently only supports str, int, or bytes types")
```

mwaskom: Also might be nice to build the error message programmatically, feels like we could easily forget to expand it in the future if we add more type support.

mwaskom (approved):
Not totally following all the maps that are happening here, seems like this could be a little simpler overall, but the bytes support itself is straightforward, thanks for adding this :+1:

modal/_serialization.py
```diff
@@ -389,23 +389,38 @@ def check_valid_cls_constructor_arg(key, obj):
         )
 
 
+def assert_bytes(obj: Any):
+    if not isinstance(obj, bytes):
+        raise ValueError(f"Expected bytes, got {type(obj)}")
```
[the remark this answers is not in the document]
freider: Yeah, that's a good point

modal/_utils/function_utils.py
freider: Turns out this error message is never normally triggered, since there was a very similar validation check that runs as part of the Cls.from_local constructor (which does have programmatic generation of the message). I refactored the two to use the same common validator now

freider:
> Not totally following all the maps that are happening here, seems like this could be a little simpler overall, but the bytes support itself is straightforward, thanks for adding this 👍

Yeah I have some plans for how to clean this up, but I want to do refactor some other parts first to make that refactor more worth while

PR #2838, opened 2025-02-05
Add deprecation warning when looking up methods

freider:
This will let us remove a bunch of method placeholder and shimming code on the backend once we prevent this fully.

---

---

</details>

## Changelog

* Adds a pending deprecation warning when looking up class methods using `Function.from_name`, e.g. `Function.from_name("some_app", "SomeClass.some_method")`. The recommended way to reference methods of classes is to look up the class instead: `RemoteClass = Cls.from_name("some_app", "SomeClass")`


modal/_functions.py
```diff
@@ -1092,6 +1092,16 @@ def from_name(
         f = modal.Function.from_name("other-app", "function")
         ```
         """
+        if "." in name:
+                (2025, 2, 5),
+                "Looking up class methods using Function.from_name will be deprecated"
+                " in a future version of Modal. Use modal.Cls.from_name() instead, e.g.\n"
+                "\n"
+                'DeployedClass = modal.Cls.from_name("some_app", "MyClass")\n'
```
mwaskom, line 1101: We should be able to translate the arguments into an exact suggestion, right?

Side note, we don't have a very consistent pattern in terms of what we suggest for naming the object that is returned by `Cls.from_name` and it's something I sort of get hung up on myself when writing code.

erikbern:
Side note but I think it would be great to support a response header such as `x-modal-warning` which we can set server-side and causes the client to output some warning.

freider:
> Side note but I think it would be great to support a response header such as `x-modal-warning` which we can set server-side and causes the client to output some warning. It would have let us implement this logic entirely server-side.

I like this idea!

freider:
> > I like this idea!
> 
> Nice! I think it's not super hard to implement server side. Client-side it would be nice to extend the proto stub generation so we can do this for all rpc calls. Otherwise we could always put it into `retry_transient_errors`

I can take a stab at it. I think we should get rid of `retry_transient_errors` and instead have everything in the modal-specific stub wrapper

erikbern:
> I can take a stab at it. I think we should get rid of `retry_transient_errors` and instead have everything in the modal-specific stub wrapper

Yep exactly, that's what I was thinking. Then adding new logic for warnings etc would immediately apply to every request.

No rush on this. I can take a look at the server side. It's a bit unclear what the SDK would look like.

freider:
@prbot approve

freider:
@prbot automerge

PR #2840, opened 2025-02-05
Don't send GPU_TYPE from client

erikbern:
This removes the use of GPU_TYPE enum.

A later change would make it possible to pass-through strings to the server without having to go through the gpu classes defined in `modal/gpu.py`. For now we still rely on those.

### DIFF
--- a/modal/gpu.py
+++ b/modal/gpu.py
@@ -9,14 +9,14 @@ from .exception import InvalidError
 
 @dataclass(frozen=True)
 class _GPUConfig:
-    count: int
     gpu_type: str
+    count: int
 
     def _to_proto(self) -> api_pb2.GPUConfig:
         """Convert this GPU config to an internal protobuf representation."""
         return api_pb2.GPUConfig(
-            count=self.count,
             gpu_type=self.gpu_type,
+            count=self.count,
         )
 
 
@@ -31,7 +31,7 @@ class T4(_GPUConfig):
         self,
         count: int = 1,  # Number of GPUs per container. Defaults to 1.
     ):
-        super().__init__(count, "T4")
+        super().__init__("T4", count)
 
     def __repr__(self):
         return f"GPU(T4, count={self.count})"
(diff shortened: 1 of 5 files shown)


erikbern:
btw a thought on pass-through strings for gpu-types:
1. should the client or the server handle the `A100` -> `A100-40GB` mapping?
2. should the client or the server make the string uppercase?

no strong feelings on either

erikbern:
and btw another question is – what do we do with the gpu_count parameter for custom string types?

`gpu="h800", gpu_count=8`?

modal/gpu.py
```diff
@@ -33,7 +31,7 @@ def __init__(
         self,
         count: int = 1,  # Number of GPUs per container. Defaults to 1.
     ):
-        super().__init__(api_pb2.GPU_TYPE_T4, count, "T4")
+        super().__init__(count, "T4")
```
mwaskom, line 34: Minorly annoying that the order of arguments is reversed from the string spelling (i.e. `"T4:{count}"`).

modal/gpu.py
```diff
@@ -150,7 +148,7 @@ class Any(_GPUConfig):
     """Selects any one of the GPU classes available within Modal, according to availability."""
 
     def __init__(self, *, count: int = 1):
-        super().__init__(api_pb2.GPU_TYPE_ANY, count, "ANY")
+        super().__init__(count, "ANY")
```
mwaskom, line 151: Still think we should probably get rid of this, or think about a more correct way of spelling it.

mwaskom (approved):

modal/gpu.py
erikbern: yeah i can flip it

erikbern:
btw here are the next few steps i think:
1. in `_parse_gpu_config`, don't return the specific classes (`T4` etc), just return the underlying `_GPUConfig`, but keep the list of gpu options in the client
2. deprecate the specific classes in the client
3. add support server-side for normalization of gpu types (at least `a100` -> `a100-40gb`, maybe also upper-casing)
4. change the code in the client to just send strings to the server

erikbern:
Btw thinking more about it, should we even move the parsing of `:n` to the server? We could just have the client send a string to the server that it parses to a `GPUConfig` proto. That would make the `GPUConfig` proto internal to the server even. Maybe that's a good thing.

mwaskom:
Yeah I'm on board with that.

erikbern:
nice, that might actually be a bit cleaner. let me think about it.

PR #2841, opened 2025-02-05
Add protos for AutoscalerSettings and request where we'll use it

mwaskom:
## Describe your changes

Adding a protobuf message to bundle the "autoscaler settings" which we'll use as a unit in a few new places.

Also added a new field using it to the `FunctionUpdateSchedulerParamsRequest`. That's one of the places.

I would also like the bundle the autoscaler settings within the `Function` definition using this method. But I'll do that in a subsequent PR. I'd just like to unblock the backend PR right now.

- Part of SVC-335

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -561,6 +561,22 @@ message Asgi {
   }
 }
 
+message AutoscalerSettings {
+  // A collection of user-configurable settings for Function autoscaling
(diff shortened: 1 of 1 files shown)


mwaskom:
@prbot approve so I can get useful CI feedback on server PR; can change any of this until that lands

gongy (commented):
:+1:

PR #2842, opened 2025-02-06
Don't construct the specific GPU classes from _parse_gpu_config

erikbern:
~Just construct `_GPUConfig` classes.~

Edit: realized we can just return proto directly.

This lets us deprecate those classes if we want to (I think we should?)

Also refactor tests to use parametrization and covering a lot more cases.

### DIFF
--- a/modal/gpu.py
+++ b/modal/gpu.py
@@ -1,6 +1,6 @@
 # Copyright Modal Labs 2022
 from dataclasses import dataclass
-from typing import Optional, Union
+from typing import Union
 
 from modal_proto import api_pb2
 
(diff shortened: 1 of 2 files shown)


modal/gpu.py
```diff
@@ -195,19 +197,16 @@ def _parse_gpu_config(value: GPU_T) -> Optional[_GPUConfig]:
...
 
-        if value.lower() == "a100-20g":
-            return A100(size="20GB", count=count)  # Triggers unsupported error underneath.
-        elif value.lower() == "a100-40gb":
-            return A100(size="40GB", count=count)
-        elif value.lower() == "a100-80gb":
-            return A100(size="80GB", count=count)
-        elif value.lower() not in STRING_TO_GPU_CONFIG:
+        if value.lower() == "a100-40gb":
+            gpu_type = "A100-40GB"
```
mwaskom, line 201: Any reason not to just have a many-to-one map in `STRING_TO_GPU_TYPE`?

mwaskom: Oh because you don't want to present it as a separate option when you error out I gather.

mwaskom (approved):
LGTM, agree that deprecating the `modal.gpu` classes makes sense.

modal/gpu.py
erikbern: yeah i guess we could, no strong feelings

erikbern:
Thanks @mwaskom 

I'll merge this later today or tomorrow – there's been a lot of GPU type changes so I just want to make sure there's no unexpected prod issue.

PR #2843, opened 2025-02-06
[CLI-316] Avoid tripping the single input cancellation signal handler repeatedly.

danielshaar:
## Describe your changes

CLI-316

Tested against repro in ticket and confirmed the cancellation is only tripped once.

### DIFF
--- a/modal/_runtime/container_io_manager.py
+++ b/modal/_runtime/container_io_manager.py
@@ -258,6 +258,7 @@ class _ContainerIOManager:
     calls_completed: int
     total_user_time: float
     current_input_id: Optional[str]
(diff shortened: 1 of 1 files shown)


freider:
Don't we need to reset this once the InputCancellation exception has bubbled up to our exception handler? Otherwise any subsequent (actual) cancellations would also get muted, right?

danielshaar:
Good catch, I forgot that we don't re-instantiate the container IO manager on each wave of inputs.

danielshaar:
Discarding in favor of more robust - https://github.com/modal-labs/modal-client/pull/2847

PR #2845, opened 2025-02-06
Add runsc_runtime_version to sandbox proto

cweld510:
Necessary for sandbox snapshots - we need a way to store and refer to the runtime version used by the sandbox which was or will be snapshotted.


cweld510:
@prbot approve

cweld510:
@prbot automerge

thundergolfer (commented):
LGTM

PR #2847, opened 2025-02-07
[SVC-331] Gracefully handle user-thrown asyncio.CancelledErrors

thecodingwizard:
This PR gracefully handles user-thrown `asyncio.CancelledError`s, correctly propagating these exceptions back to the server and marking the input as failed _unless_ we received a cancellation request for this input.

Depends on #2850

There is an additional complication: What do we do when we get a `SIGINT` signal?

This PR preserves existing behavior, sending a TERMINATED result for async inputs and suppressing cancellation traceback and doing nothing for sync inputs.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -324,7 +324,7 @@ def call_function(
 
             user_code_event_loop.run(
                 run_concurrent_inputs(),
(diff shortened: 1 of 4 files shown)


thecodingwizard:
Closing this for now unless more users end up raising `asyncio.CancelledError`s.

PR #2848, opened 2025-02-08
Deprecate old GPU classes

erikbern:
Going forward the only sanctioned way to configure GPU is through strings. This is nice because it's a lot less code.

## Changelog

* Deprecate the GPU classes (`gpu=A100(...)` etc) in favor of just using strings (`gpu="A100"` etc)

### DIFF
--- a/modal/gpu.py
+++ b/modal/gpu.py
@@ -183,6 +183,16 @@ You can see a list of Modal GPU options in the
 def my_gpu_function():
     ... # This will have 4 A100-80GB with each container
 ```
+
+**Deprecation notes**
+
+An older deprecated way to configure GPU is also still supported,
+but will be removed in future versions of Modal. Examples:
+
+- `gpu=modal.gpu.H100()` will attach 1 H100 GPU to each container
+- `gpu=modal.gpu.T4(count=4)` will attach 4 T4 GPUs to each container
+- `gpu=modal.gpu.A100()` will attach 1 A100-40GB GPUs to each container
+- `gpu=modal.gpu.A100(size="80GB")` will attach 1 A100-80GB GPUs to each container
 """
 
 GPU_T = Union[None, str, _GPUConfig]
--- a/modal/gpu.py
+++ b/modal/gpu.py
@@ -218,5 +218,5 @@ def parse_gpu_config(value: GPU_T) -> api_pb2.GPUConfig:
         return api_pb2.GPUConfig()
     else:
         raise InvalidError(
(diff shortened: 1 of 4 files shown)


modal/gpu.py
```diff
@@ -197,7 +204,9 @@ def parse_gpu_config(value: GPU_T) -> api_pb2.GPUConfig:
             gpu_type=gpu_type,
             count=count,
         )
-    elif value is None or value is False:
+    elif value is None:
         return api_pb2.GPUConfig()
     else:
-        raise InvalidError(f"Invalid GPU config: {value}. Value must be a string, a `GPUConfig` object, or `None`.")
+        raise InvalidError(
+            f"Invalid GPU config: {value}. Value must be a string or `None` (or a deprecated `_GPUConfig` object)."
```
mwaskom, line 211: Why are we pointing users towards the private version of the object?

mwaskom (commented):
One thing that we use the GPU classes for is to document the details of each GPU in the docstring for the relevant class so that it shows up here: https://modal.com/docs/reference/modal.gpu

I think we need to move most of this reference page to a "guide" page so that the information is preserved (and can be updated independently of the client in the future).

I would also add some notes on the deprecation of the GPU classes to this page, i.e. in the docstring to the `modal.gpu` module.

modal/gpu.py
erikbern: `_GPUConfig` is just the base class of all the public objects

erikbern:
> One thing that we use the GPU classes for is to document the details of each GPU in the docstring for the relevant class so that it shows up here: https://modal.com/docs/reference/modal.gpu
> 
> I think we need to move most of this reference page to a "guide" page so that the information is preserved (and can be updated independently of the client in the future).

I actually did this last week. We used to link to the reference from the guide when listing all the GPU types, but the guide now lists all the GPU types instead: https://modal.com/docs/guide/gpu (although we should probably expand this). But I'd rather have an exhaustive list of GPU types in the guide and not the reference.

> 
> I would also add some notes on the deprecation of the GPU classes to this page, i.e. in the docstring to the `modal.gpu` module.

Sure can do

modal/gpu.py
mwaskom: Yeah just seems confusing to mention it in a public error message. Maybe "or a `modal.gpu` object, although these are deprecated"?

mwaskom (approved):
LGTM modulo the error message

modal/gpu.py
erikbern: sure sgtm, i can update

erikbern:
Thanks @mwaskom!

PR #2851, opened 2025-02-10
Don't send Image.gpu

erikbern:
This has been sent along with `Image.gpu_config` for a really long time, and we did have a place really deep in the code that uses `.gpu`. But we're removing it (using `gpu_config` instead), so we should be able to merge this change which deprecates that entire field.


erikbern:
Here's a simple script to verify that this works:

```python
import modal

app = modal.App("repro-cuda-image-issue")

image = modal.Image.debian_slim()

image = image.run_commands("nvidia-smi", gpu="l4:2", force_build=True)


@app.function(image=image)
def go():
    pass
```

mwaskom (commented):
Great!

PR #2850, opened 2025-02-10
Simplify input cancellation path

thecodingwizard:
## Describe your changes

This PR simplifies the input cancellation code path to always call `io_context.cancel()`. This solves #2843 and might be helpful for #2847.

PR #2853, opened 2025-02-10
[Alternative Implementation] Gracefully handle user-thrown asyncio.CancelledErrors 

thecodingwizard:
This PR gracefully handles user-thrown `asyncio.CancelledError`s, correctly propagating these exceptions back to the server and marking the input as failed _unless_ we received a cancellation request for this input.

Depends on #2850

There is an additional complication: What do we do when we get a `SIGINT` signal?

This PR sends a FAILURE result for async input and does _not_ suppress cancellation traceback. It also does nothing for sync inputs.

Alternatively, #2847 preserves existing behavior, sending a TERMINATED result for async inputs and suppressing cancellation traceback and doing nothing for sync inputs.

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -324,7 +324,7 @@ def call_function(
 
             user_code_event_loop.run(
                 run_concurrent_inputs(),
-                # on_sigint=lambda: container_io_manager.mark_all_inputs_as_cancelled(),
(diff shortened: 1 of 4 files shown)


thecodingwizard:
Discussed with @gongy: It's probably not a good idea to send `FAILURE` results on SIGINT, just in case the worker doesn't properly drop the result.

PR #2856, opened 2025-02-10
fix: raise InvalidError on invalid Sandbox CMDs

thundergolfer:
> I'm getting an occasional exception `TypeError: bad argument type for built-in operation` when calling modal_sandbox.exec(). Is this a known issue with any other async packages I may have installed? Something else? Relatively new to the async ecosystem, so
Partial traceback: ...

https://modal-com.slack.com/archives/C069RAH7X4M/p1739219264588429

The additional `ARG_MAX` thing is something I threw in when thinking of other ways users could pass us bad data 🙂. 

## Describe your changes

---

    - Doesn't involve interaction with servers.
    -  If we ever increase `ARG_MAX` then users will need to upgrade clients to be allow to send more than 2MiB in args.

---

</details>

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -42,8 +42,26 @@ from .stream_type import StreamType
 _default_image: _Image = _Image.debian_slim()
 
 
+# The maximum number of bytes that can be passed to an exec on Linux.
+# Though this is technically a 'server side' limit, it is unlikely to change.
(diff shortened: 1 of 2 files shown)


thundergolfer:
@prbot approve

pawalt (commented):
nice catch thanks

PR #2857, opened 2025-02-10
Don't send cloud provider enum to server

erikbern:
Just send the string. Also don't verify the string client-side – the verification happens server-side.

We could remove the `CLOUD_PROVIDER` enum in the client repo now (or at least make it empty). That requires changing some server code first though. But I can do it in a subsequent PR.

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -19,7 +19,6 @@ from synchronicity.exceptions import UserCodeException
 from modal_proto import api_pb2
 from modal_proto.modal_api_grpc import ModalClientModal
 
-from ._location import parse_cloud_provider
 from ._object import _get_environment_name, _Object, live_method, live_method_gen
 from ._pty import get_pty_info
 from ._resolver import Resolver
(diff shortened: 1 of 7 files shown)


modal/_functions.py
```diff
@@ -818,8 +813,7 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
                     task_idle_timeout_secs=container_idle_timeout or 0,
                     concurrency_limit=concurrency_limit or 0,
                     pty_info=pty_info,
-                    cloud_provider=cloud_provider,  # Deprecated at some point
-                    cloud_provider_str=cloud.upper() if cloud else "",  # Supersedes cloud_provider
+                    cloud_provider_str=cloud if cloud else "",
```
mwaskom, line 816: Just double-checking that the server doesn't expect uppercase values here anymore?

mwaskom (approved):
Mentioned a couple things offline but otherwise change looks good.

modal/_functions.py
erikbern: yeah we lower case it server side anyway

PR #2858, opened 2025-02-11
Have client CI/CD publish source distribution to PyPI as well

devennavani:
Have client CI/CD publish source distribution to PyPI as well.

I ran `python setup.py sdist bdist_wheel` locally and installed modal from the `.tar.gz` file and ran a couple of examples and everything worked.

This is needed because conda-forge needs a source distribution from PyPI to build the package.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -203,8 +203,9 @@ jobs:
       - name: Build type stubs
         run: inv type-stubs
 
-      - name: Build wheel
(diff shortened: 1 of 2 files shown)


freider:
Makes sense to me. Should move to `uv`/hatchling publishing/builds soon I hope to make all this a bit cleaner :)

devennavani:
@prbot approve

mwaskom (approved):
LGTM. Might be nice to clean up the packages that we include though.

PR #2860, opened 2025-02-11
Fix automount deprecations

freider:
Also fixes a couple of outstanding issues with the `include_source` option not working for:
1. Image.run_function
2. @build() methods

This also sets pending=False on the deprecation warning, so we need to fix internal integration tests before merging this as well.

---

---

</details>

### DIFF
--- a/test/cls_test.py
+++ b/test/cls_test.py
@@ -68,7 +68,7 @@ def test_run_class(client, servicer):
     assert len(servicer.precreated_functions) == 0
     assert servicer.n_functions == 0
     with app.run(client=client):
-        method_handle_object_id = Foo.bar.object_id  # method handle object id will probably go away
+        method_handle_object_id = Foo._get_class_service_function().object_id  # type: ignore
         assert isinstance(Foo, Cls)
         assert isinstance(NoParamsCls, Cls)
         class_id = Foo.object_id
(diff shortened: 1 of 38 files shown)


test/cls_test.py
```diff
@@ -68,7 +68,7 @@ def test_run_class(client, servicer):
     assert len(servicer.precreated_functions) == 0
     assert servicer.n_functions == 0
     with app.run(client=client):
-        method_handle_object_id = Foo.bar.object_id  # method handle object id will probably go away
+        method_handle_object_id = Foo._get_class_service_function().object_id  # type: ignore
```
mwaskom, line 71: Is the subsequent assertion actually doing anything useful with this change?

test/cls_test.py
```diff
@@ -705,7 +705,7 @@ def test_web_cls(client):
         assert c.asgi.web_url == "http://asgi.internal"
 
 
-handler_app = App("handler-app")
+handler_app = App("handler-app", include_source=True)
```
mwaskom, line 708: Do you intend not to remove this later or did you just get tired of writing the comment everywhere? 😁

mwaskom (approved):
LGTM if you can get the tests to pass without any warnings

test/cls_test.py
freider: It was one of the instances where my regex didn't trigger so I made the edit manually :D

PR #2861, opened 2025-02-11
Begin sending bundled autoscaler settings in Function definition

mwaskom:
## Describe your changes

This PR adds an `autoscaler_settings` field to the `Function` definition, using the new `AutoscalerSettings` type. It also starts sending the parameter values in this message (in duplicate with the existing fields directly on the `Function` message, since we don't have backend support to read these yet).

With this change we can use a consistent data structure for both static and dynamic configuration.

Part of CLI-31 

---

---

</details>


erikbern (approved):
might want to check this in a test but either way 👍

PR #2862, opened 2025-02-11
Add retry_count to get output response item

rculbertson:
We will use the retry_count field to distinguish between outputs resulting from different attempts. This will allow us to 
1)  ignore duplicate outputs
2) ignore outputs from old attempts which arrived after newer attempts. 

---

---

</details>


rculbertson:
@prbot approve

PR #2864, opened 2025-02-12
Adds include_source on run_function

freider:
Extracted as a separate patch from https://github.com/modal-labs/modal-client/pull/2860 so we can fix the integration tests before merging the patch that will break them

---


freider:
@prbot automerge

modal/image.py
```diff
@@ -1924,6 +1924,8 @@ def run_function(
         region: Optional[Union[str, Sequence[str]]] = None,  # Region or regions to run the function on.
         args: Sequence[Any] = (),  # Positional arguments to the function.
         kwargs: dict[str, Any] = {},  # Keyword arguments to the function.
+        *,
```
mwaskom, line 1927: Would be nicer to have this right after `raw_f` but we'll need to do it gradually.

freider: Agreed :(

PR #2865, opened 2025-02-12
Allow Function.web_url to return None when Function is not a web endpoint

mwaskom:
## Describe your changes

Fixes a false-positive error that made it annoying to programmatically inspect functions for web url metadata.

I thought about refining the condition to whether the object is unhydrated, but we have a `@live_method` decorator, so that shouldn't be possible?

That said I think we would like to get rid of "live properties" as they have some gotchas. So we may need to reinstate some sort of helpful error here (ideally one that doesn't get in the way when you're on the happy path). Left a comment to that effect.

---

---

</details>

## Changelog

- `Function.web_url` will now return None (instead of raising an error) when the Function is not a web endpoint

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1258,13 +1258,10 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
     # TODO (live_method on properties is not great, since it could be blocking the event loop from async contexts)
     @property
     @live_method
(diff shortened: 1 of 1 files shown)


freider:
I assume you mean "now" and not "not" in the changelog :D

freider (approved):

PR #2866, opened 2025-02-12
Remove cloud provider in client

erikbern:
This turns the enum into a dummy enum. We have a mapping server-side to parse old client enum values, but new clients don't use this enum.


mwaskom (approved):
Truthfully I lack a little context on everything CloudProvider was used for since it's a little archaic; LGTM as far as I can see though.

erikbern:
Thanks!

PR #2869, opened 2025-02-12
Migrate from legacy `setup.py`+`setup.cfg` to `pyproject.toml`-based packaging

devennavani:
This PR updates our build system to use modern Python packaging standards by:

* Migrating from `setup.cfg` and `setup.py` to `pyproject.toml`-based configuration
* Switching from `setuptools` direct usage to the standardized build package for creating distributions

Changes

* Moved all package metadata and configuration from `setup.cfg` to `pyproject.toml` using the `[project]` table format
* Removed `setup.py` in favor of dynamic version detection via `pyproject.toml` (see `tool.setuptools.dynamic`)
* Added explicit `pip install build` step in CI workflow to install Python's standardized build frontend: https://github.com/pypa/build
* Updated CI/CD workflow to use `python -m build` instead of `python setup.py sdist bdist_wheel`

All package specifications (dependencies, entry points, package data, etc.) remain functionally equivalent to the previous configuration.


erikbern (approved):
cool

pyproject.toml
```diff
@@ -2,6 +2,61 @@
...
+Homepage = "https://modal.com"
+Source = "https://github.com/modal-labs/modal-client"
+Documentation = "https://modal.com/docs"
+"Issue Tracker" = "https://github.com/modal-labs/modal-client/issues"
+
+[project.scripts]
+modal = "modal.__main__:main"
+
+[tool.setuptools.packages.find]
+include = ["modal", "modal.*", "modal_version", "modal_proto"]
```
freider, line 50: Just curious - Is the `modal.*` here necessary and/or sufficient? Intuitively it feels like either `modal` would capture all subpackages or `modal.*` would omit anything more deeply nested than direct descendants of modal (e.g. `modal._utils.function_utils`)

pyproject.toml
```diff
@@ -2,6 +2,61 @@
...
+exclude = ["test*", "modal_docs", "modal_global_objects"]
+
+[tool.setuptools.package-data]
+modal = ["requirements/*.md", "requirements/*.txt", "requirements/*.json", "py.typed", "*.pyi"]
+modal_proto = ["*.proto", "py.typed", "*.pyi"]
+
+[tool.setuptools.dynamic]
+version = {attr = "modal_version.__version__"}
```
mwaskom, line 58: Do we still need `modal_version` here? Seems like we could be dynamically getting the version from `modal.__version__` (with that getting populated by some auto-generated file that lives under `modal/`).

pyproject.toml
devennavani: Without the `modal.*`, the subpackages within `modal`, like `_runtime` or `_utils`, weren't being included in the installation/build

mwaskom: Yeah it's super weird that `modal` is not recursive but `modal.*` is 😵‍💫

pyproject.toml
devennavani: Yeah we can probably eliminate the need for the separate package -- that change has a couple of touch points so I'll make it in a separate PR.

PR #2870, opened 2025-02-13
Export modal.FunctionCall at the top level for FunctionCall.from_id()

freider:
Noticed this was missing in the package level exports, but feels like it should be there since we have some public uses of this API in our examples

## Changelog

- `modal.FunctionCall` is now available in the top-level `modal` namespace. We recommend referencing the class this way instead of using the the fully-qualified `modal.functions.FunctionCall` name.

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -18,7 +18,7 @@ try:
     from .dict import Dict
     from .exception import Error
     from .file_pattern_matcher import FilePatternMatcher
-    from .functions import Function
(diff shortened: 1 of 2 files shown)


mwaskom (approved):
Yep I agree with this. Can we update docs / example usage? I see a reference to the fully-qualified name in the `spawn` / `_experimental_spawn` docstrings.

mwaskom (approved):

PR #2871, opened 2025-02-14
Include `modal_docs` in client installation/build

devennavani:
Include `modal_docs` in client installation/build, since our monorepo CI/CD installs the client and uses `modal_docs` for `inv docs`

PR #2872, opened 2025-02-14
Include modal_docs subpackages in client installation/build

devennavani:
Include `modal_docs` subpackages in client installation/build. https://github.com/modal-labs/modal-client/pull/2871 was unfortunately not sufficient


devennavani:
@prbot approve

mwaskom (commented):
👍 we should probably still remove this once we have a better way of incorporating the client into the monorepo build but no harm in putting it back for now

PR #2873, opened 2025-02-14
Removes double underscored deprecated field

freider:
Should be unused now (no references in code base).
This breaks type checking on mypy-proto 3.4+ otherwise


mwaskom (approved):
LGTM

PR #2876, opened 2025-02-14
Remove check for snapshot enabled locally

cweld510:
Removes a local check in `Sandbox._experimental_snapshot` that verifies snapshots are enabled for the sandbox before snapshotting. We don't correctly hydrate the field we're checking correctly, and we also verify on the server.

This fixes the issue quickly, for now. I've created a ticket to add back the local check (and add a test) once I figure out hydration.

## Changelog

- Fixes a bug where sandboxes returned from `Sandbox.list()` were not snapshottable even if they were created with `_experimental_enable_snapshot`.


cweld510:
@prbot approve

cweld510:
@prbot automerge

thundergolfer (commented):
LGTM

PR #2877, opened 2025-02-18
Revert "Removes double underscored deprecated field"

howard36:
Reverts modal-labs/modal-client#2873


ekzhang (approved):

ekzhang:
note: causing an active incident, investigating why it didn't fail tests

PR #2878, opened 2025-02-18
Start to remove a bunch of complicated 0.63 modal.Cls code

freider:
Background:
* For v0.62 lookup compatibility, we used to create `_Function` objects for every method.
* When we deprecated v0.62, we still had this code in place even though the class-bound method was no longer backed by any "real" backend function. In the client it mostly served as an information carrier to let us hydrate the (client-only) *instance bound* _Function.

This PR removes the need for this complicated information carrier, reducing the risk of us accidentally calling the class-bound method.

This will significantly help with further simplification of the class code, and once 0.66 is deprecated we should be able to remove further complicated code as the method metadata will then always be served directly by the class service function rather than the Cls.

TODO:
- [x] Test that this works with lookups of pre-0.67 classes
- [x] Run integration tests

---

---

</details>

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -362,8 +362,9 @@ class _Cls(_Object, type_prefix="cs"):
     _method_metadata: Optional[dict[str, api_pb2.FunctionHandleMetadata]] = None
 
     # These are only set where source is locally available:
-    _user_cls: Optional[type]
-    _method_partials: dict[str, _PartialFunction]
+    # TODO: wrap these in a single optional/property for consistency
+    _user_cls: Optional[type] = None
+    _method_partials: Optional[dict[str, _PartialFunction]] = None
     _callables: dict[str, Callable[..., Any]]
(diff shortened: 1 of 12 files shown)


modal/cls.py
```diff
@@ -123,12 +115,19 @@ def _deps():
...
+            partial_function.raw_f,
+            user_cls=cls._user_cls,
+            serialized=service_function.info.is_serialized(),
+        )
+
     fun._obj = service_function._obj
     fun._is_method = True
-    fun._app = class_bound_method._app
-    fun._spec = class_bound_method._spec
-    fun._is_web_endpoint = class_bound_method._is_web_endpoint
```
freider, line 131: _is_web_endpoint is a method, so this was a weird assignment...

erikbern:
Nice!!

devennavani (approved):
lgtm!

PR #2880, opened 2025-02-19
add some runtime type-checking in Image.env

thundergolfer:
## Describe your changes

Action taken in response to user report: https://modal-com.slack.com/archives/C069RAH7X4M/p1739956895311459

But if feels a bit piecemeal? Don't really want to spread lots of runtime type-checking across the code just for the purposes of nicer error messages. Alternatively I've put up a docs change on the other repo (#20024)

---

---

</details>

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -2011,13 +2011,12 @@ class _Image(_Object, type_prefix="im"):
         )
         ```
         """
+        non_str_keys = [key for key, val in vars.items() if not isinstance(val, str)]
+        if non_str_keys:
+            raise InvalidError(f"Image ENV variables must be strings. Invalid keys: {non_str_keys}")
 
         def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
-            try:
-                env_commands = [f"ENV {key}={shlex.quote(val)}" for (key, val) in vars.items()]
-            except TypeError:
-                non_str_keys = [key for key, val in vars.items() if not isinstance(val, str)]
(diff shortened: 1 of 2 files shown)


modal/image.py
```diff
@@ -2013,8 +2013,12 @@ def env(self, vars: dict[str, str]) -> "_Image":
         """
 
         def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
-            commands = ["FROM base"] + [f"ENV {key}={shlex.quote(val)}" for (key, val) in vars.items()]
-            return DockerfileSpec(commands=commands, context_files={})
+            try:
+                env_commands = [f"ENV {key}={shlex.quote(val)}" for (key, val) in vars.items()]
+            except TypeError:
+                non_str_keys = [key for key, val in vars.items() if not isinstance(val, str)]
+                raise InvalidError(f"Environment variables must be strings. Invalid keys: {non_str_keys}")
```
mwaskom, line 2020: Is there a reason you're deferring this till after catching a `TypeError`? One advantage to doing it immediately (i.e. not inside the `build_dockerfile` closure) is that the error will get attributed to the relevant line of user code; raising from the closure produces a very long and basically uninterpretable traceback.

I guess the downside is that code outside of the closure will execute on container cold starts which is undesirable. Shouldn't be too costly in most cases .... but every millisecond counts ;)

mwaskom (approved):
I think these sort of runtime type assertions have to feel piecemeal in Python, but adding one here makes sense as this is an easy mistake to make, and the current error message is not very interpretable.

Would be great to have a small unit test for this!

PR #2882, opened 2025-02-20
Remove/move is_local checks to make it possible to deploy apps with mounts from within running functions

freider:
iirc, the checks were mostly in place to prevent unnecessary logic from running in global scope of containers, but just adding pre-defined mounts to a list wouldn't account for many clock cycles, and this change still prevents auto-mounting from running within containers (soon to be deprecated anyways)

Fixes: CLI-190

---

---

</details>

## Changelog

* Fixes an issue where `modal.runner.deploy_app()` didn't work when called from within a running (remote) Modal Function

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -473,51 +473,41 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
 
         explicit_mounts = mounts
 
-        if is_local():
-            include_source_mode = get_include_source_mode(include_source)
-            if include_source_mode != IncludeSourceMode.INCLUDE_NOTHING:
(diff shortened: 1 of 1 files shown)


modal/_functions.py
```diff
@@ -473,51 +473,41 @@ def from_local(
 
         explicit_mounts = mounts
 
-        if is_local():
```
freider, line 476: This is just an unindent, with the `is_local()` getting moved to the inner auto mounts `if` condition

mwaskom (approved):
If we have an existing fixture for making unit tests appear to be running in a container, might be nice to assert on the new behavior. But thanks for addressing it!

PR #2883, opened 2025-02-20
Allow app.run inside of functions

freider:
In the same vein as #2882 

The intent of it was to prevent usage of app.run in global scope of modules that would then get loaded in containers causing an infinite recursion. This is not as likely to happen anymore now that we have the `modal run` cli instead of `if __name__ == __main__` as the main app launcher mechanic

---

</details>

## Changelog

* Allow running new ephemeral apps from **within** Modal containers using `with app.run(): ...`. Use with care, as putting such a run block in global scope of a module could easily lead to infinite app creation recursion


mwaskom (approved):

mwaskom:
>  This is not as likely to happen anymore now that we have the modal run cli

One thing that I might be missing is why the error message mentions `modal run` if the existence of `modal run` obviates the error message. But I'm on board with dropping this guardrail.

freider:
> > This is not as likely to happen anymore now that we have the modal run cli
> 
> One thing that I might be missing is why the error message mentions `modal run` if the existence of `modal run` obviates the error message. But I'm on board with dropping this guardrail.

I think it was intended to nudge people towards using `modal run` instead of having `with app.run():`  blocks in their app files

mwaskom:
Does this need a rebase?

freider:
> We could in theory set some global variable during code imports to prevent this case more selectively. Not super important though.

Good idea, I just applied that to the PR

freider:
Updated this a bit with a test (turns out there was a bug...), removing the old `is_local()` check in `from_local_python_packages` which would otherwise prevent the entrypoint mount itself from getting mounted inside containers.

I think the `is_local()` check for automounts in functions.py _should_ be enough to prevent noticable performance penalties from mount constructions inside containers anyway?

mwaskom:
My non-blocking comment here is that I'd prefer risking some users surprising themselves over introducing more complexity with this global variable / context manager.

PR #2885, opened 2025-02-20
Rename autoscaler parameters and add deprecation warnings

mwaskom:
## Describe your changes

Introduces the new autoscaler parameters and adds a deprecation warning on usage of the old names (currently `pending=True` so that we can update docs and integration tests).

- Part of CLI-31
 
## Changelog

- We're renaming several `App.function` and `App.cls` parameters that configure the behavior of Modal's autoscaler:
    - `concurrency_limit` is now `max_containers`
    - `keep_warm` is now `min_containers`
    - `container_idle_timeout` is now `scaledown_window`
- The old names will continue to work, but using them will issue a deprecation warning. The aim of the renaming is to reduce some persistent confusion about what these parameters mean. Code updates should require only a simple substitution of the new name.
- We're adding a new parameter, `buffer_containers` (previously available as `_experimental_buffer_containers`). When your Function is actively handling inputs, the autoscaler will spin up additional `buffer_containers` so that subsequent inputs will not be blocked on cold starts. When the Function is idle, it will still scale down to the value given by `min_containers`.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -29,7 +29,12 @@ from ._partial_function import (
     _PartialFunctionFlags,
 )
 from ._utils.async_utils import synchronize_api
-from ._utils.deprecation import deprecation_error, deprecation_warning, renamed_parameter
+from ._utils.deprecation import (
+    deprecation_error,
+    deprecation_warning,
+    renamed_parameter,
+    warn_on_renamed_autoscaler_settings,
+)
 from ._utils.function_utils import FunctionInfo, is_global_object, is_method_fn
 from ._utils.grpc_utils import retry_transient_errors
 from ._utils.mount_utils import validate_volumes
@@ -559,6 +564,7 @@ class _App:
 
         return wrapped
 
+    @warn_on_renamed_autoscaler_settings
     def function(
         self,
         _warn_parentheses_missing: Any = None,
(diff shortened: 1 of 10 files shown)


modal/app.py
```diff
@@ -586,17 +592,14 @@ def function(
         # Or, pass (request, limit) to additionally specify a hard limit in MiB.
         memory: Optional[Union[int, tuple[int, int]]] = None,
         ephemeral_disk: Optional[int] = None,  # Specify, in MiB, the ephemeral disk size for the Function.
+        min_containers: Optional[int] = None,  # Minimum number of containers to keep warm, even when Function is idle.
+        max_containers: Optional[int] = None,  # Limit on the number of containers that can be spun up for this Function
```
freider, line 596: Docs/comment should probably say something about *concurrent* containers

modal/app.py
```diff
@@ -586,17 +592,14 @@ def function(
         # Or, pass (request, limit) to additionally specify a hard limit in MiB.
         memory: Optional[Union[int, tuple[int, int]]] = None,
         ephemeral_disk: Optional[int] = None,  # Specify, in MiB, the ephemeral disk size for the Function.
+        min_containers: Optional[int] = None,  # Minimum number of containers to keep warm, even when Function is idle.
```
freider, line 595: In a way it could be nice if these shared a common prefix (instead of suffix) for in-editor discoverability when typing, but `containers_min` sounds so much worse that I don't know what's best...

modal/app.py
```diff
@@ -586,17 +592,14 @@ def function(
...
+        max_containers: Optional[int] = None,  # Limit on the number of containers that can be spun up for this Function
+        buffer_containers: Optional[int] = None,  # Number of additional idle containers to maintain under active load
+        scaledown_window: Optional[int] = None,  # Max amount of time a container can remain idle before scaling down.
```
freider, line 598: Not sure about this name, but the might have already been a discussion on that elsewhere?

"scaledown" feels like something more global imo, something that affects the whole pool of containers, which is only indirectly/partially true in this case.

Related - what happens if you adjust the scaledown window using `update_autoscaler` on a function with live containers - will the existing running containers have their scaledown_window updated or will they still respect the value from when they started up? (don't remember how this is implemented)

modal/cls.py
```diff
@@ -212,6 +212,41 @@ def _new_user_cls_instance(self):
...
         return user_cls_instance
 
+    async def update_autoscaler(
+        self,
+        min_containers: Optional[int] = None,
+        max_containers: Optional[int] = None,
+        scaledown_window: Optional[int] = None,
+        buffer_containers: Optional[int] = None,
+    ) -> None:
+        """Override the current autoscaler behavior for this Cls instance.
```
freider, line 222: Should we start calling a "Cls instance" a Service soon? So much easier to pronounce

freider (commented):
Some comments/questions

modal/app.py
mwaskom: Yeah we discussed extensively, it's probably my least favorite of the names here, but at this point we've made all the backend changes and I think we should value consistent frontend / backend terminology in the absence of a clear alternative.

modal/cls.py
mwaskom: Maybe? I still don't think I understand what the plan is for changes to the client object model.

modal/app.py
mwaskom: Yeah `container_floor` / `container_ceiling` is one way you could spell that but it would be sort of bizarre.

Most users won't be typing though!

modal/app.py
mwaskom: FWIW the primary motivation here was to avoid confusion arising from the fact that we don't always honor the full window as a "timeout"; containers aren't guaranteed to stay up for x seconds after becoming idle, rather, they are guaranteed not to be idle for more than x seconds.

freider (approved):
lgtm!

mwaskom:
Splitting the `.update_autoscaler` changes out to a subsequent PR as we have some open questions about it.

PR #2889, opened 2025-02-21
Quick fix for accidentally *fixing* @build deps

freider:
In version v0.73.55 we accidentally fixed one half of a bug that previously would not add dependency objects to @build methods of classes.

Fixing the other half of the bug, which is that the dependencies are not included in the app layout as provided by the workers seems a bit more involved, so for now this just makes the two dependency lists consistent (but still incorrect, causing image.imports to not work properly in @build functions)

---

---

</details>

### DIFF
--- a/modal/_runtime/user_code_imports.py
+++ b/modal/_runtime/user_code_imports.py
@@ -264,7 +264,9 @@ def import_single_function_service(
                 # The cls decorator is in global scope
                 _cls = synchronizer._translate_in(cls)
                 user_defined_callable = _cls._callables[fun_name]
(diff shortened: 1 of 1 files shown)


aksh-at (approved):

aksh-at:
Would be nice to add a test for this too, but merging this to fix for now.

freider:
Yeah I was thinking of adding a test, but also feels a bit wrong to add a test for incorrect behavior.
I'll make a follow up PR on Monday with an integration test, since this has to do with the surface between worker and client

PR #2891, opened 2025-02-21
Add `ignore_cache` config field

mwaskom:
## Describe your changes

Adds a new field to the config, `ignore_cache`. This will be included in the `ImageGetOrCreateRequest`. It will effectively ignore the Image cache (in practice, it will include the App ID and version in the cache key, although the backend implementation could change). This can be useful for testing whether Apps can run when rebuilding an Image from top-to-bottom without then forcing other Apps that depend on the same base layer(s) to rebuild.

---

---

</details>

## Changelog

- Adds a new config field, `ignore_cache` (also accessible via environment variables as `MODAL_IGNORE_CACHE=1`), which will force Images used by the App to rebuild but not clobber any existing cached Images. This can be useful for testing an App's robustness to Image rebuilds without affecting other Apps that depend on the same base Image layer(s).

### DIFF
--- a/modal/config.py
+++ b/modal/config.py
@@ -218,7 +218,7 @@ _SETTINGS = {
     "worker_id": _Setting(),  # For internal debugging use.
     "restore_state_path": _Setting("/__modal/restore-state.json"),
     "force_build": _Setting(False, transform=_to_boolean),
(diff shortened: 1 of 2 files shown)


charlesfrye:
nice work!

PR #2893, opened 2025-02-24
Draft/RFC: prep payload serialization code

freider:
Introduces a new form of payload serialization based on a proto structure (inspired by the Modal hackathon format), which could replace the current `pickle((args, kwargs))`.

Rough goals of this:
* Incorporate some of the structural proto changes/extensions needed for "container types" like lists and dicts
* Allow for optional separation of _serialization/deserialization_ and _validation_ logic, as opposed to the current class parameter serialization which entangles these concepts (class parameters *require* strict schemas via type annotations at the moment)
* Refactor code to reuse as much as possible between the "payload" and "parameter" serialization paths. In the future, we could migrate the current class parameter serialization format to this new payload format instead, but such a migration would have some forward/backward compatibility considerations since it would be a change to the serialized bytes etc.

As this PR stands, it shouldn't have any effect on actual live apps, but if we feel this is a good approach a next step would be to add support for *decoding* of these payload types into the actual client such that any functions deployed with this would be callable using proto payloads.

---

---

</details>

### DIFF
--- a/modal/_serialization.py
+++ b/modal/_serialization.py
@@ -393,7 +393,8 @@ def check_valid_cls_constructor_arg(key, obj):
 class ParamTypeInfo:
     default_field: str
     proto_field: str
-    converter: typing.Callable[[str], typing.Any]
+    encoder: typing.Callable[[Any], Any]
+    decoder: typing.Callable[[Any], Any]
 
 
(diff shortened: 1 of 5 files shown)


freider:
Hold on, bringing in the some

PR #2895, opened 2025-02-24
Exit with a non-zero exit code if we have unrecoverable errors

pawalt:
## Describe your changes

This commit propagates non-zero exit codes upwards so the runner of this code knows that the container did not exit successfully.

---

---

</details>

### DIFF
--- a/modal/_container_entrypoint.py
+++ b/modal/_container_entrypoint.py
@@ -593,12 +593,14 @@ if __name__ == "__main__":
     # The only caveat is a bunch of calls will now cross threads, which adds a bit of overhead?
     client = Client.from_env()
 
+    exited_successfully = True
     try:
         with proxy_tunnel(container_args.proxy_info):
             try:
                 main(container_args, client)
             except UserException:
                 logger.info("User exception caught, exiting")
+                exited_successfully = False
     except KeyboardInterrupt:
         logger.debug("Container: interrupted")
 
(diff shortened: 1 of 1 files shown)


pawalt:
I'm not sure of the proper way to test this change, although I feel it should be tested.

mwaskom:
I think if you look in `test/container_test.py` it should have the fixtures that you need.

modal/_container_entrypoint.py
```diff
@@ -593,12 +593,14 @@ def breakpoint_wrapper():
...
     client = Client.from_env()
 
+    exited_successfully = True
     try:
         with proxy_tunnel(container_args.proxy_info):
             try:
                 main(container_args, client)
             except UserException:
                 logger.info("User exception caught, exiting")
+                exited_successfully = False
```
mwaskom, line 603: Maybe slightly cleaner in an `else:` block?

erikbern: why not just re-raise

erikbern: that way you don't need any of the other changes, including the `exited_successfully` variable and the `exit(1)` call

pawalt:
After chatting with @cweld510 we decided to not do this right now. We'll more carefully consider these semantics at a later time, but this isn't the right moment. We need to come up with a better model for what it means when tasks shut down. I've cut a linear ticket for this work.

PR #2896, opened 2025-02-24
Add `modal.FunctionCall.gather` to replace `modal.functions.gather`

mwaskom:
## Describe your changes

Moving `gather` to be a staticmethod on `modal.FunctionCall` makes it more discoverable and clarifies its purpose. Because we now import `modal.FunctionCall` into the top-level namespace, we no longer need to recommend using names directly from `modal.functions`, which helps enforce a clearer separation between public / private API.

- Closes CLI-334

---

---

</details>

## Changelog

- We've moved the `modal.functions.gather` function to be a staticmethod on `modal.FunctionCall.gather`. The former spelling has been deprecated and will be removed in a future version.

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1597,26 +1597,35 @@ class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
         fc._is_generator = is_generator
         return fc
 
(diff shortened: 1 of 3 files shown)


freider (approved):
I like it!

PR #2899, opened 2025-02-25
Eliminate some unnecessary nesting in the reference docs

mwaskom:
Now that we're pulling `FunctionCall` up to the top-level namespace, we should also include it in the sidebar of the reference docs rather than hiding it on the page that you get to if you click `modal.Function`.

While looking into this, I noticed that a few other objects were the only public object defined in their module, but we were for some reason configuring their reference to represent the module. This doesn't seem necessary; I removed it so that the references docs are more consistently formatted across types.

e.g., before:

image

and after:

image

Also, hey buddy 👋 

image

### DIFF
--- a/modal_docs/gen_reference_docs.py
+++ b/modal_docs/gen_reference_docs.py
@@ -76,20 +76,14 @@ def run(output_dir: str = None):
 
     base_title_level = "#"
     forced_module_docs = [
-        ("modal.Function", "modal.functions"),
-        ("modal.Secret", "modal.secret"),
-        ("modal.Dict", "modal.dict"),
-        ("modal.Queue", "modal.queue"),
         ("modal.call_graph", "modal.call_graph"),
         ("modal.gpu", "modal.gpu"),
         ("modal.runner", "modal.runner"),
-        ("modal.Sandbox", "modal.sandbox"),
-        ("modal.ContainerProcess", "modal.container_process"),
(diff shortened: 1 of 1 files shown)


modal_docs/gen_reference_docs.py
```diff
@@ -76,20 +76,14 @@ def top_level_filter(module, name):
...
         ("modal.gpu", "modal.gpu"),
         ("modal.runner", "modal.runner"),
-        ("modal.Sandbox", "modal.sandbox"),
-        ("modal.ContainerProcess", "modal.container_process"),
         ("modal.io_streams", "modal.io_streams"),
         ("modal.FileIO", "modal.file_io"),
     ]
     # These aren't defined in `modal`, but should still be documented as top-level entries.
-    forced_members = {"web_endpoint", "asgi_app", "method", "wsgi_app", "forward"}
+    forced_members: set[str] = set()
```
ekzhang, line 86: Is this change intentional? Just checking that `web_endopint`, `asgi_app` and so on should still be documented somewhere.

ekzhang (approved):

modal_docs/gen_reference_docs.py
mwaskom: Yeah, I'm not sure about the context here, but these names are all in the top-level modal namespace now and get documented on my branch: 
image

PR #2900, opened 2025-02-25
Revert "Eliminate some unnecessary nesting in the reference docs (#2899)

mwaskom:
Broke the frontend build because some hardcoded cross-references were no longer valid.


mwaskom:
Seemed like there wasn’t actually any way to do it gracefully.

PR #2902, opened 2025-02-26
increase max connections to respect high allow_concurrent_inputs

samaross:
## Describe your changes

---

---

</details>


2timesjay:
1000 max is probably great for now, but can this be increased further if LLM applications running on B200s can reasonably see better performance with > 1000 max concurrent requests?

gongy:
Definitely want to revisit those existing limits when the need arises!

gongy (approved):

mwaskom:
> Definitely want to revisit those existing limits when the need arises!

May be tough to keep the client in sync with any changes made to the internal limits; can we at least add a comment about this in the server code?

PR #2906, opened 2025-02-27
Deprecate workspace from Cls.from_name and Cls.lookup

mwaskom:
## Describe your changes

This parameter is leftover from when you used to be able deploy / lookup classes to the global namepsace. The field on the proto is no longer read in the server.

---

---

</details>

## Changelog

- Adds a deprecation warning to the `workspace` parameter in `modal.Cls` lookup methods. This argument is unused and will be removed in the future.

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -504,7 +504,7 @@ class _Cls(_Object, type_prefix="cs"):
         name: str,
         namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
         environment_name: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


freider:
great! I think we could also just remove it completely and it probably wouldn't affect anyone ?

freider (approved):

mwaskom:
We can probably remove it soon (and just go straight to dropping it instead of upgrading to a deprecation error) but was alerted to it by a user question so at least someone has been using it / thinking it works (it does if you pass the workspace you're auth'd into!)

PR #2907, opened 2025-02-27
Allow Cls.with_options to lazily hydrate self and dependencies

mwaskom:
## Describe your changes

Currently `Cls.with_options` has a couple rough edges around lazy hydration:

- The Cls itself must be hydrated, so naively migrating `Cls.lookup(...).with_options(...)` to `Cls.from_name` doesn't work
- Any Volume or Secret objects that are passed are assumed to be hydrated

This change feels slightly hacky, but it appears to make simple examples work!

Fixes CLI-37

## Changelog

- `modal.Cls.with_options` now supports lazy hydration.

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -1,5 +1,6 @@
 # Copyright Modal Labs 2022
 import inspect
+import itertools
 import os
 import typing
 from collections.abc import Collection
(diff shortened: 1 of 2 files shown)


mwaskom:
Persuaded that we want this to be even lazier.

PR #2909, opened 2025-02-28
Lazy with_options

freider:
Makes `Cls.with_options()` fully lazy

This build on top of #2908 

---

---

</details>

## Changelog

- It is now possible to call `Cls.with_options` on an unhydrated Cls, e.g.

    ```python
    ModelWithGPU = modal.Cls.from_name("my-app", "Model").with_options(gpu="H100")
    ```

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -980,27 +980,30 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
             environment_name = _get_environment_name(None, resolver)
             assert parent is not None and parent.is_hydrated
 
-            volume_mounts = [
-                api_pb2.VolumeMount(
-                    mount_path=path,
-                    volume_id=volume.object_id,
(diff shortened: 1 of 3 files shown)


mwaskom (approved):
Don't totally understand it, but happy to see it anyway :D

freider:
> Don't totally understand it, but happy to see it anyway :D

We used to do a `.clone()` on the Cls eagerly to return a "copy" with the new options without having to load anything, but clone requires the target to already be hydrated. Now instead, we use the general `_from_loader()` constructor mechanics to "postpone" the hydration (also just using the parent Cls data, but then we can be "sure" that the parent is hydrated when load is called).

The slightly complicated thing is that I had to add conditions inside of "deps" to only include the "parent" cls and service function if they aren't already hydrated, otherwise the lazy resolve breaks when applied to a "local"/app-linked function (that doesn't have lazy loading internals)

PR #2913, opened 2025-03-02
Adds more verbose reprs in a couple of places

freider:
Mini cleanup

---

---

</details>

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -563,7 +563,7 @@ class _Cls(_Object, type_prefix="cs"):
             await resolver.load(self._class_service_function)
             self._hydrate(response.class_id, resolver.client, response.handle_metadata)
 
-        rep = f"Ref({app_name})"
+        rep = f"Cls.from_name({app_name!r}, {name!r})"
         cls = cls._from_loader(_load_remote, rep, is_another_app=True, hydrate_lazily=True)
 
         class_service_name = f"{name}.*"  # special name of the base service function for the class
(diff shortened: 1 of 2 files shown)


mwaskom:
I feel like we need some consistent rules about what to use for reprs; it's pretty ad hoc and inconsistent at the moment. I'm not sure that I would expect the repr to depend on what alt-constructor was used (although I could see the argument for it when the API heavily depends on alt-constructors, I guess.

freider:
> I feel like we need some consistent rules about what to use for reprs; it's pretty ad hoc and inconsistent at the moment. I'm not sure that I would expect the repr to depend on what alt-constructor was used (although I could see the argument for it when the API heavily depends on alt-constructors, I guess.

Yeah we should come up with something. This was mostly to remove the legacy "Ref" repr that doesn't makes sense at all. But having the reprs be roughly equivalent to the constructor/builder syntax is kind of nice IMO, to let people identify which object they are looking at

mwaskom:
Still kinda meh on this tbh, but we can revisit.

freider:
Yeah we can definitely revisit, the repr shouldn't really affect anyone except us and I mostly patched this since it helped me debug some internal issues more easily

mwaskom:
> the repr shouldn't really affect anyone except us

I don't think that's always true? Users will print objects to figure out what they are (and they get returned back to the REPL if you're doing something interactive). I sort of have hopes that we would include _more_ useful metadata in the repr for our objects. But since they're meant to be human-readable and not machine-readable, it's easier to change later.

I see the argument that `repr()` should reflect the alt-constructor syntax, since Python says "This representation aims to be unambiguous and, ideally, should be able to recreate the object using the representation as an argument to the eval() function.". So maybe it just looks weird since alt-constructors aren't very commonly used (or when they're used, they often can resolve down to a primary-constructor spelling). But that also kind of strikes me as a mindset that makes more sense in the original "Python is primarily a teaching language" application. Wouldn't suggest people do `eval(repr(obj))` these days ;)

PR #2914, opened 2025-03-03
Add app_id to SandboxInfo protobuf

howard36:
## Describe your changes

Part of PRD-694. After this change, `SandboxGet` and `SandboxList` can return the sandbox's app ID.

---


mwaskom (approved):
lgtm

PR #2915, opened 2025-03-03
Add runsc_runtime_version to CheckpointInfo proto

cweld510:
Required for migrating Function memory snapshots to use the new runtime versioning system.


cweld510:
@prbot approve

PR #2916, opened 2025-03-03
Remove wait_for_response parameter from web endpoint decorators

mwaskom:
We added a deprecation warning for this in May 2024 and upgraded it to an error in October 2024. Removing it will simplify the `web_endpoint` -> `fastapi_endpoint` transition a bit.

---

---

</details>

### DIFF
--- a/modal/_partial_function.py
+++ b/modal/_partial_function.py
@@ -16,7 +16,7 @@ from modal_proto import api_pb2
 
 from ._functions import _Function
 from ._utils.async_utils import synchronizer
-from ._utils.deprecation import deprecation_error, deprecation_warning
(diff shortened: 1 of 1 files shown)


erikbern (approved):

Balmasexy:
Hello world 🌎🌎 am here to introduce the invincible world individual explorer to You All.l,
I purposely revert the Metadata ton token back  and exchange where the Database and if I have to return Back the ton ecosystem metadata there must be a say to it through the community on Stack so I will release it,
warning ⚠️ never Abuse or say a negative words thanks for your support any reply goes to X or Stark waiting for reply........ @Balmasexy111.

PR #2917, opened 2025-03-03
Add `@modal.fastapi_endpoint` to replace `@modal.web_endpoint`

mwaskom:
## Describe your changes

Adds `modal.fastapi_endpoint` as an alias for `modal.web_endpoint`. (Actually I just duplicated the code; there was enough fussiness in the error messages that it was simpler to do it that way).

I migrated most tests to use `fastapi_endpoint`, while leaving a few parameterized across both spellings.

These can coexist for a little while as we update the docs etc., before `@modal.web_endpoint` begins issuing deprecation warnings.

We're doing the renaming to make it more obvious that the feature depends on `fastapi` (which must be explicitly installed in the Image as of image builder version 2024.10) and to free up the `web_endpoint` name so that we can potentially introduce a lightweight / dependency-free version of the functionality in the future.

- Part of CLI-339
 
---

---

</details>

## Changelog

- Adds `modal.fastapi_endpoint` as an alias for `modal.web_endpoint`. We will be deprecating the `modal.web_endpoint`  _name_ (but not the functionality) as part of the Modal 1.0 release.

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -23,7 +23,18 @@ try:
     from .mount import Mount
     from .network_file_system import NetworkFileSystem
     from .output import enable_output
-    from .partial_function import asgi_app, batched, build, enter, exit, method, web_endpoint, web_server, wsgi_app
+    from .partial_function import (
+        asgi_app,
+        batched,
+        build,
+        enter,
+        exit,
+        fastapi_endpoint,
+        method,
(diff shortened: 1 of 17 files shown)


mwaskom:
@prbot approve

erikbern (approved):
Gr8

PR #2918, opened 2025-03-03
Remove TaskResult

erikbern:
I don't think this is used anywhere??


erikbern:
actually it's used, i'm stupid

PR #2919, opened 2025-03-04
Drop keep_warm from `modal.method` signature

mwaskom:
## Describe your changes

Removes the `keep_warm=` parameter, which only existed to trip a deprecation warning (it was nonfunctional since we can't maintain a warm pool for a specific method since v0.63.0.

I decided to just remove it rather than upgrade it to an error since the deprecation would involve a double-migration (move to the Cls _and_ use the new `min_containers` name) which felt like it risked causing more confusion than it abated.

---

---

</details>

## Changelog

- The `keep_warm=`  parameter has been removed from the`@modal.method` decorator. This parameter has been nonfunctional since v0.63.0; all autoscaler configuration must be done at the level of the modal Cls.

### DIFF
--- a/modal/_partial_function.py
+++ b/modal/_partial_function.py
@@ -63,8 +63,8 @@ class _PartialFunction(typing.Generic[P, ReturnType, OriginalReturnType]):
     def __init__(
         self,
         raw_f: Callable[P, ReturnType],
-        *,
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot approve

erikbern (approved):

PR #2921, opened 2025-03-04
Run CI on ubuntu 24.04

mwaskom:
We've been getting brownouts on the 20.04 platform, which is being deprecated on April 1.


mwaskom:
@prbot approve

happy to iterate more if there's a better way to refresh the dependency cache but this runs so i want to unblock other prs

ekzhang (commented):
Yay!!

PR #2924, opened 2025-03-04
Provide the version where we added some 1.0 APIs in their docstrings

mwaskom:
Helps to address drift between our docs (which mostly demonstrate up-to-date patterns) and users who may have an older client installed.

This is relatively easy to do for entire functions / methods. It's harder for new parameters, because our approach to parameter documentation doesn't leave much room for exposition. That's something I'd like to fix independently...


mwaskom:
@prbot approve

thundergolfer (commented):
LGTM

PR #2931, opened 2025-03-07
Add more debug logging to `retry_transient_errors`

aksh-at:
Trying to figure out this failure within `retry_transient_errors` that we don't have a repro for.

---

---

</details>


aksh-at:
@prbot approve

gongy (commented):
🆗

PR #2934, opened 2025-03-10
[PRD-755] Record git commit info on deploy

gleb-modal:
## Describe your changes

https://linear.app/modal-labs/issue/PRD-638/display-git-hash-if-possible-in-deployment-history

Related PR in the modal repo: https://github.com/modal-labs/modal/pull/20637

## Changelog

* Include git commit info at the time of app deployment.

### DIFF
--- a/modal/_utils/git_utils.py
+++ b/modal/_utils/git_utils.py
@@ -41,11 +41,17 @@ def get_git_commit_info() -> Dict[str, Union[str, bool, int]]:
     branch = run_command(["git", "rev-parse", "--abbrev-ref", "HEAD"])
     if branch:
         git_info["branch"] = branch
+    else:
+        # If there's an error getting the branch name, bail
+        return {}
 
     # Get commit hash
     commit_hash = run_command(["git", "rev-parse", "HEAD"])
     if commit_hash:
         git_info["commit_hash"] = commit_hash
+    else:
+        # If there's an error getting the commit hash, bail
+        return {}
 
     # Check if working directory is dirty
     dirty_output = run_command(["git", "status", "--porcelain"])
@@ -55,5 +61,15 @@ def get_git_commit_info() -> Dict[str, Union[str, bool, int]]:
     repo_url = run_command(["git", "remote", "get-url", "origin"])
     if repo_url:
         git_info["repo_url"] = repo_url
+
+    # Get author name
+    author_name = run_command(["git", "show", "-s", "--format=%an", "HEAD"])
+    if author_name:
+        git_info["author_name"] = author_name
+
+    # Get author email
+    author_email = run_command(["git", "show", "-s", "--format=%ae", "HEAD"])
+    if author_email:
+        git_info["author_email"] = author_email
 
     return git_info
--- a/modal/_utils/git_utils.py
+++ b/modal/_utils/git_utils.py
@@ -20,7 +20,7 @@ async def run_command_fallible(args: list[str]) -> Optional[str]:
         return stdout_bytes.decode("utf-8").strip()
 
     except Exception as e:
-        logger.debug(f"Command {args} failed: {e}")
+        logger.debug(f"Command {args} failed: {repr(e)}")
         return None
 
 
--- a/modal/_utils/git_utils.py
+++ b/modal/_utils/git_utils.py
@@ -1,6 +1,5 @@
 # Copyright Modal Labs 2025
 import os
-import re
 import subprocess
 from typing import Dict, List, Optional, Union
 
@@ -56,24 +55,5 @@ def get_git_commit_info() -> Dict[str, Union[str, bool, int]]:
     repo_url = run_command(["git", "remote", "get-url", "origin"])
     if repo_url:
         git_info["repo_url"] = repo_url
-
-        # Generate commit URL only if we have both repo_url and commit_hash
-        if commit_hash and repo_url:
-            commit_url = None
-            if "github.com" in repo_url:
-                base_url = re.sub(r"git@github\.com:", "https://github.com/", repo_url)
-                base_url = re.sub(r"\.git$", "", base_url)
-                commit_url = f"{base_url}/commit/{commit_hash}"
-            elif "gitlab.com" in repo_url:
-                base_url = re.sub(r"git@gitlab\.com:", "https://gitlab.com/", repo_url)
-                base_url = re.sub(r"\.git$", "", base_url)
-                commit_url = f"{base_url}/-/commit/{commit_hash}"
-            elif "bitbucket.org" in repo_url:
-                base_url = re.sub(r"git@bitbucket\.org:", "https://bitbucket.org/", repo_url)
-                base_url = re.sub(r"\.git$", "", base_url)
-                commit_url = f"{base_url}/commits/{commit_hash}"
-
-            if commit_url:
-                git_info["commit_url"] = commit_url
 
     return git_info
--- a/modal/_utils/git_utils.py
+++ b/modal/_utils/git_utils.py
@@ -5,7 +5,7 @@ from typing import Optional
 from modal_proto import api_pb2
 
 
-async def run_command_async(args: list[str]) -> Optional[str]:
+async def run_command_fallible(args: list[str]) -> Optional[str]:
     try:
         process = await asyncio.create_subprocess_exec(
             *args, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
@@ -37,7 +37,7 @@ async def get_git_commit_info() -> api_pb2.CommitInfo | None:
     ]
 
     # Run commands in parallel using asyncio.gather
-    tasks = (run_command_async(cmd) for cmd in commands)
+    tasks = (run_command_fallible(cmd) for cmd in commands)
     (is_repo, log_info, branch, status, origin_url) = await asyncio.gather(*tasks)
 
     if not is_repo or not branch:
--- a/modal/_utils/git_utils.py
+++ b/modal/_utils/git_utils.py
@@ -1,80 +1,70 @@
 # Copyright Modal Labs 2025
-import os
-import subprocess
-from typing import Dict, List, Optional, Union
+import asyncio
+from typing import Optional
+
+from modal_proto import api_pb2
 
 
-def run_command(args: List[str]) -> Optional[str]:
-    """Run a command and return its output.
-
-    Args:
-        args: Command and arguments as a list
-
-    Returns:
-        Command output as string or None if command failed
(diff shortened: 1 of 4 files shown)


modal/_utils/git_utils.py
```diff
@@ -0,0 +1,79 @@
...
+        return result.stdout.strip()
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return None
+
+
+def get_git_commit_info() -> Dict[str, Union[str, bool, int]] | None:
+    """Collect git information about the current repository."""
+
+    # Check if we're in a git repository
+    if not os.path.exists(".git") and run_command(["git", "rev-parse", "--is-inside-work-tree"]) != "true":
```
gleb-modal, line 34: This runs from the current working directory, I guess it should run from the directory of the deployed app? How do I get it?

gleb-modal: I don't think we should check for .git, I guess running git should be sufficient.

modal/runner.py
```diff
@@ -7,7 +7,7 @@
 import warnings
 from collections.abc import AsyncGenerator
 from multiprocessing.synchronize import Event
-from typing import TYPE_CHECKING, Any, Optional, TypeVar
+from typing import TYPE_CHECKING, Any, Dict, Optional, TypeVar
```
mwaskom, line 10: Not necessary; we support Python 3.9+ so you can use `dict` in type annotations

modal/runner.py
```diff
@@ -182,11 +183,13 @@ async def _publish_app(
     classes: dict[str, _Cls],
     name: str = "",  # Only relevant for deployments
     tag: str = "",  # Only relevant for deployments
+    commit_info: Optional[Dict[str, Any]] = None,  # Git commit information
```
mwaskom, line 186: Perhaps this could by a `TypedDict` or a dataclass since we know the fields we'll have?

modal/runner.py
```diff
@@ -182,11 +183,13 @@ async def _publish_app(
...
 ) -> tuple[str, list[api_pb2.Warning]]:
     """Wrapper for AppPublish RPC."""
 
     definition_ids = {obj.object_id: obj._get_metadata().definition_id for obj in functions.values()}  # type: ignore
 
+    # Create the request
```
mwaskom, line 192: AI generated? :)

modal/runner.py
```diff
@@ -195,7 +198,9 @@ async def _publish_app(
         function_ids=running_app.function_ids,
         class_ids=running_app.class_ids,
         definition_ids=definition_ids,
+        commit_info=api_pb2.CommitInfo(**commit_info) if commit_info else None,
```
mwaskom, line 201: Actually seeing how you're using it here; can probably just use the `api_pb2.CommitInfo` as the primary data structure since it's all internal.

modal/_utils/git_utils.py
```diff
@@ -0,0 +1,75 @@
+# Copyright Modal Labs 2025
+import os
+import subprocess
+from typing import Dict, List, Optional, Union
+
+
+def run_command(args: List[str]) -> Optional[str]:
+    """Run a command and return its output.
```
mwaskom, line 8: Naming nit, but maybe something like `run_command_failible` to capture the "return None on any error" semantics.

modal/_utils/git_utils.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        result = subprocess.run(
+            args,
+            check=True,
+            capture_output=True,
+            text=True,
+            shell=False,
+        )
+        return result.stdout.strip()
+    except (subprocess.CalledProcessError, FileNotFoundError):
```
mwaskom, line 25: Is this every possible error class that could be raised here? IMO we probably never want to block someone running / deploying an App if this fails (but I guess the downside is that we'd never know about a bug...).

modal/_utils/git_utils.py
```diff
@@ -0,0 +1,75 @@
...
+        return None
```
mwaskom, line 26: Maybe `logger.debug` print any exception info?

modal/_utils/git_utils.py
mwaskom: Yeah, checking for `.git` would require that we're in the root of the repo, right?

modal/_utils/git_utils.py
```diff
@@ -0,0 +1,75 @@
...
+
+def get_git_commit_info() -> Dict[str, Union[str, bool, int]]:
+    """Collect git information about the current repository."""
+
+    # Check if we're in a git repository
+    if not os.path.exists(".git") and run_command(["git", "rev-parse", "--is-inside-work-tree"]) != "true":
+        return {}
+
+    git_info = {
+        "vcs": "git",
```
mwaskom, line 37: I guess it's good to be proactive about supporting other VCSs although I worry that some of the data we're collecting is fairly git-specific. Problem we can solve later, I guess.

modal/_utils/git_utils.py
```diff
@@ -0,0 +1,75 @@
...
+            text=True,
+            shell=False,
+        )
+        return result.stdout.strip()
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return None
+
+
+def get_git_commit_info() -> Dict[str, Union[str, bool, int]]:
+    """Collect git information about the current repository."""
```
mwaskom, line 30: I'm a little worried about the number of subprocess calls here adding nontrivial amount of latency to development loops. Feels like we could reduce the number of distinct calls out to `git` here?

modal/runner.py
mwaskom: Or, see below, maybe we should just pass around an `api_pb2.CommitInfo` here.

modal_proto/api.proto
```diff
@@ -292,6 +292,7 @@ message AppDeployRequest {
   string object_entity = 4;
   AppDeployVisibility visibility = 5;
   string tag = 6;
+  optional CommitInfo commit_info = 7;
```
mwaskom, line 295: We don't call `AppDeploy` anymore so no need to add this here.

modal_proto/api.proto
```diff
@@ -742,6 +745,16 @@ message CloudBucketMount {
   optional string oidc_auth_role_arn = 9;
 }
 
+message CommitInfo {
+  string vcs = 1; // Only git is supported for now
+  string branch = 2;
+  string commit_hash = 3;
+  bool dirty = 4;
```
mwaskom: Would this be better as an enum of unknown / dirty / clean? (Quite probably not, just wondering).

mwaskom (commented):
Cool feature; my main worry here is about adding latency to the App lifecycle. Had a few other thoughts too.

modal_proto/api.proto
gleb-modal: When should it be unknown? When we're not in a repo, the whole CommitInfo struct would be unset.

modal/_utils/git_utils.py
gleb-modal: Yeah. How could I get the path to the app so I can run the command from that directory? Or would that be unusual to run modal cli from outside the git repo the app is in?

mwaskom: Hm yeah I'm not sure. That information is getable but not at hand in the context where you want to be calling this. I'd probably say it's fine to define the behavior as capturing the git state from where you're running modal; it's just some optional metadata after all.

modal_proto/api.proto
mwaskom: Yeah fair point.

modal/_utils/git_utils.py
```diff
@@ -0,0 +1,80 @@
...
+    """Run a command and return its output.
+
+    Args:
+        args: Command and arguments as a list
+
+    Returns:
+        Command output as string or None if command failed
+    """
+    try:
+        result = subprocess.run(
```
ekzhang, line 17: Small thing, since Python's asynchronous executor is single-threaded we try to avoid running [synchronous code inside an asynchronous function](https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread). Starting and waiting for a `git` subprocess takes about ~20-200 milliseconds and would halt execution of concurrent / background tasks in the rest of the Modal client while it happens.

Can stay fully async with a quick API change to [create_subprocess_exec()](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.create_subprocess_exec)

ekzhang (commented):
Quick note!

modal/_utils/git_utils.py
mwaskom: Nice I didn't know there were async tools in subprocess. Could probably use that to parallelize some of the information gathering too?

modal/_utils/git_utils.py
```diff
@@ -0,0 +1,73 @@
...
+        stdout_bytes, _ = await process.communicate()
+
+        if process.returncode != 0:
+            logger.debug(f"Command {args} exited with code {process.returncode}")
+            return None
+
+        return stdout_bytes.decode("utf-8").strip()
+
+    except Exception as e:
+        logger.debug(f"Command {args} failed: {repr(e)}")
```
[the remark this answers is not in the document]
gleb-modal: Yeah, I updated to pass the exception to logger.debug via exc_info, is that what you mean?

modal/_utils/git_utils.py
ekzhang: Thanks!!

modal/runner.py
```diff
@@ -521,6 +525,9 @@ async def _deploy_app(
 
     t0 = time.time()
 
+    # Get git information to track deployment history
+    commit_info_future = get_git_commit_info()
```
ekzhang, line 529: It looks like you're trying to to start this future early to let it run concurrently with the code afterward. I think you have to wrap it in `asyncio.create_task()` to make this work. Unlike JS, Python's asyncio doesn't schedule futures unless they're being awaited, or they're wrapped in an asyncio.Task

modal_proto/api.proto
```diff
@@ -751,7 +751,7 @@ message CommitInfo {
   int64 commit_timestamp = 4;
   bool dirty = 5;
   string author_name = 6;
-  string author_email = 7;
+  optional string author_email = 7;
   optional string repo_url = 8;
```
mwaskom: Why are only some of these noted as optional?

mwaskom (approved):
LGTM once you get the tests passing!

modal_proto/api.proto
gleb-modal: Good point, I'll just remove the optional keywords here since in proto3 every field is optional already.

PR #2935, opened 2025-03-11
.map retry loop should not fail retryable errors

rculbertson:
Fixes SVC-405. 

There was a bug in the pump_inputs retry loop logic (my mistake!). Every 8 retries we stop retrying, so we can log a warning to the user. We should then continue retrying for all retryable error codes, but instead we were retrying again only for RESOURCE_EXHAUSTED. We will now retry forever for all retryable error codes.

---

---

</details>

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -73,7 +73,7 @@ RETRYABLE_GRPC_STATUS_CODES = [
 class RetryWarningMessage:
     message: str
     warning_interval: int
-    error_to_warn_for: typing.List[Status]
+    errors_to_warn_for: typing.List[Status]
 
 def create_channel(
     server_url: str,
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -73,7 +73,7 @@ RETRYABLE_GRPC_STATUS_CODES = [
 class RetryWarningMessage:
     message: str
     warning_interval: int
-    error_to_warn_for: [Status]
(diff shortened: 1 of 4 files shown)


gongy (commented):
Does this catch `StreamTerminated` `ProtocolError`? I think that's the case we need to patch.

modal/_utils/grpc_utils.py
```diff
@@ -208,6 +215,9 @@ async def retry_transient_errors(
 
             n_retries += 1
 
+            if retry_warning_message and n_retries % retry_warning_message.warning_interval == 0:
+                logger.warning(retry_warning_message.message)
+
```
rculbertson, line 220: This is the heart of the change - rather than letting the exception propagate to the caller, and logging the warning there - we now log the warning here.

rculbertson:
@gongy 🤦 yes, you're totally right. Ok, I took a different approach. Replicating the logic to catch all the retryable errors in two places get messy....so instead I changed it so `retry_transient_errors` now logs the warning. Callers can optionally pass in a warning message they want to log. This approach also has the benefit that we don't reset the backoff delay every 8 retries like the previous approach did.

rculbertson:
@gongy yes, it's used on this line:
https://github.com/modal-labs/modal-client/pull/2935/files#diff-b469cf7e0edadd8958512e9bbc682a9474018a7f523d2a4f89c1bdaf2f75b75fR147

_but_ that would probably be more clear if I used a named argument. Will push another commit changing that.

rculbertson:
@prbot automerge

PR #2938, opened 2025-03-11
Improve exit when we can't establish channel with server

mwaskom:
Fixes an issue where connectivity problems (that's sort of hand-wavy; I think they can manifest in a few different ways) when running `modal setup` or `modal token new` trigger a confusing error message.

Additionally improved the end-state error message when failing to connect in general.

---

---

</details>


erikbern (approved):
👍

PR #2940, opened 2025-03-12
add a PREVIEW image builder version

matslina:
Allows us to develop functionality for future image builder versions, without
deciding on a specific calver. Pack changes up under `PREVIEW`, then move
it over to a `YYYY.MM` when we're ready. Would be useful e.g. for the client
dependency mount project.

This image builder "version" already conceptually exists in worker as `<future>`,
so if we're happy with `PREVIEW` I'll go ahead and update worker before
merging this.

---

---

</details>


modal/image.py
```diff
@@ -56,13 +56,14 @@
...
-ImageBuilderVersion = Literal["2023.12", "2024.04", "2024.10"]
+ImageBuilderVersion = Literal["2023.12", "2024.04", "2024.10", "PREVIEW"]
 
 # Note: we also define supported Python versions via logic at the top of the package __init__.py
 # so that we fail fast / clearly in unsupported containers. Additionally, we enumerate the supported
 # Python versions in mount.py where we specify the "standalone Python versions" we create mounts for.
 # Consider consolidating these multiple sources of truth?
 SUPPORTED_PYTHON_SERIES: dict[ImageBuilderVersion, list[str]] = {
+    "PREVIEW": ["3.9", "3.10", "3.11", "3.12", "3.13"],
     "2024.10": ["3.9", "3.10", "3.11", "3.12", "3.13"],
```
mwaskom, line 67: Very satisfying that PREVIEW has the same number of characters as a YYYY.MM date 😅

matslina: aaah that is satisfying. thanks for pointing it out! =)

matslina:
> We should maybe document that PREVIEW is explicitly unstable somewhere 🤔

Any idea where? I don't really want to encourage anyone outside of Modal to use it. Maybe print an explanatory warning whenever its used?

mwaskom:
The two places we document the Image Builder Version concept here:
- The Image guide: https://modal.com/docs/guide/images#image-builder-updates
- The Image Settings config page: https://modal.com/settings/modal-labs/image-config

> I don't really want to encourage anyone outside of Modal to use it.

Yeah that's probably reasonable. In that case I would suggest that we try to suppress PREVIEW from the list of acceptable versions shown in `modal.image._get_image_builder_version` and otherwise leave it undocumented.

.github/workflows/ci-cd.yml
```diff
@@ -144,6 +144,8 @@ jobs:
             python-version: "3.9"
           - image-builder-version: "2024.10"
             python-version: "3.13"
+          - image-builder-version: "PREVIEW"
+            python-version: "3.13"
```
mwaskom, line 148: Should probably also test these on 3.9 (could be possible that we pin a new version that doesn't work with Python 3.9 (requires some additional transitive dependency, or drops support altogether).

mwaskom (approved):
LGTM thanks!

PR #2941, opened 2025-03-12
Only print warning for RESOURCE_EXHAUSTED

rculbertson:
Closes SVC-405

Fixes a bug in previous PR where we were not checking status codes in `errors_to_warn_for`. Also adds test which would have caught this.


gongy (approved):
Nice!

rculbertson:
@prbot automerge

PR #2943, opened 2025-03-13
add CMD primitives to Image

matslina:
## Describe your changes

Adds `Image.cmd()` for setting image `CMD`. Also, as of a future builder version, default `CMD` of `sleep 48h` is added to `Image.debian_slim()` and `Image.micromamba()`. This provides ease-of-use in conjunction with `Sandbox`, as a future version will run `CMD` when no `entrypoint_args` are given.

---

---

</details>

## Changelog

- Added `Image.cmd()` for setting image default entrypoint args (a.k.a. `CMD`).

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1444,6 +1444,9 @@ class _Image(_Object, type_prefix="im"):
                 f"RUN micromamba install -n base -y python={validated_python_version} pip -c conda-forge",
             ]
             commands = _Image._registry_setup_commands(tag, version, setup_commands)
+            if version > "2024.10":
+                # for convenience when launching in a sandbox: sleep for 48h
+                commands.append(f'CMD ["sleep", "{48 * 3600}"]')
             context_files = {CONTAINER_REQUIREMENTS_PATH: _get_modal_requirements_path(version, python_version)}
             return DockerfileSpec(commands=commands, context_files=context_files)
 
@@ -1833,6 +1836,9 @@ class _Image(_Object, type_prefix="im"):
             ]
             if version > "2023.12":
                 commands.append(f"RUN rm {CONTAINER_REQUIREMENTS_PATH}")
+            if version > "2024.10":
+                # for convenience when launching in a sandbox: sleep for 48h
+                commands.append(f'CMD ["sleep", "{48 * 3600}"]')
             return DockerfileSpec(commands=commands, context_files=context_files)
 
         return _Image._from_args(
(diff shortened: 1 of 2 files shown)


mwaskom:
> This provides ease-of-use in conjunction with Sandbox, as a future version will run CMD when no entrypoint_args are given.

What will the behavior be if you use, e.g. `Image.from_dockerfile` and your Dockerfile doesn't supply a CMD?

matslina:
> What will the behavior be if you use, e.g. `Image.from_dockerfile` and your Dockerfile doesn't supply a CMD?

Image will receive an empty `CMD`. Using that image in a `Sandbox` without `entrypoint_args` will run the image's `ENTRYPOINT` without args. If that's not set (or incorrectly set), container runtime will error out.

modal/image.py
```diff
@@ -2021,6 +2027,24 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             dockerfile_function=build_dockerfile,
         )
 
+    def cmd(self, cmd: list[str]) -> "_Image":
+        """Set the default entrypoint argument (`CMD`) for the image.
+
```
mwaskom, line 2032: Should we clarify how this is used in a Modal context here?

modal/image.py
```diff
@@ -2021,6 +2027,24 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        ```python
+        image = (
+            modal.Image.debian_slim().cmd(["python", "app.py"])
+        )
+        ```
+        """
+
+        cmd_str = _flatten_str_args("cmd", "cmd", cmd)
+        cmd_str = '"' + '", "'.join(cmd_str) + '"' if cmd_str else ""
```
mwaskom, line 2046: I'm not following what the concatenation with empty strings is doing  here

mwaskom (approved):

modal/image.py
```diff
@@ -2021,6 +2027,27 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             dockerfile_function=build_dockerfile,
         )
 
+    def cmd(self, cmd: list[str]) -> "_Image":
```
mwaskom, line 2030: Should we just accept `*args` and convert that to a list? Seems nicer that directly exposing the Dockerfile syntax, which is a little weird.

mwaskom: Not sure what's best really.

matslina: I personally prefer the list, as in `entrypoint()` and `shell()`. It feels more dockerfiley. But perhaps args is more python? Either way, we should do the same for all.

modal/image.py
matslina: Good question. I started doing just that, but felt it'd be weird to have it in here before the functionality (and the image builder version on which it depends) is released.

modal/image.py
mwaskom: Yeah would have preferred a pythonic interface but good point that entrypoint and shell already expose the dockerfile style; +1 for internal consistency

PR #2944, opened 2025-03-13
Add flag to enable/disable client retries

rculbertson:
Part of SVC-387

* Adds `sync_client_retries_enabled` to `FunctionMapResponse`
* Updates client to retry only if the flag is true

For now I've left the MODAL_CLIENT_RETRIES flag. This determines if we use SYNC or SYNC_LEGACY. Will be removed in follow up PRs.

---

---

</details>


rculbertson:
@prbot approve

samaross (approved):
👍

PR #2945, opened 2025-03-13
[Notebooks] Add duration field to ExecuteReply (for kernelshim)

ekzhang:


ekzhang:
very minor!

@prbot approve @luiscape

PR #2947, opened 2025-03-14
Remove defer_updates from FunctionCreateRequest

mwaskom:
Now that we no longer need backwards compatibility for <0.64 clients in the server, this field is no longer used.

Closes SVC-402


mwaskom:
@prbot approve

freider (commented):
cool

PR #2948, opened 2025-03-14
Markdown test modal "Runnable"

freider:
Quite hacky and barely tested, but only used when running doctests so I think it's ok.

`execs` the code fence like the regular fence tests (in order to inject all globals/fixtures), but keeps a reference to the globals/locals defined by it and then uses the `modal run` import logic to infer a runnable within the fence to run.

The runner then starts an ephemeral app and invokes the runnable (without arguments for now - use local entrypoint if args are needed)

I had to change the global `app` into a non-running one, so this might break some sandbox tests - will have to add a separate app fixture for a "deployed app" to use for those

---

---

</details>


charlesfrye:
Re: sandboxes. FWICT our existing examples can just be run with the default Python runner, so that should be okay.

freider:
> Re: sandboxes. FWICT our existing examples can just be run with the default Python runner, so that should be okay.

Yeah those will run with the current runner, but as part of making the new runner work I redefined the `app` global as being a non-running app (just an instance of App()) which doesn't work with Sandbox.create. I fixed this by adding a new `running_app` global which has a hydrated/running app using `App.lookup` as required by sandboxes.

charlesfrye:
Ah right sorry! Probably not a good idea to try code review before sunrise.

PR #2949, opened 2025-03-14
Proto changes to allow byte class parameters

freider:
Broke out proto changes from #2837 

---

---

</details>


freider:
@prbot approve

mwaskom (commented):
:+1:

PR #2950, opened 2025-03-14
Validate that local_python_source is passed as a string

mwaskom:
Feels like it's relatively easy to accidentally pass a module type here. While that does raise a relevant error currently, it happens inside the resolver, so the context / traceback are suppressed and it's harder to understand what the actual problem is.


mwaskom:
@prbot approve

freider (commented):
lgtm

PR #2951, opened 2025-03-14
Ignore sync_client_retries_enabled flag for now

rculbertson:
Don't honor the sync_client_retries_enabled flag until this has been rolled out on the server.
---

---

</details>


samaross (approved):
👍 but I'd expect this to break the client tests?

rculbertson:
@samaross yep, you are totally right. I marked the failing test as skipped for now.

PR #2952, opened 2025-03-14
stricter type check in Image.entrypoint() and Image.shell()

matslina:
## Describe your changes

This prevents calls to `.entrypoint()` and`.shell()` from "flattening" strings into lists of characters.
E.g., `.shell("/bin/bash")` would produce `SHELL ["/", "b", "i", "n", "/", "b", "a", "s", "h"]`

---

---

</details>

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -972,9 +972,9 @@ def test_image_docker_command_entrypoint_nonempty(builder_version, servicer, cli
 
 def test_image_docker_command_entrypoint_malformed():
     with pytest.raises(InvalidError, match="must be a list of strings"):
-        Image.debian_slim().entrypoint("this is not a list of strings")
+        Image.debian_slim().entrypoint("this is not a list of strings")  # type: ignore
     with pytest.raises(InvalidError, match="must be a list of strings"):
-        Image.debian_slim().entrypoint(4711)
+        Image.debian_slim().entrypoint(4711)  # type: ignore
 
 
 def test_image_docker_command_shell(builder_version, servicer, client, tmp_path_with_content):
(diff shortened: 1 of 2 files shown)


test/image_test.py
```diff
@@ -970,6 +970,13 @@ def test_image_docker_command_entrypoint_nonempty(builder_version, servicer, cli
         assert 'ENTRYPOINT ["/temp.sh"]' in layers[0].dockerfile_commands
 
 
+def test_image_docker_command_entrypoint_malformed():
+    with pytest.raises(InvalidError, match="must be a list of strings"):
+        Image.debian_slim().entrypoint("this is not a list of strings")  # type: ignore
+    with pytest.raises(InvalidError, match="must be a list of strings"):
+        Image.debian_slim().entrypoint(4711)  # type: ignore
```
mwaskom, line 977: FYI could use `pytest.mark.parameterize` ([docs](https://docs.pytest.org/en/stable/example/parametrize.html)) to reduce code repetition here. (Just a tip for the future; no need to fix).

test/image_test.py
```diff
@@ -1016,6 +1030,13 @@ def test_image_cmd_nonempty(builder_version, servicer, client, tmp_path_with_con
         assert 'CMD ["echo", "hello"]' in layers[0].dockerfile_commands
 
 
+def test_image_cmd_malformed(builder_version, servicer, client, tmp_path_with_content):
```
mwaskom, line 1033: Don't think these fixtures are needed FyI since the error should get raised immediately.

mwaskom (approved):
Thanks!!

erikbern (commented):
nice!

test/image_test.py
matslina: Thanks. Too fast on the copy pasta there.

PR #2953, opened 2025-03-14
Have .remote() use SYNC instead of SYNC_LEGACY

rculbertson:
Previously MODAL_CLIENT_RETRIES controlled whether you used SYNC or SYNC_LEGACY. Now we always use SYNC, and MODAL_CLIENT_RETRIES has been removed. The ability to enable/disable client-side retries is now on the server.

Do not merge until all server changes have been deployed.

---

</details>


rculbertson:
@prbot approve

gongy (approved):
LGTM!

PR #2954, opened 2025-03-14
Small updates to modal.Dict docstring

mwaskom:
User ran into an issue when trying to use a complex type as a Dict key; cloudpickle does not guarantee deterministic serialization, so it's probably best to use primitive types as keys even though we technically support any cloudpickle-able type. Also added a note about other serialization gotchas and fixed a misleading statement regarding the synchronicity-wrapped get/put methods.

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -26,8 +26,12 @@ def _serialize_dict(data):
 class _Dict(_Object, type_prefix="di"):
     """Distributed dictionary for storage in Modal apps.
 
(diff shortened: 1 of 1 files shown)


mwaskom:
@prbot approve

mwaskom:
@prbot approve

mwaskom:
@prbot approve now plz

shariqm-modal (commented):
approve

PR #2955, opened 2025-03-17
Slightly improve fastapi_endpoint docs

mwaskom:
If you write the `@fastapi_endpoint` function as an `@asgi_app` function and return a `FastAPI` instance, the error state is a pretty nasty endless recursion inside fastapi.

That's tough, but we can at least try to call out the distinction between the two decorators more clearly.


charlesfrye (approved):
👍

PR #2957, opened 2025-03-17
Use fully qualified names in un-called decorator error

mwaskom:
Updates the "suggested" invocation pattern in the error message when decorators aren't called to have a fully-qualified name, matching the style in our docs.


mwaskom:
FWIW I would like to handle this better in general, but feels like a strict marginal improvement.

mwaskom:
@prbot approve

erikbern (commented):
👍

PR #2958, opened 2025-03-17
Add @modal.concurrent decorator

mwaskom:
## Describe your changes

Adds the new `@modal.concurrent` decorator and makes some internal changes so that `max_concurrent_inputs` is the primary internal representation, with `target_concurrent_inputs` supported as an optional configuration that we can expose later.

Doesn't add a deprecation warning to `allow_concurrent_inputs` yet.

- Requires https://github.com/modal-labs/modal/pull/20971
- Part of CLI-345

---

---

</details>

## Changelog

- Adds the `@modal.concurrent` decorator, which will be replacing the beta `allow_concurrent_inputs=` parameter of `@app.function` and `@app.cls` for enabling input concurrency. Notably, `@modal.concurrent` introduces a distinction between `max_inputs` and `target_inputs`, allowing containers to burst over the concurrency level targeted by the Modal autoscaler during periods of high load.


modal/_partial_function.py
```diff
@@ -36,6 +36,7 @@ class _PartialFunctionFlags(enum.IntFlag):
     EXIT = 16
     BATCHED = 32
     CLUSTERED = 64  # Experimental: Clustered functions
+    CONCURRENT = 128
```
freider, line 39: Do we need this flag if we have the max_concurrent_input attribute on the _PartialFunction anyway? This might apply to other flags here too though...
I think the purpose of the flags was be able to look up sets of functions that need special handling at different lifecycle stages (or deploy time), which shouldn't be needed for concurrent?

mwaskom: Hm perhaps not.

modal/_partial_function.py
```diff
@@ -722,3 +726,33 @@ def wrapper(raw_f: Callable[..., Any]) -> _PartialFunction:
         )
 
     return wrapper
+
+
+def _concurrent(
+    _warn_parentheses_missing=None,
+    *,
+    max_inputs: int,  # Limit on each contianer's concurrency
```
freider, line 734: Comment typo: `contianer`

I know this is beating a dead horse but something about the name `max_inputs` still itches for me (not a blocker though, I think it's more important to just get the changes out there).

Did we ever consider `max_concurrency` as the public name for this attribute - I still have a feeling `max_inputs` could intuitively indicate a cap on total the number of inputs?

test/supports/concurrency_config.py
```diff
@@ -0,0 +1,22 @@
...
+
+@app.function()
+@modal.concurrent(max_inputs=CONFIG_VALS["NEW"])
+def has_new_concurrency_config():
+    pass
+
+
+@app.function()
+def has_no_concurrency_config():
+    pass
```
freider, line 22: We should add failing tests for:

```
@concurrent
@method
```

and

```
@method
@concurrent
```

modal/_partial_function.py
mwaskom: We considered a lot of things :) https://www.notion.so/modal-com/The-modal-concurrent-decorator-1ad1e7f169498050974ac8a1cd3a2bec?pvs=4

test/supports/concurrency_config.py
freider: And even just this, which might also give unexpected results otherwise:

```py
@app.cls()
class A:
    @concurrent
     def some_method(...):
```

modal/_partial_function.py
freider: I know, but I couldn't find `max_concurrency` in that doc 😬

modal/_partial_function.py
mwaskom: Actually actually, maybe not since `max_inputs` must be set if you use `@modal.concurrent`.

modal/_partial_function.py
mwaskom: It's equivalent to the `limit` proposal IMO.

test/supports/concurrency_config.py
mwaskom: Hm yeah we'll need to catch that in `app.cls` I guess? Pretty messy.

freider: `Cls.from_local` is probably a good place to put it. We probably have other similar cases that are odd already, like @batched. 

All partial functions on the class must either be FUNCTION, BUILD_FUNCTION, ENTER_X, EXIT-flagged to be valid, I think, anything else would indicate a mistake (the web endpoint class of annotations should add FUNCTION as well)

mwaskom: Is FUNCTION supposed to mean "METHOD" or is it valid for decorators that should go on a (regular, lowercase-f) function as well?

mwaskom:
@prbot approve

mwaskom:
@prbot approve

PR #2960, opened 2025-03-19
Class parameters: Decouple encoding/decoding from schema validation

freider:
Previously we required a predefined schema to encode or decode parameters for a class into a proto payload:

`serialize_proto_params(python_values: dict[str, Any], schema: SchemaDefinition) -> bytes`
and 
`deserialize_proto_params(param_bytes: bytes, schema: SchemaDefinition) -> dict[str, Any]`

The reason it's been intertwined is because we want the payload bytes consistent (default arguments included etc.), and to make the payload as "strictly typed" as possible to prevent issues later on during container startup - but this is very specific to the class parameter payload.

However, if we want to reuse this code for more general schema extraction (and/or function payload) encoding, it's preferable to split up the schema validation of a payload from the actual proto serialization (which would then use runtime types instead of annotations).

This PR does essentially that by changing the compounded `serialize_proto_params(python_values, schema)` call into:
```py
python_values_with_defaults = apply_defaults(python_values, schema)
validate_params(python_values_with_defaults, schema)
bytes = serialize_proto_params(python_values_with_defaults)  # no schema
```
and similarly `deserialize_proto_params(param_bytes, schema)`:
```py
python_values = deserialize_proto_params(python_values_with_defaults) # no schema
validate_params(python_values, schema)  # kind of useless tbh... remove?
```

It also refactors a few of the validation steps, exception types and error messages we emit in related code, but in general the end result of the class parameterization should be unchanged from this code change.

In followup PRs, I intend to extend the "schema" structure to support additional types (class parameters would still be restricted to these few ones) so that we can (optionally) register schemas for functions as well, which opens up a bunch of cool possibilities.

---

---

</details>


modal/_serialization.py
```diff
@@ -411,75 +412,112 @@ class ParamTypeInfo:
...
+            type=protobuf_type,
         )
-        python_value = python_params.get(schema_param.name)
-        if python_value is None:
-            if schema_param.has_default:
-                python_value = getattr(schema_param, type_info.default_field)
-            else:
-                raise ValueError(f"Missing required parameter: {schema_param.name}")
         try:
             converted_value = type_info.converter(python_value)
```
mwaskom, line 467: I'm a little confused about the role of the `converter`. If we're calling the converter during *serialization*, and the converter is typically itself the type, what type do we expect the input to have here?

modal/_serialization.py
```diff
@@ -488,11 +526,21 @@ def deserialize_params(serialized_params: bytes, function_def: api_pb2.Function,
...
+        validate_params(param_kwargs, list(function_def.class_parameter_info.schema))
     else:
         raise ExecutionError(
             f"Unknown class parameter serialization format: {function_def.class_parameter_info.format}"
         )
 
     return param_args, param_kwargs
+
+
+def validate_parameter_type(parameter_type: type) -> "api_pb2.ParameterType.ValueType":
```
mwaskom, line 541: Nit but I don't love it when "validate" functions have return values.

modal/_serialization.py
```diff
@@ -411,75 +412,112 @@ class ParamTypeInfo:
...
+    for schema_param in schema:
+        if schema_param.name not in params:
+            # we expect all values to be present - even defaulted ones (defaults are applied on payload construction)
+            raise InvalidError(f"Missing required parameter: {schema_param.name}")
+        python_value = params[schema_param.name]
+        python_type = type(python_value)
+        param_protobuf_type = validate_parameter_type(python_type)
+        if schema_param.type != param_protobuf_type:
+            expected_python_type = PROTO_TYPE_INFO[schema_param.type].type
+            raise TypeError(
```
mwaskom, line 507: I feel like we need a clear policy about when we use `InvalidError` (or other `modal.exception` types vs. builtin types).

modal/_utils/function_utils.py
```diff
@@ -99,6 +106,24 @@ def get_function_type(is_generator: Optional[bool]) -> "api_pb2.Function.Functio
     return api_pb2.Function.FUNCTION_TYPE_GENERATOR if is_generator else api_pb2.Function.FUNCTION_TYPE_FUNCTION
 
 
+def signature_to_protobuf_schema(signature: inspect.Signature) -> list[api_pb2.ClassParameterSpec]:
```
mwaskom, line 109: It's somewhat surprising to me that we have `modal.parameter()` annotations, from which we build a synthetic constructor, and then we get the schema from the signature of the synthetic constructor. Do I have that right? Would it be any simpler to just get the annotations of the class directly?

mwaskom (approved):

modal_proto/api.proto
```diff
@@ -172,6 +172,7 @@ enum ParameterType {
   PARAM_TYPE_INT = 2;
   PARAM_TYPE_PICKLE = 3;
   PARAM_TYPE_BYTES = 4;
+  PARAM_TYPE_UNKNOWN = 5;  // used in schemas to signify unrecognized or un-annotated types
```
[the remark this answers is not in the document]
freider: Not sure for standard operations, but for debugging etc. it's nice to be able to differentiate between "untintentionally missing" (proto will fill it with the unspecified = 0 value) and intentionally set to "unknown".

modal/_serialization.py
freider: Yeah I don't think this is very consistent, and I don't particularly like the converter name either. Will redo this code in a followup PR to be more like the type handling in the "polyglot" draft PR where register individual handler classes for each payload type and let those group information about serialization/deserialization logic for each type

freider: (Currently we do expect the input types to be of the expected type, and the "converters" are a way to enforce the type if the provided value is *not* of the right type, like converting strings into ints, but I think it would actually be better to crash in that case...)

modal/_serialization.py
freider: Agreed - at first it didn't return anything and this was added in a lazy refactor commit - will adjust the name :)

modal/_serialization.py
freider: Yeah... I didn't think a ton about it here, and went with something like "if the wrong type is provided, use TypeError, if something else doesn't match assumptions, use InvalidError"

modal/_utils/function_utils.py
freider: We use the `modal.parameter()` annotations to build an `inspect.Signature` representation of the constructor (bu looking at the annotations on the class directly, so we don't actually make a synthetic constructor)

Originally this was done since this way since we had some initial implementation of "strictly typed" parameters where we still used explicit constructors (but with annotated constructor args).

But this is also useful for other reasons:
* The Signature object is useful to apply defaults and bind values in the standard python way so we don't have to implement keyword/non-keyword resolution and default values ourselves (right now we only allow keyword arguments so it's not too hard to implement, but signature.bind *could* allow us to use positionals as well if we want to)
* We can reuse the same code to register schemas for functions soon since the schema extractor code uses a Signature

modal/_serialization.py
mwaskom: I don't think you need to hold up this PR on it but we should figure it out as part of 1.0 (there's a ticket already). We're pretty inconsistent about it AFAICT.

PR #2961, opened 2025-03-19
Add base images to invoke

ekzhang:
Adding base images to `inv` so they can be run more easily in development, and also to add a confirmation similar to `inv publish-base-mounts` to prevent accidental pushing to prod.

Hoping to add a new base image for notebooks soon after this.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -225,13 +225,15 @@ def _check_prod(no_confirm: bool):
 
 @task
 def publish_base_mounts(ctx, no_confirm: bool = False):
+    """Publish the client mount and other mounts."""
     _check_prod(no_confirm)
     for mount in ["modal_client_package", "python_standalone"]:
         ctx.run(f"{sys.executable} modal_global_objects/mounts/{mount}.py", pty=True)
 
 
 @task
-def publish_base_image(ctx, name: str, builder_version: str = "2024.10", no_confirm: bool = False) -> None:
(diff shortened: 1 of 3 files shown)


ekzhang:
image

Also added a table for reference

ekzhang:
@prbot approve

ekzhang:
Happy to make any follow-up changes!

tasks.py
```diff
@@ -210,21 +210,41 @@ def check_copyright(ctx, fix=False):
...
 
+@task
+def publish_base_images(ctx, name: str, builder_version: str = "2024.10", no_confirm: bool = False) -> None:
+    """Publish base images. For example, `inv publish-base-images debian_slim`."""
+    _check_prod(no_confirm)
+    ctx.run(
+        f"python -m modal_global_objects.images.base_images {name}",
+        pty=True,
+        env={
+            "MODAL_IMAGE_ALLOW_GLOBAL_DEPLOYMENT": "1",
```
mwaskom, line 242: I wonder if we should expose this via a flag to the task and make it default to False, just for extra safety?

mwaskom (commented):
Thanks for the improvements! Had one suggestion about handling the global Image failsafe.

tasks.py
ekzhang: I like that, without the flag it's kind of like a dry run. Will add!

PR #2963, opened 2025-03-19
Add base images for notebooks

ekzhang:
None of these package versions are pinned yet, this is just an MVP including common libraries. Including CUDA versions of PyTorch/JAX and transformers/diffusers.

Merging this will unlock like ~2-3x improved startup time for notebook kernels since it won't need to download notebook kernel dependencies on startup.


modal_global_objects/images/base_images.py
```diff
@@ -3,26 +3,50 @@
...
 
 
 if __name__ == "__main__":
     _, name = sys.argv
-    constructor = getattr(modal.Image, name)
+
+    if name in ("debian_slim", "micromamba"):
+        constructor = getattr(modal.Image, name)
+    elif name == "notebook":
+        constructor = notebook_base_image
```
ekzhang, line 25: Moved this here so that it doesn't clutter up `image.py` @mwaskom

modal_global_objects/images/notebook_base_image.py
```diff
@@ -0,0 +1,99 @@
...
+    ]
+
+    def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
+        return DockerfileSpec(commands=["FROM base", *(f"RUN {cmd}" for cmd in commands)], context_files={})
+
+    return Image._from_args(
+        base_images={"base": base_image},
+        dockerfile_function=build_dockerfile,
+        force_build=force_build,
+        _namespace=api_pb2.DEPLOYMENT_NAMESPACE_GLOBAL,
```
mwaskom, line 98: Ah I was wrong about this being exclusively public API since we need to pass this; that said I think it's fine to call into private APIs from this script.

Don't forget that we have a failsafe mechanism that prevents admins from deploying images to the global namespace unless they really intend to do that!

ekzhang: Yes absolutely! I included this in an `inv` task in #2961 by the way — but now that I have more context we could definitely move this into private repo

mwaskom: Yeah FWIW it's not so much the public / private distinction that I find super important w/r/t these. It's more about global deployments being something we should only be doing intentionally at this point, whereas currently we try to do a deployment on every commit but rely on caching behavior to make that a no-op in most cases.

mwaskom (approved):

ekzhang:
Thanks for reviewing @mwaskom! By the way, I just realized that the images I built were actually created in the wrong deployment namespace, will debug that.

PR #2965, opened 2025-03-20
Remove pending=True from 1.0 deprecations

mwaskom:
A little scary as these were quite noisy at one point, but I've checked recent CI logs for 

- Integration tests
- Server tests
- Client doctests

And don't see any pending deprecation warnings.

I'm not sure where to find the Guide doctest logs though 😳.


mwaskom:
Oops, rephrased an error message a bit and didn't realize we had a strict test.

freider:
Ah, looks like we have a bunch of tests for the old deprecated `-m`-less auto-module syntax that makes this PR fail - we can probably add explicit exceptions to those tests to prevent the deprecation warning from failing them?

mwaskom:
@prbot approve

PR #2966, opened 2025-03-21
[WRK-814] Client side impl Sandbox.resource_usage()

thundergolfer:
## Describe your changes

- WRK-814

---

---

</details>

## Changelog

* Adds new `Sandbox.resource_usage()` method to retrieve billable resource usage (CPU, RAM, GPU time).


modal_proto/api.proto
```diff
@@ -2373,6 +2373,17 @@ message SandboxListResponse {
   repeated SandboxInfo sandboxes = 1;
 }
 
+message SandboxGetResourceUsageRequest {
```
thundergolfer, line 2376: If the proto is approved I'll split it out into its own PR so that the server-side can be merged before merging this PR.

devennavani (approved):
LGTM. It might be good to clarify in the comments/docstring that the returned CPU and memory values are max(used, requested)

modal/sandbox.py
```diff
@@ -63,6 +64,23 @@ def _validate_exec_args(entrypoint_args: Sequence[str]) -> None:
...
+class SandboxResourceUsage:
+    """Simple data structure storing resource usage stats of a Sandbox."""
+
+    # CPU usage is in core-nanoseconds. Using 1 core for 1 second is 1e9 core ns.
+    # Using 0.5 core for 1 second is 5e8 core ns.
+    cpu_core_nanosecs: int
+    # Memory usage is in gibibyte-nanoseconds. Using 1 GiB for 1 second is 1e9 gib ns.
+    # Using 0.5 GiB for 1 second is 5e8 gib ns.
+    mem_gib_nanosecs: int
+    # GPU type, e.g. "a10g". Combine with `gpu_nanosecs` to understand billable GPU spend.
```
mwaskom, line 77: Currently we have a many-to-one mapping for some GPU types on configuration (e.g. we accept `"A10"` and `"A10g"`, plus we're case-insensitive, etc. How are we defining the schema for GPU type here? We've broken our own internal tooling recently with some migrations; are we now obligated to keep the current schema fixed moving forward?

modal/sandbox.py
```diff
@@ -63,6 +64,23 @@ def _validate_exec_args(entrypoint_args: Sequence[str]) -> None:
         )
 
 
+@dataclass(frozen=True)
+class SandboxResourceUsage:
+    """Simple data structure storing resource usage stats of a Sandbox."""
+
+    # CPU usage is in core-nanoseconds. Using 1 core for 1 second is 1e9 core ns.
+    # Using 0.5 core for 1 second is 5e8 core ns.
+    cpu_core_nanosecs: int
```
mwaskom, line 73: Random ignorable suggestion but `nanosecs` is super non-human-friendly, maybe the dataclass could have a method that presents the underlying data in a more human-readable format to save people some hassle when they're using it interactively or printing it or whatever.

mwaskom (approved):

modal/sandbox.py
thundergolfer: Yep good idea, I can do a nice `repr`.

mwaskom:
> LGTM. It might be good to clarify in the comments/docstring that the returned CPU and memory values are max(used, requested)

oh missed that as I haven’t looked closely at the backend pr, but would argue that maybe suggests a different name, not just documentation (which is also good to add).

mwaskom:
I think different people might really want to know both though (what did it actually use, eg for profiling, and what did we bill for, eg for passing on costs).

mwaskom:
And ideally we wouldn’t expose multiple variants of similar information through multiple different ad hoc methods :/

thundergolfer:
> Maybe an as_billed: bool parameter or something?

Like `resource_usage(raw: bool = True)`? 

Hmm I don't love it. I didn't realize that we applied compute resource multipliers because of region scheduling. I'd otherwise be OK with passing through the reserved value as-is, because I don't think this endpoint can ever serve as profiling data. 

I'll leave this for a bit and think about it.

modal/sandbox.py
thundergolfer: Hmm, maybe not, I like this behavior: 

> Return a string containing a printable representation of an object. **For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to [eval()](https://docs.python.org/3/library/functions.html#eval);**

modal/sandbox.py
erikbern: We have a canonical map in the backend, could just use that one. But I'm also fine not exposing it here.

mwaskom: that makes sense to me in general but do sandboxes have gpu fallbacks? or can you only request one gpu type?

mwaskom:
> Like resource_usage(raw: bool = True)? If raw==True then pass through the usage, otherwise apply the reservation and regional modifiers.

> Hmm I don't love it. I didn't realize that we applied compute resource multipliers because of region scheduling. I'd otherwise be OK with passing through the reserved value as-is, because I don't think this endpoint can ever serve as profiling data.

I am just suggesting that we allow users to query `used` or `max(used, reserved)`. I don't think this endpoint should reflect any modifiers which I see as a sort of implementation detail.

> I don't think this endpoint can ever serve as profiling data

Why not? I'd certainly try to use it for something like that based on how it's currently presented, and I expect other users would too.

modal/sandbox.py
mwaskom: Personally I've never thought that notion of `repr()` was very useful, but yeah I think you could do it with `__str__` instead of `__repr__`. I'd probably try to format the CPU/RAM/GPU usage with SI prefixes so it's readable across multiple OOMs.

PR #2967, opened 2025-03-21
Add support for generic type schema extraction

freider:
This is a pretty large PR, but most of it is somewhat self-contained in a new "type management" module + tests.

The key changes to look at are:
## Proto changes
Adds a new proto type for encoding "generic types" (types with sub types):
```proto
message GenericPayloadType {
  ParameterType base_type = 1;
  repeated GenericPayloadType sub_types = 2;  // sub-type for generic types like lists
}
```
This will allow us to record types such as `list[str]` in a protobuf structure.

We are also changing the `ClassParameterSpec`, replacing the old `type` attribute with a `full_type: GenericPayloadType` (for class parameters we'll have to fill the old attributes to not break old->new class lookups).

## Type lookups
To generalize a bit and make it easier to add support for more types without forgetting something, I made a bit of an overengineery _type_manager.py module which uses a registry pattern with decorators to handle type-related lookups (we could/should move the CLI type handling here too maybe?).

This does away with the old inlined lookup dicts and dataclasses we used to determine the logic for different parameter types.

## Practical changes
This PR should have almost no consequences for existing apps, except make any new deploys a bit more forward compatible in how they store they store class parameter schemas.

The idea is that `ClassParameterSpec` will be reusable for recording *function* schemas, and not only class parameters as we do today, and would support a much wider range of types.

For class parameters, we'll still only support scalar values for now, so the new generic types don't apply - except that the new client will add the generic type in addition to the scalar type (duplicating this info).
Once all clients are only looking at the generic type info (even for class parameters), we can remove the old `.type` attribute from the proto def.

The end goal is that class parameters and function argument can use the same type to record schema information - there is a followup PR that does this (pending some rebasing).

---

---

</details>

## Changelog

* Instantiation of classes using keyword arguments that are not defined as as `modal.parameter()` will now raise an error on the calling side rather than in the receiving container. Note that this only applies if there is at least one modal.parameter() defined on the class, but this will likely apply to parameter-less classes in the future as well.


test/function_utils_test.py
```diff
@@ -153,21 +151,3 @@ def has_global_ref():
 def test_global_variable_extraction(func):
     info = FunctionInfo(func)
     assert info.get_globals().get("GLOBAL_VARIABLE") == GLOBAL_VARIABLE
-
-
-def test_schema_extraction_unknown():
```
freider, line 158: moved to serialization_test

modal/_type_manager.py
```diff
@@ -0,0 +1,198 @@
...
+            self._proto_type_to_serde[enum_type_value] = ph
+            return ph
+
+        return deco
+
+    def get_encoder(self, python_base_type: type) -> ParameterProtoSerde:
+        try:
+            return self._py_base_type_to_serde[python_base_type]
+        except InvalidError:
+            raise KeyError(f"No decoder implemented for python type {python_base_type.__name__}")
```
mwaskom, line 49: Are the error types backwards here?

modal/_type_manager.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def encode(self, python_value: type) -> api_pb2.ClassParameterValue:
+        return self.get_encoder(type(python_value)).encode(python_value)
```
mwaskom, line 52: Are users of this class meant to interact just with `encode`/`decode` or also with `get_encoder`/`get_decoder` directly?

modal_proto/api.proto
```diff
@@ -1728,6 +1734,11 @@ message GeneratorDone {  // Sent as the output when a generator finishes running
   uint64 items_total = 1;
 }
 
+message GenericPayloadType {
+  ParameterType base_type = 1;
+  repeated GenericPayloadType sub_types = 2;  // sub-type for generic types like lists
```
mwaskom, line 1739: In a type annotation like `list[str]`, what's the "base" and what's the "sub"? "base" and "sub" are sort of synonymous so it doesn't feel super obvious.

mwaskom (approved):
I like it 👍

modal/_type_manager.py
freider: Users being us. In most cases it will be enough to just use encode/decode, but there are e couple of places where the getters are used. I'll see if I can refactor those cases that might be cleaner

modal_proto/api.proto
freider: For a generic type `base_type` is its "origin" type (`list` in this case) in python nomenclature and the "args" are the sub types.

For a non-generic/scalar type, the `base_type` is the type itself (but notably in python `get_origin(int)` is None, so it wouldn't be an origin type, which is part of the reason I chose another name.

modal/_type_manager.py
freider: lol, yes

modal/_type_manager.py
freider: I refactored to put the validation logic in the `ProtoParameterSerdeRegistry` instead, so now the get_decoder and get_encoder are both private there

modal/_resolver.py
```diff
@@ -153,7 +154,11 @@ async def loader():
                 self._deduplication_cache[deduplication_key] = cached_future
 
         # TODO(elias): print original exception/trace rather than the Resolver-internal trace
-        return await cached_future
+        try:
+            return await cached_future
+        except Exception:
+            traceback.print_exc()
```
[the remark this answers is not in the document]
ekzhang: Hey @freider — I think this line of code that was just merged is making it not possible to catch errors in Modal APIs without the traceback being printed out.

image

If it's an unrelated change to this PR, is it ok if I revert for now so we can get this API to work again without printing out the exception?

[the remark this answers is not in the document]
freider: Yeah it's fine to revert this particular change! @ekzhang

PR #2968, opened 2025-03-21
Display git info in the client deployment history

gleb-modal:
Show git info we started collecting in #2934.

## Changelog

- Adds a new "commit info" column to the `modal app history` command. It shows the short git hash at the time of deployment, with an asterisk `*` if the repository had uncommitted changes.


mwaskom (approved):
LGTM ... not sure about how to get the mock working, sorry

PR #2970, opened 2025-03-22
Add schema on function definition [opt-in via env var/config for now]

freider:
Translates type annotations of functions/methods into a protobuf structure that is stored on the function definition and returned as hydration metadata. The structure reuses the structure we already use for class parameter specification, but with extended type support and return value annotations.

This will allow retrieval of this information on function lookup, enabling things like UI and API stub generation.
A private API for retrieving schemas for a hydrated function is also added: `Function._get_schema()` which returns the raw proto schema of the function, if one exists.

For now the schema capture is opt-in using the `MODAL_FUNCTION_SCHEMAS=1` env var.

Builds upon #2967 

---

---

</details>


modal/cls.py
```diff
@@ -428,6 +428,24 @@ def _get_method_names(self) -> Collection[str]:
         # returns method names for a *local* class only for now (used by cli)
         return self._method_partials.keys()
 
+    @live_method
+    async def _get_constructor_args(self) -> typing.Sequence[api_pb2.ClassParameterSpec]:
```
freider, line 432: These are the `Cls` variants of schema getters which makes sense to have since you want this info before constructing any instance of the class (for building uis etc). It does clutter the namespace of the Cls though, for something that's quite internal at this point, so maybe the names should be a bit more mangled? `._modal_get_constructor_args()`?

modal_proto/api.proto
```diff
@@ -1711,6 +1713,12 @@ message FunctionRetryPolicy {
   uint32 retries = 18;
 }
 
+message FunctionSchema {
+  int32 schema_version = 1; // 1 | 0 for now, allows easy disambiguation between empty schema and no schema collection
```
freider, line 1717: Should this be an enum? I don't have a strong opinion... I mostly put it there to have a way to get field presence for schema to translate when moving the attribute to other protos

modal/_serialization.py
```diff
@@ -526,3 +526,27 @@ def signature_to_parameter_specs(signature: inspect.Signature) -> list[api_pb2.C
         field_spec = _signature_parameter_to_spec(param, include_legacy_parameter_fields=True)
         modal_parameters.append(field_spec)
     return modal_parameters
+
+
+def get_callable_schema(
+    callable: typing.Callable, ignore_first_argument: bool = False
+) -> typing.Optional[api_pb2.FunctionSchema]:
+    # ignore_first_argument can be used in case of unbound methods where we want to ignore the first (self) argument
```
mwaskom, line 534: Example of regret over callable methods IMO.

modal/_functions.py
```diff
@@ -648,10 +650,17 @@ def from_local(
             for method_name, partial_function in partial_functions.items():
                 function_type = get_function_type(partial_function.is_generator)
                 function_name = f"{info.user_cls.__name__}.{method_name}"
+
+                if not partial_function._is_web_endpoint():
+                    method_schema = get_callable_schema(partial_function.raw_f, ignore_first_argument=True)
```
mwaskom, line 655: Minor but might make some code easier to follow if we had a positive assertion for ... whatever kind of think you are when you're a method with a callable schema that isn't a web endpoint.

modal/cls.py
freider: Good idea. It's a bit annoying though since it needs to hook into the lazy-loader functionality (the `live_method` thing), but I'll see if I can do that in a nice way for standalone functions too...

modal_proto/api.proto
```diff
@@ -1711,6 +1713,12 @@ message FunctionRetryPolicy {
...
+  repeated ClassParameterSpec arguments = 2;
```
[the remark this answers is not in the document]
freider: I've tried to keep "parameter" as the name for the thing in parameterized classes, and "argument" here as the arguments to a function, but wanted to reuse the schema specification data structure between the two. Plan is to rename the datastructure to something more generic

modal/cls.py
freider: Wasn't so bad. Just had to wrap the helper in a synchonize_api

freider:
@prbot approve

mwaskom:
✅

PR #2972, opened 2025-03-24
Change print to logger.debug in .map

rculbertson:
## Describe your changes

Fixing issue where `print` was used instead of `logger.debug`

---

---

</details>


rculbertson:
@prbot approve

mwaskom (approved):
Thanks for the quick fix!

rculbertson:
@prbot automerge

PR #2981, opened 2025-03-26
fix: I got ARGMAX wrong

thundergolfer:
## Describe your changes

Turns out when I threw in `ARG_MAX` handling into https://github.com/modal-labs/modal-client/pull/2856 I got it wrong. 
The real limit (in prod and on my devbox) is not 2MiB. 

---

---

</details>


pawalt (approved):
Interesting thanks. Was this due to the user issue about too long args?

thundergolfer:
It was prompted by that yeh. I was surprised they ran into it because I thought I stopped it client-side.

thundergolfer:
We should also check this server-side.

PR #2983, opened 2025-03-27
Revert #2958

mwaskom:
Reverts the change introduced in #2958, which has a defect. The published version has been yanked from PyPI.


danielshaar (approved):
Thanks!

PR #2984, opened 2025-03-27
Get releases working again

mwaskom:
We've discovered an annoying circular dependency. If we yank a client release without publishing a subsequent version, client doctests will hit prod with the yanked version, get rejected, fail CI, and block the release of the fix.

But we still merge the fix to main, so _another_ PR will have the "new" client version (which wasn't actually published to PyPI, but _did_ get a Modal client mount created) and *should* 🤞  work.


mwaskom:
@prbot approve

danielshaar (commented):
sure thing

PR #2985, opened 2025-03-27
stop passing default sandbox entrypoint_args

matslina:
Stop replacing blank entrypoint_args with `sleep 48h` in `Sandbox.create()`.

For images built at version `<= 2024.10`, worker defaults to `sleep` which preserves the current
behavior. For later images, it will instead run the default entrypoint arguments, a.k.a. the image's
`CMD`.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -108,9 +108,6 @@ class _Sandbox(_Object, type_prefix="sb"):
     ) -> "_Sandbox":
         """mdmd:hidden"""
 
-        if len(entrypoint_args) == 0:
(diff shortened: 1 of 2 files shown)


matslina:
Feeling a bit paranoid about this one. I've tested it against prod, and with monorepo e2e tests. Seems to work as intended.

thundergolfer (approved):
LGTM. 

Should we update our docs for this to handle the transition.

matslina:
I don't think there's a need to update docs just yet. Behavior on `Sandbox.create()` should be identical, except for images on builder version `PREVIEW` which we're holding off on advertising publicly.

PR #2986, opened 2025-03-27
Add `@modal.concurrent` decorator (take two)

mwaskom:
## Describe your changes

Adds the new `@modal.concurrent` decorator and makes some internal changes so that `max_concurrent_inputs` is the primary internal representation, with `target_concurrent_inputs` supported as an optional configuration that we can expose later.

Doesn't add a deprecation warning to `allow_concurrent_inputs` yet.

This is a retry of https://github.com/modal-labs/modal-client/pull/2958, which needed to be reverted due to changes flagged by internal integration tests.

- Requires https://github.com/modal-labs/modal/pull/20971
- Part of CLI-345

---

---

</details>

## Changelog

- Adds the `@modal.concurrent` decorator, which will be replacing the beta `allow_concurrent_inputs=` parameter of `@app.function` and `@app.cls` for enabling input concurrency. Notably, `@modal.concurrent` introduces a distinction between `max_inputs` and `target_inputs`, allowing containers to burst over the concurrency level targeted by the Modal autoscaler during periods of high load.


modal/_partial_function.py
```diff
@@ -722,3 +730,49 @@ def wrapper(raw_f: Callable[..., Any]) -> _PartialFunction:
...
+
+    The concurrency mechanism depends on whether the function is async or not:
+    - Async functions will run inputs on a single thread as asyncio tasks.
+    - Synchronous functions will use multi-threading. The code must be thread-safe.
+
+    Input concurrency will be most useful for workflows that are IO-bound
+    (e.g., making network requests) or when running an inference server that supports
+    dynamic batching.
+
+    When `target_inputs` is set, Modal's autoscaler will try to provision resources such
```
freider, line 751: Should we mention what happens if target_inputs isn't set but max_inputs is?

freider: I.e. that max is then used as the target

freider: Also, should a target > max be prevented?

modal/_runtime/container_io_manager.py
```diff
@@ -976,6 +977,10 @@ def target_concurrency(self) -> int:
     def max_concurrency(self) -> int:
         return self._max_concurrency
 
+    @property
+    def input_concurrency_enabled(self) -> int:
+        return max(self._max_concurrency, self._target_concurrency) > 1
```
freider, line 982: Are there cases where target would be higher than max?

modal/app.py
```diff
@@ -678,6 +678,12 @@ def wrapped(
                 is_generator = f.is_generator
                 batch_max_size = f.batch_max_size
                 batch_wait_ms = f.batch_wait_ms
+                if f.max_concurrent_inputs:  # Using @modal.concurrent()
```
freider, line 681: Should there be some kind of error if both @concurrent and allow_concurrent_inputs on the decorator is used? Maybe it would be sufficiently covered by the allow_concurrent_inputs deprecation?

modal/_partial_function.py
mwaskom: You can't not set `max_inputs` when using the decorator, as it doesn't have a default argument.  And we reject `target_inputs > max_inputs` below.

modal/app.py
mwaskom: Yeah I think once we add that deprecation warning it'll be unambiguous that one replaces the other, if it's not clear now.

modal/_partial_function.py
freider: I mean if you set the `max` but not the target (that it gets defaulted to the max in that case)

freider (approved):

modal/_partial_function.py
mwaskom: Oh I misread. But I'm not following what you think is confusing about the current description. Are you worried people will not realize that modal functions autoscale if `target` is not set?

PR #2987, opened 2025-03-28
Adds `bool` type support for class parameters

freider:
## Describe your changes

---

---

</details>


modal/_functions.py
```diff
@@ -652,11 +652,11 @@ def from_local(
...
-
-                if config.get("function_schemas") and not partial_function._is_web_endpoint():
-                    method_schema = get_callable_schema(partial_function.raw_f, ignore_first_argument=True)
-                else:
-                    method_schema = None
+                method_schema = get_callable_schema(
+                    partial_function._get_raw_f(),
+                    is_web_endpoint=partial_function._is_web_endpoint(),
+                    ignore_first_argument=True,
+                )
```
mwaskom, line 659: Could be nice if `get_callable_schema` could just accept the `Function`/`PartialFunction` (or ... `Interface`? Some other correct type?) and then consistently access the underlying callable and metadata like `is_web_endpoint`. Food for thought.

modal_proto/api.proto
```diff
@@ -1718,7 +1718,11 @@ message FunctionRetryPolicy {
 }
 
 message FunctionSchema {
-  int32 schema_version = 1; // 1 | 0 for now, allows easy disambiguation between empty schema and no schema collection
```
mwaskom: I didn't mind this as an int fwiw

modal/_serialization.py
```diff
@@ -529,9 +529,13 @@ def signature_to_parameter_specs(signature: inspect.Signature) -> list[api_pb2.C
 
 
 def get_callable_schema(
-    callable: typing.Callable, ignore_first_argument: bool = False
+    callable: typing.Callable, *, is_web_endpoint: bool, ignore_first_argument: bool = False
 ) -> typing.Optional[api_pb2.FunctionSchema]:
     # ignore_first_argument can be used in case of unbound methods where we want to ignore the first (self) argument
+    if is_web_endpoint or not config.get("function_schemas"):
+        # we don't support schemas on web endpoints for now
```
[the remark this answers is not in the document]
freider: probably not - would only make a little sense for fastapi_endpoint, and even there it would be kind of odd. fastapi also has its own openapi self-description support, which probably makes more sense to use in case of http endpoints

freider:
@prbot automerge

PR #2988, opened 2025-03-28
Added changelog entry for breaking parameter change

freider:
This was a somewhat unintended (yet partially intended...) side effect of a larger refactor. Making sure it's documented so that we and customers know which revision this was introduced in.

---

---

</details>


freider:
@prbot approve

PR #2990, opened 2025-03-28
Add bug report template and links to slack for support / feature requests

mwaskom:
We generally prefer communication over slack, because it plugs into our internal ticketing system and more engineers will see it (issues in the `modal-client` repo are monitored only by the small team that work directly on the SDK).


mwaskom:
I think we'll just need to merge this to see how it looks and then iterate if anything is wrong.

ekzhang (approved):
LGTM! Love issue templates

https://github.com/google/gvisor has some templates too if you're looking for syntax examples

Issue #2620, opened 2024-12-06
Modal run with local entrypoint completes, but produces traceback


rmorgans:
Running an app, using a `local_entrypoint` that uses a starmap to run remote functions (loading data)

```
inputs = [(sensor, installation, date) for date in sequential_dates]
    for events_loaded in load_by_installation_date.starmap(inputs):
        print(events_loaded)
```

The app appears to complete normally but does throw a traceback. Not sure what's going on, thought I would report.

```etl/detections on main [!?] via 🐍 v3.13.0 took 9s 
❯ modal run load__detections.py 
✓ Initialized. View run at https://modal.com/apps/XXXX/ap-FVxCHuON0HxpTlLn1cBtBR
✓ Created objects.
├── 🔨 Created mount /Users/rick/Documents/SS/git_checkout/etl/xxx/load__xxx.py
└── 🔨 Created function load_by_installation_date.
1440
Stopping app - local entrypoint completed.
[Thread-1 (thread_inner)] 2024-12-06T21:08:19+1030 Loop attempt for _run_app.<locals>.heartbeat failed
Traceback (most recent call last):
  File "/Users/rick/.local/share/uv/tools/modal/lib/python3.11/site-packages/modal/_utils/async_utils.py", line 195, in loop_coro
    await asyncio.wait_for(async_f(), timeout=timeout)
  File "/Users/rick/.local/share/uv/python/cpython-3.11.10-macos-aarch64-none/lib/python3.11/asyncio/tasks.py", line 489, in wait_for
    return fut.result()
           ^^^^^^^^^^^^
  File "/Users/rick/.local/share/uv/tools/modal/lib/python3.11/site-packages/modal/runner.py", line 56, in _heartbeat
    await retry_transient_errors(client.stub.AppHeartbeat, request, attempt_timeout=HEARTBEAT_TIMEOUT)
  File "/Users/rick/.local/share/uv/tools/modal/lib/python3.11/site-packages/modal/_utils/grpc_utils.py", line 184, in retry_transient_errors
    raise exc
  File "/Users/rick/.local/share/uv/tools/modal/lib/python3.11/site-packages/modal/_utils/grpc_utils.py", line 178, in retry_transient_errors
    return await fn(*args, metadata=metadata, timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2991, opened 2025-03-28
[CLI-354] Fix spurious error log on successful App shutdown

thundergolfer:
## Describe your changes

This issue has been around a long while, but a customer brought it up recently which motivated me in checking if it was a quick fix. I think it is. 

- https://linear.app/modal-labs/issue/CLI-354/app-heartbeat-fails-because-app-is-app-state-stopped-race
- https://github.com/modal-labs/modal-client/issues/2620

With this changeset I can now run the new test 100+ times without a fail @ `533088eb`. Previously my test would hit the error almost 100% of the time. 

---

---

</details>

## Changelog

* Fix shutdown race bug that emitted spurious error-level logs.


erikbern (approved):

mwaskom:
Amazing, thanks for tracking this down!

PR #2992, opened 2025-03-29
Revert traceback.print_exc() in resolver

ekzhang:
Reverts a small exception handling change in #2967

See https://github.com/modal-labs/modal-client/pull/2967/files/a77171c0638189509dfd13b21b91fc289ec47b36#r2019709823

Part of SPR-65

PR #2993, opened 2025-03-29
Add `modal.experimental.ipython` extension

ekzhang:
## Describe your changes

Implement IPython line magic in a not-yet stable `modal.experimental.ipython` module. Remove the ancient `modal.extensions.ipython` module that doesn't work and has no usages on GitHub.

I didn't want to take the existing module name since that would imply 1.0 stability guarantees which I don't think is appropriate for this yet. Modal Notebooks will be configured to load the `modal.experimental.ipython` module automatically, but I think it's good for us to include any custom Jupyter extensions we write also so they can be run in normal notebook environments with Modal (And it also makes it a bit easier to ship it)

See also #2992 which improves the output by removing a traceback printer.

Part of SPR-65.

## Changelog

- Added `modal.experimental.ipython` module, which can be loaded in Jupyter notebooks with `%load_ext modal.experimental.ipython`. Currently it provides the `%modal` line magic for looking up functions:

  ```python
  %modal from main/my-app import my_function, MyClass as Foo

  # Now you can use my_function() and Foo in your notebook.
  my_function.remote()
  Foo().my_method.remote()
  ```
- Removed the legacy `modal.extensions.ipython` module from 2022.


modal/experimental/ipython.py
```diff
@@ -0,0 +1,80 @@
...
+
+from ..cls import Cls
+from ..exception import NotFoundError
+from ..functions import Function
+
+
+@magics_class
+class ModalMagics(Magics):
+    @line_magic
+    def modal(self, line):
```
mwaskom, line 18: Have you thought about potential future applications for a `%modal` magic that this is going to squat on? E.g. I'd previously prototyped an idea where you could `%modal` in a local notebook and that cell would be turned into a Function that would execute in the cloud.

Not necessarily saying that specific conflict is a problem, just pointing out that `%modal` is a valuable name and we're just using it for one fairly narrow operation here (i.e., looking up a deployed Function or Cls).

modal/experimental/ipython.py
```diff
@@ -0,0 +1,80 @@
...
+        """Lookup a deployed Modal Function or Class.
+
+        **Example:**
+
+        ```python notest
+        %modal from main/my-app import my_function, MyClass as Foo
```
mwaskom, line 24: Curious if you think there's any risks for overloading the "import" concept here. I.e. in normal Python code, doing

```python
from my_modal_module import MyClass
```

does not give you a hydrated version of that class, just a pointer to the local object that's probably registered to a (non-running) App.

mwaskom (commented):
Had a couple narrow questions, mostly just wondering how you're thinking about growing the Notebook integrations.

Bigger picture, it's not totally clear to me what the advantage of using a "magic" spelling for the lookup operation is. These seem like the same amount of code

```
%modal from main/my-app import MyClass as Foo
```

```
Foo = modal.Cls.from_name("my-app", "main")
```

With the advantage that the latter is not a notebook-specific construct. Am I missing something about what the magic invocation provides?

modal/experimental/ipython.py
ekzhang: Yes exactly — I think `%modal` as a magic is probably a good standard for a lot of operations and I'm hoping we can parse the operation from what statement they mention from `%modal`. I'm only taking `%modal from` here for lookups.

For example, other `from` usages to import a volume or dict:

```
%modal from main.volume import my-volume as MyVolume
%modal from main.dict import my-dict as MyDict
```

And if there's a future block magic for running an app, it can start with a different command:

```
%%modal app run

@app.function() ...
```

or something like that

ekzhang:
> With the advantage that the latter is not a notebook-specific construct. Am I missing something about what the magic invocation provides?

Thanks for the comments. Yes, I think that's right — the primary advantage is not having to know the `environment_name` and import objects like `Cls` and `Function`.

Totally agree it's good to have a clean API like `from_name()` in addition to any magic syntax since that API would be the clear way to lookup a function from production code. I'm just interested in trying this out (experimentally) so we can improve the ergonomics of looking up a function from notebooks — the hydration here is also nice since it shows you errors if the function was not found.

modal/experimental/ipython.py
```diff
@@ -0,0 +1,80 @@
...
+        line_without_from = line[5:]
+        env_app_part, sep, import_part = line_without_from.partition(" import ")
+        if not sep:
+            print("Invalid syntax. Missing 'import' keyword.")
+            return
+
+        # Parse environment and app from "env/app"
+        if "/" not in env_app_part:
+            print("Invalid app specification. Expected format: <env>/<app>")
+            return
```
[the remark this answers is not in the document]
ekzhang: My feeling is that if the examples show `from main/my-app` then people will pattern-match and work out what `main` is, or we include a doc explaining this. Might actually be a good way of teaching people what environments are anyway!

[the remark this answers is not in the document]
mwaskom: > I'm not sure we have a default environment concept right now, but I'm happy to add it if we do have a default environment

We don't have a way of configuring what environment to use as the default, but we do have a defined behavior for when the environment is not provided: https://github.com/modal-labs/modal/blob/2bfefa96a467c55453c895f644d0de97db6a8fb1/server/modal_server/objects/environment.py#L210-L219

modal/experimental/ipython.py
mwaskom: > The syntax is "magic" so I'd believe that people probably wouldn't extrapolate to normal behavior from this one

Not all users will know that `%modal` makes the line magic right? Especially if we don't require you to always include `{env}/` it feels possible to mistake that for a Python import statement even if it's doing something different.

ekzhang:
> > Syntax convenience: ease of remembering how to import something in a quick notebook session
> 
> IMO this goal is somewhat undermined by introducing secondary syntax that conflicts with the way you'd do it "normally". Is the concern that notebook users won't typically have `modal` imported?

Yeah — I think that's right. Notebook users won't typically have `modal` imported, and also they probably won't have `from modal import Function, Cls` either. We have an implicit auth scope that lets you access functions in notebooks, but then you need to connect the dots and realize that you can import these objects from the `modal` client library.

modal/experimental/ipython.py
ekzhang: Hmm I think it's pretty clear that `%modal` makes the line magic. It's an unusual syntax for Python for people who are unfamiliar with notebooks, and it's familiar to people who are familiar with notebooks.

modal/experimental/ipython.py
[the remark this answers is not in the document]
ekzhang: Oh interesting! That's a neat fallback. Yeah I think that definitely seems a little bit error-prone to expose as a default in notebooks then, we can keep it explicit until notebooks actually have environments attached (if they do).

ekzhang:
> Similarly, my main concern here is about potentially making it harder to understand Modal's normal behaviors, i.e. that looking up a deployed instance of your Function or Cls is different from just referencing the object as defined locally (without "running the App").

I understand that it could make it harder to understand Modal client library behaviors. I guess, my primary goal is to make notebooks feel like a first-party product that's integrated with Modal and to improve the experience _within_ notebooks right now while we're still in beta — that's why this is a less stable part of the client library. I feel like any alternative to prevent us from _exploring_ this route would need to have an equally strong client UX for notebooks.

If people find it difficult to understand Modal client library behaviors then I would suggest revisiting this, since we have a lot fewer users of notebooks we can change it as needed.

But that's a bit far in the future — we don't know if people will even use this yet! And if people don't use it then it's a bit of a non-issue I think.

mwaskom:
> and also they probably won't have from modal import Function, Cls either

FWIW this isn't necessary as those names will be available in the top-level namespace

```
import modal
f = modal.Function("my-app", "f")
```

And this is the style we promote in all our docs these days.

mwaskom:
> we don't know if people will even use deployed functions in notebooks yet! And if people turn out to not use it, then it's a bit of a non-issue I think

I'm not totally sure I follow this line of reasoning since it would suggest to me that we hold off on introducing a new approach to function lookup that is optimized for notebooks until we learn that looking up deployed Functions is something (a) they do often and (b) find cumbersome to do via the usual client API. AFAICT the only notebooks-specific thing here is wanting to avoid `import modal` before you look up a Function. Is there something else that I'm missing?

ekzhang:
> optimized for notebooks until we learn that looking up deployed Functions is something (a) they do often and (b) find cumbersome to do via the usual client API

Yes, this is a part of the design of [notebooks](https://www.notion.so/modal-com/Modal-Notebooks-1571e7f16949809f877fd826d765879b) and an area we're exploring to use. Without user feedback and testing it out, it's inefficient to make progress on this, and I think it's important to have the ability to experiment with new features while we're at an early stage.

If you'd prefer that this would not be in the client library, I can publish a new library to PyPI like `modal-experimental` that we use here, but otherwise I'm blocked on being able to make progress on this feature.

ekzhang:
Thanks for all the concerns raised, super helpful — for now I'm going to add this into `experimental` to unblock the notebooks beta but let's discuss more

@prbot approve

mwaskom:
> AFAICT the only notebooks-specific thing here is wanting to avoid import modal before you look up a Function. Is there something else that I'm missing?

still interested in the answer to this question

erikbern:
Also somewhat in favor of promoting the "normal" Modal syntax as a way to "train" people on how to use the core Modal primitives, and to make it consistent.

PR #2994, opened 2025-03-31
pin all GitHub Actions versions by hash

charlesfrye:
## Describe your changes

Pins all GitHub Actions by their hash, which is immutable, instead of their version tag, which is not.

These versions must also be allow-listed in Actions settings, unless the Actions are provided by GitHub.

[SEC-48](https://linear.app/modal-labs/issue/SEC-48/pin-all-github-actions-that-we-use-to-a-specific-hash)


mwaskom (approved):
Didn't double check that these are all correct, but 👍 on the change

PR #2995, opened 2025-03-31
Less ambitious partial_functions refactor

mwaskom:
Second attempt at #2973. This approach has more annoying code repetition, but avoids some of the type annotation issues I was running into while still improving the implementation of decorator stacking and validation.

TBH I don't fully understand the intention of making the PartialFunction class generic over the wrapped signature elements so there's some copy-paste from the existing code without that much thought behind it!

The main goal here is to (in a consistent way with minimal code duplication):
- support decorator stacking
- impose constraints on stacking (and on decorator <->wrapped compatibility)

## Changelog

- Improved handling of cases where `@modal.concurrent` is stacked with other decorators.

### DIFF
--- a/modal/_partial_function.py
+++ b/modal/_partial_function.py
@@ -94,15 +94,6 @@ class _PartialFunction(typing.Generic[P, ReturnType, OriginalReturnType]):
     params: _PartialFunctionParams
     wrapped: bool  # TODO rename ("registered?")
 
-    # --- # TODO delete these after other typing issues are sorted out
-    webhook_config: Optional[api_pb2.WebhookConfig]
-    is_generator: bool
-    batch_max_size: Optional[int]
-    batch_wait_ms: Optional[int]
-    cluster_size: Optional[int]  # Experimental: Clustered functions
-    build_timeout: Optional[int]
-    force_build: bool
-
     def __init__(
         self,
         wrapped_obj: Callable[P, ReturnType],
@@ -119,10 +110,6 @@ class _PartialFunction(typing.Generic[P, ReturnType, OriginalReturnType]):
     def raw_f(self) -> Callable[P, ReturnType]:
         # TODO(michael): temporary to avoid needing changes elsewhere in the library
         return self.wrapped_obj
-
-    def __getattr__(self, name: str) -> Any:
-        # TODO(michael): temporary to avoid needing changes elsewhere in the library
-        return getattr(self.params, name)
 
     @property
     def wrapped_function(self) -> Callable[P, ReturnType]:
@@ -198,9 +185,9 @@ class _PartialFunction(typing.Generic[P, ReturnType, OriginalReturnType]):
         return self.wrapped_obj
 
     def _is_web_endpoint(self) -> bool:
-        if self.webhook_config is None:
+        if self.params.webhook_config is None:
             return False
-        return self.webhook_config.type != api_pb2.WEBHOOK_TYPE_UNSPECIFIED
+        return self.params.webhook_config.type != api_pb2.WEBHOOK_TYPE_UNSPECIFIED
 
     def __get__(self, obj, objtype=None) -> _Function[P, ReturnType, OriginalReturnType]:
         # to type checkers, any @method or similar function on a modal class, would appear to be
--- a/modal/_partial_function.py
+++ b/modal/_partial_function.py
@@ -3,11 +3,13 @@ import enum
 import inspect
 import typing
 from collections.abc import Coroutine, Iterable
+from dataclasses import asdict, dataclass
 from typing import (
     Any,
     Callable,
     Optional,
     Union,
+    cast,
 )
 
 import typing_extensions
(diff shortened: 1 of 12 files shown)


modal/_partial_function.py
```diff
@@ -29,80 +30,163 @@
...
+    cluster_size: Optional[int] = None
+    max_concurrent_inputs: Optional[int] = None
+    target_concurrent_inputs: Optional[int] = None
+    build_timeout: Optional[int] = None
+
+    def update(self, other: "_PartialFunctionParams") -> None:
+        """Update self with params set in other."""
+        for key, val in asdict(other).items():
+            if val is not None:
+                setattr(self, key, val)
```
freider, line 85: Should we raise an error if a not-None-valued key is set (again)? This should never be allowed to happen I think?

modal/_partial_function.py
```diff
@@ -29,80 +30,163 @@
...
 
 class _PartialFunction(typing.Generic[P, ReturnType, OriginalReturnType]):
-    """Intermediate function, produced by @enter, @build, @method, @web_endpoint, or @batched"""
+    """Object produced by a decorator in the `modal` namespace
+
+    The object will eventually by consumed by an App decorator.
+    """
 
-    raw_f: Callable[P, ReturnType]
+    obj: Callable[P, ReturnType]  # function, method, or class
```
freider, line 101: Wait, when do we allow `obj` to be a class? I think that would break a lot of assumptions in _PartialFunction and typing

modal/_partial_function.py
```diff
@@ -29,80 +30,163 @@
...
+        for key, val in asdict(other).items():
+            if val is not None:
+                setattr(self, key, val)
+
 
 P = typing_extensions.ParamSpec("P")
 ReturnType = typing_extensions.TypeVar("ReturnType", covariant=True)
 OriginalReturnType = typing_extensions.TypeVar("OriginalReturnType", covariant=True)
+NullaryFuncOrMethod = Union[Callable[[], Any], Callable[[Any], Any]]
+NullaryMethod = Callable[[Any], Any]
```
freider, line 92: Nice

modal/_partial_function.py
```diff
@@ -29,80 +30,163 @@
...
+    BUILD = 1  # Deprecated, will be removed
+    ENTER_PRE_SNAPSHOT = 2
+    ENTER_POST_SNAPSHOT = 4
+    EXIT = 8
+    # Interface flags
+    CALLABLE_INTERFACE = 16
+    WEB_INTERFACE = 32
+    # Service decorator flags
+    # It's, unclear if we need these, as we can also generally infer based on some params being set
+    # In the current state where @modal.batched is used _insead_ of `@modal.method`, we need to give
```
freider, line 43: insead typo

modal/_partial_function.py
mwaskom:
```
@app.cls()
@modal.concurrent()
class MyConcurrentService:
    ...
```

modal/_partial_function.py
mwaskom: talking about https://www.insead.edu obviously

modal/_partial_function.py
mwaskom: you mean like
```
        for key, val in asdict(other).items():
            if val is not None:
                if getattr(self, key, None) is not None:
                    raise InvalidError(f"Setting {key} twice")
                setattr(self, key, val)
```
?

freider: Yeah, but it's not super important. Maybe there are use cases for double-decoration for overriding what someone has already set too, so maybe it's a bad idea

modal/_partial_function.py
```diff
@@ -29,80 +30,165 @@
...
+        return self.obj
 
     def _is_web_endpoint(self) -> bool:
-        if self.webhook_config is None:
+        if self.params.webhook_config is None:
             return False
-        return self.webhook_config.type != api_pb2.WEBHOOK_TYPE_UNSPECIFIED
+        return self.params.webhook_config.type != api_pb2.WEBHOOK_TYPE_UNSPECIFIED
 
     def __get__(self, obj, objtype=None) -> _Function[P, ReturnType, OriginalReturnType]:
```
freider, line 193: **If** we want `obj` to possibly be a class, we might this to fail in that case? (as it's expected to only be used on functions)

Also this code gets quite hard to grok now since there is both `self.obj` , `obj` and `objtype` here, where the latter two are expected to refer to Modal class/instances (which we also call Obj) whereas `self.obj` in this case *must not* be a Cls, but is a raw python function 🤯

mwaskom: First rule of Python: "everything is an obj" :trollface:

modal/_partial_function.py
mwaskom: OK i can revert it too 🙃

mwaskom:
@prbot approve

mwaskom:
@prbot approve

PR #2996, opened 2025-03-31
Small fixes to type checking issues

mwaskom:
Been running the type checks a lot locally, just fixing a few things that pop up when you do that.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -123,8 +123,49 @@ def lint_protos(ctx):
 
 
 @task
+def type_stubs(ctx):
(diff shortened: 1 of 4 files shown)


tasks.py
```diff
@@ -123,8 +123,49 @@ def get_first_lineno_with_prefix(text: str, prefix: str) -> int:
 
 
 @task
+def type_stubs(ctx):
```
mwaskom, line 126: Just moving this above the `type_check` task so we can reference the dependency the "correct" invoke way (I was hoping it would do caching like Make, but apparently not).

PR #2997, opened 2025-04-02
Add read_only flag for VolumeMounts

gimaker:
This is preparatory work to be able to make the backend changes needed to support mounting volumes read-only.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2875,6 +2875,7 @@ message VolumeMount {
   string volume_id = 1;
   string mount_path = 2;
   bool allow_background_commits = 3;
+  bool read_only = 4;
 }
 
 message VolumePutFilesRequest {


modal_proto/api.proto
```diff
@@ -2875,6 +2875,7 @@ message VolumeMount {
   string volume_id = 1;
   string mount_path = 2;
   bool allow_background_commits = 3;
```
mwaskom, line 2877: Is this deprecated btw?

modal_proto/api.proto
```diff
@@ -2875,6 +2875,7 @@ message VolumeMount {
...
+  bool read_only = 4;
```
mwaskom, line 2878: Just to overthink things, is there any more general concept of "mode" or "permissions" that could end up being more useful than just having a boolean?

mwaskom (approved):
Added a discussion question, but not expecting it to require any changes.

modal_proto/api.proto
gimaker: I think so. Not sure if you can force disable it tbh.

modal_proto/api.proto
gimaker: I don't think there are any related non-orthogonal concerns. It's either read-only or read-write (there's no write-only mode).

There are other mount options you could theoretically want to set, like `NoExec`. But as long as the mount options aren't duplicated in a ton of different proto messages just adding another `bool` for that seems fine to me.

I think we can cross that bridge when we get there.

modal_proto/api.proto
gimaker: The backend still respects this setting if you set it to `false`. So we can't remove it just quite yet.

PR #2998, opened 2025-04-02
Note that context_mount is deprecated in Image docstrings

mwaskom:
Adds comments to Image docstrings that have a `context_mount`. We fire a runtime warning about this being deprecated (we now infer the context from the Dockerfile commands so theres no obligation to provide an explicit Mount). A user pointed out that it wasn't clear from the docs that they shouldn't go down this path.

I noticed also that `from_dockerfile` has `context_mount` as the second, non-kwarg-required parameter. That's going to make it annoying to fully remove.


mwaskom:
@prbot approve, just a tiny docs change

freider (commented):
👍

PR #2999, opened 2025-04-03
Api change to allow server to configure max outstanding inputs

rculbertson:
Part of SVC-442

Adds `max_inputs_outstanding` to `FunctionMapResponse` so from the server we can control the number outstanding inputs in the client during a .map call.

---

</details>


rculbertson:
@prbot approve

gongy:
thanks!

PR #3000, opened 2025-04-03
Honor max_inputs_outstanding field in FunctionMapResponse

rculbertson:
Closes SVC-442

There is a semaphore which controls the maximum number of inputs that can be in progress - either queued to be sent, or waiting for completion. If this limit is reached, we will block sending more inputs to the server until some of the existing inputs are completed.

This PR makes it so we can configure this value server-side. 

---

</details>

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -70,6 +70,7 @@ class _OutputValue:
     value: Any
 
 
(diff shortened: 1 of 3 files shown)


gongy (approved):
LGTM!

rculbertson:
@prbot automerge

PR #3001, opened 2025-04-03
Add experimental functions for building Image without client dependencies

mwaskom:
Adds `modal.experimental.raw_dockerfile_image` and `modal.experimental.raw_registry_image`.

These build a `modal.Image` directly from a provided Dockerfile/registry tag with no additional embellishment (i.e. no installation of modal client dependencies. Such Images will only work directly with Sandboxes (although it should be possible to manually install the client dependencies in subsequent image layers and then use the Image in a Function).

This is intended as a stop-gap helper until we ship Image Builder 2025.04 which will change how we include the client dependencies in our images.

Closes CLI-359

### DIFF
--- a/modal/experimental/__init__.py
+++ b/modal/experimental/__init__.py
@@ -2,7 +2,7 @@
 import os
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Optional, Union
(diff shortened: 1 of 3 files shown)


mwaskom:
@prbot approve

aksh-at (commented):
neat!

PR #3002, opened 2025-04-04
Add context_dir to Docker-oriented Image methods

mwaskom:
Re-establishes some indirection between where you're building the Image and where you'd like to COPY files from, which was previously afforded via the `context_mount` concept.

Closes CLI-360

</details>

## Changelog

- Adds a `context_dir` parameter to `modal.Image.from_dockerfile` and `modal.Image.dockerfile_commands`. This parameter can be used to provide a local reference for relative COPY commands.


freider (approved):
Makes sense! Should add some tests

PR #3003, opened 2025-04-05
sandboxes: reduce timeouts within while loops; add repr

thundergolfer:
## Describe your changes

This is intended to lower the maximum time these loops can be 'stuck' waiting on an existing RPC. We should rework the RPCs themselves to better support this long-polling, but in the short term I don't see a downside to using a much lower but still large timeout. 

---

---

</details>


pawalt (approved):
Should we do `io_streams` while we're in the area? Something like the sandbox log reading or the exec log reading could benefit from this.

PR #3004, opened 2025-04-07
[FS-209] modal_proto: Add VolumePutFiles2 RPC

dflemstr:
## Describe your changes

Adds a proposed schema for the new API to upload files into a version 2 volume.

The idea is that the client will call `VolumePutFiles2` repeatedly, until the server no longer reports any missing chunks, at which point the upload is complete.

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2880,11 +2880,17 @@ message VolumeMount {
 }
 
 message VolumePutFiles2Request {
-  // Id of volume to put files into.
-  string volume_id = 1;
-
   // List of files, sorted lexicographically by `path`.
-  repeated File files = 2;
+  repeated File files = 1;
+
+  // The last time the client called `VolumePutFiles2` for this file, it was
+  // told that some blocks were missing. This field contains information
+  // about the client having uploaded those missing blocks.
+  repeated NewBlock new_blocks = 2;
+
+  // If set to true, prevent overwriting existing files. (Note that we don't
+  // allow overwriting existing directories with uploaded files regardless.)
+  bool disallow_overwrite_existing_files = 3;
 
   message File {
     // Destination path of the file to be uploaded, including any parent dirs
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -2879,6 +2879,70 @@ message VolumeMount {
   bool read_only = 4;
 }
 
+message VolumePutFiles2Request {
+  // Id of volume to put files into.
+  string volume_id = 1;
+
+  // List of files, sorted lexicographically by `path`.
+  repeated File files = 2;
```
gimaker, line 2887: Do we need the sorting requirement? Are we planning to enforce it?

gimaker (approved):

modal_proto/api.proto
```diff
@@ -2879,6 +2879,70 @@ message VolumeMount {
...
+    // The raw bytes of the body that was returned from the HTTP PUT request
+    // when the client made a request for the `put_url` returned in the
+    // previous `VolumePutFiles2Response`.
+    bytes put_response = 2;
+  }
+}
+
+message VolumePutFiles2Response {
+  // Id from the original `VolumePutFiles2Request`
+  string volume_id = 1;
```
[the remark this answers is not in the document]
dflemstr: I'll remove it :)

modal_proto/api.proto
```diff
@@ -2879,6 +2879,70 @@ message VolumeMount {
...
+  message NewBlock {
+    // The index of the block, ie. offset into the file in units of 8MiB.
+    uint64 index = 1;
+
+    // The raw bytes of the body that was returned from the HTTP PUT request
+    // when the client made a request for the `put_url` returned in the
+    // previous `VolumePutFiles2Response`.
+    bytes put_response = 2;
+  }
+}
```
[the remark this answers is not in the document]
dflemstr: I can add that here for completeness.

modal_proto/api.proto
dflemstr: I want to enforce it because it means the caller is aware of the order, which is essential since we refer to files by index in the response. It is not strictly necessary, though.

modal_proto/api.proto
```diff
@@ -2879,6 +2879,70 @@ message VolumeMount {
...
+    uint64 size = 2;
+
+    // SHA-256 checksum of each 8MiB block of the file's contents, in binary
+    // (ie 32 raw bytes) format for compactness.
+    repeated bytes blocks_sha256 = 3;
+
+    // The last time the client called `VolumePutFiles2` for this file, it was
+    // told that some blocks were missing. This field contains information
+    // about the client having uploaded those missing blocks.
+    repeated NewBlock new_blocks = 4;
```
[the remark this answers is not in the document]
dflemstr: We'd still get the SHA256's by doing `files[file_index].blocks_sha256[block_index]`

modal_proto/api.proto
dflemstr: I want to apply the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle) here, that we should suggest a desired format, but be liberal in what we accept. Hence, I think it makes sense to keep this comment but to tolerate unsorted input on the server side.

modal_proto/api.proto
```diff
@@ -2879,6 +2879,71 @@ message VolumeMount {
   bool read_only = 4;
 }
 
+message VolumePutFiles2Request {
+  // List of files, sorted lexicographically by `path`.
```
freider, line 2883: Curious as to why the list needs to be sorted here?

modal_proto/api.proto
```diff
@@ -2879,6 +2879,71 @@ message VolumeMount {
...
+    // Destination path of the file to be uploaded, including any parent dirs
+    // etc.; for example "foo/bar/baz.txt"
+    string path = 1;
+
+    // The total size of the file, in bytes.
+    uint64 size = 2;
+
+    // SHA-256 checksum of each 8MiB block of the file's contents, in binary
+    // (ie 32 raw bytes) format for compactness.
+    repeated bytes blocks_sha256 = 3;
```
freider, line 2905: Could this be a `repeated Block = 3` instead where 
Block
```
message Block {
    block_sha256 = 1;
    optional bytes put_response = 2; // only set for new blocks when requested
}
```
That would get rid of all the file/block index stuff if I'm not mistaken?

freider (commented):
Makes sense! Added some inline comments with thoughts

modal_proto/api.proto
dflemstr: I think it is good practice to always have a well-defined structure of requests/responses, see earlier discussion here: https://github.com/modal-labs/modal-client/pull/3004#discussion_r2032750234

PR #3006, opened 2025-04-07
Address race condition in .map

rculbertson:
Closes SVC-444

## Issue 
There is a race condition in .map which causes us to make an extra call to `FunctionGetOutputs`, even though we've already received all the outputs. We then wait 55 seconds for the `FunctionGetOuputs` call to time out. Then .map see that we have all the outputs and returns.

Our HFMs hit this because the .map calls do no work, and return immediately. Our HFMs timeout after 20 seconds, so they are reporting this as a failure. For a user, the call would succeed, it would just take 55 seconds longer than it should. 

This happens if we receive all the outputs before `pump_inputs` has a chance to mark the `have_all_inputs` flag as true. 

Here is the sequence of events:
1. `have_all_inputs` is false
2. `pump_inputs` pushes all inputs to the server, but due to async behavior, the function does not complete and yields to someone else
3. the server process all inputs and sends outputs
4. `get_all_outputs` sees that `have_all_outputs` is false, and makes another call to `FunctionGetOutputs`
5. `pump_inputs` resumes, sees there are no more inputs to consume, and marks `have_all_inputs` as False
6. After 55 seconds, `FunctionGetOuputs` returns and .map completes 

## How to address it
This PR makes it so we set `have_all_inputs` to True in `drain_input_generator` instead of `pump_inputs`. This means we set the flag much sooner, reducing the chances of hitting the race condition. 

From a few hours of testing this change, it greatly improves things, but the issue can still happen. It looks to me like fixing it completely would mean non-trivial changes. I'm not sure if it's worth the effort at this point, given this seems like an edge case. I propose we go with this fix for now and see what impact it has on the HFMs.

The top box is without the fix, and the bottom box is with it.
image

#

## Changes to tests
There were a few tests which took advantage of the extra call to FunctionGetOutputs to change state in the tests. The changes to the tests are so that the tests no longer rely on this issue. 

#

---

</details>

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -156,11 +156,12 @@ async def _map_invocation(
 
         # close queue iterator
         await input_queue.put(None)
+        have_all_inputs = True
         yield
 
     async def pump_inputs():
         assert client.stub
-        nonlocal have_all_inputs, inputs_created, inputs_sent
+        nonlocal inputs_created, inputs_sent
         async for items in queue_batch_iterator(input_queue, max_batch_size=MAP_INVOCATION_CHUNK_SIZE):
             # Add items to the manager. Their state will be SENDING.
             await map_items_manager.add_items(items)
@@ -182,7 +183,6 @@ async def _map_invocation(
                 f"Successfully pushed {len(items)} inputs to server. "
                 f"Num queued inputs awaiting push is {input_queue.qsize()}."
             )
-        have_all_inputs = True
         yield
 
     async def retry_inputs():
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -62,6 +62,11 @@ class VolumeFile:
     data_blob_id: str
     mode: int
 
+@dataclasses.dataclass
+class GrpcErrorAndCount:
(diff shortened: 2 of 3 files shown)


gongy:
This change looks sound to me -- although the delay can still happen as you noted.

IIUC it can also cause delays in non-trivial maps, if `have_all_inputs=True` happens during an ongoing `FunctionGetOutputs` call.

Thinking aloud, would a more direct fix involve interrupting `FunctionGetOutputs` when we have both all inputs and outputs?

rculbertson:
> IIUC it can also cause delays in non-trivial maps, if have_all_inputs=True happens during an ongoing FunctionGetOutputs call.

That sounds like what's happening in this case (but let me know if I'm misunderstanding)? We've consumed all the outputs, but `have_all_inputs` is False, so we call `FunctionGetOutputs`. While that's waiting, `have_all_inputs` is set to True. But we need to wait the full 55 seconds for `FunctionGetOutputs` to return. 

Cancelling the call to FunctionGetOuputs is a good idea. That would fix the delay. I wonder if it's possible to prevent making the unnecessary call to `FunctionGetOuputs` altogether. Maybe something like:

PR #3007, opened 2025-04-07
Add version added note to modal.concurrent docstring

mwaskom:
Small docs change.

Technically we added this a few versions earlier but there were a few kinks to shake out. The provided version will work better than earlier ones.


mwaskom:
@prbot approve

PR #3012, opened 2025-04-08
Format python version in a more structured way

erikbern:
Some weird Python versions have a plus at the end, idk why. E.g. `"3.10.6+"`. Our backend expects this header to be a proper version string.

### DIFF
--- a/modal/client.py
+++ b/modal/client.py
@@ -2,6 +2,7 @@
 import asyncio
 import os
 import platform
+import sys
 import warnings
 from collections.abc import AsyncGenerator, AsyncIterator, Collection, Mapping
 from typing import (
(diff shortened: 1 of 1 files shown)


rohansingh:
https://github.com/python/cpython/issues/99968

mwaskom (approved):
Been a few years since I've seen `%` interpolation in Python; fun throwback 😁

erikbern:
> Been a few years since I've seen `%` interpolation in Python; fun throwback 😁

I used it in this case because it's more strict with the types. `%d` requires the argument to be an integer. Vs an f-string which would have happily accepted any type.

But realized I guess I could have used f-strings with `:d` also. Idk. Maybe this is cleaner?

```python
python_version = f"{sys.version_info.major:d}.{sys.version_info.minor:d}.{sys.version_info.micro:d}"
```

PR #3013, opened 2025-04-08
Improve error messages for failed Function and Cls lookups

mwaskom:
When Function or Cls lookups fail with a NotFound status, the server doesn't include the name of the object that was not found. This can make it hard to diagnose issues if you are looking up an object programatically. We have that information on the client side, so we can enrich the Python exception that we raise:

image


thundergolfer (approved):
LGTM. Thanks for fixing this.

PR #3014, opened 2025-04-09
Add `return_exceptions` argument to `modal.FunctionCall.gather`.

danielshaar:
## Describe your changes

This is possible in the async context by using asyncio.gather directly, but we want to expose this in a sync way for users as well.

## Changelog

- Adds a `return_exceptions` argument to `modal.FunctionCall.gather` to allow gathering failed and successful results from a collection of `FunctionCall` objects.

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1768,7 +1768,7 @@ class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
         `Function.spawn()`).
 
         By default, this will raise an exception from the first failing function call. If `return_exceptions=True`, it
-        will instead return a list of results, with exceptions aggregated in the results list.
+        will instead aggregate exceptions in the results list.
 
         Examples:
 
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1760,7 +1760,10 @@ class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
         return fc
 
     @staticmethod
-    async def gather(*function_calls: "_FunctionCall[T]") -> typing.Sequence[T]:
+    async def gather(
+        *function_calls: "_FunctionCall[T]",
+        return_exceptions: bool = False,  # propagate exceptions (False) or aggregate them in the results list (True)
+    ) -> typing.Sequence[T | BaseException]:
         """Wait until all Modal FunctionCall objects have results before returning.
 
         Accepts a variable number of `FunctionCall` objects, as returned by `Function.spawn()`.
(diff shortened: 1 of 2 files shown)


mwaskom (approved):
Would be great if we had a test but not sure how much effort it would be to update the relevant fixtures to support it. (Our remote function call mocking is pretty thin right now).

test/function_test.py
```diff
@@ -496,6 +496,24 @@ def test_function_exception(client, servicer):
...
+    failure_modal = app.function()(failure)
+    with app.run(client=client):
+        with pytest.raises(CustomException) as excinfo:
+            FunctionCall.gather(failure_modal.spawn(), failure_modal.spawn())
+        assert "foo!" in str(excinfo.value)
+
+    with app.run(client=client):
+        results = FunctionCall.gather(failure_modal.spawn(), failure_modal.spawn(), return_exceptions=True)
+        for result in results:
+            assert isinstance(result, UserCodeException)
```
mwaskom, line 512: What is this `UserCodeException` type?

test/function_test.py
```diff
@@ -496,6 +496,24 @@ def test_function_exception(client, servicer):
...
+    servicer.function_body(failure)
+    failure_modal = app.function()(failure)
+    with app.run(client=client):
+        with pytest.raises(CustomException) as excinfo:
+            FunctionCall.gather(failure_modal.spawn(), failure_modal.spawn())
+        assert "foo!" in str(excinfo.value)
+
+    with app.run(client=client):
+        results = FunctionCall.gather(failure_modal.spawn(), failure_modal.spawn(), return_exceptions=True)
+        for result in results:
```
mwaskom, line 511: Might be nice to assert that you can have a mixture of exceptions with expected successful results.

modal/_functions.py
```diff
@@ -1760,13 +1760,15 @@ async def from_id(function_call_id: str, client: Optional[_Client] = None) -> "_
         return fc
 
     @staticmethod
-    async def gather(*function_calls: "_FunctionCall[T]") -> typing.Sequence[T]:
```
mwaskom, line 1763: This is going to mean that the return type is "wrong" in the case of an exception right? Any idea how `asyncio.gather` handles this? Might need to `@overload`

freider: +1 - add an @overload which differentiates the return type based on a Literal[True] value for the return_exceptions argument, and otherwise only returns Sequence[`T`] - that should fix the existing type assertion

test/function_test.py
freider: After some more investigation, the problem might be worse than I thought, apparently the client has some explicit wrapping using UserCodeException too in _process_result, which seems like the more likely culprit here 🤯

mwaskom: We need to get a 😵‍💫 option for GitHub reactions 😁

freider: Ohhhh no, `synchronicity.UserCodeException` is apparently being returned by `.map(return_exceptions=True)` as well :(

Now I'm not sure what to do - will have to discuss in client team...

PR #3018, opened 2025-04-11
Don't use an extra rpc to fetch serialized function

freider:
Note: this can only be merged once backend changes are live to ensure this data is delivered in the container args file

---

---

</details>

## Changelog

* Internal change that makes containers for functions/classes with `serialized=True` start up *slightly* faster than before


freider:
@prbot approve

modal/_container_entrypoint.py
```diff
@@ -394,7 +394,17 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
...
+
+            if function_def.function_serialized:
+                ser_fun = deserialize(function_def.function_serialized, _client)
+            else:
+                ser_fun = None
+
+            if function_def.class_serialized:
+                ser_usr_cls = deserialize(function_def.class_serialized, _client)
+            else:
+                ser_usr_cls = None
```
mwaskom, line 407: I really dislike all these parallel paths for function / class. But I guess we have to live with it.

freider: Parts of it we'll have to live with, but as soon as we stop supporting `@build` methods we can remove a bunch of it (the parts that support having *both* a function and a class 😵 )

PR #3019, opened 2025-04-11
Adds user_code_imports to pyright allowlist

freider:
* Adds pyright testing of user_code_imports.py
* Fixes a bunch of existing type errors that surfaced some possible bugs
* Refactors some of the `@build` function special case logic to prevent these bugs
* Removes the use of app fallbacks in the main container_entrypoint.py and makes the import code responsible for all app inference
* Removes the use of class param values for single-function service creation, since the only time it uses classes is for `@build` methods and those don't support non-default parameters

---
We'll be able to remove *a lot* of the messy code in user_code_imports.py as soon as we deprecate `@build` methods.
I also realized a bug with `@build` methods that is currently present (but probably not worth fixing since we are deprecating these methods soon) - it does not respect `serialized=True` if set on the class - the method will still be imported. This also means it won't work with notebooks...

## Describe your changes

---

---

</details>


modal/_runtime/user_code_imports.py
```diff
@@ -147,6 +148,7 @@ def get_finalized_functions(
         finalized_functions = {}
         for method_name, _partial in self._partial_functions.items():
             user_func = _partial.raw_f
+            assert user_func
```
mwaskom, line 151: Any chance this could be null with `_partial.user_cls` being not null? 🫣

modal/_runtime/user_code_imports.py
```diff
@@ -187,7 +189,7 @@ def get_user_class_instance(_cls: modal.cls._Cls, args: tuple, kwargs: dict[str,
 
     TODO: Could possibly change this to use an Obj to clean up the data model? would invalidate isinstance checks though
     """
-    cls = synchronizer._translate_out(_cls)  # ugly
+    cls = typing.cast(modal.cls.Cls, synchronizer._translate_out(_cls))  # ugly
```
mwaskom, line 192: maybe add an exclamation point to the comment 😁

modal/_runtime/user_code_imports.py
```diff
@@ -226,11 +226,15 @@ def import_single_function_service(
     """
     user_defined_callable: Callable
     service_deps: Optional[Sequence["modal._object._Object"]] = None
-    active_app: Optional[modal.app._App] = None
+    active_app: modal.app._App
+
+    user_cls_or_cls: typing.Union[None, type, modal.cls.Cls]
```
mwaskom, line 231: Why Union here rather than Optional? (jw if there's a reason)

mwaskom (approved):
I definitely don't follow all of this but 👍 great to get more type checking coverage.

modal/_runtime/user_code_imports.py
freider: (I still prefer to use Optional[] when the arg type is a non-union, if that makes sense

test/supports/user_code_import_samples/func.py
```diff
@@ -1,7 +1,11 @@
 # Copyright Modal Labs 2024
+
 from modal import App
 
-app = App(include_source=True)  # TODO: remove include_source=True when automount is disabled by default
+# TODO: remove include_source=True when automount is disabled by default
+
+print("IMPORT", __name__)
```
mwaskom: Missed one?

PR #3022, opened 2025-04-12
Support for modal run:ing class methods defined with modal.parameter

freider:
A bit of code repetition across different cases in order to have the same internal signature representation work for both classes and functions.

I am itching to rewrite this CLI signature stuff using the `_type_manager.py` we use for schemas and parameter serde, but won't do that right now

---

---

</details>

## Changelog

* Fixes a bug where `modal run` didn't recognize `modal.parameter()` class parameters

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -10,7 +10,7 @@ import time
 import typing
 from dataclasses import dataclass
 from functools import partial
-from typing import Any, Callable, Optional, get_type_hints
+from typing import Any, Callable, Optional
 
 import click
 import typer
@@ -77,7 +77,7 @@ class CliRunnableSignature:
 
 def safe_get_type_hints(func_or_cls: typing.Union[Callable[..., Any], type]) -> dict[str, type]:
     try:
-        return get_type_hints(func_or_cls)
+        return typing.get_type_hints(func_or_cls)
     except Exception as exc:
         # E.g., if entrypoint type hints cannot be evaluated by local Python runtime
         msg = "Unable to generate command line interface for app entrypoint. See traceback above for details."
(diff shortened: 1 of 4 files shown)


modal/cli/run.py
```diff
@@ -223,7 +224,10 @@ def _get_click_command_for_cls(app: App, method_ref: MethodReference):
     cls = method_ref.cls
     method_name = method_ref.method_name
 
-    cls_signature = _get_signature(cls._get_user_cls())
+    user_cls = cls._get_user_cls()
+    type_hints = safe_get_type_hints(user_cls)
+    sig: inspect.Signature = _get_class_constructor_signature(user_cls)
```
freider, line 229: We already had this utility for getting the signature of a modal class (regardless if it has an explicit or implicit constructor), so with some refactoring we could use it here

freider:
@prbot approve

freider:
@prbot automerge

modal/cli/run.py
```diff
@@ -77,10 +77,10 @@ class CliRunnableSignature:
 
 def safe_get_type_hints(func_or_cls: typing.Union[Callable[..., Any], type]) -> dict[str, type]:
     try:
-        return get_type_hints(func_or_cls)
+        return typing.get_type_hints(func_or_cls)
     except Exception as exc:
         # E.g., if entrypoint type hints cannot be evaluated by local Python runtime
-        msg = "Unable to generate command line interface for app entrypoint. See traceback above for details."
+        msg = "Unable to generate command line interface for app entrypoint due to unparseable type hints:\n" + str(exc)
```
mwaskom, line 83: This is tough to communicate about in general, but "unparseable" might be misleading in some contexts right? There are type annotations that can be "parsed" in some narrow sense but that we can't turn into a CLI option.

PR #3024, opened 2025-04-13
Add support for multiple gRPC stubs

thecodingwizard:
This PR adds support for a Modal GRPC stub to accept different server URLs. This should have no impact on production.

## Why?

The distributed input plane needs to route function invocations to different server URLs. Specifically, the `_Invocation` class will make RPCs against a particular server URL.

This PR is a prerequisite to doing that.

## Background

Our gRPC stack works as follows:

- A gRPC `Channel` represents one connection to the server
- `inv protoc` generates a grpclib stub, which accepts a `Channel` argument
- `inv protoc` also generates a Modal wrapper around this grpclib stub. Our code interacts with this Modal wrapper, not the grpclib stub

We have a Modal wrapper because:

1. When the process forks, we seamlessly re-establish a connection to the server in the new process
2. When the client is restored from a snapshot, we re-initialize the client and the connection to the server
3. We want to run the RPC in our asyncio task cancellation context

How do we re-establish the connection to the server? `_Client` tracks both `_stub` (Modal wrapper) and `_grpclib_stub`. Whenever an RPC is made, our Modal wrapper intercepts the RPC and tells the client to handle it. If `_Client` needs to re-establish the connection, it will _recreate_ `_grpclib_stub`, passing in a new `Channel`.

## Proposed API

```py
stub = await client.get_stub(server_url)

# use stub as you would normally
```

## Implementation

Only `UnaryUnaryWrapper`, `UnaryStreamWrapper`, and the protobuf plugin interacts with the "grpclib stub". Everything else, including the Client, is only aware of our Modal stub wrapper.

The `Client` is responsible for exposing a `get_channel(server_url: str) -> grpclib.Channel` method. This method returns cached channels when possible, and will be called on every RPC. If the process is forked, it will recreate the channel.

Important change: when the connection to the server is reestablished, `Unary[...]Wrapper` overrides the _existing_ grpclib method's `channel` directly, rather than recreating a grpclib stub. Refer to the [grpclib code](https://github.com/vmagamedov/grpclib/blob/master/grpclib/client.py#L844) to see what I mean. Since overriding the channel is not explicitly supported by the grpclib code, it's possible this will break in a future version of grpclib. However, the grpclib code doesn't appear to change very often.

As an alternative to overriding the `channel`, we could recreate the grpclib stub with a new channel. However, this will require us to track both the Modal wrapper stub and the grpclib stub, for every `(Stub, sever_url)` combination. (Actually, if we're committed to using a single `Stub` type, we can reduce some complexity)

<details>
<summary>One alternative Implementation</summary>

```py
class ModalStubWrapper(ABC):
    _grpclib_stub_type: Any
    _api_endpoint: str

    @abstractmethod
    def __init__(self, grpclib_stub: Any, client: "modal.client._Client", api_endpoint: str) -> None:
        self._api_endpoint = api_endpoint


ModalStubType = TypeVar("ModalStubType", bound=ModalStubWrapper)


class ConnectionPool:
    def __init__(self, client: "modal.client._Client", metadata: dict[str, str] = {}):
        self._client = client
        self._metadata = metadata
        self._channels: dict[str, grpclib.client.Channel] = {}
        self._grpclib_stubs: dict[tuple[str, Any], Any] = {}
        self._wrapper_stubs: dict[tuple[str, Any], Any] = {}

    async def _get_or_create_channel(self, server_url: str) -> grpclib.client.Channel:
        if server_url not in self._channels:
            self._channels[server_url] = create_channel(server_url, self._metadata)
            try:
                await connect_channel(self._channels[server_url])
            except OSError as exc:
                raise ConnectionError("Could not connect to the Modal server.") from exc
        return self._channels[server_url]

    async def _get_grpclib_stub(self, api_endpoint: str, cls: Any) -> Any:
        key = (api_endpoint, cls)
        if key not in self._grpclib_stubs:
            channel = await self._get_or_create_channel(api_endpoint)
            self._grpclib_stubs[key] = cls(channel)
        return self._grpclib_stubs[key]

    async def get_stub(self, api_endpoint: str, cls: typing.Type[ModalStubType]) -> ModalStubType:
        key = (api_endpoint, cls)
        if key not in self._wrapper_stubs:
            grpclib_stub = await self._get_grpclib_stub(api_endpoint, cls._grpclib_stub_type)
            self._wrapper_stubs[key] = cls(grpclib_stub, self._client, api_endpoint)
        return self._wrapper_stubs[key]
    def close(self):
        for channel in self._channels.values():
            channel.close()
        self._channels.clear()
        self._grpclib_stubs.clear()
```

Another alternative would be to have the Modal wrapper track and recreate the grpclib stub.

</details>


modal/client.py
```diff
@@ -284,31 +282,33 @@ async def _reset_on_pid_change(self):
...
-        return getattr(self._grpclib_stub, method_name)
+        stub = await self._connection_pool.get_grpclib_stub(api_endpoint)
+        return getattr(stub, method_name)
 
     @synchronizer.nowrap
     async def _call_unary(
         self,
         method_name: str,
         request: Any,
+        api_endpoint: str,
```
erikbern, line 306: Pretty confused by the terminology here. Do you mean hostname?

Endpoint to me is the url of a specific resource i.e. in this case a grpc method handler

thecodingwizard: Good point. I renamed `api_endpoint` to `server_url`. Thanks!

modal/client.py
```diff
@@ -366,30 +362,37 @@ async def __call__(
         metadata: Optional[_MetadataLike] = None,
     ) -> ResponseType:
         if self.client._snapshotted:
-            logger.debug(f"refreshing client after snapshot for {self._wrapped_method_name}")
+            logger.debug(f"refreshing client after snapshot for {self.name.rsplit('/', 1)[1]}")
             self.client = await _Client.from_env()
```
thecodingwizard, line 366: unrelated to this PR, but:

- `await _Client.from_env()` is called once _for every unique RPC_ that is made after snapshotting
- I don't fully understand why we need to completely recreate the client -- isn't it sufficient to just reestablish connections to the server, similar to what we do when the process forks?

modal/_functions.py
```diff
@@ -135,9 +135,11 @@ async def create(
         client: _Client,
         function_call_invocation_type: "api_pb2.FunctionCallInvocationType.ValueType",
     ) -> "_Invocation":
-        assert client.stub
+        io_plane_url = client.server_url  # TODO(nathan): replace with per-function endpoint
+        stub = await ModalClientModal.create(client, io_plane_url)  # TODO(nathan): decide if this is too expensive
```
thecodingwizard, line 139: Also wanted to call this out. This is `O(# of gRPC methods)`. I'm assuming this is OK, but it would be good to get confirmation from others.

modal/client.py
```diff
@@ -366,30 +362,37 @@ async def __call__(
...
-        return await self.client._call_unary(self._wrapped_method_name, req, timeout=timeout, metadata=metadata)
+        self.wrapped_method.channel = await self.client._get_channel(self.server_url)
```
thecodingwizard, line 367: This is the part that might be too sketchy: We override the grpclib method's channel (see [grpclib's code](https://github.com/vmagamedov/grpclib/blob/62f968a4c84e3f64e6966097574ff0a59969ea9b/grpclib/client.py#L844)). I think this is fine since grpclib's code doesn't seem to change very much, but we can also implement this differently if we aren't comfortable with this.

`client._get_channel` will return a cached channel, so this is fast unless the process was forked.

modal/_functions.py
```diff
@@ -136,8 +136,10 @@ async def create(
         function_call_invocation_type: "api_pb2.FunctionCallInvocationType.ValueType",
     ) -> "_Invocation":
         assert client.stub
+        stub = client.stub
```
thecodingwizard, line 139: Eventually, this will be

```py
        stub = await ModalInputPlaneModal.create(client, function._input_plane_server_url)
```

This constructor is `O(# of gRPC methods)`; for now, there are only two gRPC methods. I'm assuming this is OK, but it would be good to get confirmation from others.

If this is too slow, we can add a `(StubType, server URL) -> Stub` cache, but personally I think it's not necessary.

erikbern:
Thinking about it some more, I wonder if it's cleaner if the client just connects to a single host (like an "access point") which then handles potential forwarding to other places?

thecodingwizard:
That's a great question! We considered that approach [here](https://www.notion.so/modal-com/Multiple-API-endpoints-for-distributed-input-plane-1c11e7f1694980b3943ddcd01f7f8dfc) but decided not to go with it since we thought the server implementation for forwarding requests would require substantially more effort than having the client choose which server URL to connect to.

freider:
I've been considering removing the whole ClientClosed thing, since I think it causes more issues than it fixes. If we also remove the whole "call safely" check for running in the correct event loop (which is mostly to detect regressions in this code which I think would be better done in test code if possible), we eliminate the entire need of the `_call_safely` stuff.

Do you think such a change would impact how you structure this code? Seems like it would simplify a lot and almost remove the need of our own plugin/wrapper (although we have talked about moving endpoint retries etc. into this wrapper, so there may be other usages for it)

thecodingwizard:
I'm not sure if removing ClientClosed would reduce the need for our own plugin / wrapper. I think the custom wrapper is necessary if we want to automatically recreate Channels after a client has been snapshotted or after the process has been forked. We basically need to run some code before every gRPC call to recreate the channel if necessary, which we do by having a custom wrapper around the grpclib stub. Maybe there's an easier approach I'm missing though?

freider:
> I'm not sure if removing ClientClosed would reduce the need for our own plugin / wrapper. I think the custom wrapper is necessary if we want to automatically recreate Channels after a client has been snapshotted or after the process has been forked. We basically need to run some code before every gRPC call to recreate the channel if necessary, which we do by having a custom wrapper around the grpclib stub. Maybe there's an easier approach I'm missing though?

Yeah you might be right - there is t

modal/client.py
freider: For the second bullet - It could have something to do with the auth credentials possibly changing from the snapshot to the resume (iirc I didn't work on this directly, but I have a vague memory of something of this effect being an issue)

modal/_functions.py
freider: Probably fine, but something about it itches especially given that the Invocation itself will only use a handful of the rpc endpoints

modal/_utils/grpc_utils.py
```diff
@@ -77,15 +77,32 @@ class RetryWarningMessage:
     errors_to_warn_for: typing.List[Status]
 
 
+class ConnectionPool:
```
freider, line 80: nit: This name could be a bit confusing since this isn't a connection pool in the traditional sense of an interchangeable pool of connections.

modal/client.py
freider: Might want to add this comment here into the codebase for future reference :D

freider: I might not have this fully understood, but isn't there (at least superficially) a risk of a concurrency race here where the channel of the wrapped method gets swapped out by two concurrent callers using the same underlying method?

thecodingwizard: That's a great question. If you look at `ServiceMethod` from grpclib's repo:

```py
    def open(
        self,
        *,
        timeout: Optional[float] = None,
        metadata: Optional[_MetadataLike] = None,
    ) -> Stream[_SendType, _RecvType]:
        return self.channel.request(self.name, self._cardinality,
                                    self.request_type, self.reply_type,
                                    timeout=timeout, metadata=metadata)
```

`self.channel` is only accessed once and there are no async yields, so I don't think it's possible for `self.channel` to be changed while it is being called. If `self.channel` is changed _after_ `.open()` is called, the old channel / request will continue working as normal since it has no dependency on `self`. Also, `ConnectionPool` should return the same channel for concurrent callers, so `self.channel` will be replaced with an identical value.

modal/_functions.py
thecodingwizard: Yeah :/ We can split it into `ModalInputPlaneClient` and `ModalControlPlaneClient` but it seems undesirable to end up with a large number of stubs / services. We can always refactor / revisit this later.

modal/_utils/grpc_utils.py
thecodingwizard: good point. I updated to `ConnectionManager`

modal/client.py
freider: I see. I still feel it's a bit brittle that if either grpclib, or more likely us, add another async yield somewhere between the channel assignment and method usage (like an await inside _call_unary), it would potentially lead to the wrong input plane being used.

As long as we have this top of mind it's unlikely to happen, but these things tend to be forgotten over time, and such a "bug" would have a high likelyhood of passing silently into production and cause issues if it did happen.

Maybe I'm missing something here, but if this is indeed the case I'd prefer if we could solve this in a slightly safer way without having to modify state/internals of the method.

Having the Modal stub be the go-to place for all rpc usage and have that internally create new grpclib stubs when needed sounds compelling to me and at least superficially might keep the complexity low? But it's possible it might be tricky for some reason I haven't thought of

freider (approved):
lgtm!

Separate from this implementation, I do have some more general concerns about potential "staleness" of server_url for a function - what if it changes after a function was looked up? That's probably more of an issue for the `.remote()` implementation etc. - i.e. we'd have to get some kind of error from the backend if you try to submit an input to the wrong control plane, letting the function rehydrate itself with a new server_url?

thecodingwizard:
That's right, old clients will have old `server_url`'s. This is something we will support in the server: A function will have one "input queue / container scaling pool" per `server_url`. `min_containers` will only be respected on the "current" server_url's container scaling pool. Old clients running on the old `server_url` will be operating on a different input queue than new clients running on a new `server_url`.

An alternative implementation would be to have the server "redirect" old clients to call the new input queue, but this might be tricky.

PR #3025, opened 2025-04-14
improve cluster proto msg

thundergolfer:
## Describe your changes

Turns out I don't want to expose `region` because it's not valid to say that a cluster only has one region. That's _currently_ the case, but it's not a restriction we want binding us in the future (e.g. DiLiCo cluster with GPUs spread across multiple regions in a cloud). 

Because nothing is consuming this proto I'll just swap out the name for a field I do need, `cluster_id`. Realized I was stupid and I need the ID in the list response. 

---

---

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -776,7 +776,7 @@ message ClusterListResponse {
 message ClusterStats {
   string app_id = 1;
   repeated string task_ids = 2;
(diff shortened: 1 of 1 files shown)


thundergolfer:
@prbot approve

pawalt (commented):
sg

PR #3026, opened 2025-04-14
[CLI-269] limit output batch to 20

gongy:
## Describe your changes

---

---

</details>

## Changelog

- Fix bug where containers hang with batch sizes above 100 (with `@modal.batched`).
- Fix bug where containers can fail with large outputs and batch sizes above 49 (with `@modal.batched`)


mwaskom (approved):
> - Fix issue with large batch sizes (with @modal.batched).

Can we provide a little more information about how the issue manifested here (is there a specific error type, etc.?) so that users running into it can see from the Changelog that they can get a fix by upgrading?

PR #3027, opened 2025-04-15
Catch modification of Mount contents between build start and upload

mwaskom:
## Describe your changes

Some users have run into issues when they modify the local files included in their Modal App during a build (e.g., changing the git branch). Because some steps might need to run before we checksum the files and upload them to the Modal server, it's possible for the App to end up including the wrong contents.

Solving this in general is fairly challenging, but we can try to catch it when it happens. The approach here for the Resolver object to track when it was instantiated and to cross-reference the modification time of any files that we Mount against this baseline.

This is a little scary. It might make sense to include a config options that turns it into a warning? (Or ignores it altogether?)

- Fixes CLI-277
  
---

---

</details>

### DIFF
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2023
 import asyncio
 import contextlib
+import time
 import typing
 from asyncio import Future
 from collections.abc import Hashable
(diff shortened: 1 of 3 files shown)


erikbern:
Nice!!

aksh-at (commented):
In theory `_get_file_upload_spec` could happen quite a bit before the file is actually uploaded (I think you mentioned this on Slack). What about instead of passing `mtime`, passing the file path and then checking it _after_ `MountPutFileRequest` has gone through. If the file changed, we error then. That way any change between build start and upload would be caught?

aksh-at (approved):
Nice lgtm

PR #3031, opened 2025-04-15
Add experimental update_autoscaler function

mwaskom:
A few users have expressed interested in expanded autoscaler overrides, which are fully built on the backend but pending in the client on some decisions about the interface.

In the short-term, we can ship the functionality in the experimental namespace: `modal.experimental.update_autoscaler`. While the final form is likely to be a method, this is a standalone function that can be called on either a Function or an Obj.


mwaskom:
@prbot approve

freider (commented):
cool

PR #3032, opened 2025-04-15
Send the last batch index when retrying ContainerExecGetOutput

saltzm:
## Describe your changes

This prevents retries from accidentally processing the same batch twice.

---

---

</details>


cweld510 (approved):
LGTM; this double-layered iterator is confusing, though

PR #3033, opened 2025-04-16
volume: Add support for VolumePutFiles2 RPC

dflemstr:
## Describe your changes

- Adds support for version 2 `volume.batch_upload()` via the new block-based API used by `VolumePutFiles2`

---

---

</details>

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -1,15 +1,25 @@
 # Copyright Modal Labs 2022
 import asyncio
 import dataclasses
+import functools
 import hashlib
-import io
 import os
 import platform
 import time
 from collections.abc import AsyncIterator
 from contextlib import AbstractContextManager, contextmanager
+from io import BytesIO, FileIO
 from pathlib import Path, PurePosixPath
-from typing import TYPE_CHECKING, Any, BinaryIO, Callable, Optional, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    BinaryIO,
+    Callable,
+    ContextManager,
+    Optional,
+    Union,
+    cast,
+)
 from urllib.parse import urlparse
 
 from modal_proto import api_pb2
@@ -42,6 +52,9 @@ DEFAULT_SEGMENT_CHUNK_SIZE = 2**24
 # well, but the limit will never be raised.
 # TODO(dano): remove this once we stop requiring md5 for blobs
 MULTIPART_UPLOAD_THRESHOLD = 1024**3
+
+# For block based storage like volumefs2: the size of a block
+BLOCK_SIZE: int = 8 * 1024 * 1024
 
 
 @retry(n_attempts=5, base_delay=0.5, timeout=None)
@@ -94,7 +107,7 @@ async def _upload_to_s3_url(
 
 
 async def perform_multipart_upload(
-    data_file: Union[BinaryIO, io.BytesIO, io.FileIO],
+    data_file: Union[BinaryIO, BytesIO, FileIO],
     *,
     content_length: int,
     max_part_size: int,
@@ -112,9 +125,9 @@ async def perform_multipart_upload(
     # Give each part its own IO reader object to avoid needing to
     # lock access to the reader's position pointer.
     data_file_readers: list[BinaryIO]
-    if isinstance(data_file, io.BytesIO):
+    if isinstance(data_file, BytesIO):
         view = data_file.getbuffer()  # does not copy data
-        data_file_readers = [io.BytesIO(view) for _ in range(len(part_urls))]
+        data_file_readers = [BytesIO(view) for _ in range(len(part_urls))]
     else:
         filename = data_file.name
         data_file_readers = [open(filename, "rb") for _ in range(len(part_urls))]
@@ -174,7 +187,7 @@ async def _blob_upload(
     upload_hashes: UploadHashes, data: Union[bytes, BinaryIO], stub, progress_report_cb: Optional[Callable] = None
 ) -> str:
     if isinstance(data, bytes):
-        data = io.BytesIO(data)
+        data = BytesIO(data)
 
     content_length = get_content_length(data)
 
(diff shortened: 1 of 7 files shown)


modal/_utils/blob_utils.py
```diff
@@ -366,6 +377,113 @@ def source():
...
+
+                if not chunk:
+                    break
+
+                num_bytes_read += len(chunk)
+                sha256_hash.update(chunk)
+
+        return sha256_hash.digest()
+
+    with asyncio.TaskGroup() as tg:
```
freider, line 484: Sadly, TaskGroup was added in version 3.11 - all client code needs to support 3.9+ 😬

modal/_utils/blob_utils.py
```diff
@@ -366,6 +377,113 @@ def source():
...
+        return await asyncio.gather(*map(lambda idx: tg.spawn_task(hash_block_sha256(idx)), range(num_blocks)))
```
freider, line 485: nit: It's usually more idiomatic python to use list or generator comprehensions instead of the global `map` (you should be able to just skip the `map` part here and keep the asterisk in front of the parentheses)

modal/_utils/blob_utils.py
```diff
@@ -366,6 +377,113 @@ def source():
...
+async def hash_blocks_sha256(
+    source: _FileUploadSource2,
+    size: int,
+) -> list[bytes]:
+    def ceildiv(a: int, b: int) -> int:
+        return -(a // -b)
+
+    num_blocks = ceildiv(size, 8 * 1024 * 1024)
+
+    async def hash_block_sha256(block_idx: int) -> bytes:
```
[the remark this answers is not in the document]
dflemstr: I was in the middle of refactoring it to be blocking again, as discussed on Slack; I removed the `async` keyword here now.

modal/_utils/blob_utils.py
```diff
@@ -366,6 +377,113 @@ def source():
...
+
+
+    @staticmethod
+    async def from_fileobj(
+        source_fp: BinaryIO,
+        mount_filename: PurePosixPath,
+        mode: int
+    ) -> "FileUploadSpec2":
+        def source():
+            new_fd = os.dup(source_fp.fileno())
```
dflemstr, line 422: Even though the type system doesn't allow it, do we have some users passing in non-file-backed objects here? How to best cater for that?

gimaker: Yeah I think we need to support passing `io.BytesIO` here. We support that today.

Python is weird - `BytesIO` objects have the `fileno` function but it blows up at runtime:
```
In [18]: x = io.BytesIO(b"foo")

In [19]: "fileno" in dir(x)
Out[19]: True

In [20]: x.fileno()
---------------------------------------------------------------------------
UnsupportedOperation                      Traceback (most recent call last)
Cell In[20], line 1
----> 1 x.fileno()

UnsupportedOperation: fileno
```

dflemstr: I want to tweak the type signature for these functions to indicate that we accept more types of file-like objects, not only `BinaryIO`.

It made me suspicious as to why the tests are passing that test for this functionality:

```
test/volume_test.py::test_volume_batch_upload[2] PASSED
```

Something is broken with the test setup, I think.

dflemstr: Fixed the test setup now, PTAL

erikbern:
Nice with the unit tests!

modal/_utils/blob_utils.py
```diff
@@ -43,6 +54,9 @@
 # TODO(dano): remove this once we stop requiring md5 for blobs
 MULTIPART_UPLOAD_THRESHOLD = 1024**3
 
+# For block based storage like volumefs2: the size of a block
+BLOCK_SIZE: int = 8 * 1024 * 1024
```
freider, line 58: Do we want this to be determined by the client or the server? What happens if a user disregards this value?

modal/_utils/blob_utils.py
```diff
@@ -368,6 +382,122 @@ def source():
...
+    @staticmethod
+    async def from_fileobj(
+        source_fp: Union[BinaryIO, BytesIO],
+        mount_filename: PurePosixPath,
+        mode: int
+    ) -> "FileUploadSpec2":
+        try:
+            fileno = source_fp.fileno()
+            def source():
+                new_fd = os.dup(fileno)
```
freider, line 429: Hope this works on windows  😬 Are there test that check the parallel blob hashing (by having enough data)?

modal/volume.py
```diff
@@ -499,7 +513,14 @@ async def batch_upload(self, force: bool = False) -> "_VolumeUploadContextManage
             batch.put_file(io.BytesIO(b"some data"), "/foobar")
         ```
         """
-        return _VolumeUploadContextManager(self.object_id, self._client, force=force)
+        from modal_proto.api_pb2 import VolumeFsVersion
```
freider, line 516: This import can probably be global?

modal/volume.py
```diff
@@ -527,7 +548,35 @@ async def rename(
         await retry_transient_errors(obj._client.stub.VolumeRename, req)
 
 
-class _VolumeUploadContextManager:
+Volume = synchronize_api(_Volume)
+
+# TODO(dflemstr): Find a way to add ABC or AbstractAsyncContextManager superclasses while keeping synchronicity happy.
```
freider, line 553: What were the issues that popped up when you tried?

dflemstr:
Odd that hydration started failing on Ubuntu, it doesn't seem Linux related as I can run the tests locally... I'll look into this

modal/volume.py
```diff
@@ -678,9 +727,209 @@ async def _upload_file(self, file_spec: FileUploadSpec) -> api_pb2.MountFile:
...
+            with missing_block_file_spec.source() as source_fp:
+                data = b""
+                num_bytes_read = 0
+
+                source_fp.seek(missing_block_start)
+
+                while num_bytes_read < BLOCK_SIZE:
+                    # Most of the time, this will be a single read() call, so don't optimize too much for this
+                    # loop having more than one iteration.
+                    chunk = source_fp.read(BLOCK_SIZE - num_bytes_read)
```
freider, line 911: Not sure how much we need to worry about this read being blocking. I suppose in the general case, lets say you are reading from a network drive, it could potentially be bad since it would lock up the event loop, but for really small reads on local ssd it feels like putting this in a to_thread might make things actually go slower. wdyt?

test/conftest.py
```diff
@@ -221,7 +228,7 @@ def __init__(self, blob_host, blobs, credentials):
         self.task_result = None
 
         self.nfs_files: dict[str, dict[str, api_pb2.SharedVolumePutFileRequest]] = defaultdict(dict)
-        self.volume_files: dict[str, dict[str, VolumeFile]] = defaultdict(dict)
+        self.volumes: dict[str, Volume] = {}
```
freider, line 231: nice

test/conftest.py
```diff
@@ -1997,7 +2090,7 @@ async def async_main():
     thread = threading.Thread(target=run_server_other_thread)
     thread.start()
     started.wait()
-    yield host, blobs
+    yield host, blobs, blocks
```
freider, line 2093: neat

freider (commented):
didn't check all the finer details, but overall looks sane, and great with the tests. some inline comments

modal/volume.py
dflemstr: I didn't see an existing pattern of importing proto messages into the global scope, so I wanted to limit the blast radius. I'll happily hoist it up if you think it is a good idea.

modal/_utils/blob_utils.py
dflemstr: Yes, there are tests (both multi-block on-disk file and multi-block in-memory byte buffer)

modal/_utils/blob_utils.py
dflemstr: We have equivalent safety checks on the server side, so you will get "bad request" type error messages.

modal/volume.py
dflemstr: Synchronicity will try to resolve the parent classes using the blocking naming convention, ie. it re-wrote the parent classes to `BlockingABC` and `BlockingAbstractAsyncContextManager`

dflemstr:
Determined with @freider that we will have to make an excursion and add proper support for `VolumeMetadata` instead of the hack in this PR due to how lazy hydration needs to work: https://github.com/modal-labs/modal-client/pull/3035

modal/volume.py
dflemstr: (Please confirm if you still want me to do that given my reply above)

freider: The established pattern is to do `from modal_proto import api_pb2` and then reference the proto messages using `api_pb2.X`

dflemstr: Yeah, I used to have that before but felt like the line got too long, hence the import. But I can move back to it, sounds good!

modal/volume.py
```diff
@@ -678,9 +744,204 @@ async def _upload_file(self, file_spec: FileUploadSpec) -> api_pb2.MountFile:
...
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        if not exc_val:
+            # Flatten all the uploads yielded by the upload generators in the batch
+            def gen_upload_providers():
+                for gen in self._uploader_generators:
+                    yield from gen
+
+            async def gen_file_upload_specs() -> list[FileUploadSpec2]:
+                uploads = [asyncio.create_task(fut()) for fut in gen_upload_providers()]
```
freider, line 781: iirc, there is some pre-existing "issue" here that @gimaker mentioned when this was originally conceived, where creating the `uploads` list here will in itself be waiting for all files to be completely hashed before we start any kind of uploads.

This is at best just a suboptimal concurrency issue, and since it's pre-existing we could look into it as a separate issue. Maybe worth doing a quick sanity check that we don't have any related issues that would potentially explode memory usage etc by keeping all files in memory or something, but I'm hopeful we don't do that :D

freider: Btw, there seems to be a lot of overlap here with the v1 uploader so there is some risk of implementation drift if we update this logic in one of the places - are there any quick opportunities for code reuse across the two? (I'm not a DRY absolutist though, just mostly this part that I feel we might want to re-implement anyways)

modal/volume.py
dflemstr: When using this signature:

```
class _AbstractVolumeUploadContextManager(AbstractAsyncContextManager):
```

...I get this:
```
modal/volume.pyi:281: error: Class modal.volume._AbstractVolumeUploadContextManager has abstract attributes "__aexit__"  [misc]
modal/volume.pyi:281: note: If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass
modal/volume.pyi:303: error: Name "contextlib.BlockingAbstractAsyncContextManager" is not defined  [name-defined]
Found 2 errors in 1 file (checked 308 source files)
```

...and when using:
```
class _AbstractVolumeUploadContextManager(AbstractAsyncContextManager, ABC):
```

I get this:
```
modal/volume.pyi:282: error: Class modal.volume._AbstractVolumeUploadContextManager has abstract attributes "__aexit__"  [misc]
modal/volume.pyi:282: note: If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass
modal/volume.pyi:304: error: Name "contextlib.BlockingAbstractAsyncContextManager" is not defined  [name-defined]
modal/volume.pyi:304: error: Name "abc.BlockingABC" is not defined  [name-defined]
Found 3 errors in 1 file (checked 308 source files)
```

I guess I could hand craft a bunch of superclasses, where there's a common ABC with only `put_file` and `put_directory`, then have two subclasses (blocking and non blocking) that inherit from that common ABC + the right `ContextManager` ABC. I think that would lead to nice types but would be a lot of boiler plate? WDYT?

In other words, I would write this manually:
```python
class _AbstractVolumeUploadContextManager:
    def put_file(
        self,
        local_file: Union[Path, str, BinaryIO, BytesIO],
        remote_path: Union[PurePosixPath, str],
        mode: Optional[int] = None,
    ):
        ...

    def put_directory(
        self,
        local_path: Union[Path, str],
        remote_path: Union[PurePosixPath, str],
        recursive: bool = True,
    ):
        ...

class AbstractVolumeUploadContextManager(AbstractAsyncContextManager, ABC):
    pass

class BlockingAbstractVolumeUploadContextManager(AbstractContextManager, ABC):
    pass
```

dflemstr: I'll try this for now to get a feel for what it looks like

modal/volume.py
dflemstr: > Btw, there seems to be a lot of overlap here with the v1 uploader so there is some risk of implementation drift if we update this logic in one of the places - are there any quick opportunities for code reuse across the two? (I'm not a DRY absolutist though, just mostly this part that I feel we might want to re-implement anyways)

I looked at this and didn't find anything that stands out; do you have any pointers for what we can de-duplicate?

Some things of note:

* Old upload path deals with hashing and uploads where a file is the unit, while the new code uses blocks as its unit. We could rewrite the old code paths to take `(file, offset, len)` params everywhere, but I didn't think it was worth it, especially because the duplicated lines of code are like 10-20 lines.
* We never create "blob entities" in the new code path, but we never inline data in the gRPC request, so those code paths disappear entirely. However, we *do* make put requests to the clobber worker. Hence, v1 uses upload mechanisms A and B while v2 uses upload mechanism C with basically nothing shared between them (well, I guess the progress bar code is shared...)

modal/volume.py
freider: Yeah this is not super important

modal/volume.py
freider: Ah yeah it was the two aexit methods I was thinking about, but yeah you are right they are quite different now that there is a single phase of uploading

freider (approved):
lots of code, but lgtm!

PR #3036, opened 2025-04-17
Add experimental_options passthrough and handle server warnings

mwaskom:
## Describe your changes

Adds `experimental_options` to `@app.function` and `@app.cls` (external) and `api_pb2.Function` (internal).

We'll use this going forward for testing features that depend only on server-side configuration. The advantage of passing through the options is that we don't need to tie server-side features to specific client upgrades, nor to we need to do the housekeeping involved with adding (and eventually removing) specific options to the decorators / proto messages.

This PR also adds a concept of server-side warnings for Function configuration:

image

It's relevant here because we want to be able to indicate when "experimental" options are deprecated server-side (and anyone using them should migrate to whatever stable API we've added, if relevant). But it could be generally useful for other cases where we want to warn based on some server side logic / limits / etc. without actually stopping the build.

We're not currently using this for anything but I might migrate some of the `_experimental_*` flags to this system in subsequent PRs.

- Client side of CLI-369

---

---

</details>

## Changelog

- The `@app.function` and `@app.cls` decorators now support `experimental_options`, which we'll use going forward when testing experimental functionality that depends only on server-side configuration.

### DIFF
--- a/modal/__main__.py
+++ b/modal/__main__.py
@@ -2,7 +2,7 @@
 import sys
 
 from ._traceback import reduce_traceback_to_user_code
-from .cli._traceback import highlight_modal_deprecation_warnings, setup_rich_traceback
+from .cli._traceback import highlight_modal_warnings, setup_rich_traceback
 from .cli.entry_point import entrypoint_cli
 from .cli.import_refs import _CliUserExecutionError
 from .config import config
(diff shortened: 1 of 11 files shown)


gongy (approved):
Server warnings is really quite a nice abstraction :)

PR #3041, opened 2025-04-19
chore: use consistent synchronicity version

davidxia:
in ci-cd.yml Github worklow as the one specified in `pyproject.toml` to avoid any surprises in the future.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -157,7 +157,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r modal/requirements/${{ matrix.image-builder-version }}.txt
-          pip install synchronicity
+          pip install synchronicity~=0.9.10
 
       - name: Compile protos
         run: |


mwaskom:
I think the risk of surprises here is low; this check just asserts that the modal runtime package can be imported with the restricted set of dependencies we include in our container images.

davidxia:
I see. Is `synchronicity` even needed to be installed here? Looks like the later build steps don't use it?

mwaskom:
> Is synchronicity even needed to be installed here?

Yes, it's a dependency of the modal runtime package. But it's not part of the published container requirements because the latest version of synchronicity is automatically included in the Modal containers through a distinct mechanism (actually we're likely to migrate all client dependency inclusion to this mechanism soon).

davidxia:
I see. Does it make sense to remove this line entirely then?

mwaskom:
Not entirely, since it won't be possible to import modal without having it installed.

PR #3047, opened 2025-04-22
Add support for calling input plane server

thecodingwizard:
This PR adds support for using `modal-client` to call an input plane server if the `input_plane_url` setting is set.

Our immediate goal is to get a synmon running against the input plane server. Rather than implementing logic for `FunctionGet` to return the input plane URL, we'll take a shortcut for now and just configure the input plane URL through environment variables we'll manually set when running the synmon. 

We plan to remove this environment variable before releasing this to users in any capacity.

This depends on #3024.

### DIFF
--- a/modal/config.py
+++ b/modal/config.py
@@ -251,6 +251,7 @@ _SETTINGS = {
     "cuda_checkpoint_path": _Setting("/__modal/.bin/cuda-checkpoint"),  # Used for snapshotting GPU memory.
     "function_schemas": _Setting(False, transform=_to_boolean),
     "build_validation": _Setting("error", transform=_check_value(["error", "warn", "ignore"])),
+    "input_plane_url": _Setting(None),  # For internal use only.
 }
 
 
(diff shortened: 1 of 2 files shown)


modal/config.py
```diff
@@ -242,6 +242,7 @@ class _Setting(typing.NamedTuple):
     "cuda_checkpoint_path": _Setting("/__modal/.bin/cuda-checkpoint"),  # Used for snapshotting GPU memory.
     "function_schemas": _Setting(False, transform=_to_boolean),
     "build_validation": _Setting("error", transform=_check_value(["error", "warn", "ignore"])),
+    "internal_experimental_input_plane_url": _Setting(None),  # For internal/experimental use
```
mwaskom, line 245: We have a few other "internal" config settings above that we don't distinguish based on the name. We should probably clean that up somehow but I think better to be consistent with existing conventions. (The config options are public in some sense, but the documentation is opt-in).

thecodingwizard: makes sense, I renamed to `input_plane_url`. thanks!

thecodingwizard:
Closing in favor of #3095

PR #3049, opened 2025-04-22
Support relative paths in ignore files

### LINKED ISSUE #3048: `ignore=[modal.FilePatternMatcher.from_file(...)]` uses absolute file paths for matching
### Summary

Following up from [this discussion](https://modallabscommunity.slack.com/archives/C069RAH7X4M/p1745253991939809)

`FilePatternMatcher` matches on absolute paths instead of relative paths, so rules in a `.dockerignore` like `*.txt` will not trigger (but `**/*.txt` would). This behavior is different to passing in rules directly to ignore. See [this repo for a reproducible example](https://github.com/N-Demir/basic_modal_bug).

### Version

0.74.14

### App ID

_No response_

mwaskom:
## Describe your changes

Currently we pass absolute paths into `ignore` functions, so relative patterns like `*.txt` don't properly ignore `.txt` files in the top-level of the context dir.

Not 100% sure about the consequences here. I think we want to do this for all string-based ignore patterns (i.e. `ignore=["*.txt"]` _and _ when a `dockerignore` file is in use). But this changes behavior also for when users pass an ignore _function_. OTOH I think we basically want some consistency between pattern-based and function-based ignoring. So maybe we just live with the change?

Needs a thoughtful changelog at least...

- Fixes CLI-373
- Fixes #3048

---

---

</details>

## Changelog

- Fixes the behavior of `ignore=` in `modal.Image` methods. Exclusion patterns are now correctly interpreted as relative to the directory being added (e.g., `*.json` will now ignore all json files in the top-level of the directory). The fix also affects the interpretation of patterns in `.dockerignore` files used by `modal.Image.from_dockerfile`. **This involves two breaking changes:**
    - When providing a custom function to `ignore=`, file paths passed into the function will now be _relative_, rather than absolute.
    - When providing ignore patterns (either as strings or in a `dockerignore` file), an error will be raised if any of the pattens are absolute paths.

### DIFF
--- a/modal/file_pattern_matcher.py
+++ b/modal/file_pattern_matcher.py
@@ -94,6 +94,8 @@ class FilePatternMatcher(_AbstractPatternMatcher):
                     raise ValueError('Illegal exclusion pattern: "!"')
                 new_pattern.exclusion = True
                 pattern = pattern[1:]
+            if os.path.isabs(pattern):
+                raise ValueError("Ignore patterns cannot be absolute paths")
             # In Python, we can proceed without explicit syntax checking
             new_pattern.cleaned_pattern = pattern
             new_pattern.dirs = pattern.split(os.path.sep)
--- a/modal/image.py
+++ b/modal/image.py
@@ -279,8 +279,9 @@ def _create_context_mount(
     include_fn = FilePatternMatcher(*copy_patterns)
 
     def ignore_with_include(source: Path) -> bool:
-        relative_source = source.relative_to(context_dir)
-        if not include_fn(relative_source) or ignore_fn(relative_source):
+        if source.is_absolute():
+            source = source.relative_to(context_dir)
+        if not include_fn(source) or ignore_fn(source):
             return True
 
         return False
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -339,8 +339,8 @@ class _Mount(_Object, type_prefix="mo"):
         return _Mount._new()._extend(
             _MountDir(
                 local_dir=local_path,
-                ignore=ignore,
                 remote_path=remote_path,
+                ignore=ignore,
                 recursive=True,
             ),
         )
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -153,9 +153,9 @@ class _MountDir(_MountEntry):
 
         for local_filename in gen:
             local_path = Path(local_filename)
-            if not self.ignore(local_path):
-                local_relpath = local_path.expanduser().absolute().relative_to(local_dir)
-                mount_path = self.remote_path / local_relpath.as_posix()
+            rel_local_path = local_path.relative_to(local_dir)
+            if not self.ignore(rel_local_path):
+                mount_path = self.remote_path / rel_local_path.as_posix()
                 yield local_path.resolve(), mount_path
 
     def watch_entry(self):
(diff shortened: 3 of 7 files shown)


freider:
It feels a bit scary to me to change the contract of the ignore/condition function to pass in relative paths, I'm pretty sure there could be users relying on the current behavior.
As an alternative, I wonder if we could just inspect the input patterns that are passed and transform them into "absolute patterns" if they don't have an "absolute prefix", by adding the current working directory as a prefix to the pattern?

mwaskom:
> I'm pretty sure there could be users relying on the current behavior.
Yeah. Worst case is that mounts would be more inclusive right?

> I wonder if we could just inspect the input patterns that are passed and transform them into "absolute patterns" if they don't have an "absolute prefix", by adding the current working directory as a prefix to the pattern?
I suspect that the code to do this would be pretty complicated, given how distributed this logic is 🫤

freider:
> > I'm pretty sure there could be users relying on the current behavior.
> > Yeah. Worst case is that mounts would be more inclusive right?

They could also be more exclusive if people are currently doing `if path == Path(__file__).parent / "blah.txt"` etc. :( 

> > I wonder if we could just inspect the input patterns that are passed and transform them into "absolute patterns" if they don't have an "absolute prefix", by adding the current working directory as a prefix to the pattern?
> > I suspect that the code to do this would be pretty complicated, given how distributed this logic is 🫤

Not necessarily - couldn't we just add something to the FilePatternMatcher that does this when adding patterns?
`if not pattern.startswith("/"): pattern = cwd + "/" + pattern`

mwaskom:
> couldn't we just add something to the FilePatternMatcher

But also, currently the `FilePatternMatcher` doesn't know about the `context_dir` (and while we instantiate the `FilePatternMatcher` in image.py, we don't resolve the context dir until mount.py. So that's what I mean about the logic being distributed.

freider:
> > couldn't we just add something to the FilePatternMatcher
> 
> But also, currently the `FilePatternMatcher` doesn't know about the `context_dir` (and while we instantiate the `FilePatternMatcher` in image.py, we don't resolve the context dir until mount.py. So that's what I mean about the logic being distributed.

Hmm, yeah maybe we'd want the logic to make the patterns absolute outside of the FIlePatternMatcher somehow 🤔 

> That wouldn't work for **/ patterns right? I'm not sure it would work for ./ patterns either. I'm not sure what else.

Wouldn't a **/ pattern just become `<cwd>/**/pattern` which would be what I expect?

An explicitly local pattern with `./` prefix would become `<cwd>/./pattern` which should either just work (haven't tested) or we'd have to strip the `./` before concatenation. Feels to me like this would work?

mwaskom:
@prbot approve

freider:
Some (late) thoughts, which shouldn't be blocking for the concept of relative paths, but more about making sure we have thought about/have tests for the various cases of what the paths are *relative to*:

* We would by default make this relative to the cwd of the process - this makes sense and is consistent with docker
* However, this can be overridden in *some* cases - namely `from_dockerfile()` and `dockerfile_commands()` which take a `context_dir` argument - do we know if paths *to the ignore function* are then relative to this context_dir? (in addition to the COPY source being relative to this path)
* There is also the case of auto-dockerignore support (when no ignore argument is provided), but I *think* that the dockerignore file is only allowed to be in the context dir, so it should be fine (otherwise we might have to make the predicate paths relative to where that file is 🤯 )
* When using `add_local_dir()` and `add_local_python_source()` it's a bit unclear to me what expected relative-to behavior for `ignore=` should be. Relative to the added directory, or to the cwd? (It's currently the latter, which makes things like `ignore=[`*.txt`]` not work :( )

PR #3050, opened 2025-04-22
Fix lazy hydration of Cls

mwaskom:
## Describe your changes

Fixes CLI-374

Not sure what the best approach is here? It's all a little 😵‍💫

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -970,20 +970,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         parent = self
 
         async def _load(param_bound_func: _Function, resolver: Resolver, existing_object_id: Optional[str]):
-            try:
-                identity = f"{parent.info.function_name} class service function"
-            except Exception:
-                # Can't always look up the function name that way, so fall back to generic message
-                identity = "class service function for a parametrized class"
-            if not parent.is_hydrated:
-                if parent.app._running_app is None:
-                    reason = ", because the App it is defined on is not running"
-                else:
-                    reason = ""
-                raise ExecutionError(
-                    f"The {identity} has not been hydrated with the metadata it needs to run on Modal{reason}."
-                )
-
+            await parent.hydrate()
             assert parent._client and parent._client.stub
 
             if (
--- a/test/cls_test.py
+++ b/test/cls_test.py
@@ -679,6 +679,10 @@ def test_cls_lookup_keep_warm(client, servicer):
         @method()
         def bar(self): ...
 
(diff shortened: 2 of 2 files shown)


modal/_functions.py
```diff
@@ -970,20 +970,7 @@ def _bind_parameters(
...
-            if not parent.is_hydrated:
-                if parent.app._running_app is None:
-                    reason = ", because the App it is defined on is not running"
-                else:
-                    reason = ""
-                raise ExecutionError(
-                    f"The {identity} has not been hydrated with the metadata it needs to run on Modal{reason}."
-                )
-
+            await parent.hydrate()
```
freider, line 973: Very satisfying patch!

Yeah I think this is mostly old legacy stuff from pre-lazily-hydratable classes.

There is still one code path where classes aren't lazily hydratable, which we should make sure has ok error message: when someone tries to use an in-project defined class without having run/deployed the app, e.g.:

```py

@app.cls()
class A:
     ...

if __name__ == "__main__":
     # no app.run() session etc. makes this usage illegal and it shoud have a good error message,
     # similar to the old `App it is defined on is not running`
     A().keep_warm(...)  # not inside a 

```

However this error message would probably belong somewhere in the *base* function hydration step (i.e. within `parent.hydrate()` here)

test/cls_test.py
```diff
@@ -669,6 +669,26 @@ def bar(self): ...
...
+        arg: str = modal.parameter(default="")
+
+        @method()
+        def bar(self): ...
+
+    deploy_app(app, name, client=client)
+
+    C = Cls.from_name(name, "ClsWithMethod")
+    obj = C()
+    obj.keep_warm(3)
```
freider, line 686: we probably already have tests for this, but one thing I've run into in the past when refactoring this code (which is why I think we should have tests for it by now...) is that it's easy to accidentally make stuff re-hydrate on every call. Should make sure that two calls to `.keep_warm()` don't cause a second `FunctionBindParams` request (or `FunctionGet` on the base function) to go out

modal/_functions.py
mwaskom: Yep this still triggers from the underlying `.hydrate` call. The error message is more generic:

```
modal.exception.ExecutionError: Function has not been hydrated with the metadata it needs to run on Modal, because the App it is defined on is not running.
```

rather than the more custom

```
modal.exception.ExecutionError: The T.* class service function has not been hydrated with the metadata it needs to run on Modal, because the App it is defined on is not running.
```

I think that's acceptable since "class service function" isn't really a public-facing concept anyway.

freider (approved):

PR #3051, opened 2025-04-22
Experimentally allow RDMA on functions

pawalt:
This will allow us to hand out a limited beta of RDMA for customers.

## Describe your changes

Enable Functions to specify that they need RDMA to run.

---

---

</details>


pawalt:
we're going to do this with an experimental option

PR #3052, opened 2025-04-22
[SVC-421] Implement primitive for spawning many calls

gongy:
## Describe your changes

Added a method for spawning many calls in parallel using `async_map` under the hood.

Approximate performance at time of testing: ~1000 calls/s. To increase this performance, we can later add a `async_foreach` method which foregoes `async_merge`.

</details>

## Changelog

- Added a method for spawning many calls in parallel.

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -142,31 +142,33 @@ class _Invocation:
         function_id = function.object_id
         item = await _create_input(args, kwargs, client, method_name=function._use_method_name)
 
-        transient_error_kwargs = {}
-        if from_spawn_map:
-            transient_error_kwargs = {
-                "max_retries": None,
-                "max_delay": 30.0,
-                "retry_warning_message": RetryWarningMessage(
-                    message="Warning: `.spawn_map(...)` for function `{self._function_name}` is waiting to create"
-                    "more function calls. This may be due to hitting rate limits or function backlog limits.",
-                    warning_interval=10,
-                    errors_to_warn_for=[Status.RESOURCE_EXHAUSTED],
-                ),
-                "additional_status_codes": [Status.RESOURCE_EXHAUSTED],
-            }
-
         request = api_pb2.FunctionMapRequest(
             function_id=function_id,
             parent_input_id=current_input_id() or "",
             function_call_type=api_pb2.FUNCTION_CALL_TYPE_UNARY,
             pipelined_inputs=[item],
             function_call_invocation_type=function_call_invocation_type,
-            from_spawn_map=from_spawn_map,
         )
-        response = await retry_transient_errors(client.stub.FunctionMap, request, **transient_error_kwargs)
-        function_call_id = response.function_call_id
 
+        if from_spawn_map:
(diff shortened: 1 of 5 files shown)


modal/_functions.py
```diff
@@ -1343,6 +1352,47 @@ async def _map(
...
+
+        Synchronicity-wrapped spawn_map implementation.
+
+        To avoid executing slow iterators inside synchronicity, we use a _SynchronizedQueue
+        instance fed by higher level functions like _spawn_map_sync() and _spawn_map_async().
+
+        _SynchronizedQueue is used instead of asyncio.Queue so that the main thread can put
+        items in the queue safely.
+
+        Worker pattern is 10% faster than creating a new asyncio task for every input.
```
freider, line 1366: Interesting!

modal/parallel_map.py
```diff
@@ -450,36 +473,68 @@ async def _map_async(
...
+    async for res in _spawn_map_generator(self, *input_iterators, kwargs=kwargs):
+        return res
+
+
+def _spawn_map_sync(self, *input_iterators, kwargs={}) -> None:
+    return list(
+        AsyncOrSyncIterable(
+            _spawn_map_generator(self, *input_iterators, kwargs=kwargs),
+            nested_async_message=(
+                "You can't run Function.spawn_map() from an async function. Use Function.map.aio() instead."
```
freider, line 532:
```suggestion
                "You can't run Function.spawn_map() from an async function. Use Function.spawn_map.aio() instead."
```

mwaskom: Why is this the only place we disarm this footgun?

freider (commented):
Mostly makes sense to me, but I think I'd prefer not using AsyncOrSyncIterable here since the return value isn't expected to be iterated

modal/parallel_map.py
```diff
@@ -450,36 +473,68 @@ async def _map_async(
...
             yield output
 
 
+async def _spawn_map_generator(self, *input_iterators, kwargs={}) -> typing.AsyncGenerator[None, None]:
+    """mdmd:hidden
+    This runs in an event loop on the main thread. It consumes inputs
+    from the input iterators and creates async function calls for each.
+
+    We create a common generator that is shared by `_spawn_map_async` and `_spawn_map_sync` because
+    it is easier to wrap async generators than coroutines.
```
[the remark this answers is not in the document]
gongy: Thank you! I'm going to get this going. This is exactly the feedback I needed ❤️

modal/_functions.py
```diff
@@ -1343,6 +1352,47 @@ async def _map(
...
+        """
+
+        MAX_CONCURRENT_SPAWNS = 256
```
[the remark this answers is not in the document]
gongy: From benchmarking, `128` doesn't saturate our server rate limits for clients, and there is limited improvement above 300. I am somewhat nervous about request concurrency (from within a container, that is currently limited to `500`).

gongy:
@prbot approve

PR #3053, opened 2025-04-22
Improve UX for container heartbeat warnings

thundergolfer:
## Describe your changes

Status quo is this, which I think is not as readable and understandable as it could be

```
[Thread-1 (thread_inner)] 2025-04-22T20:56:05+0000 Heartbeat attempt failed (time_elapsed=156.61637557400036, error=ConnectionError('Deadline exceeded'))
```

---

    -   logging change only
    -   logging change only

---

</details>

### DIFF
--- a/modal/_runtime/container_io_manager.py
+++ b/modal/_runtime/container_io_manager.py
@@ -352,29 +352,42 @@ class _ContainerIOManager:
         await self._client.stub.ContainerHello(Empty())
 
     async def _run_heartbeat_loop(self):
+        t_last_success = time.monotonic()
         while 1:
             t0 = time.monotonic()
             try:
                 if await self._heartbeat_handle_cancellations():
-                    # got a cancellation event, fine to start another heartbeat immediately
-                    # since the cancellation queue should be empty on the worker server
-                    # however, we wait at least 1s to prevent short-circuiting the heartbeat loop
-                    # in case there is ever a bug. This means it will take at least 1s between
-                    # two subsequent cancellations on the same task at the moment
+                    # Got a cancellation event, so it is fine to start another heartbeat immediately
+                    # since the cancellation queue should be empty on the worker server.
+                    # However, we wait at least 1s to prevent short-circuiting the heartbeat loop
+                    # in case there is ever a bug.
+                    # This means it will take at least 1s between two subsequent cancellations on the
+                    # same task at the moment.
                     await asyncio.sleep(1.0)
                     continue
+                t_last_success = time.monotonic()
             except ClientClosed:
                 logger.info("Stopping heartbeat loop due to client shutdown")
                 break
             except Exception as exc:
                 # don't stop heartbeat loop if there are transient exceptions!
-                time_elapsed = time.monotonic() - t0
+                attempt_dur = time.monotonic() - t0
+                time_since_heartbeat_success = time.monotonic() - t_last_success
                 error = exc
-                logger.warning(f"Heartbeat attempt failed ({time_elapsed=}, {error=})")
+                logger.warning(
+                    f"Heartbeat attempt failed ({attempt_dur=:.2f}, {time_since_heartbeat_success=:.2f}, {error=})"
+                )
+                if time_since_heartbeat_success > HEARTBEAT_INTERVAL * 50:
+                    trouble_mins = time_since_heartbeat_success / 60
+                    logger.warning(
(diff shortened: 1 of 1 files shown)


thundergolfer:
@prbot approve

modal/_runtime/container_io_manager.py
```diff
@@ -352,29 +352,43 @@ async def hello(self):
...
-                logger.warning(f"Heartbeat attempt failed ({time_elapsed=}, {error=})")
+                logger.warning(
+                    f"Heartbeat attempt failed ({attempt_dur=:.2f}, {time_since_heartbeat_success=:.2f}, {error=})"
+                )
+                if time_since_heartbeat_success > HEARTBEAT_INTERVAL * 50:
+                    trouble_mins = time_since_heartbeat_success / 60
+                    logger.warning(
+                        f"Heartbeat attempts have been failing for over {trouble_mins:.2f} minutes. "
+                        "Container will eventually be marked unhealthy. "
+                        "See https://modal.com/docs/guide/troubleshooting#heartbeat-timeout. "
```
mwaskom, line 386: Is this going to wrap properly?

modal/_runtime/container_io_manager.py
```diff
@@ -352,29 +352,43 @@ async def hello(self):
...
+                time_since_heartbeat_success = time.monotonic() - t_last_success
                 error = exc
-                logger.warning(f"Heartbeat attempt failed ({time_elapsed=}, {error=})")
+                logger.warning(
+                    f"Heartbeat attempt failed ({attempt_dur=:.2f}, {time_since_heartbeat_success=:.2f}, {error=})"
+                )
+                if time_since_heartbeat_success > HEARTBEAT_INTERVAL * 50:
+                    trouble_mins = time_since_heartbeat_success / 60
+                    logger.warning(
+                        f"Heartbeat attempts have been failing for over {trouble_mins:.2f} minutes. "
```
mwaskom, line 384: Nit but seems odd to report a duration in decimal minutes.

modal/_runtime/container_io_manager.py
```diff
@@ -352,29 +352,43 @@ async def hello(self):
...
                 break
             except Exception as exc:
                 # don't stop heartbeat loop if there are transient exceptions!
-                time_elapsed = time.monotonic() - t0
+                attempt_dur = time.monotonic() - t0
+                time_since_heartbeat_success = time.monotonic() - t_last_success
                 error = exc
-                logger.warning(f"Heartbeat attempt failed ({time_elapsed=}, {error=})")
+                logger.warning(
+                    f"Heartbeat attempt failed ({attempt_dur=:.2f}, {time_since_heartbeat_success=:.2f}, {error=})"
```
mwaskom, line 379: This is definitely more helpful but potentially could make the warning messages look more like something that is intended to be user-facing; I think part of why they're currently confusing is that users don't understand where they're coming from or exactly what they're trying to communicate.

Also could we gloss what the "heartbeat"? "Modal client->server heartbeat" or something?

mwaskom (approved):
LGTM, feel free to merge, left a few thoughts for the next time we take a pass here.

modal/_runtime/container_io_manager.py
thundergolfer: Yep agree. Will do that.

modal/_runtime/container_io_manager.py
thundergolfer: Why? Given the timescale involved it's the most natural and readable to me.

modal/_runtime/container_io_manager.py
thundergolfer: What's the alternative? Should I add `\n`. I don't have much idea about what do here. I suspect adding newlines could make the logs even less readable in our system.

modal/_runtime/container_io_manager.py
mwaskom: Not a big deal either way, I agree that reporting minutes rather than seconds makes sense, just don't usually see fractional minutes (or fractional hours, etc.). Would probably be fine just to truncate and report integral minutes since you say "more than" anyway.

modal/_runtime/container_io_manager.py
mwaskom: Yeah genuinely not sure, I just know we have some long logging messages (e.g. the function queuing status) where the more relevant information tends to get cut off. Probably best to leave it like this and fix that elsewhere.

PR #3057, opened 2025-04-24
Reapply #3050 (lazy Cls hydration)

### LINKED ISSUE #2979: keep_warm raises exception
According to the documentation, I should be able to do this:
```
modal.Cls.from_name('myapp', 'mycls')().keep_warm(1)
```

However, I get the following exception:
```
modal.exception.ExecutionError: The app has not been assigned on the function at this point
```

I was able to work around this by doing the following:
```
cls = modal.Cls.from_name('myapp', 'mycls')
cls.hydrate()
cls().keep_warm(1)
```

mwaskom:
Now passes the internal regression test. Also cleans up some apparently uneccessary code that was part of why the test broke.

Fixes #2979

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -970,7 +970,9 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         parent = self
 
         async def _load(param_bound_func: _Function, resolver: Resolver, existing_object_id: Optional[str]):
-            await parent.hydrate()
+            if not parent.is_hydrated:
+                # While the base Object.hydrate() method appears to be idempotent, it's not always safe
+                await parent.hydrate()
             assert parent._client and parent._client.stub
 
             if (
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -973,6 +973,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
             if not parent.is_hydrated:
                 # While the base Object.hydrate() method appears to be idempotent, it's not always safe
                 await parent.hydrate()
+
             assert parent._client and parent._client.stub
 
             if (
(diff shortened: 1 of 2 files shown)


modal/_functions.py
```diff
@@ -970,20 +970,9 @@ def _bind_parameters(
...
             if not parent.is_hydrated:
-                if parent.app._running_app is None:
-                    reason = ", because the App it is defined on is not running"
-                else:
-                    reason = ""
-                raise ExecutionError(
-                    f"The {identity} has not been hydrated with the metadata it needs to run on Modal{reason}."
-                )
-
+                # While the base Object.hydrate() method appears to be idempotent, it's not always safe
```
freider, line 974: Should we try to add a unit test that checks the exercises the same path that failed in the integration test?

modal/_object.py
```diff
@@ -258,24 +258,15 @@ async def hydrate(self, client: Optional[_Client] = None) -> Self:
 
         *Added in v0.72.39*: This method replaces the deprecated `.resolve()` method.
         """
-        if self._is_hydrated:
-            if self.client._snapshotted and not self._is_rehydrated:
-                # memory snapshots capture references which must be rehydrated
-                # on restore to handle staleness.
-                logger.debug(f"rehydrating {self} after snapshot")
-                self._is_hydrated = False  # un-hydrate and re-resolve
+        if not self._is_hydrated:
```
freider, line 261: Oooh, I like this simplification! Maybe run the full integration test suite before mergin though 😬

mwaskom: Had to roll this back as it broke one of the client unit tests. The current diff passes a number of relevant integration test modules (though I haven't run against all of them 😶‍🌫️)

mwaskom: Reapplied it to debug and it seems to be passing tests now... Weird!

mwaskom:
@prbot approve

freider (approved):

PR #3058, opened 2025-04-24
Add `modal secret delete` CLI

mwaskom:
## Changelog

- Added a `modal secret delete` command to the CLI.

### DIFF
--- a/modal/cli/secret.py
+++ b/modal/cli/secret.py
@@ -9,10 +9,11 @@ import click
 import typer
 from rich.console import Console
 from rich.syntax import Syntax
+from typer import Argument
 
 from modal._utils.async_utils import synchronizer
 from modal._utils.grpc_utils import retry_transient_errors
-from modal.cli.utils import ENV_OPTION, display_table, timestamp_to_local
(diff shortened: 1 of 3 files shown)


modal/cli/secret.py
```diff
@@ -91,6 +92,28 @@ def some_function():
...
+    if not yes:
+        typer.confirm(
+            f"Are you sure you want to irrevocably delete the modal.Secret '{secret_name}'?",
+            default=False,
+            abort=True,
+        )
+    client = await _Client.from_env()
+
+    # TODO: replace with API on `modal.Secret` when we add it
+    await client.stub.SecretDelete(api_pb2.SecretDeleteRequest(secret_id=secret.object_id))
```
erikbern, line 114: Should we just add it to the sdk as well?

erikbern (approved):

modal/cli/secret.py
mwaskom: Yeah eventually for sure, but I think we're still split on the pattern to use for "object management" methods like this. Didn't want to hold it up on that discussion.

PR #3059, opened 2025-04-25
volumes: Fix upload for version 2

gimaker:
Use POST instead of PUT for uploading blocks.

Doing a minimal fix for now - leaving the door open for changing the server-side to use PUT instead.

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -933,7 +933,7 @@ async def _put_missing_blocks(
                 progress_report_cb=functools.partial(progress_cb, progress_task_id)
             )
 
-            async with ClientSessionRegistry.get_session().put(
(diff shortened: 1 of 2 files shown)


modal/volume.py
```diff
@@ -933,7 +933,7 @@ async def put_missing_block(
                 progress_report_cb=functools.partial(progress_cb, progress_task_id)
             )
 
-            async with ClientSessionRegistry.get_session().put(
+            async with ClientSessionRegistry.get_session().post(
                 missing_block.put_url,
```
gimaker, line 937: might want to `s/put/post/g` in a couple of places if we change to `POST`

gimaker:
Closing because we decided to change the server-side to PUT instead.

PR #3061, opened 2025-04-25
fix usage typo in `batched` docstring

jvmncs:
This seems like a typo in the usage part of the docstring for the `modal.batched` decorator.


mwaskom:
Indeed. Thanks!

mwaskom (approved):

PR #3063, opened 2025-04-25
Increase ASGI proxy client timeout

### LINKED ISSUE #2752: Allow Timeout Configuration for web_server_proxy
The web_server_proxy function in Modal, designed to proxy requests to a local web server, currently uses a hardcoded timeout of 3600 seconds (1 hour) within its internal aiohttp.ClientSession. This fixed timeout presents challenges when proxying requests to services that might require longer processing times, such as those involving machine learning model training or extensive data operations.

This lack of configurability can lead to premature termination of requests that exceed the one-hour limit, disrupting long-running tasks and negatively impacting user experience.

**Proposed Solution:**

Introduce a mechanism to allow customization of the `aiohttp.ClientTimeout` used within `web_server_proxy`.
**Introduce a global or app-level configuration option:** This would allow setting a default timeout for all `web_server_proxy` instances within an application.

https://github.com/modal-labs/modal-client/blob/b61aed4330feada82bee7e2b27190b2c07925fe1/modal/_asgi.py#L323

ekzhang:
Resolves #2752. Thank you @mwaskom for raising this.

Edit: I actually realized that I can just pass `None` to disable it fully. https://docs.aiohttp.org/en/stable/client_quickstart.html#timeouts

### DIFF
--- a/modal/_runtime/asgi.py
+++ b/modal/_runtime/asgi.py
@@ -468,7 +468,8 @@ async def _proxy_lifespan_request(base_url, scope, receive, send) -> None:
                 session = aiohttp.ClientSession(
                     base_url,
                     cookie_jar=aiohttp.DummyCookieJar(),
-                    timeout=aiohttp.ClientTimeout(total=3600),
+                    # "Disable" this by setting to max possible input timeout. Timeouts are handled on input level instead.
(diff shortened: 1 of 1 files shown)


mwaskom (approved):
Thanks!

PR #3066, opened 2025-04-28
Require setting serialized=True when overriding function name

mwaskom:
When you set `name=` in `@app.function`, we try to import that name from the module where the function is defined during container startup. This will fail (or worse, pull in the some other object).

This isn't a problem for the standard pattern using `name=` along with `serialized=True` (i.e., defining functions "on the fly"), because then we don't try to import the name from the app module.

In theory we could track these pieces of information separately, but that would be a larger change, and it's not obvious that overriding the function name while not also setting `serialized=True` has much feature value. In the meantime, we'll help users avoid an annoying runtime error by blocking the incompatible parameterization at build time.

- Fixes CLI-383

---

---

</details>

## Changelog

- When setting a custom `name=` in `@app.function()`, an error is now raised unless `serialized=True` is also set.

### DIFF
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -140,7 +140,7 @@ class FunctionInfo:
     def __init__(
         self,
         f: Optional[Callable[..., Any]],
-        serialized=False,
+        serialized: bool = False,
         name_override: Optional[str] = None,
         user_cls: Optional[type] = None,
     ):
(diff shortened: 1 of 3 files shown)


freider:
I think there are legitimate cases for when `name` would good to have for non-serialized functions too, but as you say it's non trivial to fix it, and until we do it would be good to have a good error message when people try it (instead of having imports fail in containers)

freider (approved):

freider:
Btw, we should add `name` for app.cls() too!

PR #3067, opened 2025-04-28
Support modal run against Cls.with_options

mwaskom:
## Describe your changes

- Fixes CLI-384

## Changelog

- Fixed a bug that prevented doing `modal run` against an entrypoint defined by `Cls.with_options`.

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -395,6 +395,7 @@ class _Cls(_Object, type_prefix="cs"):
 
     def _initialize_from_other(self, other: "_Cls"):
         super()._initialize_from_other(other)
+        self._app = other._app
         self._user_cls = other._user_cls
         self._class_service_function = other._class_service_function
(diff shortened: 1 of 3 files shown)


modal/cls.py
```diff
@@ -395,6 +395,7 @@ def _initialize_from_empty(self):
 
     def _initialize_from_other(self, other: "_Cls"):
         super()._initialize_from_other(other)
+        self._app = other._app
```
mwaskom, line 398: Any downsides to this? Not sure why we didn't have it previously ... makes me a little nervous.

freider: I *think* it should be fine. I really dislike how many member vars we have on classes and functions and I think we should try to group them into distinct sets based on which are expected to be set at any given time. Would make it so much easier to reason about these things

mwaskom: Yes plz

PR #3068, opened 2025-04-29
Break the environment cache we use in CI

mwaskom:
This is an atheoretical attempt to resolve the persistent flakiness we've seen in this action.


mwaskom:
@prbot approve @freider

freider (commented):
😢

mwaskom:
This seemed to work! Haven't had a build flake since.

PR #3069, opened 2025-04-29
Make dummy function in base image app serialized

mwaskom:
#3066 broke the "base image deployment" script, which creates a function on the fly with `name=`. Because this function is never actually executed, it doesn't run into the error that #3066 is intended to help avoid. But the consequence is that our CI is currently being marked as failing.

This reminds me that we don't need these workflows running on every client release _anyway_. But we need a reliable way to trigger them when they should.


mwaskom:
@prbot approve @freider — well I guess that's one usecase for it!

mwaskom:
@prbot approve

mwaskom:
good bit, bot

PR #3070, opened 2025-04-29
Allow users to restrict Function's modal resource access

pawalt:
## Describe your changes

This commit adds a new `restrict_modal_access` field which prevents functions from performing any operations other than getting inputs and responding to those inputs.

---

---

</details>

## Changelog

- A new parameter, `restrict_modal_access`, can be provided on a Function to prevent it from interacting with other resources in your Modal Workspace like Queues, Volumes, or other Functions. This can be useful for running user-provided or LLM-written code in a safe way.

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -607,6 +607,7 @@ class _App:
         region: Optional[Union[str, Sequence[str]]] = None,  # Region or regions to run the function on.
         enable_memory_snapshot: bool = False,  # Enable memory checkpointing for faster cold starts.
         block_network: bool = False,  # Whether to block network access
+        restrict_modal_access: bool = False,  # Whether to allow this function access to other Modal resources
         # Maximum number of inputs a container should handle before shutting down.
         # With `max_inputs = 1`, containers will be single-use.
         max_inputs: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


modal/app.py
```diff
@@ -607,6 +607,7 @@ def function(
         region: Optional[Union[str, Sequence[str]]] = None,  # Region or regions to run the function on.
         enable_memory_snapshot: bool = False,  # Enable memory checkpointing for faster cold starts.
         block_network: bool = False,  # Whether to block network access
+        restrict_modal_access: bool = False,  # Whether to allow this function to access other Modal resources
```
mwaskom, line 610:
```suggestion
        restrict_modal_access: bool = False,  # Whether to allow this function access to other Modal resources
```

tripped over the double infinitive the first time i read it (if you want to make this change, good to change in the cls signature too)

mwaskom (approved):
cool; would be great to add a simple unit test to make sure we set the internal definition field the way we expect

mwaskom:
> this should be an exception: got exception: Received :status = '401'

What's the actual python exception here?

pawalt:
@mwaskom It's an `AuthError` which sounds right to me

pawalt:
@mwaskom I added a test btw

PR #3073, opened 2025-04-29
Warn when creating a new token that will be ignored

mwaskom:
If the `MODAL_TOKEN_ID` environment variable is set, it will take precedence over any tokens set in the `.modal.toml` config. We can catch this when the user does `modal token new` or `modal token set`:

image

This should help avoid some confusion, as the clear user intent here is "use this new token", but that won't happen until they unset the environment variable.

### DIFF
--- a/modal/token_flow.py
+++ b/modal/token_flow.py
@@ -158,6 +158,12 @@ async def _set_token(
         f"[magenta]{profile}[/magenta].[/green]"
     )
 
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot approve

mwaskom:
@prbot approve

charlesfrye (commented):
lgtm

PR #3075, opened 2025-05-01
Add guidance about not using the CLI implementation as API

mwaskom:
Trying to improve our clarity about what is "public API" in the Modal client. In general, user code should not call directly into any functions defined in the `modal.cli` subpackage.

Closes CLI-43


mwaskom:
@prbot approve

mwaskom:
Happy to tweak the text in a followup!

PR #3076, opened 2025-05-02
Add logger name prefix to client logs

freider:
---

---

</details>

## Changelog

* Changes the log format of the modal client's default logger. Instead of `[%(threadName)s]`, the client now logs `[modal-client]` as the log line prefix.
* Adds a configuration option (MODAL_LOG_PATTERN) to the modal config for setting the log formatting pattern, in case users want to customize the format. To get the old format, use `MODAL_LOG_PATTERN='[%(threadName)s] %(asctime)s %(message)s'` (or add this to your `.modal.toml` in the `log_pattern` field).

### DIFF
--- a/modal/_utils/logger.py
+++ b/modal/_utils/logger.py
@@ -32,7 +32,7 @@ def configure_logger(logger: logging.Logger, log_level: str, log_format: str):
     else:
         if not (log_format_pattern := config.get("log_pattern")):
             # TODO: use `%(name)s` instead of `modal-client` as soon as we unify the loggers we use
-            log_format_pattern = "modal-client [%(threadName)s] %(asctime)s %(message)s"
+            log_format_pattern = "[modal-client] %(asctime)s %(message)s"
 
         ch.setFormatter(logging.Formatter(log_format_pattern, datefmt=datefmt))
 
--- a/modal/_utils/logger.py
+++ b/modal/_utils/logger.py
@@ -45,4 +45,3 @@ log_format = os.environ.get("MODAL_LOG_FORMAT", "STRING")
 
 logger = logging.getLogger("modal-utils")
 configure_logger(logger, log_level, log_format)
-logger.warning("hello")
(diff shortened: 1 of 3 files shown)


modal/_utils/logger.py
```diff
@@ -4,35 +4,45 @@
...
+    datefmt = "%Y-%m-%dT%H:%M:%S%z"
     if log_format.upper() == "JSON":
         # This is primarily for modal internal use.
         # pythonjsonlogger is already installed in the environment.
         from pythonjsonlogger import jsonlogger
 
-        json_formatter = jsonlogger.JsonFormatter(
-            fmt=(
+        if not (log_format_pattern := config.get("log_pattern")):
+            log_format_pattern = (
```
mwaskom, line 20: Would users be expected to specify something like this in their `.modal.toml`?

freider: Typically no, but they could, similar to how they could set log_format="JSON" there today

freider: I mostly added it as a configurable fallback option in case there are users relying on the current logging format for log ingestion or similar

mwaskom (approved):
Should maybe mention this in the docstring of `modal.config` too? Since it's the source of these docs: https://modal.com/docs/reference/modal.config

modal/_utils/logger.py
```diff
@@ -4,33 +4,42 @@
...
+        json_formatter = jsonlogger.JsonFormatter(
+            fmt=log_format_pattern,
+            datefmt=datefmt,
+        )
         ch.setFormatter(json_formatter)
     else:
-        ch.setFormatter(logging.Formatter("[%(threadName)s] %(asctime)s %(message)s", datefmt="%Y-%m-%dT%H:%M:%S%z"))
+        if not (log_format_pattern := config.get("log_pattern")):
+            # TODO: use `%(name)s` instead of `modal-client` as soon as we unify the loggers we use
+            log_format_pattern = "modal-client [%(threadName)s] %(asctime)s %(message)s"
```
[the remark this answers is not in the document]
freider: It can be useful when debugging to know if a client operation happened in the synchronicity thread or a user thread, but at this point I would hope we have no more "leakage" between those so I agree we could maybe remove it.

PR #3077, opened 2025-05-02
Draft/RFC: Adds a Service mixin that can be used to get static type safety for classes

freider:
Very rough "implementation", just adding some static type check tests that this actually provides static type support. 

This doesn't actually implement these methods, and it's a bit TBD if the *actual* implementation should be in in the mixin or the `Obj` (probably the latter?)

Preferably, the same implementation can be reused through composition in Function

This is also pending the discussion of whether to put these methods on the class directly or under a `.service` attribute or similar, but IMO this is the most promising and consistent way of doing it...

---

---

</details>

### DIFF
--- a/test/supports/type_assertions.py
+++ b/test/supports/type_assertions.py
@@ -37,14 +37,12 @@ async def async_typed_func(b: bool) -> str:
     return ""
 
 
-async_typed_func
-
 should_be_str = async_typed_func.remote(False)  # should be blocking without aio
 assert_type(should_be_str, str)
 
 
 @app.cls()
-class Cls:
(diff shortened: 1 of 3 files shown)


test/supports/type_assertions.py
```diff
@@ -37,14 +37,12 @@ async def async_typed_func(b: bool) -> str:
...
 
-async_typed_func
-
 should_be_str = async_typed_func.remote(False)  # should be blocking without aio
 assert_type(should_be_str, str)
 
 
 @app.cls()
-class Cls:
+class UserCls(modal.Service):
```
freider, line 45: So beautiful!

PR #3079, opened 2025-05-04
Add type hint for all Secret.from*

op:
## Describe your changes

Add missing return typing hints for all `Secret.from_*` methods.

## Changelog

Add missing return typing hints for all Secret.from methods.

### DIFF
--- a/test/supports/type_assertions.py
+++ b/test/supports/type_assertions.py
@@ -1,7 +1,7 @@
 # Copyright Modal Labs 2024
 import typing
-
 from importlib.util import find_spec
+
 from typing_extensions import assert_type
 
 import modal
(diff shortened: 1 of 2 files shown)


dflemstr (approved):

op:
> Good stuff! If you want, you can add "tests" for this too in `test/supports/type_assertions.py`, .e.g

@freider cheers and pushed! Can you have a look at those tests before merging?

test/supports/type_assertions.py
```diff
@@ -100,3 +102,23 @@ async def async_block() -> None:
...
 assert_type(file_io2.readlines(), typing.Sequence[bytes])
+
+# check secrets
+secret = modal.Secret.from_dict({})
+assert_type(secret, modal.Secret)
+
+secret = modal.Secret.from_local_environ(["PATH"])
+assert_type(secret, modal.Secret)
+
+try:
```
freider, line 113: I think you don't have to do all of the "support" stuff here - the file should never be executed, it's only type checked

freider (approved):

PR #3080, opened 2025-05-05
volume: Limit hash/put concurrency by a semaphore

dflemstr:
## Describe your changes

- This limits the concurrency of block hashing/uploads for huge files, where it might be too expensive to open all blocks simultaneously. This avoids issues with running out of file descriptors or similar.

---

---

</details>

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -399,6 +399,7 @@ class FileUploadSpec2:
     async def from_path(
         filename: Path,
         mount_filename: PurePosixPath,
+        hash_semaphore: asyncio.Semaphore,
         mode: Optional[int] = None,
     ) -> "FileUploadSpec2":
         # Python appears to give files 0o666 bits on Windows (equal for user, group, and global),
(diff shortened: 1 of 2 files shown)


modal/_utils/blob_utils.py
```diff
@@ -497,7 +503,11 @@ def hash_block_sha256(block_idx: int) -> bytes:
 
         return sha256_hash.digest()
 
-    tasks = (asyncio.to_thread(functools.partial(hash_block_sha256, idx)) for idx in range(num_blocks))
+    async def hash_block_sha256(block_idx: int) -> bytes:
+        async with hash_semaphore:
+            return await asyncio.to_thread(functools.partial(blocking_hash_block_sha256, block_idx))
```
op, line 508: Unrelated to this PR but do you know why we don't pass args directly to `to_thread`?

op (commented):
Lgtm! Don't dare to approve tho :(

modal/_utils/blob_utils.py
dflemstr: Ah nice, didn't know that was a thing!

gimaker (approved):

PR #3084, opened 2025-05-05
Test & fix propagating starmap iterator exception

gongy:
Kind of a yak-shave refactor, but I think `parallel_map.py` becomes easier to read and reason about. Removes some duplicate code as well.

The initial goal was to fix error surfacing in `.starmap` -- the bug was a result of having this much duplicate code.

Added a core function `_map_helper` which powers `_map_async`, `_starmap_async` and `_for_each`.


freider (commented):
Great stuff! I assume the bug was not listening to the `feed_queue ` task potential exceptions, since it wasn't part of the `async_merge()` as in the map case?

I think this will also help with my "future work" of removing the temporary event loop from the sync map implementations

gongy:
@prbot approve @freider

freider (commented):
thanks for the fix!

PR #3085, opened 2025-05-05
Add async_foreach method

gongy:


modal/_utils/async_utils.py
```diff
@@ -771,7 +833,13 @@ async def worker() -> AsyncGenerator[V, None]:
         while True:
             item = await queue.get()
             if isinstance(item, ValueWrapper):
-                yield await async_mapper_func(item.value)
+                if isinstance(item.value, int) and item.value % 500 == 0:
+                    print(f"{item.value} found at time {time.time() - t0}")
+                r = await async_mapper_func(item.value)
+                if yield_nothing:
```
freider, line 839: Would this actually make any difference? If the mapper function doesn't have a result it would return None, and if it did have a result it would be passed by reference and wouldn't be slower than passing None afaik?

modal/_utils/async_utils.py
```diff
@@ -771,7 +833,13 @@ async def worker() -> AsyncGenerator[V, None]:
         while True:
             item = await queue.get()
             if isinstance(item, ValueWrapper):
-                yield await async_mapper_func(item.value)
+                if isinstance(item.value, int) and item.value % 500 == 0:
```
freider, line 836: printy printty

modal/_utils/async_utils.py
```diff
@@ -748,19 +761,68 @@ async def gather_cancel_on_exc(*coros_or_futures):
...
     async def producer() -> AsyncGenerator[V, None]:
         async for item in input_generator:
-            await queue.put(ValueWrapper(item))
+            queue.put_nowait(ValueWrapper(item))
 
         for _ in range(concurrency):
-            await queue.put(STOP_SENTINEL)
+            queue.put_nowait(STOP_SENTINEL)
+
+        print(f"producer: {time.time() - t0:.2f}s")
```
freider, line 825: printy printy

mwaskom:
Did we end up not needing this for `.spawn_map`?

gongy:
Yeah I think it's not worth the complexity.

PR #3087, opened 2025-05-05
Add input_plane_url to FunctionHandleMetadata proto

rculbertson:
Part of SVC-492

Add `input_plane_url` to `FunctionHandleMetadata` proto so that we can return it in `FunctionGet` and `FunctionCreate`. This is how the client knows which input plane to connect to.

---

---

</details>


thecodingwizard (approved):

thecodingwizard (commented):
Actually, should we make this optional? So it's clear that some functions aren't using the new input plane system

rculbertson:
Yea that sounds good. Added.

thecodingwizard (approved):

freider (commented):
lgtm

freider (approved):

PR #3090, opened 2025-05-06
Store the server url during the token flow if needed

erikbern:
Super minor 5 min thing that makes it easier to bootstrap tokens when in Modal dev environments.


erikbern:
hopefully uncontroversial

freider (commented):
nice!

PR #3091, opened 2025-05-06
Add RDMA as a flag to the clustered decorator

pawalt:
RDMA makes sense here since it's only going to be allowed for clustered functions, and as we ship to clients, we need a better interface than experimental options.

## Describe your changes

Adds an `rdma` field to experimental functions.

---

---

</details>

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -475,6 +475,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         i6pn_enabled: bool = False,
         # Experimental: Clustered functions
         cluster_size: Optional[int] = None,
+        rdma: Optional[bool] = None,
         max_inputs: Optional[int] = None,
         ephemeral_disk: Optional[int] = None,
         # current default: first-party, future default: main-package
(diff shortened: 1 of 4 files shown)


mwaskom:
Will users not generally want `RDMA` enabled on their clusters? (Wondering about the default).

thundergolfer (commented):
Can we validate the dumb easy stuff like CPU-only not being usable with RDMA. Fine with this also being server-side validation, because that's easy to evolve. 

Really `rdma=True` should only be allowed with H100s currently.

thundergolfer:
I think it's a good question @mwaskom. I think most users will be using with RDMA enabled. If we make default `True` then the default `modal.Function` won't work with the default `clustered` decorator because CPU-only can't use RDMA.

mwaskom:
The mutually-incompatible-individually-optimal defaults is indeed annoying though :(

pawalt:
@thundergolfer this is already true. RDMA requires `H100:8` with server-side validation.

mwaskom:
Not sure if it's a good idea but could the default be `True` with it being ignorable in the case of otherwise incompatible resources?

pawalt:
> the default modal.Function won't work with the default clustered decorator because CPU-only can't use RDMA.

I don't think we should make `rdma=True` the default because it's a very sparse resource. That said, this isn't true. CPU tasks technically can use RDMA. RDMA can be used for transferring regular memory not just GPU memory.

thundergolfer:
> That said, this isn't true. CPU tasks technically can use RDMA. RDMA can be used for transferring regular memory not just GPU memory.

Yeh I meant in practice. It would be very un-economical for us.

thundergolfer (approved):

PR #3092, opened 2025-05-06
Add 'modal cluster' subcommand

thundergolfer:
## Describe your changes

Minimal convenient CLI functionality for multi-node users.

```
 Usage: modal cluster [OPTIONS] COMMAND [ARGS]...

 Manage and connect to running multi-node clusters.

╭─ Options ─────────────────────────────────────────────────────────────────────╮
│ --help          Show this message and exit.                                   │
╰───────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ────────────────────────────────────────────────────────────────────╮
│ list    List all clusters that are currently running.                         │
│ shell   Open a shell to a multi-node cluster node.                            │
╰───────────────────────────────────────────────────────────────────────────────╯
```

image

---

---

</details>

## Changelog

* Experimental `modal cluster` subcommand is added.

### DIFF
--- a/modal/cli/entry_point.py
+++ b/modal/cli/entry_point.py
@@ -10,6 +10,7 @@ from modal._utils.async_utils import synchronizer
 
 from . import run
 from .app import app_cli
+from .cluster import cluster_cli
 from .config import config_cli
 from .container import container_cli
 from .dict import dict_cli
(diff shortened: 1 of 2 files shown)


modal/cli/entry_point.py
```diff
@@ -92,6 +93,8 @@ async def setup(profile: Optional[str] = None):
 # Deployments
 entrypoint_cli_typer.add_typer(app_cli, rich_help_panel="Deployments")
 entrypoint_cli_typer.add_typer(container_cli, rich_help_panel="Deployments")
+# TODO: cluster is hidden while multi-node is in beta/experimental
```
thundergolfer, line 96: I was interested in having some config-based way to enable experimental functionality, like `experimental=true` in `~/.modal.toml`. But this didn't seem trivial so I skipped if for this change set.

mwaskom:
If you're sharing a workspace where people are running a few clusters, is this output going to make it easy to know which one is "yours"? Looks like we associate it only with the App ID which, while semi-public, isn't obvious or prominent.

cweld510 (approved):
Looks reasonable. I think it would be nice to include the app name as well as the app ID so users can distinguish between their clusters.

thundergolfer:
>  Looks like we associate it only with the App ID which, while semi-public, isn't obvious or prominent.

Yeh I think this is a UX issue. It would be nice to see App name. Can fix it by updating RPCs, but not high priority IMO.

PR #3093, opened 2025-05-06
Deprecate passing data directly to Dict constructors

mwaskom:
## Describe your changes

This has its uses, but it's an odd one-off pattern with Modal objects, and the semantics in the case of `modal.Dict.from_name` are poorly defined.

Closes CLI-390

---

---

</details>

## Changelog

- Adds a deprecation warning when data is passed directly to `modal.Dict.from_name` or `modal.Dict.ephemeral`. Going forward, it will be necessary to separate `Dict` population from creation.

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -73,7 +73,7 @@ class _Dict(_Object, type_prefix="di"):
     @asynccontextmanager
     async def ephemeral(
         cls: type["_Dict"],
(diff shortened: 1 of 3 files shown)


mwaskom:
@prbot approve

danielshaar (approved):
🙏

PR #3094, opened 2025-05-06
Support full dict.update API in modal.Dict

mwaskom:
Noticed that this didn't work as expected.

---

---

</details>

## Changelog

- `modal.Dict.update` now also accepts a positional Mapping, like Python's `dict` type:

    ```python
    d = modal.Dict.from_name("some-dict")
    d.update({"a_key": 1, "another_key": "b"}, some_kwarg=True)
    ```


mwaskom:
@prbot approve

PR #3095, opened 2025-05-07
Add support for calling input plane server

rculbertson:
This PR adds support for calling input plane functions.

Calls to FunctionGet and FunctionCreate return the input_plane_url, so the client can then connect to the input plane in the correct region for the function.

Closes SVC-492

---

---

</details>


thecodingwizard (approved):
awesome!!

rculbertson:
@mwaskom Would you like to review this PR, or should we go ahead and merge?

mwaskom:
No strong opinions here; feel free to merge!

rculbertson:
Thanks!

PR #3097, opened 2025-05-07
volume: Add fast path for downloading files into file-like object

dflemstr:
## Describe your changes

- This resurrects the old `Volume.read_file_into_fileobj` with correct write semantics.
- It is beneficial to use this method for CLI `volume get` since it will concurrently download blocks and do scattered writes to disk in a scalable way.

---

---

</details>

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -528,9 +528,9 @@ class _Volume(_Object, type_prefix="vo"):
 
             return num_bytes_written
 
-        tasks = (asyncio.create_task(download_block(idx, url)) for idx, url in enumerate(response.get_urls))
+        coros = [download_block(idx, url) for idx, url in enumerate(response.get_urls)]
 
-        total_size = sum(await asyncio.gather(*tasks))
+        total_size = sum(await asyncio.gather(*coros))
         fileobj.seek(start_pos + total_size)
 
         return total_size
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -447,6 +447,96 @@ class _Volume(_Object, type_prefix="vo"):
 
 
     @live_method
+    async def read_file_into_fileobj(
+        self,
+        path: str,
+        fileobj: typing.IO[bytes],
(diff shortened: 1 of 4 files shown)


gimaker (approved):

test/conftest.py
```diff
@@ -2015,7 +2015,7 @@ async def VolumePutFiles2(self, stream):
                         assert len(block_data) == BLOCK_SIZE
                     # If this is the last block, it must be at most BLOCK_SIZE
                     if block_index + 1 == len(file.blocks):
-                        assert len(block_data) < BLOCK_SIZE
+                        assert len(block_data) <= BLOCK_SIZE
```
dflemstr, line 2018: Test fixture had a bug here

modal/volume.py
```diff
@@ -446,6 +446,96 @@ async def _read_file2(self, path: str) -> AsyncIterator[bytes]:
...
+                            n = fileobj.write(chunk)
+
+                        num_chunk_bytes_written += n
+                        progress_cb(advance=n)
+
+                    num_bytes_written += len(chunk)
+
+            return num_bytes_written
+
+        tasks = (asyncio.create_task(download_block(idx, url)) for idx, url in enumerate(response.get_urls))
```
erikbern, line 531: i think you can just create coros here as opposed to tasks, i.e.

```python
coros = [download_block(idx, url) for idx, url in enumerate(response.get_urls)]
```

(also a bit better i think to explicitly unwind the generator into a list, otherwise it will be lazy evaluated later)

creating tasks is a bit sketchy because python won't garbage collect them once you goes out of scope – you need something like `TaskContext` for that. but coros is probably better

dflemstr: Makes sense, I'll change. I don't think the list-vs-no-list makes a difference since we do `asyncio.gather(*tasks)` after but I might as well change it to avoid foot-guns in future refactors.

PR #3098, opened 2025-05-07
Upgrade warning on passing flagged options to Image.pip_install

mwaskom:
This was a pretty old deprecation; I think it's fine to make it an error now.

Note that we have similar issues in a few other "list of arguments" Image methods but I think we might need to live with them for 1.0.

---

---

</details>

## Changelog

- Passing flagged options to the `Image.pip_install` package list will now raise an error. Use the `extra_options`  parameter for additional flags that aren't exposed through the `Image.pip_install` signature:
  ```python
  image.pip_install("flash-attn", "--no-build-isolation")  # This is now an error!
  image.pip_install("flash-attn", extra_options="--no-build-isolation")  # Correct spelling
  ```

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1011,13 +1011,16 @@ class _Image(_Object, type_prefix="im"):
         pkgs = _flatten_str_args("pip_install", "packages", packages)
         if not pkgs:
             return self
+        elif not _validate_packages(pkgs):
+            raise InvalidError(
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot approve @charlesfrye

charlesfrye (approved):
Nice, thanks!

PR #3099, opened 2025-05-07
Update modal.Dict docstring to reflect new Dict behavior.

danielshaar:
## Describe your changes

SVC-519

Won't submit until all new dicts have the new backend enabled.

## Changelog

- New `modal.Dict`s (forthcoming on 2025-05-20) use a new durable storage backend that is more "cache-like" - items expire after 7 days of inactivity (no reads or writes). Previously created `modal.Dict`s will continue to use the old backend, but support will eventually be dropped.

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -38,6 +38,10 @@ class _Dict(_Object, type_prefix="di"):
     An individual Dict entry will expire after 7 days of inactivity (no reads or writes). The
     Dict entries are written to durable storage.
 
+    Legacy Dicts (created before 2025-05-21) will still have entries expire 30 days after being
+    last added. Additionally, data are stored in memory on the Modal server and could be lost due to
+    unexpected server restarts. Eventually, these Dicts will be fully sunset.
+
     **Usage**
 
     ```python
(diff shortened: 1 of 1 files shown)


mwaskom:
Do you not want to include the old / new behavioral differences in the docstring? I guess some chance that people with an old Dict get surprised by this?

mwaskom (approved):
I'm fine with this although there's also an argument to be made that these details are not really "API reference" and should go in the Guide. That would also make it easier to sync backend changes with documentation changes.

danielshaar:
> I'm fine with this although there's also an argument to be made that these details are not really "API reference" and should go in the Guide. That would also make it easier to sync backend changes with documentation changes.

I agree in principle, but we do this for https://modal.com/docs/reference/modal.Queue as well and I feel like folks def rely on these details being here and not just in the guide

mwaskom:
Could also fold this into the other PR?

danielshaar:
Sure, deleting this PR in favor of merging the update into https://github.com/modal-labs/modal-client/pull/3101

PR #3100, opened 2025-05-07
Add proto support for future put if_not_exists Dict feature.

danielshaar:
## Describe your changes

SVC-519

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1112,9 +1112,11 @@ message DictPopResponse {
 message DictUpdateRequest {
   string dict_id = 1 [ (modal.options.audit_target_attr) = true ];
   repeated DictEntry updates = 2;
+  bool if_not_exists = 3;
 }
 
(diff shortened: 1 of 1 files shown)


mwaskom (approved):

modal_proto/api.proto
```diff
@@ -1112,9 +1112,11 @@ message DictPopResponse {
 message DictUpdateRequest {
   string dict_id = 1 [ (modal.options.audit_target_attr) = true ];
   repeated DictEntry updates = 2;
+  bool if_not_exists = 3;
 }
 
 message DictUpdateResponse {
+  bool created = 1;
```
mwaskom, line 1119: You're not adding support for this in `.update` right now and I think that's fine, but this _may_ cause issues in the future if we want to do that, since presumably the updates would be optional at the level of individual keys rather than whole payloads.

mwaskom (approved):

PR #3101, opened 2025-05-07
Add support for Dict put skip_if_exists.

danielshaar:
## Describe your changes

SVC-519

---

---

## Changelog

- New `modal.Dict`s (forthcoming on 2025-05-20) use a new durable storage backend with more "cache-like" behavior - items expire after 7 days of inactivity (no reads or writes). Previously created `modal.Dict`s will continue to use the old backend, but support will eventually be dropped.
- `modal.Dict.put` now supports an `skip_if_exists` flag that can be used to avoid overwriting the value for existing keys:

  ```
  item_created = my_dict.put("foo", "bar", skip_if_exists=True)
  assert item_created
  new_item_created = my_dict.put("foo", "baz", skip_if_exists=True)
  assert not new_item_created
  ```

  Note that this flag only works for `modal.Dict` objects with the new backend (forthcoming on 2025-05-20) and will raise an error otherwise.

### DIFF
--- a/test/dict_test.py
+++ b/test/dict_test.py
@@ -64,9 +64,9 @@ def test_dict_update(servicer, client):
         assert sorted(items) == [("bar", 2), ("baz", 4), ("foo", 3)]
 
 
-def test_dict_put_if_not_exists(client):
+def test_dict_put_skip_if_exists(client):
     with Dict.ephemeral(client=client, _heartbeat_sleep=1) as d:
-        assert d.put("foo", 1, if_not_exists=True)
-        assert not d.put("foo", 2, if_not_exists=True)
+        assert d.put("foo", 1, skip_if_exists=True)
+        assert not d.put("foo", 2, skip_if_exists=True)
         items = list(d.items())
         assert items == [("foo", 1)]
--- a/test/dict_test.py
+++ b/test/dict_test.py
@@ -62,3 +62,11 @@ def test_dict_update(servicer, client):
         d.update({"foo": 1, "bar": 2}, foo=3, baz=4)
         items = list(d.items())
         assert sorted(items) == [("bar", 2), ("baz", 4), ("foo", 3)]
+
+
+def test_dict_put_if_not_exists(client):
+    with Dict.ephemeral(client=client, _heartbeat_sleep=1) as d:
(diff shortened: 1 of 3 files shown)


test/dict_test.py
```diff
@@ -62,3 +62,11 @@ def test_dict_update(servicer, client):
         d.update({"foo": 1, "bar": 2}, foo=3, baz=4)
         items = list(d.items())
         assert sorted(items) == [("bar", 2), ("baz", 4), ("foo", 3)]
+
+
+def test_dict_put_if_not_exists(client):
+    with Dict.ephemeral(client=client, _heartbeat_sleep=1) as d:
+        assert d.put("foo", 1, if_not_exists=True)
+        assert not d.put("foo", 2, if_not_exists=True)
```
mwaskom, line 70: I think you'll need to update the mock servicer to implement this, if you haven't.

mwaskom (approved):

freider:
I can't quite put my finger on it, but `if_not_exists=True` feels a bit odd as a boolean arg in my head. An `overwrite=False` (defaulting to True) would feel more intuitive to me, but maybe making separate method `setdefault` to mimic the python builtin dict would be even better? Arguably a bit weird since we don't mimic it very well otherwise with `put`... (`putdefault`?)

test/dict_test.py
danielshaar: done

danielshaar:
> I can't quite put my finger on it, but `if_not_exists=True` "sounds" a bit odd as a boolean arg in my head. An `overwrite=False` (defaulting to True) would feel more intuitive to me, but maybe making separate method `setdefault` to mimic the python builtin dict would be even better? Arguably a bit weird since we don't mimic it very well otherwise with `put`... (`putdefault`?)

I think if_not_exists being used in SQL, Redis, and S3 makes it a natural enough choice. I don't like setdefault as I don't want to force the user to do a get if the value is already there. I'm okay with a separate method, but I kinda hate the setdefault naming / API, it's quite confusing

mwaskom:
> I think if_not_exists being used in SQL, Redis, and S3 makes it a natural enough choice

Just flagging that @freider is having the same reaction as me: in these other contexts its a _flag_, here it's a parameter that accepts a boolean.

PR #3102, opened 2025-05-08
Add id to TaskLogs proto

annacarey:
Resolves PRD-21

This PR adds id to the TaskLogs proto. This supports work to enable opening a specific log with a unique URL.

It's required for this PR: https://github.com/modal-labs/modal/pull/21926


modal_proto/api.proto
```diff
@@ -2769,6 +2769,7 @@ message TaskLogs {
   TaskProgress task_progress = 9;
   string function_call_id = 10;
   string input_id = 11;
+  string id = 12;
```
mwaskom, line 2772: what kind of id is this?

annacarey: It's a unique identifier but the value is something like `{task_id}-{timestamp_nanoseconds}`

annacarey:
@prbot approve

PR #3106, opened 2025-05-09
Add `Cls.with_concurrency` and `Cls.with_batching`

mwaskom:
## Describe your changes

- Closes CLI-364

---

---

</details>

## Changelog

- Introduced `modal.Cls.with_concurrency` and `modal.Cls.with_batching` for runtime configuration of functionality that is exposed through the `@modal.concurrent` and `@modal.batched` decorators.
  ```python
  model = Model.with_options(gpu="H100").with_concurrency(max_inputs=100)()
  ```
- Added a deprecation warning when using `allow_concurrent_inputs` in `modal.Cls.with_options`.
- Added `buffer_containers` to `modal.Cls.with_options`.
- *Behavior change:* when `modal.Cls.with_options` is called multiple times on the same object, the configurations will be merged rather than using the most recent.

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -75,14 +75,18 @@ def _get_class_constructor_signature(user_cls: type) -> inspect.Signature:
 
 @dataclasses.dataclass()
 class _ServiceOptions:
-    secrets: typing.Collection[_Secret]
-    resources: Optional[api_pb2.Resources]
-    retry_policy: Optional[api_pb2.FunctionRetryPolicy]
-    concurrency_limit: Optional[int]
-    timeout_secs: Optional[int]
-    task_idle_timeout_secs: Optional[int]
-    validated_volumes: typing.Sequence[tuple[str, _Volume]]
-    target_concurrent_inputs: Optional[int]
+    secrets: typing.Collection[_Secret] = ()
+    validated_volumes: typing.Sequence[tuple[str, _Volume]] = ()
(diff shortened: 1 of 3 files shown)


mwaskom:
This is working against prod now:

```
import time
import modal

app = modal.App()

@app.cls()
class C:
    @modal.method()
    def f(self):
        time.sleep(20)


@app.local_entrypoint()
def main():
    c = C.with_concurrency(max_inputs=100)()
    fcs = [c.f.spawn() for _ in range(5)]
    time.sleep(10)
    stats = c.f.get_current_stats()
    assert stats.backlog == 0
    assert stats.num_total_runners == 1
```

mwaskom:
@prbot approve @freider

modal/cls.py
```diff
@@ -421,7 +425,7 @@ class _Cls(_Object, type_prefix="cs"):
     """
 
     _class_service_function: Optional[_Function]  # The _Function (read "service") serving *all* methods of the class
-    _options: Optional[_ServiceOptions]
+    _options: _ServiceOptions
```
freider, line 428: I think it used to be the case that we used the presence of `_options` to determine if we needed to load an instance service function (FunctionBindParams) or not - i.e. if we had to do `FunctionBindParams` or could just use the class service function. I *think* we have tests for this "optimization" but could be worth double checking

PR #3109, opened 2025-05-09
Reapply changes to ignore pattern interpretation

mwaskom:
## Describe your changes

-Closes CLI-393

Second attempt at #3049 

---

---

</details>

## Changelog

- Fixes the behavior of `ignore=` in `modal.Image` methods, including when `.dockerignore` files are implicitly used in docker-oriented methods. This may result in Image rebuilds with different final inventories:
  - When using `modal.Image.add_local_dir`, exclusion patterns are now correctly interpreted as relative to the directory being added (e.g., `*.json` will now ignore all json files in the top-level of the directory).
  - When using `modal.Image.from_dockerfile`, exclusion patterns are correctly interpreted as relative to the context directory.
  - As in Docker, leading or trailing path delimiters are stripped from the ignore patterns before being applied.
  - **Breaking change**: When providing a custom function to `ignore=`, file paths passed into the function will now be _relative_, rather than absolute.

### DIFF
--- a/test/file_pattern_matcher_test.py
+++ b/test/file_pattern_matcher_test.py
@@ -272,11 +272,6 @@ def test_match():
             assert FilePatternMatcher(pattern)._matches(text) is expected
 
 
-def test_absolute_path_error():
-    with pytest.raises(ValueError, match="cannot be absolute paths"):
-        FilePatternMatcher("/home/data/")
-
-
 def __helper_get_file_paths(tmp_path: Path) -> list[Path]:
     file_paths = []
     for root, _, files in os.walk(tmp_path):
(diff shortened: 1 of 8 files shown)


test/file_pattern_matcher_test.py
```diff
@@ -272,11 +272,6 @@ def test_match():
             assert FilePatternMatcher(pattern)._matches(text) is expected
 
 
-def test_absolute_path_error():
```
freider: should we keep this test but assert that absolute paths work as we expect them to - as a regression test?

freider (approved):

test/file_pattern_matcher_test.py
mwaskom: Hm yeah I added handling of leading / trailing slashes to the basic `test_match` test in `file_pattern_matching_test`

PR #3111, opened 2025-05-12
proto changes for client dependency mounting

matslina:
## Describe your changes

Flag on Function for requesting dependency mounts at runtime. Additional
associated metadata on ImageMetadata.

---

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1369,13 +1369,17 @@ message Function {
   // Note the value type as string. Internally we'll coerce all values to string with str().
   // On the server, it's necessary to convert back to the most natural type (e.g. int) when relevant.
   map<string, string> experimental_options = 81;
+
+  // If set, client deps will be mounted into the container, and are
+  // no longer expected to exist in the image itself.
+  bool mount_client_dependencies = 82;
 }
 
 message FunctionAsyncInvokeRequest {
   string function_id = 1;
   string parent_input_id = 2;
   FunctionInput input = 3;
-} 
+}
 
 message FunctionAsyncInvokeResponse {
   bool retry_with_blob_upload = 1;


mwaskom:
What's going to be setting the field on Function here? In theory we should be able to figure this out sever-side right?

BTW while your trick with the pythonpath saves us from needing to mount each individual dependency separately (yay) I do think it would be useful to design towards a low-cardinality set of dependency mounts with some optionality — mainly so that we can reintroduce "automatic" provisioning of fastapi when `@modal.fastapi_endpoint` is used (although we could also then make the `aiohttp` stack optionally dependent on `@modal.asgi_app`, as I believe that's the only time those are strictly needed.

mwaskom:
> I think it's best to do this client side, since some functions (like the "HTTP containers") will not want the client mounts. In a later change, that flag would e.g. not be set for the http forwarding case.

Hm yeah if "HTTP containers" will be defined with an `api_pb2.Function` then I could see how that would make sense. OTOH if there were some other way to distinguish them, it might be simpler to just assume "all Functions get client dependencies, no HTTP containers don't" server side? Mostly I just have some preference for pushing more Image definition logic over to the server.

> I don't know that I fully understand the issues with fastapi, but I still think that should work pretty well with the proposed design. It'd be a new mount for fastapi, and as long as it goes into the same /__modal/.client_pkgs it should just work?

Yeah I think it should work the same way and agree it doesn't need to be fully implemented now. The main thing to care about is whether a single boolean will be expressive enough / extensible in the future for where you might want to specify, e.g. that a container should be provisioned with the core client dependencies, but not the asgi_app dependencies (but maybe with fastapi).

matslina:
> might be simpler to just assume "all Functions get client dependencies, no HTTP containers don't" server side? Mostly I just have some preference for pushing more Image definition logic over to the server.

I was about to write it roughly that way, but felt that this made it all easier to follow. A whole lot of these runtime flags are already on Function, and I think this one fits pretty well there too. Otherwise, if done in server, there'd need to be a check for image builder version (a concept which is primarily present client side), which I wouldn't mind avoiding. Not hard to do, but less clean than this, at least in my eyes. LMK what you think.

> The main thing to care about is whether a single boolean will be expressive enough / extensible

One option is to do a flag per package (or passing a list of packages) to mount. Another is to use this bool for the "core dependencies", and add a new bool for any additional dependency. I'm leaning towards the latter because it seems simpler to me, but also open to doing it differently if you have a proposal?

matslina:
@prbot approve @mwaskom

PR #3112, opened 2025-05-12
Update various comments in dict.py.

danielshaar:
Use Dict more consistently instead of dict or dictionary.

Update len's comment to include the new Dict limitation and remove the part about expired keys being misreported for legacy Dict. I don't think it's too worth keeping that comment accurate for legacy Dict.

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -39,8 +39,8 @@ class _Dict(_Object, type_prefix="di"):
     Dict entries are written to durable storage.
 
     Legacy Dicts (created before 2025-05-20) will still have entries expire 30 days after being
-    last added. Additionally, data is stored in memory on the Modal server and could be lost due to
-    unexpected server restarts. Eventually, these Dicts will be fully sunset.
+    last added. Additionally, contents are stored in memory on the Modal server and could be lost
+    due to unexpected server restarts. Eventually, these Dicts will be fully sunset.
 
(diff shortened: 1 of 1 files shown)


modal/dict.py
```diff
@@ -238,7 +238,10 @@ async def contains(self, key: Any) -> bool:
 
     @live_method
     async def len(self) -> int:
-        """Return the length of the dictionary, including any expired keys."""
+        """Return the length of the Dict.
+
+        Note: This is an expensive operation and will return at most 100,000.
```
mwaskom, line 243: Is this a change relative to the previous Dict implementation?

modal/dict.py
```diff
@@ -39,7 +39,7 @@ class _Dict(_Object, type_prefix="di"):
     Dict entries are written to durable storage.
 
     Legacy Dicts (created before 2025-05-20) will still have entries expire 30 days after being
-    last added. Additionally, data are stored in memory on the Modal server and could be lost due to
+    last added. Additionally, data is stored in memory on the Modal server and could be lost due to
```
mwaskom, line 42: Debatable.....

mwaskom (approved):

modal/dict.py
danielshaar: Let's kill the controversy

PR #3114, opened 2025-05-12
Fix `modal app history --json` output

mwaskom:
## Describe your changes

We had a bug in `modal app history --json` when the history contained a mixture of versions with/without a tag or git commit.

Given that we collect git commits by default going forward (unless collection fails) I think it's fine to always include that column in this output. For tag (which is optional) I preserved the "only show when used" behavior but used a null placeholder when one wasn't provided.

## Changelog

- Fixed the behavior of `modal app history --json` when the history contains versions with and without commit information or "tag" metadata. Commit information is now always included (with a `null` placeholder when absent), while tag metadata is included only when there is at least one tagged release (other releases will have a `null` placeholder).

### DIFF
--- a/modal/cli/app.py
+++ b/modal/cli/app.py
@@ -224,10 +224,10 @@ async def history(
         "Time deployed",
         "Client",
         "Deployed by",
+        "Commit",
+        "Tag",
     ]
     rows = []
-    deployments_with_tags = False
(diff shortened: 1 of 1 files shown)


mwaskom:
@prbot approve

gleb-modal:
LGTM thank you

PR #3115, opened 2025-05-12
Prepare CI-CD workflow for 1.0 release model

mwaskom:
## Describe your changes

Closes CLI-392

Going forward, we'll still release every (main) commit to PyPI, but we want to be able to distinguish between "development" versions and true releases. This PR updates the auto-version-incrementing logic to always produce a *dev* version, and to run only when the precipitating commit did not modify the version.

The idea is that you trigger a true release by updating the version manually (to a non-development number).

The process of creating a release should also involve cleaning up the changelog file. We're keeping the auto PR description => CHANGELOG.md step, but it's going to have per-development-version updates. Our true releases should have a changelog for that release that is organized, prioritizing the most important updates, etc. TBD on what the best intermediate format is (i.e. should we still have section headers for development versions?) but avoiding changing too much of that now.

A few other things are getting changed here. We'll also only create a new git tag for true releases.

I would sort of like to move the version _into_ the `modal` package itself. Having a distinct `modal_version` package is vestigial and a little weird, but we do depend on that setup internally in a few small ways.


tasks.py
```diff
@@ -303,26 +303,65 @@ def publish_base_images(
...
-build_number = {new_build_number}  # git: {git_sha}
-"""
-        )
+    git_hash = ctx.run("git rev-parse --short=8 HEAD", hide="out").stdout.rstrip()
+
+    if v.is_prerelease:
+        if not v.is_devrelease:
+            raise RuntimeError("We only know how to auto-bump dev versions")
+        # For dev releases, increment the dev suffix
+        next_version = f"{v.major}.{v.minor}.{v.micro}.dev{v.dev + 1}+{git_hash}"
```
freider, line 342: til that `.dev` is an actual component of Version!

mwaskom:
@prbot approve @freider what's the worst that can happen

mwaskom:
@prbot approve @freider

freider (commented):
I'm not sure I follow completely - where is the logic that determines whether to publish a pre vs a non-pre release?

mwaskom:
> I'm not sure I follow completely - where is the logic that determines whether to publish a pre vs a non-pre release?

There's actually no push-side logic for that; it's all in the version specifier. If you include a `.dev*`, `.a*`, `.b*`, or `.rc*` suffix, then it's treated as a pre-release.

PR #3118, opened 2025-05-13
Fix the version filename when pushing back to main

mwaskom:
Followup to #3115; forgot to update the name of the file that we explicitly add when updating the version.


freider (commented):
👍

PR #3119, opened 2025-05-13
Manually update development version correctly

mwaskom:
Followup to #3118 

Fussy fussy; when manually updating the development version need to remember to set it to the _next_ version now.

### DIFF
--- a/modal_version/__init__.py
+++ b/modal_version/__init__.py
@@ -1,4 +1,4 @@
 # Copyright Modal Labs 2025
 """Supplies the current version of the modal client library."""
 
-__version__ = "0.76.4.dev0"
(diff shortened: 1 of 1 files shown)


mwaskom:
@prbot approve @freider

freider:
How often do we need to update the dev version manually?

mwaskom:
> How often do we need to update the dev version manually?

In the stable state it shouldn't be something we need to do, but I needed to do it a few times here while updating the workflows, because bugs made them fail between updating the version and actually doing the release, which the simple automated system isn't robust to (and then in this case, I had previously manually set it to the wrong thing).

TBD on exactly how often flaky CI + merge race conditions cause issues that need manual fixing like this.

PR #3121, opened 2025-05-13
Fix base image checkout

mwaskom:
Followup to #3115; I think this might be the final thing to fix.


mwaskom:
@prbot approve @freider

freider (commented):
😵

PR #3123, opened 2025-05-13
Fix hash used for releases

mwaskom:
Fixes to the CI-CD workflow to ensure that we check out the correct commit when we do the release.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -25,7 +25,7 @@ jobs:
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     outputs:
-      git-hash: ${{ steps.metadata.outputs.git_hash }}
+      release-hash: ${{ steps.release-data.outputs.hash }}
 
     steps:
       - name: Generate token for Github PR Bot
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -50,20 +50,22 @@ jobs:
...
-         echo "git_hash=`git rev-parse HEAD`" >> "$GITHUB_OUTPUT"
+      - name: Get release tag
+        id: tag
+        run: echo "release_tag=`inv get-release-tag`" >> "$GITHUB_OUTPUT"
 
       - uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9
         with:
           add: modal_version/__init__.py CHANGELOG.md
-          tag: ${{ steps.metadata.outputs.release_tag }}
+          tag: ${{ steps.tag.outputs.release_tag }}
```
mwaskom, line 60: There's still a problem here which is that this doesn't push the tag when there's nothing to commit. Not sure if that's a configuration I can set or if we're going to need to drop down to manual git (which seems fine in principle but I expect it will involve more thrashing with CI to get it right 😢).

mwaskom:
@prbot approve @freider

freider:
I think I've lost track of all the changes now (also not a github actions expert, so a lot of this is new to me) 😵 Will try to review the full CI flow instead

mwaskom:
Yeah most of these are just fixes to dumb issues that were hard to catch without running the workflows. Happy to walk you through the final state if it would be helpful!

PR #3125, opened 2025-05-13
Add a `--from-env` option to `modal secret create`

ehdr:
## Describe your changes

Resolves CLI-381 (at least partially - could easily add a `--from-file` too).

Can be combined with `key=val`s as before:
```
ENVKEY=val modal secret create env-secrets --force --from-env ENVKEY
ENVKEY=val modal secret create env-secrets --force key=val --from-env ENVKEY
ENVKEY1=val ENVKEY2=val modal secret create env-secrets --force key=val --from-env ENVKEY1 --from-env ENVKEY2
```

I chose to go with the default Typer multi-value option behavior of having to repeat the `--from-env`, but could as well go for `--from-env=ENVKEY1,ENVKEY2` if you think that is cleaner.

</details>

## Changelog

- Added a `--from-env` flag to `modal secret create`, which allows creating secrets by reading the secret value from an environment variable. This is especially useful for applications (like CI) where your code is checked in to version control.

### DIFF
--- a/modal/cli/secret.py
+++ b/modal/cli/secret.py
@@ -48,13 +48,14 @@ async def list_(env: Optional[str] = ENV_OPTION, json: bool = False):
 @synchronizer.create_blocking
 async def create(
     secret_name,
-    keyvalues: list[str] = typer.Argument(..., help="Space-separated KEY=VALUE items"),
+    keyvalues: Optional[list[str]] = typer.Argument(default=None, help="Space-separated KEY=VALUE items"),
     env: Optional[str] = ENV_OPTION,
-    force: bool = typer.Option(False, "--force", help="Overwrite the secret if it already exists."),
+    from_env: Optional[list[str]] = typer.Option(default=None, help="Name of environment variable holding the secret"),
+    force: bool = typer.Option(False, "--force", help="Overwrite secrets that already exists."),
 ):
     env = ensure_env(env)
     env_dict = {}
-    for arg in keyvalues:
+    for arg in keyvalues or []:
         if "=" in arg:
             key, value = arg.split("=", 1)
             if value == "-":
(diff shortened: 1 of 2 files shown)


ehdr:
My first Modal PR, so please feel free to shoot this down if something doesn't make sense!

modal/cli/secret.py
```diff
@@ -48,13 +48,14 @@ async def list_(env: Optional[str] = ENV_OPTION, json: bool = False):
 @synchronizer.create_blocking
 async def create(
     secret_name,
-    keyvalues: list[str] = typer.Argument(..., help="Space-separated KEY=VALUE items"),
+    keyvalues: Optional[list[str]] = typer.Argument(default=None, help="Space-separated KEY=VALUE items"),
     env: Optional[str] = ENV_OPTION,
-    force: bool = typer.Option(False, "--force", help="Overwrite the secret if it already exists."),
+    from_env: Optional[list[str]] = typer.Option(default=None, help="Name of environment variable holding the secret"),
```
freider, line 53: nit: potentially confusing that we have an `--env` flag specifying the destination *modal environment* whereas the `--from-env` specifies a variable name in the *local environment*

Naming is so hard :(

test/cli_test.py
```diff
@@ -109,6 +109,28 @@ def test_secret_create_list_delete(servicer, set_env_client):
...
+        (["key=val"], [], 0),
+        (["key=val"], ["ENV_KEY1"], 0),
+        ([], ["ENV_KEY1"], 0),
+        (["key=val"], ["ENV_KEY1", "ENV_KEY2"], 0),
+        (["key1=val1", "key2=val2"], ["ENV_KEY1", "ENV_KEY2"], 0),
+        (["key=val"], ["MISSING_ENV_KEY"], 2),
+        ([], ["MISSING_ENV_KEY"], 2),
+    ],
+)
+def test_secret_create_from_env(servicer, set_env_client, keyvalues, from_envs, expected_exit_code):
```
freider, line 125: nice tests!

freider (approved):
looks good - not sure if you have some alternative ideas for naming that would be less ambiguous?

Also, let's not merge before 1.0

ehdr:
Yeah, also thought of naming, and also concluded it's hard! :D Open to suggestions, and yeah no rush merging obviously.

modal/cli/secret.py
ehdr: Maybe `--from-env-var`? But (even more) verbose...

ehdr: FWIW [Vercel has similar challenges](https://vercel.com/docs/cli/env) 😄 

image

mwaskom (approved):
Comment is definitely non-blocking either way.

ehdr:
I suggest closing this and going with #3185 instead, but open to input @freider @mwaskom !

ehdr:
Yes exactly! Closing this for now, thanks all!

PR #3126, opened 2025-05-13
Allow empty commit when tagging releases

mwaskom:
Should avoid the problem we ran into where because "release" commits manually update the version and changelog, there's nothing changed in the `client-versioning` workflow so the `add-and-commit` action bails out early and fails to tag the release.


mwaskom:
@prbot approve @freider

freider (commented):
😵

PR #3127, opened 2025-05-13
Abandon use of local info (git hash) in dev versions

mwaskom:
Dropping the `+{git-hash}` from our development version scheme; didn't realize this was blocked by PyPI. Since you're either getting the client from PyPI or using it via a git checkout, it doesn't seem like it adds much info.

It does seem to be a relatively common scheme, but now I'm not sure why; I guess if you log your version you get the get metadata as a bonus. We can revisit it the reverse lookup from dev version to commit feels a lot harder going forward. (Previously we had a tag for every commit, so it was a bit more straightforward).

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -325,7 +325,7 @@ def bump_dev_version(ctx, dry_run: bool = False, force: bool = False):
     """
     version_file = "modal_version/__init__.py"
     commit_files = ctx.run("git diff --name-only HEAD~1 HEAD", hide="out").stdout.splitlines()
-    if version_file in commit_files:
+    if version_file in commit_files and not force:
         print(f"Aborting: {version_file} was modified by the most recent commit")
         return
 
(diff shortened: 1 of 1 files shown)


mwaskom:
@prbot approve @freider

freider:
Not following exactly - is pypi preventing us from having the +githash suffix?

Also - why can't we tag the dev release hashes?

mwaskom:
> is pypi preventing us from having the +githash suffix?

Yeah it rejected the upload. Documented [here](https://peps.python.org/pep-0440/#local-version-identifiers):

> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server, but MAY be used to identify private builds created directly from the project source. Local version identifiers SHOULD be used by downstream projects when releasing a version that is API compatible with the version of the upstream project identified by the public version identifier, but contains additional changes (such as bug fixes). As the Python Package Index is intended solely for indexing and hosting upstream projects, it MUST NOT allow the use of local version identifiers.

PR #3131, opened 2025-05-13
Update client version in channel tests

mwaskom:
I don't really understand why these were hardcoded, but they broken when I set the version to `1.0.0.dev0`.


mwaskom:
@prbot approve @freider

PR #3133, opened 2025-05-14
modal_proto: Make dedicated RPCs for remaining volume v2 ops

dflemstr:
## Describe your changes

- We want to route all remaining v2 volume ops completely differently from v2 volume ops, so we create new dedicated RPCs for consistency.

---

## Compatibility checklist

---

## Release checklist

---

</details>


modal_proto/api.proto
```diff
@@ -2909,6 +2909,11 @@ message VolumeCopyFilesRequest {
   bool recursive = 4;
 }
 
+message VolumeDelete2Request {
```
gimaker, line 2912: Don't think this and `VolumeRenameRequest` should be duplicated for v2 (will divulge details/discuss out of band)

gimaker (approved):

PR #3134, opened 2025-05-14
Add `--timestamps` flag to `modal app logs`, that prepends timestamps

ehdr:
## Describe your changes

Resolves CLI-387.

## Changelog

- Added a `--timestamps` flag to `modal app logs`, that prepends the timestamp to each log line.

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -32,6 +32,7 @@ from rich.progress import (
 from rich.spinner import Spinner
 from rich.text import Text
 
+from modal._utils.time_utils import timestamp_to_local
 from modal_proto import api_pb2
 
 from ._utils.grpc_utils import RETRYABLE_GRPC_STATUS_CODES, retry_transient_errors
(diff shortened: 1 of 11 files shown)


freider (approved):
nice! How did it behave when used with something that has escape sequences like tqdm or torch progress? (since it's a flag I think it's fine if it doesn't behave well there, but still interesting to know)

freider:
I think this works well as a first step - my feeling is that it would take a lot of work (something like a custom terminal emulator with sub-windows) to make this work perfectly reliably with escape sequence things

modal/_output.py
```diff
@@ -81,22 +82,27 @@ def download_progress_bar() -> Progress:
     )
 
 
-class LineBufferedOutput(io.StringIO):
+class LineBufferedOutput:
```
ehdr, line 85: It wasn't quite a `StringIO` even before, and definitely isn't one with this change of `write`, and I don't think we need it to be?

freider: Yeah I think it makes sense to remove the base class

freider (approved):

freider:
On thought is that we might want to add the same flag to `modal run` for concistency (and possibly `modal deploy` which sometimes prints image build logs)

modal/_utils/time_utils.py
```diff
@@ -0,0 +1,14 @@
...
+def timestamp_to_local(ts: float, isotz: bool = True) -> str:
+    if ts > 0:
+        locale_tz = datetime.now().astimezone().tzinfo
+        dt = datetime.fromtimestamp(ts, tz=locale_tz)
+        if isotz:
+            return dt.isoformat(sep=" ", timespec="seconds")
+        else:
+            return f"{datetime.strftime(dt, '%Y-%m-%d %H:%M')} {locale_tz.tzname(dt)}"
+    else:
+        return None
```
mwaskom, line 14: Should the return time be `Optional[str]`? (Or should we keep str but use `""` as the null return value?)

mwaskom (approved):
> On thought is that we might want to add the same flag to modal run for consistency (and possibly modal deploy which sometimes prints image build logs)

Good point, but also the proliferation of random flags is always pretty annoying. Maybe this makes more sense as a config option?

IMO fine to ship for `app logs` and not immediately for the other commands; those are showing logs "in real time" so there's less ambiguity.

Otherwise LGTM!

PR #3136, opened 2025-05-14
Remove Object.resolve method

mwaskom:
## Describe your changes

- Closes CLI-399
 
---

## Compatibility checklist

---

</details>

## Changelog

- Removed the `.resolve()` method on Modal objects. This method had not been publicly documented, but where used it can be replaced straightforwardly with `.hydrate()`. Note that explicit hydration should rarely be necessary: in most cases you can rely on lazy hydration semantics (i.e., objects will be hydrated when the first method that requires server metadata is called).

### DIFF
--- a/modal/_object.py
+++ b/modal/_object.py
@@ -10,7 +10,6 @@ from typing_extensions import Self
 
 from ._resolver import Resolver
 from ._utils.async_utils import aclosing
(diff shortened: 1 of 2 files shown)


mwaskom:
@prbot approve @freider

freider (commented):
This removes the resolver right? 🤡

PR #3137, opened 2025-05-14
Remove support for parameters in `@modal.asgi_app` / `@modal.wsgi_app`

mwaskom:
## Describe your changes

- Closes CLI-398

---

## Compatibility checklist

---

</details>

## Changelog

- Functions decorated with `@modal.asgi_app` or `@modal.wsgi_app` are now required to be nullary. Previously, we only warned in the case where a function was defined with all parameters having default arguments.


danielshaar (commented):
sick

PR #3138, opened 2025-05-14
Remove or re-type antiquated deprecation errors

mwaskom:
## Describe your changes

Some of these are old deprecation errors that we can just remove entirely, others are cases where we issued a deprecation error because behavior had changed, but the prior context is no longer relevant.

None of these were major changes.

- Part of CLI-401

## Compatibility checklist

</details>

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -762,23 +762,3 @@ class _Sandbox(_Object, type_prefix="sb"):
 
 
 Sandbox = synchronize_api(_Sandbox)
-
-
-def __getattr__(name):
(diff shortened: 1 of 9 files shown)


modal/sandbox.py
```diff
@@ -762,23 +762,3 @@ async def list(
 
 
 Sandbox = synchronize_api(_Sandbox)
-
-
-def __getattr__(name):
```
freider, line 767: omg, I didn't know about this one :D

freider (approved):
love it!

PR #3139, opened 2025-05-14
Remove `show_progress=` from `App.run`

mwaskom:
## Describe your changes

- Part of CLI-401

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Removed the `show_progress` parameter from `modal.App.run`. This parameter has been replaced by the `modal.enable_output` context manager:
  ```python
  with modal.enable_output(), app.run():
    ...  # Will produce verbose Modal output
  ```

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -328,7 +328,6 @@ class _App:
         self,
         *,
         client: Optional[_Client] = None,
-        show_progress: Optional[bool] = None,
         detach: bool = False,
         interactive: bool = False,
         environment_name: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


modal/app.py
```diff
@@ -374,16 +373,6 @@ async def run(
...
 
-        # See Github discussion here: https://github.com/modal-labs/modal-client/pull/2030#issuecomment-2237266186
-
-        if show_progress is True:
-            deprecation_error(
-                (2024, 11, 20),
-                "`show_progress=True` is no longer supported. Use `with modal.enable_output():` instead.",
-            )
-        elif show_progress is False:
-            deprecation_warning((2024, 11, 20), "`show_progress=False` is deprecated (and has no effect)")
```
mwaskom, line 385: Just removing this entirely instead of making it a `deprecation_error` since it had no effect anyway.

mwaskom:
There's still a `show_progress` parameter in `modal.enable_output`. I don't really understand why ... when would you do `modal.enable_output(show_progress=False)`? But I'm not sure that we should touch it here.

PR #3140, opened 2025-05-14
Remove automount concept

mwaskom:
## Describe your changes

- Closes CLI-400

---

## Compatibility checklist

---

</details>

## Changelog

- We've made some breaking changes to Modal's "automounting" behavior:
  - Previously, Modal containers would automatically include the source for local Python packages that were imported by your Modal App. Going forward, it will be necessary to explicitly include such packages in the Image (i.e., with `modal.Image.add_local_python_source`). If you've not already adapted your source code in response to warnings about changes in the automount behavior, Apps built with 1.0+ will have different files included and may not run as expected.
  - Support for the `automount` configuration (`MODAL_AUTOMOUNT`) has been removed; this environment variable will no longer have any effect.
  - Modal will continue to automatically include the Python module or package where the Function is defined. This is narrower in scope than the old automounting behavior: it's limited to at most a single package, and it includes only `.py` files. The limited automounting can also be disabled in cases where your Image definition already includes the package defining the App: set `include_source=False` in the `modal.App` constructor or `@app.function` decorator.

### DIFF
--- a/test/function_test.py
+++ b/test/function_test.py
@@ -22,7 +22,7 @@ from modal_proto import api_pb2
 from test.conftest import GrpcErrorAndCount
 from test.helpers import deploy_app_externally
 
-app = App(include_source=True)  # TODO: remove include_source=True when automount is disabled by default
+app = App()
 
 
 if os.environ.get("GITHUB_ACTIONS") == "true":
@@ -758,7 +758,7 @@ def test_from_id_iter_gen(client, servicer, is_generator):
         assert rehydrated_function_call.get() == "hello"
 
 
-lc_app = App(include_source=True)  # TODO: remove include_source=True when automount is disabled by default
+lc_app = App()
 
 
 @lc_app.function()
@@ -929,7 +929,7 @@ def test_calls_should_not_unwrap_modal_objects_gen(servicer, client):
     assert len(servicer.client_calls) == 1
 
 
-def test_function_deps_have_ids(client, servicer, monkeypatch, test_dir, set_env_client, disable_auto_mount):
+def test_function_deps_have_ids(client, servicer, monkeypatch, test_dir, set_env_client):
     monkeypatch.syspath_prepend(test_dir / "supports")
     app = App()
     app.function(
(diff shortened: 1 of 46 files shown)


freider:
changelog nit - it's technically the file/package where the *function* is defined (not the app) that's automatically included

test/function_test.py
```diff
@@ -1159,12 +1150,7 @@ def f():
     (tmp_path / "mod.py").touch()  # some file
     entrypoint_file.write_text(src)
 
-    monkeypatch.delenv("MODAL_AUTOMOUNT")
-    if config_automount is not None:
-        env = {**os.environ, "MODAL_AUTOMOUNT": config_automount}
-    else:
-        env = {**os.environ}
-    output = deploy_app_externally(servicer, credentials, str(entrypoint_file), env=env)
+    output = deploy_app_externally(servicer, credentials, str(entrypoint_file), env={**os.environ})
```
freider, line 1153: one cool thing is that `deploy_app_externally` might not be needed after this change (lets do that separately though) - iirc it's mostly in place because we needed a clean separate interpreter to not have tests interact through automount side effects when they import modules

test/mounted_files_test.py
```diff
@@ -117,46 +87,13 @@ def test_mounted_files_package(supports_dir, env_mount_files, servicer, server_u
...
-        "/root/a.py",
-        "/root/b/c.py",
-        "/root/b/e.py",
-        "/root/script.py",
-        "/root/pkg_b/__init__.py",
-        "/root/pkg_b/f.py",
-        "/root/pkg_b/g/h.py",
-    }
+    # No automount, so we only have the entrypoint file
+    assert files == {"/root/script.py"}
```
freider, line 96: such a lonely script - I'm not sure we need this test anymore - the whole `sys_prefix` thing is because of auto mount logic determining which modules should be included and not based on that

freider:
Lgtm!
Should also be able to remove a bunch of code in `mount.py`, e.g. `get_sys_modules_mounts`, but we can do that in another PR

mwaskom:
@prbot approve @freider

PR #3141, opened 2025-05-14
Update id to timestamp_ns for TaskLog

annacarey:
We decided to use `timestamp_ns` instead of `id` here. (`id` was a concatenation of task_id and timestamp_ns but now we'll just send timestamp_ns and concatenate a unique on the FE.)


gleb-modal (approved):
Should be fine since we haven't used this field yet

PR #3143, opened 2025-05-14
Accept auth token in initial metadata too

thecodingwizard:
Auth tokens can be sent in either the initial metadata (from Rust) or the trailing metadata (from Python). This PR adds support for accepting auth tokens from initial metadata.

I think we should consider sending auth tokens exclusively in initial metadata at some point.


thecodingwizard:
@prbot approve

mwaskom (approved):
👍

PR #3148, opened 2025-05-15
Remove modal.Stub

mwaskom:
## Describe your changes

Makes it an error to reference `modal.Stub`, whereas previously we had kept `Stub` as an alias for `App`. It's an `AttributeError`, but we have a special error message pointing the user towards `modal.App`.

---

## Compatibility checklist

</details>

## Changelog

- Referencing the deprecated `modal.Stub` object will now raise an `AttributeError`, whereas previously it was an alias for `modal.App`. This is a simple name change.

### DIFF
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -11,7 +11,7 @@ from modal_version import __version__
 try:
     from ._runtime.execution_context import current_function_call_id, current_input_id, interact, is_local
     from ._tunnel import Tunnel, forward
-    from .app import App, Stub
+    from .app import App
     from .client import Client
     from .cloud_bucket_mount import CloudBucketMount
     from .cls import Cls, parameter
@@ -54,6 +54,7 @@ except Exception:
     print()
     raise
 
+
 __all__ = [
     "__version__",
     "App",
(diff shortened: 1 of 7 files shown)


modal/__init__.py
```diff
@@ -99,3 +99,11 @@
     "web_server",
     "wsgi_app",
 ]
+
+
+def __getattr__(name):
```
erikbern, line 104: wow you can have a `__getattr__` on a module level 🤯

erikbern: will this also catch `from modal import Stub`?

mwaskom: > will this also catch from modal import Stub?

unfortunately not:

```
$ python -c 'from modal import Stub'
  File "<string>", line 1, in <module>
ImportError: cannot import name 'Stub' from 'modal' (/home/ec2-user/client/modal/__init__.py)
```

mwaskom: Another reason to avoid the `from x import y` spelling in documentation IMO.

mwaskom:
@prbot approve

modal/__init__.py
erikbern: not super important but in theory we could keep the `Stub` class and turn the warning into an error. But I suspect this usage is uncommon

PR #3155, opened 2025-05-16
Fix Sandbox.list having `returncode==0` for running sandboxes [CLI-406]

freider:
## Changelog

* Fixes a bug where objects returned by `Sandbox.list` had `returncode==0` for *running* sandboxes


thundergolfer:
Gonna merge this so we can get back to the user 👍

PR #3158, opened 2025-05-16
URL encode platform metadata

ehdr:
Otherwise it fails with `Invalid metadata value` when it's non-ASCII.

---

## Compatibility checklist

- [x] Client+Server: this change is compatible with old servers: 

</details>

### DIFF
--- a/test/client_test.py
+++ b/test/client_test.py
@@ -3,6 +3,7 @@ import platform
 import pytest
 import subprocess
 import sys
+import urllib.parse
 
 from google.protobuf.empty_pb2 import Empty
 from grpclib import GRPCError, Status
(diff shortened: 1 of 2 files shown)


test/client_test.py
```diff
@@ -26,7 +27,7 @@ def test_client_type(servicer, client):
 
 def test_client_platform_string(servicer, client):
     client.hello()
-    platform_str = servicer.last_metadata["x-modal-platform"]
+    platform_str = urllib.parse.unquote(servicer.last_metadata["x-modal-platform"])
```
mwaskom, line 30: Is it worth mocking some sort of weird unicode output from the `platform` module here?

mwaskom:
Does this need any server-side changes?

mwaskom (approved):
Not sure how critical the testing is; otherwise lgtm

test/client_test.py
ehdr: Yes good point, added [an explicit test](https://github.com/modal-labs/modal-client/pull/3158/files#diff-c3f9238c73a319c67ee49a92bcd534772c62b66a7a3c3b61da1ad630d178488bR37) now!

PR #3159, opened 2025-05-16
Client API Changes for HTTP/2 Tunnels Support

xpbowler:
[WRK-940](https://linear.app/modal-labs/issue/WRK-940/support-http2-in-relays), [ Design doc](https://www.notion.so/modal-com/H2-Tunnel-Support-1f81e7f16949806cb0c8d34fb299c7c2?pvs=4)

Enable tunnels to support H2 (HTTP/2 with TLS)

There will be 4 sequential PRs to roll out this feature. In order to being merged. This is the Client API changes PR.

1. `modal-client` [Protobuf changes](https://github.com/modal-labs/modal-client/pull/3177)
2. [Relay changes](https://github.com/modal-labs/modal/pull/22697)
3. [Worker changes](https://github.com/modal-labs/modal/pull/22874)
4. `modal-client` [Client API changes](https://github.com/modal-labs/modal-client/pull/3159)

## Changelog

- Added a new `h2_ports` parameter to `Sandbox.create` for exposing encrypted ports using HTTP/2. The following example will create an H2 port on 5002 and an HTTPS over HTTP/1.1 port on 5003.
```
sb = modal.Sandbox.create(app=app, h2_ports = [5002], encrypted_ports = [5003])
```


xpbowler:
> Should there be a changelog for this?

Created the changelog, let me know if I am missing anything else

modal/_tunnel.py
```diff
@@ -51,7 +51,9 @@ def tcp_socket(self) -> tuple[str, int]:
 
 
 @asynccontextmanager
-async def _forward(port: int, *, unencrypted: bool = False, client: Optional[_Client] = None) -> AsyncIterator[Tunnel]:
+async def _forward(
+    port: int, *, unencrypted: bool = False, tunnel_type: str = None, client: Optional[_Client] = None
```
mwaskom, line 55: I think I'm missing how `tunnel_type: str` is used in the function. It looks like the argument passed to the RPC is hardcoded as `tunnel_type=api_pb2.TUNNEL_TYPE_H2`.

xpbowler: oops, I forgot to pass it in

modal/_tunnel.py
```diff
@@ -51,7 +51,13 @@ def tcp_socket(self) -> tuple[str, int]:
 
 
 @asynccontextmanager
-async def _forward(port: int, *, unencrypted: bool = False, client: Optional[_Client] = None) -> AsyncIterator[Tunnel]:
+async def _forward(
+    port: int,
+    *,
+    unencrypted: bool = False,
+    tunnel_type: Optional["api_pb2.TunnelType.ValueType"] = None,
```
mwaskom, line 58: Hm is this meant to be an `api_pb2` (protobuf) type? We shouldn't be exposing those in public API, which I think this function is.

thundergolfer: I think this also makes the API clunky, because it allows for invalid states. `unecrypted=False, h2=True` is invalid but specifiable. An enum (str form or Enum form) would be cleanest imo, but that's backwards incompatible. 

Perhaps this is an ok compromise:

```
protocol: "http1" | "http2" = "http1"
unencrypted: bool = False,
```

(the new param can go above because this function uses `*` to force keywords, and I think above is nicer)

Perhaps we could defer this to a later PR.

cweld510: Yeah I think it makes sense to defer this for now, or to a separate PR

cweld510 (approved):

PR #3163, opened 2025-05-19
mount client dependencies at runtime, as of builder version 'PREVIEW'

matslina:
## Describe your changes

For image builder version > "2024.10", stop adding modal client dependencies directly to container images. Instead, pass a flag on Function to request the dependencies get mounted at runtime.

---

## Compatibility checklist

- [ x Client forward compatibility: this change ensures client can accept data intended for later versions of itself

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1442,7 +1442,9 @@ class _Image(_Object, type_prefix="im"):
             if version > "2024.10":
                 # for convenience when launching in a sandbox: sleep for 48h
                 commands.append(f'CMD ["sleep", "{48 * 3600}"]')
-            context_files = {CONTAINER_REQUIREMENTS_PATH: _get_modal_requirements_path(version, python_version)}
+            context_files = {}
+            if version <= "2024.10":
+                context_files = {CONTAINER_REQUIREMENTS_PATH: _get_modal_requirements_path(version, python_version)}
             return DockerfileSpec(commands=commands, context_files=context_files)
 
         return _Image._from_args(
@@ -1517,12 +1519,15 @@ class _Image(_Object, type_prefix="im"):
 
         # Note: this change is because we install dependencies with uv in 2024.10+
         requirements_prefix = "python -m " if builder_version < "2024.10" else ""
-        modal_requirements_commands = [
-            f"COPY {CONTAINER_REQUIREMENTS_PATH} {CONTAINER_REQUIREMENTS_PATH}",
-            f"RUN python -m pip install --upgrade {_base_image_config('package_tools', builder_version)}",
-            f"RUN {requirements_prefix}{_get_modal_requirements_command(builder_version)}",
-        ]
-        if builder_version > "2023.12":
+        modal_requirements_commands = []
+        if builder_version <= "2024.10":
+            # past 2024.10, client dependencies are mounted at runtime
+            modal_requirements_commands.extend([
+                f"COPY {CONTAINER_REQUIREMENTS_PATH} {CONTAINER_REQUIREMENTS_PATH}",
+                f"RUN python -m pip install --upgrade {_base_image_config('package_tools', builder_version)}",
+                f"RUN {requirements_prefix}{_get_modal_requirements_command(builder_version)}",
+            ])
+        if  "2024.10" >= builder_version > "2023.12":
             modal_requirements_commands.append(f"RUN rm {CONTAINER_REQUIREMENTS_PATH}")
 
         return [
@@ -1585,7 +1590,9 @@ class _Image(_Object, type_prefix="im"):
 
         def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             commands = _Image._registry_setup_commands(tag, version, setup_dockerfile_commands, add_python)
-            context_files = {CONTAINER_REQUIREMENTS_PATH: _get_modal_requirements_path(version, add_python)}
+            context_files = {}
+            if version <= "2024.10":
+                context_files = {CONTAINER_REQUIREMENTS_PATH: _get_modal_requirements_path(version, add_python)}
             return DockerfileSpec(commands=commands, context_files=context_files)
 
         return _Image._from_args(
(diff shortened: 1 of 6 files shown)


modal/image.py
```diff
@@ -1810,22 +1819,35 @@ def debian_slim(python_version: Optional[str] = None, force_build: bool = False)
...
-                f"COPY {CONTAINER_REQUIREMENTS_PATH} {CONTAINER_REQUIREMENTS_PATH}",
+            ]
+            if version <= "2024.10":
+                commands.extend([
+                    f"COPY {CONTAINER_REQUIREMENTS_PATH} {CONTAINER_REQUIREMENTS_PATH}",
+                ])
+            commands.extend([
                 "RUN apt-get update",
                 "RUN apt-get install -y gcc gfortran build-essential",
                 f"RUN pip install --upgrade {_base_image_config('package_tools', version)}",
```
mwaskom, line 1839: Just checking that it's intentional you're keeping this here but removing it from other constructors. I think it could make sense (since `debian_slim` is more overtly a "Modal-provided base Image") but allows for more inconsistent behavior between Image recipes that users might find surprising.

modal_global_objects/mounts/modal_client_dependencies.py
```diff
@@ -0,0 +1,87 @@
...
+
+        with tempfile.NamedTemporaryFile() as sitecustomize:
+            sitecustomize.write(
+                "import sys; sys.path.append('/__modal/.client_pkgs')\n".encode("utf-8"),
+            )
+            sitecustomize.flush()
+
+            python_mount = python_mount.add_local_file(
+                sitecustomize.name,
+                remote_path=REMOTE_SITECUSTOMIZE_PATH,
```
mwaskom, line 68: TIL that individual Mounts can be distributed across multiple remote paths!

mwaskom (approved):
Nice!

modal_global_objects/mounts/modal_client_dependencies.py
```diff
@@ -0,0 +1,87 @@
...
+from modal.config import config
+from modal.exception import NotFoundError
+from modal.image import ImageBuilderVersion
+from modal.mount import (
+    PYTHON_STANDALONE_VERSIONS,
+    Mount,
+)
+from modal_proto import api_pb2
+
+REMOTE_PACKAGES_PATH = "/__modal/.client_pkgs"
```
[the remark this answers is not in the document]
matslina: Saw other stuff already in `/__modal/.bin` etc, and picked something similar for consistency. Happy to change it.

modal_global_objects/mounts/modal_client_dependencies.py
matslina: Pretty neat

modal/image.py
mwaskom: That's a reasonable mindset. It will be easier to understand what the final answer should be on installing them in the base image once we have a better sense of how we're going to solve the general "make sure that `pip` and `uv` are available when you need them" solution for arbitrary Images.

modal_global_objects/mounts/modal_client_dependencies.py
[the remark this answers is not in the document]
mwaskom: I guess I'd expect to see them in roughly the same location as the client and first-party synchronicity dependency, although I guess since we prioritize `/pkg` on sys.path that won't work?

My expectation would be that a directory named `__modal/.something` would be totally private and not exposed to users but I do think this will be "user facing" in some sense, e.g. the path will show up in tracebacks. Obviously the `__modal` thing is not a "secure" way of hiding things from users but that will make it more exposed for better or for worse.

It shouldn't be disruptive to change later but for the fact that once a named Mount is created I'm not sure we have a way of overwriting it...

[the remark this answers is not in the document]
matslina: Yeah needs to be different from `/pkg`. Might as well switch to something we like before merging so we don't have to delete and recreate mounts in prod later on.

There's also the option of versioning this mount, like for the client. Might be a bit more future proof?

modal/image.py
matslina: Ok. We should probably at least have a plan for this before merging. I'll try to ping you on slack towards US hours.

modal_global_objects/mounts/modal_client_dependencies.py
```diff
@@ -0,0 +1,87 @@
...
+REMOTE_SITECUSTOMIZE_PATH = "/pkg/sitecustomize.py"
+
+def create_client_dependencies(
+    builder_version: ImageBuilderVersion,
+    python_version: str,
+    platform: str,
+    arch: str,
+):
+    profile_environment = config.get("environment")
```
thundergolfer, line 23: Do we ever deploy to a non-`main` Environment? Wondering if we could assert here

modal_global_objects/mounts/modal_client_dependencies.py
```diff
@@ -0,0 +1,87 @@
...
+                "install",
+                "-r",
+                requirements,
+                "--compile-bytecode",
+                "--target",
+                tmpd,
+                "--python-platform",
+                f"{arch}-{platform}",
+                "--python-version",
+                python_version,
```
thundergolfer, line 51: Does adding `--strict` work?

modal_global_objects/mounts/modal_client_dependencies.py
```diff
@@ -0,0 +1,87 @@
...
+            check=True,
+            capture_output=True,
+        )
+        print(f"🌐 Downloaded and unpacked packages to {tmpd}.")
+
+        python_mount = Mount._from_local_dir(tmpd, remote_path=REMOTE_PACKAGES_PATH)
+
+        with tempfile.NamedTemporaryFile() as sitecustomize:
+            sitecustomize.write(
+                "import sys; sys.path.append('/__modal/.client_pkgs')\n".encode("utf-8"),
```
thundergolfer, line 62: I'd write documentation into this file helping the end-user understand what's being done here and why.

modal_global_objects/mounts/modal_client_dependencies.py
```diff
@@ -0,0 +1,87 @@
...
+                mount_name,
+                api_pb2.DEPLOYMENT_NAMESPACE_GLOBAL,
+                environment_name=profile_environment,
+            )
+            print(f"✅ Deployed mount {mount_name} to global namespace.")
+
+
+def main(client=None):
+    for python_version in PYTHON_STANDALONE_VERSIONS:
+        for platform in ["manylinux_2_17"]:
```
thundergolfer, line 82: wondering if there's an implicit TODO here about subsequent glibc versions, e.g. `manylinux_2_28`? Do we add new ones as we discover that customers need them?

modal_global_objects/mounts/modal_client_dependencies.py
matslina: If so we should probs to so for the other publish scripts as well.

modal_global_objects/mounts/modal_client_dependencies.py
matslina: Seems to work!

PR #3165, opened 2025-05-19
[input-plane] Client should retry inputs that are unfinished after task ends

rculbertson:
Part of SVC-595

If a task ends, and there are unfinished inputs, the client should retry them. This has always be done by the python server, but in the new input plane, the client will handle the retries. This is because the input plane does not support server side retries at all. 

For now we will hard code the number of retries to 8 - which is what the python server has always done - but we may want to make this value configurable server side in the future, perhaps before releasing this to users.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -415,6 +415,7 @@ class _InputPlaneInvocation:
                     input=self.input_item,
                     attempt_token=self.attempt_token,
                 )
+                # TODO(ryan): Add exponential backoff?
                 retry_response = await retry_transient_errors(self.stub.AttemptRetry, retry_request)
                 self.attempt_token = retry_response.attempt_token
 
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -399,9 +399,10 @@ class _InputPlaneInvocation:
             )
 
             try:
-                return await _process_result(
-                    await_response.output.result, await_response.output.data_format, self.stub, self.client
-                )
+                if await_response.HasField("output"):
+                    return await _process_result(
+                        await_response.output.result, await_response.output.data_format, self.stub, self.client
+                    )
             except InternalFailure as e:
                 internal_failure_count += 1
                 # Limit the number of times we retry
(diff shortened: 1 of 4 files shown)


rculbertson:
@prbot approve

modal/_functions.py
```diff
@@ -365,36 +370,54 @@ async def create(
...
             )
-            response: api_pb2.AttemptAwaitResponse = await retry_transient_errors(
+            await_response: api_pb2.AttemptAwaitResponse = await retry_transient_errors(
                 self.stub.AttemptAwait,
-                request,
+                await_request,
                 attempt_timeout=OUTPUTS_TIMEOUT + ATTEMPT_TIMEOUT_GRACE_PERIOD,
             )
 
-            if response.HasField("output"):
```
thecodingwizard, line 394: we still want to check if the output field is set right?

modal/_utils/function_utils.py
```diff
@@ -466,7 +466,10 @@ async def _process_result(result: api_pb2.GenericResult, data_format: int, stub,
 
     if result.status == api_pb2.GenericResult.GENERIC_STATUS_TIMEOUT:
         raise FunctionTimeoutError(result.exception)
-    elif result.status == api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE:
+    elif result.status in [
+        api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE,
+        api_pb2.GenericResult.GENERIC_STATUS_TERMINATED,
```
thecodingwizard, line 471: We also use terminated status code when the app stops. In that case, we don't want the client to retry. Should worker preemptions return INTERNAL_FAILURE instead or something?

thecodingwizard: Alternatively we can stop returning TERMINATED when the function is stopped?

rculbertson: Yikes, my bad for having this retry on TERMINATED! As just discussed IRL, that was not my intent - will fix, thanks for catching!

modal/_functions.py
rculbertson: Ah right, I see that's it's optional. Ok, yea will add back in.

PR #3168, opened 2025-05-20
Dont use internal synchronicity UserCodeException type

freider:
This turned out to be a bit of a rabbit hole.

* The client has abused the fact that synchronicity unwraps synchronicity.UserCodeExceptions and wrapped any exceptions returned by containers
* I believe the original intent of this was to piggy back off synchronicity's use of this to exclude certain internal frames in tracebacks
* More recently, we've also started using it for client-side retries, since it has been a way for us to disambiguate internal errors from errors in user remote containers (things that should be retried)
* Synchronicity only unwraps these exceptions when they are *raised*, so an unfortunate side effect has been that `map(return_exceptions=True)` has returned actual synchronicity.UserCodeException objects instead of the actual exception (!)

This PR removes the usage of this internal synchronicity exception type in favor of an explicit modal-specific exception wrapper type, `_ContainerException` (open to other names). This means we also need to explicitly unwrap these whenever exceptions can be returned to users.

Unfortunately, this is is a breaking change for `.map(return_exceptions=True)` in that it will no longer wrap the user exceptions... But I believe that has never actually been the intent of the functionality, and it's quite ugly that it has been returning a *synchronicity* exception type (!).

Closes CLI-419

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -15,8 +15,8 @@ import typing_extensions
 from google.protobuf.message import Message
 from grpclib import GRPCError, Status
 from synchronicity.combined_types import MethodWithAio
-from synchronicity.exceptions import UserCodeException
 
+from modal.exception import _ContainerException
 from modal_proto import api_pb2
 from modal_proto.modal_api_grpc import ModalClientModal
 
@@ -100,6 +100,7 @@ if TYPE_CHECKING:
     import modal.partial_function
 
 MAX_INTERNAL_FAILURE_COUNT = 8
+
 
 @dataclasses.dataclass
 class _RetryContext:
@@ -258,6 +259,7 @@ class _Invocation:
 
     async def _get_single_output(self, expected_jwt: Optional[str] = None) -> Any:
         # waits indefinitely for a single result for the function, and clear the outputs buffer after
+        # raises wrapped internal _ContainerExceptions if they occur
         item: api_pb2.FunctionGetOutputsItem = (
             await self.pop_function_call_outputs(
                 timeout=None,
@@ -285,15 +287,16 @@ class _Invocation:
         while True:
             try:
                 return await self._get_single_output(ctx.input_jwt)
-            except (UserCodeException, FunctionTimeoutError) as exc:
+            except (_ContainerException, FunctionTimeoutError):
                 delay_ms = user_retry_manager.get_delay_ms()
                 if delay_ms is None:
-                    raise exc
+                    raise
                 await asyncio.sleep(delay_ms / 1000)
             except InternalFailure:
                 # For system failures on the server, we retry immediately,
                 # and the failure does not count towards the retry policy.
                 pass
+
             await self._retry_input()
 
     async def poll_function(self, timeout: Optional[float] = None):
@@ -311,9 +314,12 @@ class _Invocation:
         elif len(response.outputs) == 0:
             raise TimeoutError()
 
-        return await _process_result(
-            response.outputs[0].result, response.outputs[0].data_format, self.stub, self.client
-        )
+        try:
+            return await _process_result(
+                response.outputs[0].result, response.outputs[0].data_format, self.stub, self.client
+            )
+        except _ContainerException as container_exception:
+            raise container_exception.unwrap()
 
     async def run_generator(self):
         items_received = 0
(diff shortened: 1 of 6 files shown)


modal/_utils/function_utils.py
```diff
@@ -459,6 +459,15 @@ def exc_with_hints(exc: BaseException):
 
 
 async def _process_result(result: api_pb2.GenericResult, data_format: int, stub, client=None):
+    try:
+        return await _process_result_wrap_exceptions(result, data_format, stub, client)
+    except _ContainerException as exc:
+        raise exc.unwrap()
+
+
+async def _process_result_wrap_exceptions(result: api_pb2.GenericResult, data_format: int, stub, client=None):
```
freider, line 468: Another option is to use a boolean flag to wrap exceptions, but that feels a bit more smelly to me

test/function_test.py
```diff
@@ -357,6 +356,35 @@ async def test_function_future_async(client, servicer):
         assert future.object_id not in servicer.cleared_function_calls  # keep results around a bit longer for futures
 
 
+def test_function_spawn_exception(client, servicer):
```
freider, line 359: Kind of scary that we didn't have any tests for spawn or generator calls with exceptions...

test/function_test.py
```diff
@@ -533,7 +561,7 @@ def test_map_exceptions(client, servicer):
 
         res = list(custom_function_modal.map(range(6), return_exceptions=True))
         assert res[:4] == [0, 1, 4, 9] and res[5] == 25
-        assert type(res[4]) is UserCodeException and "bad" in str(res[4])
+        assert type(res[4]) is CustomException and "bad" in str(res[4])
```
freider, line 564: Breaking change!

modal/_functions.py
```diff
@@ -256,45 +258,50 @@ async def _retry_input(self) -> None:
...
-        ctx = self._retry_context
-        if (
-            not ctx
-            or not ctx.retry_policy
-            or ctx.retry_policy.retries == 0
-            or ctx.function_call_invocation_type != api_pb2.FUNCTION_CALL_INVOCATION_TYPE_SYNC
-            or not ctx.sync_client_retries_enabled
-        ):
-            return await self._get_single_output()
+        try:
```
freider, line 274: This is the only place where we actually "need" the exception wrapping (in order to detect retryable errors), so it's explicitly using `_get_single_output_wrap_exceptions` and then unwrapping it in an outer try/except

freider:
Related thing I noticed - `run_generator` currently calls `run_function`, but `run_function` is unnecessarily complex with retry logic that is never used for generators so maybe we could have it just call `_get_single_output` instead? wdyt @rculbertson ?

rculbertson (approved):

modal/_functions.py
```diff
@@ -256,45 +258,50 @@ async def _retry_input(self) -> None:
...
+                        raise
+                    await asyncio.sleep(delay_ms / 1000)
+                except InternalFailure:
+                    # For system failures on the server, we retry immediately,
+                    # and the failure does not count towards the retry policy.
+                    pass
+
+                await self._retry_input()
+        except _ContainerException as exc:
+            raise exc.unwrap()
```
thomasjpfan, line 304: Nit: In an effort to try to reduce the amount of indenting can the `try`, `except` be a decorator? Something like this:

```python
def _unwrap_container_exception(func):
    @functools.wraps(func)
    async def wrapper(*args, **kwargs):
        try:
        except _ContainerException as exc:
            raise exc.unwrap()
    return wrapper

@_unwrap_container_exception
async def run_function(self) -> Any:
    ...
```

modal/exception.py
```diff
@@ -22,6 +22,19 @@ class Error(Exception):
     """
 
 
+class _ContainerException(Exception):
```
thomasjpfan, line 25: I prefer this name a little more:

```suggestion
class _WrappedException(Exception):
```

At a glance, I thought "Container" meant "in the runtime container".

test/function_test.py
thomasjpfan: Although this is breaking, I'll say this is more of a bug fix.

freider:
Superseded by https://github.com/modal-labs/modal-client/pull/3226

PR #3169, opened 2025-05-20
Adds event loop monitor for containers

freider:
* Crude detection if main thread (user code) event loop is delayed by blocking calls - simply a background loop that checks if it gets a callback at the expected cadence
* Configurable by three settings:
    * event_loop_monitor: bool   # enable/disable [enabled by default]
    * event_loop_monitor_period_ms: int  # period of the checker loop
    * event_loop_monitor_warning_threshold_ms: # if the loop is *at least* this delayed for every iteration of the loop, a warning is emitted

This is implemented as an async context manager that is set up and torn down for every async input that is run, as opposed to something that is "always on" in the container. This is in order to support both async and sync functions running in the container entrypoint, in particular classes that could run a mix of both.

TODO:
* Figure out a better way to tweak/disable the warning. Using modal's config/env vars is quite clunky for container-level settings since the option doesn't transfer unless you add a secret that forwards it explicitly to containers.
* This only *detects* the issue - what should we tell users to do in order to pinpoint/fix it?
* What are good default values for the check period and warning threshold? Currently set at 200ms each

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

* Modal containers for Functions and Classes will now log warnings if an async function blocks the event loop for more than 200 ms by running non-async code, helping users pinpoint issues that could lead to inefficient concurrency when using `@modal.concurrent()` with async function definitions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,11 @@ This changelog documents user-facing updates (features, enhancements, fixes, and
 
 <!-- NEW CONTENT GENERATED BELOW. PLEASE PRESERVE THIS COMMENT. -->
 
-#### 1.0.1.dev0 (2025-05-16)
+### 1.0.1 (2025-05-19)
 
-* Fixes a bug where objects returned by `Sandbox.list` had `returncode == 0` for *running* sandboxes. Now those sandboxes will have a null returncode.
-
+- Added a `--timestamps` flag to `modal app logs` that prepends a timestamp to each log line.
+- Fixed a bug where objects returned by `Sandbox.list` had `returncode == 0` for *running* Sandboxes. Now the return code for running Sandboxes will be `None`.
+- Fixed a bug affecting systems where the `sys.platform.node` name includes unicode characters.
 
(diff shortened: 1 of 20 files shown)


mwaskom:
> helping users pinpoint issues that could lead to inefficient concurrency when using @modal.concurrent()

Is this relevant only when input concurrency is being used?

modal/_runtime/event_loop_monitor.py
```diff
@@ -0,0 +1,58 @@
...
+        yield
+        return
+
+    def callback(delay: float) -> None:
+        # TODO: add hints for user to find the problem?
+        logger.warning(
+            f"Detected an asyncio event loop delay of {delay:.2f}s.\n"
+            f"This indicates something non-async is running in an async method for a "
+            f"which could prevent functioning concurrency.\n"
+            f"If this is intentional, you can deactivate this warning with MODAL_EVENT_LOOP_MONITOR=0"
```
mwaskom, line 52: How are we propagating this local config to the container environment?

PR #3170, opened 2025-05-20
volume: Retry put_missing_block a few times

dflemstr:
## Describe your changes

- Retry block uploads a few times when they fail for HTTP reasons.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_utils/bytes_io_segment_payload.py
+++ b/modal/_utils/bytes_io_segment_payload.py
@@ -52,12 +52,16 @@ class BytesIOSegmentPayload(BytesIOPayload):
         self._value.seek(self.initial_seek_pos)
 
     @contextmanager
-    def reset_on_error(self):
+    def reset_on_error(self, subtract_progress: bool = False):
         try:
             yield
         except Exception as exc:
             try:
-                self.progress_report_cb(reset=True)
+                if subtract_progress:
+                    negative_progress = -self.num_bytes_read
+                    self.progress_report_cb(advance=negative_progress)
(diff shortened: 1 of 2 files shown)


danielnorberg:
"In the cloud, all errors are ephemeral" - Old Jungle Saying

modal/_utils/bytes_io_segment_payload.py
```diff
@@ -52,12 +52,16 @@ def reset_state(self):
         self._value.seek(self.initial_seek_pos)
 
     @contextmanager
-    def reset_on_error(self):
+    def reset_on_error(self, subtract_progress: bool = False):
```
danielnorberg, line 55: Do we need intra-block progress?

dflemstr: In my opinion, it would be good to keep using the existing `BytesIOSegmentPayload` instead of trying to create a new thing.

danielnorberg (approved):

modal/volume.py
```diff
@@ -825,7 +826,7 @@ def __init__(
         progress_cb: Optional[Callable[..., Any]] = None,
         force: bool = False,
         hash_concurrency: int = multiprocessing.cpu_count(),
-        put_concurrency: int = multiprocessing.cpu_count(),
+        put_concurrency: int = 128,
```
danielnorberg, line 829: Not sure if this is intentional but I'm not opposed 😄

PR #3171, opened 2025-05-20
fix: simulate_preemption should not accept zero

thundergolfer:
## Describe your changes

Reported by a user after they tried `0` and found that doing that breaks https://docs.python.org/3/library/signal.html#signal.alarm

---

## Compatibility checklist

---

## Release checklist

- [x] Version file (`modal_version/__init__.py`) has been updated with the next logical version
- [x] Changelog has been cleaned up and given an appropriate subhead

---

</details>


mwaskom (approved):
Seems like a fine change but I still think this should be a snippet in some doc not part of our public API.

thundergolfer:
I'm fine to have it removed 🤷

PR #3172, opened 2025-05-20
Report non-function Image build failures better

mwaskom:
## Describe your changes

When Image builds fail, we try to print the `result.exception`. Unless the failing step is a `.run_function`, this will be empty, so we print an unhelpful message like

```
Image build for im-DEHjPnPEudJASdtbmWLtgW failed with the exception:\n'
```

This PR handles the empty exception case better, directing the user to the build logs.

In the case where output is not enabled, we add a hint suggesting `modal.enable_output`. Otherwise there won't be any obvious "build logs" for the user to refer to.

This is still a little unhelpful (i.e., in the case where the build failed without output enabled, you need to run the build again, and that might take a while). I think we can potentially capture more failure information from the worker and propagate it back to the client. But it will take some more changes.

- Closes CLI-412

---

## Compatibility checklist

---

</details>

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -647,7 +647,13 @@ class _Image(_Object, type_prefix="im"):
                     metadata = resp.metadata
 
             if result.status == api_pb2.GenericResult.GENERIC_STATUS_FAILURE:
-                raise RemoteError(f"Image build for {image_id} failed with the exception:\n{result.exception}")
+                if result.exception:
+                    raise RemoteError(f"Image build for {image_id} failed with the exception:\n{result.exception}")
+                else:
+                    msg = f"Image build for {image_id} failed. See build logs for more details."
+                    if not _get_output_manager():
+                        msg += " Use `modal.enable_output()` to print logs from the process building the Image."
+                    raise RemoteError(msg)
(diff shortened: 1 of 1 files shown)


modal/image.py
```diff
@@ -1522,12 +1528,14 @@ def _registry_setup_commands(
         modal_requirements_commands = []
         if builder_version <= "2024.10":
             # past 2024.10, client dependencies are mounted at runtime
-            modal_requirements_commands.extend([
-                f"COPY {CONTAINER_REQUIREMENTS_PATH} {CONTAINER_REQUIREMENTS_PATH}",
-                f"RUN python -m pip install --upgrade {_base_image_config('package_tools', builder_version)}",
-                f"RUN {requirements_prefix}{_get_modal_requirements_command(builder_version)}",
-            ])
-        if  "2024.10" >= builder_version > "2023.12":
+            modal_requirements_commands.extend(
```
mwaskom, line 1531: Looks like this pulled in some unrelated ruff changes...

mwaskom:
@ehdr suggested on Slack that the error message could include a link to app logs which is a nice idea — but I don't think we have immediate access to that URL without adding more global state to the Resolver, so will have to punt on it for now.

PR #3173, opened 2025-05-21
Add support for specifying timezone on `Cron` schedules

ehdr:
## Describe your changes

---

## Compatibility checklist

---

</details>

## Changelog

- Added support for specifying a timezone on `Cron` schedules, helpful if you e.g. want a `Function` to run at 6am local time, regardless of whether daylight saving is in effect or not, for example:
  ```
  import modal
  app = modal.App()
  
  
  @app.function(schedule=modal.Cron("* 6 * * *"), timezone="America/New_York")
  def f():
      print("This function will run every day at 6am New York time.")
  ```

### DIFF
--- a/modal/schedule.py
+++ b/modal/schedule.py
@@ -30,15 +30,28 @@ class Cron(Schedule):
     We can specify different schedules with cron strings, for example:
 
     ```python
-    modal.Cron("5 4 * * *")  # run at 4:05am every night
-    modal.Cron("0 9 * * 4")  # runs every Thursday 9am
+    modal.Cron("5 4 * * *")  # run at 4:05am UTC every night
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -2586,6 +2586,7 @@ message SandboxWaitResponse {
 message Schedule {
   message Cron {
     string cron_string = 1;
+    string timezone = 2;
```
freider, line 2589: We usually try to split out the proto changes into a separate PR first before adding the user facing SDK changes that use the new protos

freider (approved):
looks good. but lets release the protos first, then make sure the backend works with those, and then release the SDK changes so we know they work immediately

ehdr:
Got it, thanks! I split the proto change out into #3176 now, and rebased this (so holding off on merging this).

PR #3174, opened 2025-05-21
_map_helper was using synchronizer wrapped function within event loop code

freider:
This technically locked the event loop, although for a very short time since there isn't any actual IO in SynchronizedQueue.init()

It only becomes a real problem when you try to use a ".map style" call from within the synchronizer event loop which is not something that should usually happen

---

## Compatibility checklist

---

## Release checklist

---

</details>


mwaskom (approved):
😵‍💫

PR #3178, opened 2025-05-22
Mention in Sandbox.create() docstring that the container is created asynchronously

cweld510:
We never say this explicitly in the docs -- seems worth adding.


cweld510:
@prbot approve @freider

mwaskom:
Thanks for adding (but, nit, per PEP8 guidelines the docstring summary should generally fit on a single line, with additional context or information added in paragraphs below).

PR #3179, opened 2025-05-23
emit warning if client dependency mount is not expected to exist

matslina:
## Describe your changes

Prints some warnings if client dependency mounts are needed, but none are expected to be found for the chose image. This is for image builder version > `2024.10` only.


mwaskom:
Superseded by https://github.com/modal-labs/modal/pull/24305 I think

PR #3180, opened 2025-05-23
perf: make Sandbox.terminate faster

thundergolfer:
## Describe your changes

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- `Sandbox.terminate` not longer waits for container shutdown completion, but still ensures that a terminated container will shutdown imminently.


cweld510 (approved):
👍
i suspect it's fine to break backwards-compatibility here; it would only affect very specific workloads

PR #3181, opened 2025-05-24
Fix comment on workdir (very confusing)

ekzhang:
Extremely confusing field called `workdir` on `ImageMetadata` which is not actually used by any of our systems, `ImageManifest.workdir` is the actual workdir. This is wrong in 99% of images.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1945,13 +1945,21 @@ message ImageJoinStreamingResponse {
 message ImageMetadata {
   // The output of `python -VV. Not set if missing
   optional string python_version_info = 1;
+
   // Installed python packages, as listed by by `pip list`.
   // package name -> version. Empty if missing
   map<string, string> python_packages = 2;
-  // The work directory of the image, defaulting to "/". Not set if missing
+
+  // The working directory of the image, defaulting to "/". Not set if missing.
+  //
+  // Note that this is NOT the working directory for most runner tasks if
+  // `image.manifest.workdir` is not set, which defaults to `/root`, but this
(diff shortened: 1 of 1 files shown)


mwaskom:
Not sure I know the initial intention here or exactly the nature of the confusion, but I think this is meant to convey what the workdir state would be for subsequent builder tasks, (i.e. so that we can know where `.add_local_*` layers would put their files with a relative remote location. @freider might know more.

ekzhang:
> I'm not sure I'm completely following, but is it right that what's confusing here is that we have a different default workdir for builder tasks vs runner tasks?

That's not quite it, the confusing but is that this field is just useless and not used anywhere. The correct field is `manifest.workdir`. I can update the comment to make that clearer

ekzhang:
See https://github.com/modal-labs/modal/pull/22955 for comment in other location of the code

freider:
The context here is that _for some planned future work_ the client needs to receive the image workdir in the response data from images. We need to fill this metadata field with whatever the best information available - be that from the manifest or otherwise.

Also, why is it wrong for 99% of images? (disregarding that we override it for runner tasks)

ekzhang:
Thanks for the insight @freider, I can update the comment / document better what it means when `manifest.workdir` is missing. The missing value is actually semantic versus a `/` value.

I think for now any workdir should return a `PurePosixPath | None` or `str | None` value determining whether it is set, since `None` is meaningful (and it's the case for 99% of images mentioned above) — overriding as `/root` only applies when None. Does that make sense? Just from a runtime perspective.

PR #3184, opened 2025-05-26
Fix for aiohttp 3.12.1 enforcing BytesIO type

freider:
Aiohttp 3.12.1 relies on `getbuffer()` to exist on the file like object passed to BytesIOPayload, whereas it previously worked with more general `BinaryIO` objects too (like `open("file", "rb")` handles

This "fixes" the issue by making our own Payload implementation a subclass of the Payload ABC rather than BytesIOPayload.

It may be that we could refactor our code to use aiohttp 3.12's reusable payload objects instead of having our own custom payload handler, but that's a larger project.

---

## Compatibility checklist

---

## Release checklist

---

</details>


freider:
@prbot approve

freider:
@prbot automerge

PR #3185, opened 2025-05-27
Add `--from-dotenv` and `--from-json` to `modal secret create`

ehdr:
Similarly to [`from_dotenv` in the `modal.Secret` API](https://modal.com/docs/reference/modal.Secret#from_dotenv).

## Describe your changes

Resolves CLI-381 through documenting the `modal secret create my-credentials username=john password="$PASSWORD"` pattern, and adding `--from-dotenv` and `--from-json`.

---

</details>

## Changelog

- Added `--from-dotenv` and `--from-json` flags to `modal secret create`, which allows creating secrets by reading the secret value from local files. This is useful for applications (like CI) where your code is checked in to version control.

### DIFF
--- a/modal/cli/secret.py
+++ b/modal/cli/secret.py
@@ -1,7 +1,9 @@
 # Copyright Modal Labs 2022
+import json
 import os
 import platform
 import subprocess
+from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Optional
 
@@ -48,14 +50,17 @@ async def list_(env: Optional[str] = ENV_OPTION, json: bool = False):
 @secret_cli.command("create", help="Create a new secret.")
 @synchronizer.create_blocking
 async def create(
-    secret_name,
-    keyvalues: list[str] = typer.Argument(..., help="Space-separated KEY=VALUE items"),
+    secret_name: str,
+    keyvalues: Optional[list[str]] = typer.Argument(default=None, help="Space-separated KEY=VALUE items."),
     env: Optional[str] = ENV_OPTION,
+    from_dotenv: Optional[Path] = typer.Option(default=None, help="Path to a .env file to load secrets from."),
+    from_json: Optional[Path] = typer.Option(default=None, help="Path to a JSON file to load secrets from."),
     force: bool = typer.Option(False, "--force", help="Overwrite the secret if it already exists."),
 ):
     env = ensure_env(env)
     env_dict = {}
-    for arg in keyvalues:
+
+    for arg in keyvalues or []:
         if "=" in arg:
             key, value = arg.split("=", 1)
             if value == "-":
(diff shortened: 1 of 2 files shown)


modal/cli/secret.py
```diff
@@ -48,32 +50,69 @@ async def list_(env: Optional[str] = ENV_OPTION, json: bool = False):
 @secret_cli.command("create", help="Create a new secret.")
 @synchronizer.create_blocking
 async def create(
-    secret_name,
-    keyvalues: list[str] = typer.Argument(..., help="Space-separated KEY=VALUE items"),
+    secret_name: str,
+    keyvalues: Optional[list[str]] = typer.Argument(default=None, help="Space-separated KEY=VALUE items."),
     env: Optional[str] = ENV_OPTION,
+    from_dotenv: Optional[Path] = typer.Option(default=None, help="Path to a .env file to load secrets from."),
```
mwaskom, line 56: Should we try to mirror the `Secret.from_dotenv` method and allow this to be used just as a flag `(modal secret create whatevrer --from-dotenv` with auto-discovery of the file? It's not a huge problem if we don't do that, but I think it's how I'd expect it to work.

modal/cli/secret.py
```diff
@@ -48,32 +50,69 @@ async def list_(env: Optional[str] = ENV_OPTION, json: bool = False):
...
+                env_dict.update(json.load(f))
+        except Exception as e:
+            raise click.UsageError(f"Could not parse JSON file at {from_json}: {e}")
+
     if not env_dict:
         raise click.UsageError("You need to specify at least one key for your secret")
 
+    for k, v in env_dict.items():
+        if not isinstance(k, str) or not k:
+            raise click.UsageError(f"Invalid key for secret '{k}'")
```
mwaskom, line 112: Nit, but I think I'd otherwise misread this as reporting the secret's _name_.

```suggestion
            raise click.UsageError(f"Invalid key: '{k}'")
```

mwaskom (commented):
Just minor comments! 

Curious if someone asked for `--from-json` too?

modal/cli/secret.py
```diff
@@ -48,32 +50,69 @@ async def list_(env: Optional[str] = ENV_OPTION, json: bool = False):
...
+        if not isinstance(v, str):
+            raise click.UsageError(f"Non-string value for secret '{k}'")
```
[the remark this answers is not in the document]
ehdr: I personally tend to prefer being somewhat strict on what inputs you accept, at least in a case like this with JSON where people in general are used to thinking about types (as opposed to e.g. .csv)?

modal/cli/secret.py
```diff
@@ -48,32 +50,69 @@ async def list_(env: Optional[str] = ENV_OPTION, json: bool = False):
...
+modal secret create my-credentials username=john password="$PASSWORD"
 """
             )
 
+    if from_dotenv:
+        if not from_dotenv.is_file():
+            raise click.UsageError(f"Could not read .env file at {from_dotenv}")
+
+        try:
+            from dotenv import dotenv_values
```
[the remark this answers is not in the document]
mwaskom: Cool, we should probably think about moving this support internally, but agree it's not a new dependency for now.

modal/cli/secret.py
ehdr: Discussed on Slack, and this turned out to be harder than expected, so pushing it to a potential follow-up.

mwaskom:
> Added --from-dotenv and --from-json flags to modal secret create, which allows creating secrets by reading the secret value from local files. **This is useful for applications (like CI) where your code is checked in to version control.**

Wondering about the implication of this changelog comment; wouldn't it also require that the `.env` file is checked in too?

erikbern:
nice!

ehdr:
> > Added --from-dotenv and --from-json flags to modal secret create, which allows creating secrets by reading the secret value from local files. **This is useful for applications (like CI) where your code is checked in to version control.**
> 
> Wondering about the implication of this changelog comment; wouldn't it also require that the `.env` file is checked in too?

Yeah good catch, I brought it over from the last PR but I think you're right, so removed now!

PR #3188, opened 2025-05-28
Improves error messages for local file & dir mounts

thomasjpfan:
## Describe your changes

Improves error message for `Image.add_local_file` and `Image.add_local_dir` when files or directories does not exist: https://linear.app/modal-labs/issue/CLI-367/suppressed-traceback-for-missing-add-local-file-makes-error-difficult
 
---

## Compatibility checklist

Client only changes

---

## Release checklist

---

</details>

## Changelog

Improves error message for `Image.add_local_file` and `Image.add_local_dir` when files or directories does not exist.

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -115,7 +115,8 @@ class _MountFile(_MountEntry):
     def get_files_to_upload(self):
         local_file = self.local_file.resolve()
         if not local_file.exists():
-            raise FileNotFoundError(local_file)
+            msg = f"Local file: {os.fspath(local_file)} does not exists"
+            raise FileNotFoundError(msg)
 
         rel_filename = self.remote_path
         yield local_file, rel_filename
@@ -144,10 +145,12 @@ class _MountDir(_MountEntry):
         local_dir = self.local_dir.expanduser().absolute()
 
         if not local_dir.exists():
-            raise FileNotFoundError(local_dir)
+            msg = f"Local dir: {os.fspath(local_dir)} does not exist"
+            raise FileNotFoundError(msg)
 
         if not local_dir.is_dir():
-            raise NotADirectoryError(local_dir)
+            msg = f"Local dir: {os.fspath(local_dir)} is not a directory"
+            raise NotADirectoryError(msg)
 
         if self.recursive:
             gen = (os.path.join(root, name) for root, dirs, files in os.walk(local_dir) for name in files)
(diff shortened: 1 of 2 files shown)


modal/mount.py
```diff
@@ -115,7 +115,8 @@ def description(self) -> str:
     def get_files_to_upload(self):
         local_file = self.local_file.resolve()
         if not local_file.exists():
-            raise FileNotFoundError(local_file)
+            msg = f"Local file: {os.fspath(local_file)} does not exists"
```
mwaskom, line 118: Nit
```suggestion
            msg = f"Local file {os.fspath(local_file)} does not exist"
```

mwaskom (commented):
Nice, this looks good, although I think there's also some benefit to checking existence more eagerly in the `Image` layers so that users get better tracebacks. WDYT?

mwaskom:
> Nice, this looks good, although I think there's also some benefit to checking existence more eagerly in the Image layers so that users get better tracebacks. WDYT?

Actually that may be bad advice! Thinking about it a little more, I think we'll throw that exception in containers (unless you put the file in the same place). We might have to live with the lack of a clear traceback.

modal/mount.py
```diff
@@ -144,10 +145,12 @@ def get_files_to_upload(self):
         local_dir = self.local_dir.expanduser().absolute()
 
         if not local_dir.exists():
-            raise FileNotFoundError(local_dir)
+            msg = f"Local dir: {os.fspath(local_dir)} does not exist"
```
mwaskom, line 148: Super nit but I think the colon is grammatically wrong here

test/mount_test.py
```diff
@@ -71,15 +72,26 @@ def condition(fn):
     assert repr(Path(local_dir)) in repr(m)
 
 
-def test_create_mount_file_errors(servicer, tmpdir, client):
-    m = Mount._from_local_dir(Path(tmpdir) / "xyz", remote_path="/xyz")
-    with pytest.raises(FileNotFoundError):
+def test_create_mount_file_errors(servicer, tmp_path, client):
+    invalid_dir = tmp_path / "xyz"
+    m = Mount._from_local_dir(invalid_dir, remote_path="/xyz")
```
mwaskom, line 77: Could we write the tests to assert about the public API (`Image.add_local_file` etc.) rather than the internal `Mount` object? Now that the latter is private it's possible we'll change the implementation for the public `Image` API and it's more important to be sure that we get the user-facing errors we expect from the public methods.

PR #3190, opened 2025-05-29
Raises an error when using interactive without enable_output

thomasjpfan:
## Describe your changes

This PR converts the warning into a more informative error when `enable_output` is not set.

Fixes https://linear.app/modal-labs/issue/CLI-315/warn-if-using-interactive=true-without-modalenable-output
 
---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

Raises error earlier when using interactive mode without using the `modal.enable_output` context manager.


modal/runner.py
```diff
@@ -296,12 +295,8 @@ async def _run_app(
 
     output_mgr = _get_output_manager()
     if interactive and output_mgr is None:
-        warnings.warn(
-            "Interactive mode is disabled because no output manager is active. "
-            "Use 'with modal.enable_output():' to enable interactive mode and see logs.",
-            stacklevel=2,
-        )
-        interactive = False
+        msg = "Interactive mode requires the context manager: `with modal.enable_output():`"
```
mwaskom, line 298: Mentioned on Slack but I think this error message (and the previous waring) could mislead people who are using the CLI (`modal run --interactive --quiet`). Unfortunately I don't think we can tell (in any obvious way) which way Modal is being invoked here.

mwaskom (commented):
Do you think there's any risk related to upgrading the warning into an error for people who are using interactive mode but _not_ coming across a `modal.interact()` in their code? 

I'm also still a little confused about whether this would have avoided the original confusion, since it seemed like they weren't seeing any warning at all.

modal/runner.py
```diff
@@ -296,12 +295,8 @@ async def _run_app(
 
     output_mgr = _get_output_manager()
     if interactive and output_mgr is None:
-        warnings.warn(
-            "Interactive mode is disabled because no output manager is active. "
-            "Use 'with modal.enable_output():' to enable interactive mode and see logs.",
-            stacklevel=2,
-        )
-        interactive = False
+        msg = "Use the `modal.enable_output()` context manager to enable interactive mode."
```
mwaskom: Phrasing is a little confusing here; setting `interactive=True` is what directly "enables" interactive mode but there's an additional condition that output must be enabled. Might articulate that more directly, like "Interactive mode requires output to be enabled (use the `modal.enable_output()` context manager)" or something.

mwaskom (approved):
I think it makes sense to turn the warning into an error here, although something's not sitting quite right with me about it. It feels like we're asking users to jump through a lot of hoops. There'd previously been some discussion about whether we could get rid of `interactive=` as an App-level switch (and just always have a pty attached) which is maybe a good question to revisit.

See one comment about the phrasing of the error message.

PR #3191, opened 2025-05-29
Add hyperlinks in .spawn doc

thundergolfer:
## Describe your changes

As a user it's annoying that these aren't links. 

image

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -101,6 +101,7 @@ if TYPE_CHECKING:
 
 MAX_INTERNAL_FAILURE_COUNT = 8
 
+
 @dataclasses.dataclass
 class _RetryContext:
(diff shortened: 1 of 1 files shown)


thundergolfer:
@prbot approve

mwaskom:
I don't disagree, but having to put the urls directly into the markdown compromises the plain text readability and is quite brittle (this will break the docs build if you get it wrong but i don't think we have any way to catch that in client CI) so i'm very hopeful we can get something working to do these cross-links automatically at some point.

thundergolfer:
> (this will break the docs build if you get it wrong but i don't think we have any way to catch that in client CI)

Yeh I manually verify with `inv docs` and inv just-frontend`.

>  but having to put the urls directly into the markdown compromises the plain text readability

Yeh it's unfortunate but when linking users to the docs I find the lack of linking user hostile.

PR #3192, opened 2025-05-29
add --no-deps and --no-cache when building client dep mounts

matslina:
With this change, `google/_upb/_message.abi3.so` is created. Before, it was not. Seems important that it is.


matslina:
@prbot approve @mwaskom

mwaskom (commented):
Still confused why these flags appear to solve the issue you mention, but they make sense to have either way.

PR #3196, opened 2025-05-30
move dep mount creation logic for testability

matslina:
This change allows for client dependency mounts to be created from Modal's internal integration tests.


modal/mount.py
```diff
@@ -810,3 +810,110 @@ def _is_modal_path(remote_path: PurePosixPath):
         if is_modal_path:
             return True
     return False
+
```
matslina, line 813: Tried moving this and the client package mounts into a separate module, but ended up in import loop hell and backed away from it for now.

modal/mount.py
```diff
@@ -810,3 +810,110 @@ def _is_modal_path(remote_path: PurePosixPath):
...
+REMOTE_SITECUSTOMIZE_PATH = "/pkg/sitecustomize.py"
+
+SITECUSTOMIZE_CONTENT = f"""
+# This file is automatically generated by Modal.
+# It ensures that Modal's python dependencies are available in the Python PATH,
+# while prioritizing user-installed packages.
+import sys; sys.path.append('{REMOTE_PACKAGES_PATH}')
+""".strip()
+
+async def _create_single_mount(
```
matslina, line 824: Making these async appears to be necessary for integration tests to work, but I'm not sure why, and could be mistaken. Doesn't hurt, either way I suppose.

modal/mount.py
```diff
@@ -810,3 +810,110 @@ def _is_modal_path(remote_path: PurePosixPath):
...
+    import subprocess
+    import tempfile
+
+    profile_environment = config.get("environment")
+
+    abi_tag = "cp" + python_version.replace(".", "")
+    mount_name = f"{builder_version}-{abi_tag}-{platform}-{arch}"
+    uv_python_platform = uv_python_platform or f"{arch}-{platform}"
+
+    if check_if_exists:
```
matslina, line 842: I'm not sure why, but in the integration test setup this `from_name` seems to block the event loop and everything croaks. It's strictly not needed in there, so passing this flag was a quick fix to get around it.

modal/mount.py
mwaskom: Yeah I think I've run into similar issues before! Since we no longer have any public API defined in this module I think it's tolerable to have more of this internal stuff.

modal/mount.py
```diff
@@ -810,3 +810,109 @@ def _is_modal_path(remote_path: PurePosixPath):
...
+    import tempfile
+
+    profile_environment = config.get("environment")
+    abi_tag = "cp" + python_version.replace(".", "")
+    mount_name = f"{builder_version}-{abi_tag}-{platform}-{arch}"
+    uv_python_platform = uv_python_platform or f"{arch}-{platform}"
+
+    if check_if_exists:
+        try:
+            Mount.from_name(mount_name, namespace=api_pb2.DEPLOYMENT_NAMESPACE_GLOBAL).hydrate(client)
```
mwaskom, line 843: If this is an async function I think you'll want to be using the `.aio` interface on `.hydrate`.

PR #3197, opened 2025-06-01
Improve Workdir comment after cleanup

ekzhang:
Reflects backend change https://github.com/modal-labs/modal/pull/23133 and hopefully this reduces future confusion

By the way, I tried to search through the code and wasn't able to find anywhere this field is used currently — but I'm probably missing something!


ekzhang:
@prbot approve

PR #3199, opened 2025-06-02
A few async_utils deadlock safety improvements

freider:
Some general improvements for reducing risk of async "freezes":
* cancellation of async_merge tasks are now concurrent instead of sequential, making cancellation faster
* adds `prevent_cancellation_abortion` util
* adds cancellation timeout for async_merge (default 10s) used for cancellations + exceptions
* async_map now wraps the mapper function in `prevent_cancellation_abortion` to ensure the workers don't infinite-loop (would add some task-wrapping overhead, hopefully not too bad?)
* similar cancellation timeout for async_map (just propagates to the internal async_merge)
* bumping synchronicity dependency to 0.9.14 which includes additional thread safety fixes for cancellation

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

* Fixed a rare race condition that could cause `.spawn_map` to never finish in case of exceptions or async cancellations.


freider:
@gongy I have a vague recollection of you doing some benchmarking of async_map previously - do you happen to still have whatever script you used for that? I hope that adding some asyncio.create_task-wrapping of calls here won't add noticeable overhead but you never know.

An alternative approach would be to rewrite `async_map` completely, using a less risky structure than `while True`-workers...

mwaskom (approved):
✅

PR #3202, opened 2025-06-03
Fixes issue where an empty map would stall for 55s before completing [CLI-416]

freider:
Empty maps would seemingly "stall forever" (although technically 55s, I think) due to FunctionGetOutputs longpolling not getting aborted when the map completion criteria is met (=all inputs have been created + number of emitted outputs match the number of created inputs).

While this scenario could _possibly_ happen at other points in a map lifecycle (under very rare conditions), the empty input iterator would with almost 100% likelyhood reproduce this since the initial FunctionGetOutputs gets sent immediately.

The fix is to set up an event that is triggered whenever the map is complete, which can be concurrently listened to while waiting for FunctionGetOutputs. This also gave me the opportunity to refactor some of the `nonlocal` mess and bring the most important of those state changes in under a common `update_state()` function, which would also be responsible for setting the completion criteria event.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

* Fixed an issue where `Function.map()` and similar methods would stall for 55 seconds when passed an empty iterator as input, instead of completing immediately.

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -243,9 +243,12 @@ async def _map_invocation(
             retried_outputs
 
         last_entry_id = "0-0"
-
+        print("HAVE ALL", have_all_inputs, outputs_completed < inputs_created)
         while not have_all_inputs or outputs_completed < inputs_created:
-            logger.debug(f"Requesting outputs. Have {outputs_completed} outputs, {inputs_created} inputs.")
+            logger.debug(
+                f"Requesting outputs. Have {outputs_completed} outputs, {inputs_created} inputs. {repr(have_all_inputs)}"
+            )
+            print()
             # Get input_jwts of all items in the WAITING_FOR_OUTPUT state.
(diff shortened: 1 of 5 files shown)


modal/parallel_map.py
```diff
@@ -147,9 +144,27 @@ async def input_iter():
                 break
             yield raw_input  # args, kwargs
 
-    async def drain_input_generator():
-        nonlocal have_all_inputs
+    def update_state(set_have_all_inputs=None, set_inputs_created=None, set_outputs_completed=None):
```
rculbertson, line 147: nice way to handle all these state variables!

rculbertson:
Nice work! I looked over this quickly and it seems good, but I'd like to take a closer look tomorrow to understand some of the async logic better.
Just for my info, is this urgent to get in?

Also, I ran a test job on your branch which maps over 1M inputs and has been good at catching race conditions - finished successfully, no issues!

PR #3203, opened 2025-06-03
Parallelize client dependency mount creation

mwaskom:
We build these at the start of every integration test run, so making it faster would be helpful.


matslina (approved):
🥳

PR #3204, opened 2025-06-03
Version updates for next Image Builder version

mwaskom:
## Describe your changes

Requires opt-in to IMAGE_BUILDER_VERSION=PREVIEW; also we'll need to re-publish the dependency mount.

Should get some more testing before we consider this final. I ran some integration tests which surfaced a couple things, but not all of them.

Closes CLI-420

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -115,7 +115,7 @@ def test_python_version_validation(builder_version):
 def test_dockerhub_python_version(builder_version):
     assert _dockerhub_python_version(builder_version, "3.9.1") == "3.9.1"
 
-    expected_39_full = {"2023.12": "3.9.15", "2024.04": "3.9.19", "2024.10": "3.9.20", "PREVIEW": "3.9.20"}[
+    expected_39_full = {"2023.12": "3.9.15", "2024.04": "3.9.19", "2024.10": "3.9.20", "PREVIEW": "3.9.22"}[
         builder_version
     ]
     assert _dockerhub_python_version(builder_version, "3.9") == expected_39_full


modal/requirements/PREVIEW.txt
```diff
@@ -1,16 +1,17 @@
...
-hpack==4.0.0
-hyperframe==6.0.1
+aiohappyeyeballs==2.6.1
+aiohttp==3.12.7
+aiosignal==1.3.2
+async-timeout==5.0.1 ; python_version < "3.11"
+attrs==25.3.0
+certifi==2025.4.26
+frozenlist==1.6.0
+grpclib==0.4.8
```
mwaskom, line 8: Note we currently pin 0.4.7 as a library dependency; we should update that

modal/requirements/PREVIEW.txt
```diff
@@ -1,16 +1,17 @@
...
+hpack==4.1.0
+hyperframe==6.1.0
 idna==3.10
-multidict==6.1.0
-protobuf>=3.20,<6
-typing_extensions==4.12.2
-yarl==1.13.1
+multidict==6.4.4
+propcache==0.3.1
+protobuf==6.31.1
```
mwaskom, line 15: Since this will just be a "fallback" version it feels safe to go back to pinning — any App that requires a certain protobuf version _should_ use that one.

thomasjpfan: Do we test with `protobuf==3.20` in the CI to make sure we do not break those users?

modal/requirements/PREVIEW.txt
thomasjpfan: Can we bump the version in this PR?

mwaskom: We could or we could do it separately; that change would affect all client upgraders immediately, whereas this PR won't affect anyone who's not opting into the PREVIEW image builder (basically nobody)

modal/requirements/PREVIEW.txt
mwaskom: There is one entry in the text matrix that uses an old protobuf yeah: https://github.com/modal-labs/modal-client/blob/96592a6c68d3e4efd0240aa023ee2900891935f9/.github/workflows/ci-cd.yml#L105

.github/workflows/ci-cd.yml
```diff
@@ -182,7 +182,7 @@ jobs:
 
       - name: Check entrypoint import
         run: |
-          python -c 'import modal._container_entrypoint'
+          python -c 'import modal._container_entrypoint; import modal._runtime.asgi'
```
mwaskom, line 185: A while back we made this lazily importable so we don't pull in the `aiohttp` stack unless necessary; but it's important to verify that this _can_ work with the container requirements.

modal/requirements/PREVIEW.txt
mwaskom: Probably could improve this a bit though!

thomasjpfan (approved):

.github/workflows/ci-cd.yml
ekzhang: Nice optimization!

PR #3206, opened 2025-06-04
Upgrade to synchronicity 0.9.13 [CLI-415]

freider:
Fixes thread race condition that could sometimes make .map calls deadlock


freider:
@prbot approve @mwaskom

PR #3207, opened 2025-06-04
Improve experimental dependency-free Image function docs

mwaskom:
Realized that these docstrings were slightly misleading; it's possible to use the Images they create directly for a Function if the underlying Image already has client dependencies.


thomasjpfan (commented):
LGTM

PR #3208, opened 2025-06-04
Add python_versions to create_client_dependency_mounts

ekzhang:
It's on the hot path of integration tests so we can speed it up by just deploying 1 version

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -10,7 +10,7 @@ import typing
 import warnings
 from collections.abc import AsyncGenerator
 from pathlib import Path, PurePosixPath
-from typing import Callable, Optional, Sequence, Union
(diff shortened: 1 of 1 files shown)


mwaskom:
@mats and I discussed this but decided there was a risk that integration tests could define an Image with a Python version that differs from the version used to run the tests. Is there something to guarantee that won't happen?

BTW I have a WIP PR that further parallelizes this so it's much less obtrusive.

mwaskom:
Looks like `image_e2e_test.py::test_image_metadata` is parameterized by Python version independently of the testing environment Python.

ekzhang:
Thanks for the heads-up! I'll make sure we deploy the other versions for the split that runs test specifically.

mwaskom:
That seems a little brittle, no?

PR #3210, opened 2025-06-04
Overwrite mount for preview image builder and speed up creation

mwaskom:
## Describe your changes

- Allow Mount deployment to overwrite an existing Mount of the same name
- Request this behavior when deploying the PREVIEW image builder so we can iterate on it
- Speed up client dependency mount creation by using async subprocess call

Requires a server change to work as expected.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -290,6 +290,7 @@ class _Mount(_Object, type_prefix="mo"):
     _deployment_name: Optional[str] = None
     _namespace: Optional[int] = None
     _environment_name: Optional[str] = None
+    _allow_overwrite: bool = False
     _content_checksum_sha256_hex: Optional[str] = None
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


mwaskom:
Not sure if there are any reasons for concern that we'd be updating the PREVIEW mount every time we merge a client commit. Certainly it will be the case that the vast majority of client changes do not affect this mount, so it's a bit wasteful. I suppose that we could add a file filter to the trigger for the CI job?

modal/mount.py
```diff
@@ -826,16 +835,16 @@ def _is_modal_path(remote_path: PurePosixPath):
...
-async def _create_single_mount(
+async def _create_single_client_dependency_mount(
     client: _Client,
     builder_version: str,
     python_version: str,
     platform: str,
     arch: str,
     uv_python_platform: str = None,
     check_if_exists: bool = True,
+    allow_overwrite: bool = False,
```
mwaskom, line 846: Kind of annoying that you need to set both to get the overwrite behavior.

matslina (approved):

PR #3211, opened 2025-06-05
Support recursive copy in the CLI

schmee-td:
This is not supported in the backend for V1 volumes but it works for V2 volumes.

I didn't update the docstring for this since it's not supported for V1 volumes, so it seems like something we don't want to advertise until V2 becomes mainstream.

## Compatibility checklist

---

### DIFF
--- a/test/volume_test.py
+++ b/test/volume_test.py
@@ -380,7 +380,7 @@ async def test_volume_copy_1(client, tmp_path, servicer, version):
         object_id = vol.object_id
 
         # copy file from src_path to dst_path
-        vol.copy_files([src_path], dst_path)
+        vol.copy_files([src_path], dst_path, False)
 
     assert servicer.volumes[object_id].files.keys() == {src_path, dst_path}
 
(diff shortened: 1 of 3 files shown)


test/volume_test.py
```diff
@@ -380,7 +380,7 @@ async def test_volume_copy_1(client, tmp_path, servicer, version):
         object_id = vol.object_id
 
         # copy file from src_path to dst_path
-        vol.copy_files([src_path], dst_path)
+        vol.copy_files([src_path], dst_path, False)
```
dflemstr, line 383: Do we want a test where this is `True`?

schmee-td: The in-memory volume used for testing is a flat structure, so I'd have to rewrite the entire thing which I think is out of scope for this PR.

dflemstr: Maybe this can be done in a follow-up PR then?

dflemstr:
Created a notion doc where we can collect stuff: https://www.notion.so/modal-com/volumefs2-release-plan-chores-2091e7f1694980bd8a8ef22cd8aced13?source=copy_link

PR #3213, opened 2025-06-05
[SVC-677] Add ability for clients to attempt blob uploads to multiple urls on failure

CyrilSharma:
Allows the client to accept a list of bucket-urls for different providers (S3, R2, etc.) from the server, and choose a different provider if one of them is unavailable.

## Compatibility checklist

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -4,6 +4,7 @@ import dataclasses
 import hashlib
 import os
 import platform
+import random
 import time
 from collections.abc import AsyncIterator
 from contextlib import AbstractContextManager, contextmanager
@@ -54,6 +55,8 @@ MULTIPART_UPLOAD_THRESHOLD = 1024**3
 
 # For block based storage like volumefs2: the size of a block
 BLOCK_SIZE: int = 8 * 1024 * 1024
+
+HEALTHY_R2_UPLOAD_PERCENTAGE = 0.95
 
 
 @retry(n_attempts=5, base_delay=0.5, timeout=None)
@@ -182,6 +185,22 @@ def get_content_length(data: BinaryIO) -> int:
     return content_length - pos
 
 
+async def _blob_upload_with_fallback(items, blob_ids, callback):
+    for idx, (item, blob_id) in enumerate(zip(items, blob_ids)):
+        # We want to default to R2 95% of the time and S3 5% of the time.
+        # To ensure the failure path is continuously exercised.
+        if idx == 0 and len(items) > 1 and random.random() > HEALTHY_R2_UPLOAD_PERCENTAGE:
+            continue
+        try:
+            await callback(item)
+            return blob_id
+        except Exception as _:
+            # Ignore all errors except the last one, since we're out of fallback options.
+            if idx == len(items) - 1:
+                raise
+    raise ExecutionError("Failed to upload blob")
+
+
 async def _blob_upload(
     upload_hashes: UploadHashes, data: Union[bytes, BinaryIO], stub, progress_report_cb: Optional[Callable] = None
 ) -> str:
@@ -197,17 +216,23 @@ async def _blob_upload(
     )
     resp = await retry_transient_errors(stub.BlobCreate, req)
 
-    blob_id = resp.blob_id
+    if resp.WhichOneof("upload_types_oneof") == "multiparts":
 
-    if resp.WhichOneof("upload_type_oneof") == "multipart":
-        await perform_multipart_upload(
-            data,
-            content_length=content_length,
-            max_part_size=resp.multipart.part_length,
-            part_urls=resp.multipart.upload_urls,
-            completion_url=resp.multipart.completion_url,
-            upload_chunk_size=DEFAULT_SEGMENT_CHUNK_SIZE,
-            progress_report_cb=progress_report_cb,
+        async def upload_multipart_upload(part):
+            return await perform_multipart_upload(
+                data,
+                content_length=content_length,
+                max_part_size=part.part_length,
+                part_urls=part.upload_urls,
+                completion_url=part.completion_url,
+                upload_chunk_size=DEFAULT_SEGMENT_CHUNK_SIZE,
+                progress_report_cb=progress_report_cb,
+            )
+
+        blob_id = await _blob_upload_with_fallback(
+            resp.multiparts.items,
+            resp.blob_ids,
+            upload_multipart_upload,
         )
     else:
         from .bytes_io_segment_payload import BytesIOSegmentPayload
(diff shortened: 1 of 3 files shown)


modal_version/__init__.py
```diff
@@ -1,4 +1,4 @@
 # Copyright Modal Labs 2025
 """Supplies the current version of the modal client library."""
 
-__version__ = "1.0.4.dev4"
+__version__ = "1.0.4.dev5"
```
mwaskom: You don't need to change this manually for dev releases; it happens automatically when a PR is merged (https://www.notion.so/modal-com/Client-deployment-workflow-1f51e7f1694980d381d5c5c0f4f1d8fc?source=copy_link)

mwaskom (commented):
so now the client has a "bucket list" ha ha send tweet

modal/_utils/blob_utils.py
```diff
@@ -222,6 +213,93 @@ async def _blob_upload(
...
+
+    content_length = get_content_length(data)
+
+    req = api_pb2.BlobCreateRequest(
+        content_md5=upload_hashes.md5_base64,
+        content_sha256_base64=upload_hashes.sha256_base64,
+        content_length=content_length,
+    )
+    resp = await retry_transient_errors(stub.BlobCreate, req)
+    is_legacy_server = len(resp.blob_ids) == 0
```
mwaskom: What's a "legacy server" in this context?

CyrilSharma: Servers which do not have the logic to send multiple urls. The default PR mentioned newer clients should maintain compatibility with older servers, hence the need for this.

modal/_utils/blob_utils.py
```diff
@@ -222,6 +214,104 @@ async def _blob_upload(
...
+    resp: api_pb2.BlobCreateResponse,
+    data: Union[bytes, BinaryIO],
+    upload_hashes: UploadHashes,
+    stub,
+    progress_report_cb: Optional[Callable] = None,
+) -> str:
+    if resp.WhichOneof("upload_types_oneof") == "multiparts":
+        for idx, (part, blob) in enumerate(zip(resp.multiparts.items, resp.blob_ids)):
+            # We want to ensure the failure path is tested.
+            # This is a port of what we were doing in the server.
```
devennavani: Can we leave a more informative comment here? E.g. explain we want to default to R2 95% of the time but choose S3 the other 5% to ensure the failure path is continuously exercised.

modal/_utils/blob_utils.py
```diff
@@ -222,6 +214,104 @@ async def _blob_upload(
...
+                    completion_url=part.completion_url,
+                    upload_chunk_size=DEFAULT_SEGMENT_CHUNK_SIZE,
+                    progress_report_cb=progress_report_cb,
+                )
+                return blob
+            except Exception as e:
+                if idx == len(resp.multiparts) - 1:
+                    raise e
+                logger.warn(f"Error when uploading inputs: {e}")
+                logger.warn("Trying backup url...")
```
devennavani: Shouldn't all the `logger.warn`s be `logger.warning`s?

devennavani: Also don't think we need the `Error when uploading inputs:` log here

modal/_utils/blob_utils.py
```diff
@@ -222,6 +214,104 @@ async def _blob_upload(
...
+                    url,
+                    payload,
+                    # for single part uploads, we use server side md5 checksums
+                    content_md5_b64=upload_hashes.md5_base64,
+                )
+                return blob
+            except Exception as e:
+                if idx == len(resp.upload_urls.items) - 1:
+                    raise e
+                logger.warn(f"Error when uploading inputs: {e}")
```
devennavani: ditto, dont need the `Error when uploading inputs:` log here

modal/_utils/blob_utils.py
```diff
@@ -380,8 +470,10 @@ def source():
         mode,
     )
 
+
```
devennavani, line 416: are the extraneous spaces added here and below intentional? if not, let's eliminate them from the diff

modal/_utils/blob_utils.py
```diff
@@ -222,6 +214,104 @@ async def _blob_upload(
...
+async def _blob_upload_with_fallback(
+    content_length: int,
+    resp: api_pb2.BlobCreateResponse,
+    data: Union[bytes, BinaryIO],
+    upload_hashes: UploadHashes,
+    stub,
+    progress_report_cb: Optional[Callable] = None,
+) -> str:
+    if resp.WhichOneof("upload_types_oneof") == "multiparts":
```
devennavani: Is there a clean way we can avoid the redundant code in the `if` branch and the `else` branch? e.g. the RNG and error catching and for loop iteration is redundant

modal/_utils/blob_utils.py
```diff
@@ -222,6 +214,104 @@ async def _blob_upload(
...
+    else:
+        from .bytes_io_segment_payload import BytesIOSegmentPayload
+
+        payload = BytesIOSegmentPayload(
+            data, segment_start=0, segment_length=content_length, progress_report_cb=progress_report_cb
+        )
+        for idx, (url, blob) in enumerate(zip(resp.upload_urls.items, resp.blob_ids)):
+            # We want to ensure the failure path is tested.
+            # This is a port of what we were doing in the server.
+            if idx == 0 and random.random() < 0.05:
```
devennavani: could we use a global constant here? like `HEALTHY_R2_UPLOAD_PERCENTAGE = 95` on the server

modal/_utils/blob_utils.py
```diff
@@ -222,6 +214,104 @@ async def _blob_upload(
             content_md5_b64=upload_hashes.md5_base64,
         )
 
+    return resp.blob_id
+
+
+async def _blob_upload_with_fallback(
```
devennavani: think this function is missing a `return` statement

modal/_utils/blob_utils.py
```diff
@@ -222,6 +214,104 @@ async def _blob_upload(
...
+                    max_part_size=part.part_length,
+                    part_urls=part.upload_urls,
+                    completion_url=part.completion_url,
+                    upload_chunk_size=DEFAULT_SEGMENT_CHUNK_SIZE,
+                    progress_report_cb=progress_report_cb,
+                )
+                return blob
+            except Exception as e:
+                if idx == len(resp.multiparts) - 1:
+                    raise e
```
devennavani: this should be `raise` instead of `raise e` to preserve the stack trace

devennavani (commented):
left comments

modal/_utils/blob_utils.py
mwaskom: We should generally be able to know that we're not deploying a new client change until all servers with the required backend support have rolled out (and won't need to be rolled back) — not sure we need to add complexity in the client just to support older _and no longer deployed_ servers.

mwaskom: Let me know if I'm missing something about the rollout plan here, though.

modal/_utils/blob_utils.py
CyrilSharma: Ruff seems to be adding them automatically.

modal/_utils/blob_utils.py
CyrilSharma: Makes sense, I was being overly conservative.

modal/_utils/blob_utils.py
mwaskom: yeah this looks like valid pep8 (with new spaces). i think one of the fs engineers who works on this file does not have a properly configured local python dev environment :D

devennavani:
@CyrilSharma can you modify `BlobCreate` in `MockClientServicer` to reflect the server-side changes?

PR #3214, opened 2025-06-06
Adds `uv sync` support to Image

thomasjpfan:
## Describe your changes

This PR adds `uv_sync` support `uv.lock` files. The assumption this makes is that a `uv.lock` user would want their local environment, Modal remote environment, and other remote environments to have the same.

I have tested this with all the image builder versions and 

Fixes https://linear.app/modal-labs/issue/CLI-429/add-first-class-support-for-using-uv-in-modalimage-builds

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Add `uv_sync` to image builder to build images with a `uv.lock` file. This feature is in beta as we improve it based on feedback. During the beta period, updating modal may cause image rebuilds.

```python
import modal

image = modal.Image.debian_slim().uv_sync()
```


modal/image.py
```diff
@@ -1299,6 +1299,67 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+
+            uv_root = "/.uv"
+            context_files = {"/.uv.lock": lockfile_, "/.pyproject.toml": pyproject_toml_}
+
+            uv_sync_args = [
+                f"--project={uv_root}",
+                "--frozen",  # Do not update `uv.lock`
+                "--no-install-workspace",  # Do not install the root project or any "uv workspace"
+                "--no-cache",  # Cache is not persisted, so we disable it
+                "--no-managed-python",  # Use the system python interpreter to create venv
```
thomasjpfan: Using `-no-managed-python` requires that the local python environment or the version in `debian_slim(python_version=)` to be compatible with the `pyproject.toml` file:

```toml
[project]
requires-python = ">=3.11"
```

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+                "--compile-bytecode",
+            ]
```
mwaskom, line 1513: `uv sync` has like a million other options; how can we expose them for users who need to do something specific?

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+            uv_sync_args_joined = " ".join(uv_sync_args)
+
+            # TODO Make user configurable through a kwargs
+            UV_VERSION = "0.7.11"
+
+            commands = [
+                "FROM base",
+                f"RUN python -m pip install uv=={UV_VERSION}",
```
mwaskom: It will be rare, but users may not have `pip` installed in their Image going forward. What should they do?

What about the "temporarily mount Astral's `uv` image" approach you were exploring earlier?

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             gpu_config=parse_gpu_config(gpu),
         )
 
+    def uv_sync(
+        self,
+        lockfile: str = "./uv.lock",  # Path to lockfile
+        # Path to pyproject.toml. If not provided, then use file in the same directory as the lockfile
+        pyproject_toml: Optional[str] = None,
```
mwaskom: j/w but how are you thinking about the relative importance of the project file vs the lockfile here?

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        *,
+        force_build: bool = False,  # Ignore cached builds, similar to 'docker build --no-cache'
+        secrets: Sequence[_Secret] = [],
+        gpu: GPU_T = None,
+    ) -> "_Image":
+        """Install dependencies from a local `uv.lock` file using `uv sync`.
+
+        Note that only dependencies are installed. Please use `add_local_python_source` to include local python source
+        files. Please ensure that `modal` is a dependency by running `uv add modal`.
```
mwaskom: I thought we were going to limit the availability of this method to image builders `>=2024.10` in which case this advice wouldn't be necessary right?

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        self,
+        lockfile: str = "./uv.lock",  # Path to lockfile
+        # Path to pyproject.toml. If not provided, then use file in the same directory as the lockfile
+        pyproject_toml: Optional[str] = None,
+        *,
+        force_build: bool = False,  # Ignore cached builds, similar to 'docker build --no-cache'
+        secrets: Sequence[_Secret] = [],
+        gpu: GPU_T = None,
+    ) -> "_Image":
+        """Install dependencies from a local `uv.lock` file using `uv sync`.
```
mwaskom: Technically this does more than just "install dependencies" right?

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        files. Please ensure that `modal` is a dependency by running `uv add modal`.
+        """
+
+        def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
+            lockfile_ = os.path.expanduser(lockfile)
+            if pyproject_toml is None:
+                parent_dir = os.path.dirname(lockfile_)
+                pyproject_toml_ = os.path.join(parent_dir, "pyproject.toml")
+                if not os.path.exists(pyproject_toml_):
+                    msg = "pyproject.toml does not exist in the same directory as the uv.lock file"
```
mwaskom: Would be user-friendly for the error message to show the directory it's trying to use.

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             gpu_config=parse_gpu_config(gpu),
         )
 
+    def uv_sync(
+        self,
+        lockfile: str = "./uv.lock",  # Path to lockfile
```
mwaskom: Can we clarify that these are *local* paths?

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        # Path to pyproject.toml. If not provided, then use file in the same directory as the lockfile
+        pyproject_toml: Optional[str] = None,
+        *,
+        force_build: bool = False,  # Ignore cached builds, similar to 'docker build --no-cache'
+        secrets: Sequence[_Secret] = [],
+        gpu: GPU_T = None,
+    ) -> "_Image":
+        """Install dependencies from a local `uv.lock` file using `uv sync`.
+
+        Note that only dependencies are installed. Please use `add_local_python_source` to include local python source
```
mwaskom: How would people that are fully bought-into the `uv sync` workflow normally be handling "local" dependencies?

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+                "--frozen",  # Do not update `uv.lock`
+                "--no-install-workspace",  # Do not install the root project or any "uv workspace"
+                "--no-cache",  # Cache is not persisted, so we disable it
+                "--no-managed-python",  # Use the system python interpreter to create venv
+                "--compile-bytecode",
+            ]
+            uv_sync_args_joined = " ".join(uv_sync_args)
+
+            # TODO Make user configurable through a kwargs
+            UV_VERSION = "0.7.11"
```
mwaskom: This is tricky; any hard-coded version (even as a default value for a kwarg) is going to be impossible to update in the future without forcing Image rebuilds.

If we want to lock a specific uv version we'd at least want to make it part of the image builder version so that it can be updated in the future. But is there a good reason to default to a specific version, rather than the "latest"?

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+
+            commands = [
+                "FROM base",
+                f"RUN python -m pip install uv=={UV_VERSION}",
+                f"COPY /.pyproject.toml {uv_root}/pyproject.toml",
+                f"COPY /.uv.lock {uv_root}/uv.lock",
```
mwaskom: Does the `uv.lock` file get modified outside of users adding new dependencies? I'm worried about this workflow having more image rebuilds than users "need". I guess that's also a consideration w/r/t the `pyproject.toml`, if users do something like bump the version of their project (without changing any dependencies) it will force a rebuild of the modal Image.

J/w if we can be creative about trying to avoid such issues, since we're explicitly limiting this operation to _dependency_ provisioning.

modal/image.py
mwaskom: Does this mean that `Image.uv_sync` will work only when the Image is using Python 3.11+?

test/supports/uv_lock_project/pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+name = "uv-lock-project"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.13"
+dependencies = [
+    "modal>=1.0.3",
```
mwaskom, line 8: Note that I'm still pretty unhappy about this "requirement" since the modal library dependencies do not correspond directly to the dependencies required in a Modal container.

modal/image.py
```diff
@@ -1299,6 +1299,68 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+                    raise InvalidError(msg)
+            else:
+                pyproject_toml_ = os.path.expanduser(lockfile)
+
+            uv_root = "/.uv"
+            context_files = {"/.uv.lock": lockfile_, "/.pyproject.toml": pyproject_toml_}
+
+            uv_sync_args = [
+                f"--project={uv_root}",
+                "--frozen",  # Do not update `uv.lock`
```
mwaskom: Still a little uncertain whether this is always the desired behavior.

mwaskom:
> I have tested this with all the image builder versions

Including `PREVIEW`? 😇

test/supports/uv_lock_project/pyproject.toml
thomasjpfan: The typical uv workflow is becoming `uv run modal deploy ...` and let uv manage your local environment. For that to work `modal` needs to be a requirement in the local environment.

I think it's really strange that the library dependencies do not match the dependencies in the Modal container. How would a user know what the actual dependencies are and that it does not conflict with their own code?

modal/image.py
thomasjpfan: Yea, this is the behavior when using `--no-managed-python`. If we remove it, then uv would download it's own interpreter and would not care what the base image has.

modal/image.py
thomasjpfan: Most likely a `extra_options` like `pip_install`. Do we add it now or wait for a user to request it?

modal/image.py
thomasjpfan: I'm okay with latest if we allow the user to specify the uv version in `uv_sync` if they want stability.

modal/image.py
thomasjpfan: > What about the "temporarily mount Astral's uv image" approach you were exploring earlier?

That works, but our image builder does not support "--mount" yet.

The short term solution is to get the uv's installer to work: https://docs.astral.sh/uv/getting-started/installation/#standalone-installer

modal/image.py
thomasjpfan: uv.lock does not change outside of users adding a new dependency. `pyproject.toml` would change more often like changing the license.

I'm okay with not optimizing for this use case. I think once you get `pyproject.toml` configured, you do not really modify it much unless you add a new dependency.

modal/image.py
thomasjpfan: For image building, I think it's the only option. I would not want my image builder to change any of my dependencies (that I use in other projects).

modal/image.py
thomasjpfan: We do not have to require it, but I would still recommend it. I think it's very confusing to have `import modal` in your code but not have `modal` as a requirement in your pyproject.toml or lockfile.

modal/image.py
thomasjpfan: There are three flows for `uv sync`:

1. Applications, where local python files are assumed to be next to their `uv.lock` files. (This is what `add_local_python_source` does)
2. Packages, where you want to invoke a build system first, like hatchling or setuptools. This is useful if you have an entrypoint. This is tough to get right for an image builder, because build systems assume you have the complete CWD to trigger correctly (including the `.git` folder if you are using `setuptools_scm`)

The uv default is 1. Application, which is consist with the install methods for Modal.

modal/image.py
thomasjpfan: If you mean that it corrects the venv, then it does do more than install dependencies. I'll update the doc

modal/image.py
thomasjpfan: They are equally important. `uv sync` requires both the `pyproject.toml` and `uv.lock` present.

modal/image.py
thomasjpfan: If the fix to work with older image builders is to run `uv add modal`, then I'm okay to supporting the other image builders. (I think most `uv.lock` users are okay with that)

modal/image.py
thomasjpfan: Turns out `curl` is not installed in our base environment. For now, I'm okay with requiring `pip` to be installed. (I added a line in the docs)

mwaskom: > Turns out curl is not installed in our base environment

Yeah — and regardless, we have to accept "arbitrary" images and can't assume they have such tools available unfortunately.

What about something like

```
COPY --from=ghcr.io/astral-sh/uv:latest /uv ...
```

would that work?

modal/image.py
mwaskom: I'm not sure if it matters but in my naive understanding, `uv.lock` is derived from `pyproject.toml` so i would consider the latter to be the "primary" target to build against. Could be wrong!

modal/image.py
mwaskom: > I think it's very confusing to have import modal in your code but not have modal as a requirement in your pyproject.toml or lockfile.

Maybe, but that is generally the case for all of our Image recipes right? `uv sync` aside, we don't suggest that users do `.pip_install("modal")`; the assumption is that modal (and its dependencies) are just provided when you're running on Modal.

modal/image.py
mwaskom: > if we allow the user to specify the uv version in uv_sync if they want stability

Not saying this is a bad idea at all, but i worry about letting inconsistency between `uv` and `pip` (and `poetry` for that matter) creep into the API. We don't otherwise expose the version of the underlying tool in these wrapping methods.

modal/image.py
mwaskom: For "Applications" users mostly won't need to do `add_local_python_source` on the Modal source App, even with this `uv_sync` flow right? Wouldn't want that to change.

modal/image.py
mwaskom: Ugh that seems really annoying?  We need to at least raise a clear error message here, and document this constraint. Would love if we didn't have it though.

modal/image.py
mwaskom: Don't see any reason not to add it here. Are there any other commonly-set options that make sense to elevate into the method signature? We may also want to deconflict options we're hardcoding so users don't get confusing errors.

modal/image.py
thomasjpfan: Yup, nothing is changing on with `uv_sync` with "Applications" compared to `pip_install`. For `uv_sync` or `pip_install`, one still needs to run `add_local_python_source` to have other modules available during runtime.

In that case, I'll remove the docstring.

modal/image.py
mwaskom: Maybe another thought here ... what if the parameterization were something like `uv_project_dir` and then we expected `pyproject.toml` and `uv.lock` files to live there? Are there any uv workflows where the lockfile does not end up next to the project file?

(I also wonder if we need to mandate the existence of a `uv.lock` file. It looks like I can run `uv sync` in a directory with just the `pyproject.toml` and it sets things up correctly (including generating the locfile). That might be nice for users who want to use this as a sort of "uv version of `pip_install_from_pyproject.toml`. But not sure if that violates various assumptions of the opinionated uv workflow.

modal/image.py
mwaskom: Actually I remembered we have this issue with `poetry` where we pin on 1.x and they released a 2.x that we need to make it possible to update to. So we should probably try to solve these at the same time in a consistent way.

I'm not sure what's the best default. I'm both slightly worried about doing something like `latest` and a future release breaking all modal image builds, and also worried about pinning a specific version such that two years in the future people are by default using a rather old version of uv that lacks various functionality. Seems like something important to get right...

modal/image.py
mwaskom: Spoke with @elias about this a bit and we were both still a little nervous that it won't be obvious that any changes to `pyproject.toml` will invalidate the whole Image. OTOH it's not obvious to me that there's a better solution.

thomasjpfan:
> "using one uv environment locally to deploy a Modal App that contains multiple different images"?

Use [uv dependency groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups), and then:

```python
image1 = image.Image.debian_slim().uv_sync(extra_options="--group llm")

image2 = image.Image.debian_slim().uv_sync(extra_options="--group preprocess")
```

We can have a `groups` kwargs, but I prefer having less API if `extra_options` works.

modal/image.py
thomasjpfan: > Are there any uv workflows where the lockfile does not end up next to the project file?

I've always seen `uv.lock` and `pyproject.toml` next to each other. I'm okay with `uv_project_dir` being the parameterization.

I think it's a little strange to use `uv_sync` with only a `pyproject.toml`, but I can get behind it.

modal/image.py
thomasjpfan: Yup that works!

modal/image.py
thomasjpfan: I added `extra_options` here for now. If we watch to match `poetry`, the biggest feature to elevate is "uv dependency grousps"

thomasjpfan: But they are so many "dependency group" parameters, I rather let a user decide with `extra_options`:

> Once groups are defined, the --all-groups, --no-default-groups, --group, --only-group, and --no-group options can be used to include or exclude their dependencies.

modal/image.py
thomasjpfan: I'm okay with either option as long as the user can pin the version themselves.

modal/image.py
thomasjpfan: I updated the PR to check for `modal` as a dependency when the image builder is `<= 2024.10` and raise an error.

thomasjpfan:
Updated PR with:

- Use `uv_project_dir` to specify the project directory
- Support projects with only `pyproject.toml`.
- When using an image builder that is `<= 2024.10` and modal is not specified, then raise an error to state that modal must be installed.

modal/image.py
mwaskom: But that's back to this question of whether most uses would consider installing from our PyPI mirror rather than PyPI would be "changing their dependencies" right? Or is the issue that if your `pyproject.toml` doesn't have exact pins that it could end up pulling in different dependencies from the last time you locked locally? Just want to make sure I understand all the reasoning here since I think you're assuming a very specific opinionated workflow.

modal/image.py
```diff
@@ -1299,6 +1299,103 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        ```
+        """
+
+        def _check_pyproject_toml(pyproject_toml: str, version: ImageBuilderVersion):
+            if not os.path.exists(pyproject_toml):
+                msg = f"Expected {pyproject_toml} to exist in {uv_project_dir}"
+                raise InvalidError(msg)
+
+            if version > "2024.10":
+                # For builder version > 2024.10, modal is mounted at the end and is not
```
mwaskom: Nit: `modal` itself is always provided to containers with a runtime mount. The difference with the forthcoming Image builder version is that the client *dependencies* get added in the runtime mount rather than in the base image.

modal/image.py
```diff
@@ -1299,6 +1299,103 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+                # a requirement in `uv.lock`
+                return
+
+            import toml
+
+            with open(pyproject_toml) as f:
+                pyproject_toml_content = toml.load(f)
+            dependencies = pyproject_toml_content["project"]["dependencies"]
```
mwaskom, line 1476: How does this parsing interact with the use of dependency groups?

modal/image.py
```diff
@@ -1299,6 +1299,103 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        self,
+        uv_project_dir: str = "./",  # Path to local uv mananged project
+        *,
+        force_build: bool = False,  # Ignore cached builds, similar to 'docker build --no-cache'
+        extra_options: str = "",  # Extra options to pass to `uv sync`
+        uv_version: Optional[str] = None,  # uv version to use
+        secrets: Sequence[_Secret] = [],
+        gpu: GPU_T = None,
+    ) -> "_Image":
+        """Creates a virtual environment with the dependencies in `uv.lock` file using `uv sync`.
```
mwaskom: Should we update this description if `uv.lock` is not strictly required?

modal/image.py
mwaskom: Hm I tried running this App:

```python
import sys
import modal

app = modal.App(image=modal.Image.debian_slim("3.10").uv_sync())

@app.function()
def f():
    print(sys.version)
    print(sys.prefix)
```
and it seems to have run fine, using `py3.10 with `sys.prefix` reporting `/.uv/.venv`, which seems to be the same as with 3.11. What am I missing here?

modal/image.py
```diff
@@ -1299,6 +1299,103 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+
+            context_files["/.pyproject.toml"] = pyproject_toml
+            commands.append(f"COPY /.pyproject.toml {uv_root}/pyproject.toml")
+
+            uv_lock = os.path.join(uv_project_dir_, "uv.lock")
+            if os.path.exists(uv_lock):
+                context_files["/.uv.lock"] = uv_lock
+                commands.append(f"COPY /.uv.lock {uv_root}/uv.lock")
+
+                # Do not update `uv.lock` when we have one
```
mwaskom: Would be helpful to explain in the comment why we do this (pypi mirror etc.)

modal/image.py
```diff
@@ -1299,6 +1299,103 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+                "--no-install-workspace",  # Do not install the root project or any "uv workspace"
+                "--no-cache",  # Cache is not persisted, so we disable it
+                "--no-managed-python",  # Use the system python interpreter to create venv
+                "--compile-bytecode",
+                extra_options,
+            ]
+            commands = ["FROM base"]
+
+            if uv_version is None:
+                commands.append("COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv")
```
mwaskom: Is `/usr/local/bin` guaranteed to exist (and be on the path) for arbitrary images?

modal/image.py
mwaskom: Is it fair to say that `--group` is the main option there that would be useful for the "one `pyproject.toml` -> many image" setup that we'd like to support?

I'm also a little confused about the distinction between "extras" and "dependency groups"; the uv docs describe the latter as being primarily relevant to *development dependencies* but I'm not sure exactly how that would map onto a modal App, seems kind of aligned with the "non-published data science project" usecase they mention too.

modal/image.py
mwaskom: Actually having completely borked my local setup I've now learned about the "uv workspace" concept and I think that does provide a counterexample to the question of 

> Are there any uv workflows where the lockfile does not end up next to the project file?

since i gather that uv [expects](https://docs.astral.sh/uv/concepts/projects/workspaces/#workspace-layouts) the lockfile to be one directory up from the pyproject.toml in this setup.

FWIW this sort of thing is why I've been reticent to "support" uv in modal in ways that require modal to bake in uv's assumptions about project layout etc. as it feels like they're still experimenting with various ideas. Not sure exactly what the actionable implication is though.

mwaskom (commented):
We're getting there! Sorry about all the comments but I think this is all pretty complex and a bit risky in terms of shipping something that turns out not to be what users want / expect.

modal/image.py
thomasjpfan: > Or is the issue that if your pyproject.toml doesn't have exact pins that it could end up pulling in different dependencies from the last time you locked locally? 

This. It'll basically not respecting your lock file and update your dependencies when `uv sync` is called.

Note, to support `pyproject.toml` without a `uv.lock`, there is no `--frozen`, because the `uv.lock` file is always generated during build time.

In any case, I'm okay with introducing a `frozen` kwargs so one can opt out of the frozen behavior.

modal/image.py
thomasjpfan: If you set your pyproject.toml to be:

```python
[project]
name = "my-app"
version = "0.1.0"
description = "An example of uv workspace"
readme = "README.md"
requires-python = ">=3.12"
dependencies = ["modal"]
```

then using `Image.debian_slim("3.10")` will fail with this error message:

```bash
=> Step 4: RUN uv sync --project=/.uv --no-install-workspace --no-cache --no-managed-python --compile-bytecode  --frozen
error: No interpreter found for Python >=3.12 in search path
```

`requires-python = ">= 3.12"` is not compatible with ``Image.debian_slim("3.10")``.

modal/image.py
thomasjpfan: With uv workspaces with a root `pyproject.toml`:

```python
[project]
name = "my-app"
version = "0.1.0"
description = "An example of uv workspace"
readme = "README.md"
requires-python = ">=3.9"
dependencies = ["modal"]

[tool.uv.sources]
my-app = { workspace = true }

[tool.uv.workspace]
members = ["packages/*"]
```

It expects a file structure like this:

```bash
.
├── packages
│   ├── mod1
│   │   └── pyproject.toml
│   └── mod2
│       └── pyproject.toml
├── pyproject.toml
└── uv.lock
```

For us to properly support this, we'll have to parse the root `pyproject.toml` and understand to upload the other `pyproject.toml` and 

For the first iteration, I'll like to not support this and `[tool.uv.workspace]` members and error if it exist.

modal/image.py
thomasjpfan: The "development dependency" is just a dependency group named "dev". It's the most common dependency group so they made it a first class citizen.

I can see one using `[project.optional-dependencies]` or "extras" to manage their different images. 

```toml
[project]
name = "my-app"
version = "0.1.0"
description = "Some project"
readme = "README.md"
requires-python = ">=3.12"
dependencies = ["modal"]

[project.optional-dependencies]
extra1 = ["scipy>=1.15.3"]

[dependency-groups]
group1 = ["pydantic>=2.11.5"]

```

So one can either do `uv sync --group group1` or `uv sync --extra extra1`.

The biggest difference between them is that `[project.optional-dependencies]` is used to publish a package such that `my-lib[extra1]` works.

modal/image.py
mwaskom: Oh I misread your original example entirely! Got it now.

modal/image.py
mwaskom: I'm getting that from [these docs](https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies):

> Development dependencies can be divided into multiple groups, using the --group flag.

modal/image.py
thomasjpfan: With docker the `/usr/local/bin` will be created with COPY. Just in case `/usr/local/bin` is not on the path, I'll use the full `/usr/local/bin/uv` to call `uv`.

mwaskom: Since we're putting other uv-related stuff in `/.uv` and now referencing an absolute path when we call it, it might be cleaner to do `/.uv/uv`? Just a thought to better isolate the mutations from this method.

modal/image.py
thomasjpfan: You can do `uv add --group dev pytest` or `uv add --dev pytest`. They mean the same thing.

mwaskom: Right I get the point about `--dev` aliasing, but that documentation about "development dependencies can be divided into multiple groups" makes me think they construe all `--groups` as "development" dependencies. Doesn't necessarily mean it's misaligned with use in a Modal App, just trying to make sure I understand how they're thinking about these features.

modal/image.py
```diff
@@ -1299,6 +1299,140 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             gpu_config=parse_gpu_config(gpu),
         )
 
+    def uv_sync(
+        self,
+        uv_project_dir: str = "./",  # Path to local uv mananged project
+        *,
+        force_build: bool = False,  # Ignore cached builds, similar to 'docker build --no-cache'
+        group: Optional[str] = None,  # Dependency group to install using `uv sync --group`
```
mwaskom: Should this also accept a list? Or does `uv sync` only allow at most one `--group` at a time?

modal/image.py
```diff
@@ -1299,6 +1299,140 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        optional: Optional[str] = None,  # Optional dependencies to install using `uv sync --optional`
```
mwaskom: AFAICT there's no `--optional` option for `uv sync` ... I see one for `uv add` and that places the added dependency in the `optional.dependency` table but not super clear to me how to request this at the `uv sync` time. Seems closest to `--extra` maybe?

thomasjpfan: Yea, it's `extra`. Thank you for catching this!

modal/image.py
```diff
@@ -1299,6 +1299,155 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             gpu_config=parse_gpu_config(gpu),
         )
 
+    def uv_sync(
+        self,
+        uv_project_dir: str = "./",  # Path to local uv managed project
+        *,
+        force_build: bool = False,  # Ignore cached builds, similar to 'docker build --no-cache'
+        group: Optional[list[str]] = None,  # Dependency group to install using `uv sync --group`
```
mwaskom: Should the parameter name be plural if the argument type is a list?

modal/image.py
```diff
@@ -1299,6 +1299,155 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+
+            with open(pyproject_toml) as f:
+                pyproject_toml_content = toml.load(f)
+
+            if (
+                "tool" in pyproject_toml_content
+                and "uv" in pyproject_toml_content["tool"]
+                and "workspace" in pyproject_toml_content["tool"]["uv"]
+            ):
+                raise InvalidError("uv workspaces are not support")
```
mwaskom:
```suggestion
                raise InvalidError("uv workspaces are not supported")
```

Typo?

modal/image.py
```diff
@@ -1299,6 +1299,155 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+            if uv_version is None:
+                commands.append("COPY --from=ghcr.io/astral-sh/uv:latest /uv /.uv/uv")
+            else:
+                commands.append(f"COPY --from=ghcr.io/astral-sh/uv:{uv_version} /uv /.uv/uv")
+
+            context_files = {}
+
+            _check_pyproject_toml(pyproject_toml, version)
+
+            context_files["/.pyproject.toml"] = pyproject_toml
```
mwaskom, line 1533: jw but why make the context paths "private"?

mwaskom (approved):

modal/image.py
thomasjpfan: I'm trying to stay consistent with the pattern here:

https://github.com/modal-labs/modal-client/blob/c430972947bf7e471bfece99f3d56657ab1e953b/modal/image.py#L1251

thomasjpfan: I suspect this is because of the edge case where `pyproject.toml` already exist in the base image.

mwaskom: Not really a big deal either way but they _key_ in the `context_files` dict can't collide with any files in the image right? We end up copying it from that context to `/{UV_ROOT}/pyproject.toml`?

thomasjpfan: Yup, the key of `context_files` should not collide with files in the image.

> We end up copying it from that context to

Yea. It gets copied over, otherwise the RUN commands can not see the file. (I initially tried to add have the context files keyed at `{UV_ROOT}/pyproject.toml`, but it did not work)

I guess I can place them in `/tmp/pyproject.toml` which feels slightly better.

mwaskom: i don't think the "context" location matters much was really just a bit curious. seems fine to stick with what's working and done elsewhere

PR #3216, opened 2025-06-06
Introduce `Image.uv_pip_install` with volume caching

thomasjpfan:
## Describe your changes

Adds `uv pip install` to `modal.Image`. With this PR, a user can provide a cache that is mounted during build time:

```python
cache_volume = modal.Volume.from_name(name="uv-cache-volume", create_if_missing=True)
image = modal.Image.debian_slim().uv_pip_install("torch", cache_volume=cache_volume)
```

This makes it fast to iterate on a project when adding additional packages when you already built an image.

Quick benchmarking: (install PyTorch + some random packages)
- uv pip install with a warm volume cache: 24s
- uv pip install without cache: 55s
- with normal `pip_install`: 93s

I have considered taking a string as a `cache_volume` (with a default value) to create the volume cache for the user, but I think it's a better to be explicit and `cache_volume` to be passed in.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

Adds `uv_pip_install` to install packages with `uv`. You can also pass in a Volume for `uv` to use as a build cache:

```
import modal

cache_volume = modal.Volume.from_name(name="uv-cache-volume", create_if_missing=True)
image = modal.Image.debian_slim().uv_pip_install("torch", cache_volume=cache_volume)
```

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -331,11 +331,10 @@ def _uv_pip_install(pkgs: list[str], extra_options_args: list[str]):
     """Run uv pip install with pkgs and extra_options_args."""
     from subprocess import run
 
-    package_args = sorted(pkgs)
     run(
         ["uv", "pip", "install", "--python", sys.executable, "--compile-bytecode", "--link-mode", "copy"]
         + extra_options_args
-        + package_args
+        + pkgs
     )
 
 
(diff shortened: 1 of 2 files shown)


thomasjpfan:
Marking this as a draft while we decide if we want caching.

See https://github.com/modal-labs/modal-client/pull/3230 for the standard way to do this.

PR #3217, opened 2025-06-08
Prune out ignored directories early in `MountDir.get_files_to_upload()`

ehdr:
## Describe your changes

If you e.g. `image.add_local_dir("dir", ignore=["**/venv"])`, this will prune out any `venv` directories early when evaluating which files to include, avoiding searching through all files within.

CLI-433

## Changelog

- Optimized handling of the `ignore` parameter to `Image.add_local_dir` and similar functions. If you e.g. `image.add_local_dir("dir", ignore=["**/venv"])`, we now prune out any `venv` directories early when evaluating which files to include, avoiding traversing through all files within.

### DIFF
--- a/modal/file_pattern_matcher.py
+++ b/modal/file_pattern_matcher.py
@@ -47,12 +47,33 @@ class _AbstractPatternMatcher:
         return super().__repr__()
 
     @abstractmethod
+    def can_prune_directories(self) -> bool:
+        """
+        Returns True if this pattern matcher allows safe early directory pruning.
+
+        Directory pruning is safe when matching directories can be skipped entirely
+        without missing any files that should be included.
+
+        An example where pruning is not safe is for inverted patterns, like "!**/*.py".
+        """
+        ...
+
+    @abstractmethod
     def __call__(self, path: Path) -> bool: ...
 
 
 class _CustomPatternMatcher(_AbstractPatternMatcher):
     def __init__(self, predicate: Callable[[Path], bool]):
         self._predicate = predicate
+
+    def can_prune_directories(self) -> bool:
+        """
+        Custom pattern matchers (like negated matchers) cannot safely prune directories.
+
+        Since these are arbitrary predicates, we cannot determine if a directory
+        can be safely skipped without evaluating all files within it.
+        """
+        return False
 
     def __call__(self, path: Path) -> bool:
         return self._predicate(path)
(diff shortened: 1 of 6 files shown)


modal/mount.py
```diff
@@ -139,6 +139,12 @@ class _MountDir(_MountEntry):
     def description(self):
         return str(self.local_dir.expanduser().absolute())
 
+    def _walk_and_prune(self, top_dir: Path) -> Generator[str, None, None]:
+        for root, dirs, files in os.walk(top_dir, topdown=True):
+            dirs[:] = [d for d in dirs if not self.ignore(Path(os.path.join(root, d)).relative_to(top_dir))]
```
mwaskom, line 147: 😵‍💫 was not obvious to me why this worked (maybe this is a common idiom I'm not familiar with). Could err on the side of over-description and explain in a comment?

mwaskom (approved):
Nice! Thanks for catching that this was an issue, makes a lot of sense.

mwaskom:
This should maybe get a changelog entry .. it shouldn't change what ends up on an Image but should speed up the App creation process for some people, so it's useful to know about.

ehdr:
I reworked this to account for the fact the inverted/negated ignore patterns (e.g. `ignore="!*.py"` or `~FilePatternMatcher("**/*.py")`) make it impossible to prune directories early, so now we don't try if they are present. Would appreciate a quick re-review since this logic feels a bit involved!

modal/file_pattern_matcher.py
```diff
@@ -54,6 +66,15 @@ class _CustomPatternMatcher(_AbstractPatternMatcher):
     def __init__(self, predicate: Callable[[Path], bool]):
         self._predicate = predicate
 
+    def can_prune_directories(self) -> bool:
+        """
+        Custom pattern matchers (like negated matchers) cannot safely prune directories.
```
mwaskom, line 71: I'm a little confused but this but is negation the only way we instantiate this class?

modal/file_pattern_matcher.py
```diff
@@ -173,6 +194,16 @@ def _matches(self, file_path: str) -> bool:
 
         return matched
 
+    def can_prune_directories(self) -> bool:
+        """
+        Returns True if this pattern matcher allows safe early directory pruning.
+
+        Directory pruning is safe when matching directories can be skipped entirely
+        without missing any files that should be included. This is only safe when
+        there are no exclusion patterns (patterns starting with '!').
```
mwaskom: 😵‍💫 I think that what we're locally calling an "exclusion" pattern ends up being an _inclusion_ pattern in context since we match files to **ignore**?

(That terminology predates your diff just noticing it for the first time).

modal/file_pattern_matcher.py
ehdr: AFAICT yes, that's currently the only way we use it. But since it takes a `predicate` that could do anything, in the general case we can never say that it's safe for sure, thus setting it to `False`. If we ever use this for another specific purpose we should probably refactor this to allow overriding `can_prune_directories` or something like that.

modal/file_pattern_matcher.py
ehdr: Yeah, I think that terminology is from the initial https://github.com/moby/patternmatcher code, that was just for matching (not specifically ignoring). But there's no need to leak that terminology into the comment here anyway, so changed it! 🙏

PR #3218, opened 2025-06-09
Make the batch limits more ergonomic.

danielshaar:
The server already supports these limits, let's avoid making 999 and 10min - 1ms the limits for users. 1k and 10min is simpler.


mwaskom (approved):
Looks like you need to update the expected error in the tests but otherwise LGTM 👍

PR #3221, opened 2025-06-09
Move blob uploads on async function calls to the client

CyrilSharma:
Makes the client perform blob uploads that previously the server was doing (on the async path).


modal/_utils/blob_utils.py
```diff
@@ -252,8 +253,11 @@ async def upload_to_s3_url(url):
                 content_md5_b64=upload_hashes.md5_base64,
             )
 
+        print("HELLO!")
```
mwaskom: :wave:

PR #3224, opened 2025-06-10
file_io: raise watch() exception in calling task

matslina:
## Describe your changes

Exceptions from `watch()` were previously raised in the `_consume_watch_output` task, which swallowed them and prevented user from handling.

## Changelog

- Updated `Sandbox.watch()` so that exceptions are raised in (and can be caught by) the calling task.


modal/file_io.py
```diff
@@ -144,11 +144,12 @@ class _FileIO(Generic[T]):
     _task_id: str = ""
     _file_descriptor: str = ""
     _client: _Client
-    _watch_output_buffer: list[Optional[bytes]] = []
+    _watch_output_buffer: list[Union[Optional[bytes],Exception]] = []
 
     def __init__(self, client: _Client, task_id: str) -> None:
         self._client = client
         self._task_id = task_id
+        self._watch_output_buffer = []
```
matslina, line 152: This was missing, which I believe caused events to be passed around completely at random.

saltzm: I don't know python well enough to have any idea why that would happen, but would be nice to have a better explanation of it before moving on I think

matslina: @mwaskom WDYT?

mwaskom: Yeah it looks like a bug to have default values for the class variables. This would have basically shared a list across all instances of the `FileIO`. I think basically all of the default assignment should move from the class scope into the constructor (although this is the nastiest one since it's mutable and only used mutably afaict).

matslina: alright. I'll look over these and follow up in a separate PR.

PR #3225, opened 2025-06-11
Support specified poetry version and update default spec in next image builder

mwaskom:
## Describe your changes

- Closes CLI-441

Also improved the tests a bit.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Added a `poetry_version` parameter to `modal.Image.poetry_install_from_file`, which supports installing a specific version of `poetry`. It's also possible to set `poetry_version=None` to skip the install step, i.e. when poetry is already available in the Image.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1227,14 +1227,15 @@ class _Image(_Object, type_prefix="im"):
         poetry_lockfile: Optional[str] = None,  # Path to lockfile. If not provided, uses poetry.lock in same directory.
         *,
         ignore_lockfile: bool = False,  # If set to True, do not use poetry.lock, even when present
-        # If set to True, use old installer. See https://github.com/python-poetry/poetry/issues/3336
-        old_installer: bool = False,
         force_build: bool = False,  # Ignore cached builds, similar to 'docker build --no-cache'
         # Selected optional dependency groups to install (See https://python-poetry.org/docs/cli/#install)
         with_: list[str] = [],
         # Selected optional dependency groups to exclude (See https://python-poetry.org/docs/cli/#install)
         without: list[str] = [],
         only: list[str] = [],  # Only install dependency groups specifed in this list.
+        poetry_version: Optional[str] = None,  # Version of poetry to install; default depends on image builder version
+        # If set to True, use old installer. See https://github.com/python-poetry/poetry/issues/3336
+        old_installer: bool = False,
         secrets: Sequence[_Secret] = [],
         gpu: GPU_T = None,
     ) -> "_Image":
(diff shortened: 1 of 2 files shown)


modal/image.py
```diff
@@ -1250,7 +1251,11 @@ def poetry_install_from_file(
         def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             context_files = {"/.pyproject.toml": os.path.expanduser(poetry_pyproject_toml)}
 
-            commands = ["FROM base", "RUN python -m pip install poetry~=1.7"]
+            if poetry_version is None:
+                poetry_spec = "~=1.7" if version <= "2024.10" else "~=2.0"
```
thomasjpfan: What do you think of having a central place to store the mapping between `ImageBuilderVersion` and poetry version at the top of the file?

```python
DEFAULT_POETRY_VERSIONS = {
    "2023.12": "~=1.7",
    "2024.04": "~=1.7",
    "2024.10": "~=1.7",
    "PREVIEW": "~=2.0",
}
```

A little more verbose, but makes the configuration more "data centered".

mwaskom: Yeah true, I'd moved a lot of this to https://github.com/modal-labs/modal-client/tree/main/modal/requirements so it was even more "static" and easier to see what varied with image builder version, but this doesn't exactly fit in "base-images" and i'm reluctant to make a new json just for it.

I guess we might be far enough away from whatever bug there is on <1.6 that we should simplify things and default to `pip install poetry` going forward? Micromanaging the versions feels a little bit paternalistic; if people are using poetry they probably know what they're doing.

modal/image.py
```diff
@@ -1227,14 +1227,15 @@ def poetry_install_from_file(
...
         ignore_lockfile: bool = False,  # If set to True, do not use poetry.lock, even when present
-        # If set to True, use old installer. See https://github.com/python-poetry/poetry/issues/3336
-        old_installer: bool = False,
         force_build: bool = False,  # Ignore cached builds, similar to 'docker build --no-cache'
         # Selected optional dependency groups to install (See https://python-poetry.org/docs/cli/#install)
         with_: list[str] = [],
         # Selected optional dependency groups to exclude (See https://python-poetry.org/docs/cli/#install)
         without: list[str] = [],
         only: list[str] = [],  # Only install dependency groups specifed in this list.
+        poetry_version: Optional[str] = None,  # Version of poetry to install; default depends on image builder version
```
mwaskom: I'm not sure if it makes sense to provide some way for users to request "My Image recipe will manage the poetry version; don't add a `pip install poetry` command when using this method. Could be nice for any weird edge cases where users have an Image without `pip`. Maybe that's not worth worrying about?

modal/image.py
thomasjpfan: > I guess we might be far enough away from whatever bug there is on <1.6 that we should simplify things and default to pip install poetry going forward? 

I'll be happy with `pip install poetry` and not micromanage anymore.

I'm guessing for image build stability, we'll only do this for the PREVIEW image build

modal/image.py
thomasjpfan: I can see setting the default to be `poetry_version="latest"` which installs the latest. Then `None` can mean "use `poetry` already in your image".

mwaskom: The annoying thing here is that for anyone on Image builder 2024.10 or earlier "latest" would actually mean `~=1.7` which is a lot harder to explain.

thomasjpfan: How about:

- `poetry_version=None` use "latest" for new builder and ~=1.7 for old ones
- `poetry_version="from_image"` for poetry in the image.

mwaskom: `from_image` is pretty clumsy IMO. But I thought about it more and I can live with `None/"latest"/{specific}` as the parameterization, even though it requires a little explanation in the docstring of the image builder version nuance.

thomasjpfan (approved):

PR #3226, opened 2025-06-11
First step towards removing UserCodeException wrapping [CLI-419]

freider:
* Don't use synchronicity.UserCodeException wrapping from *modal* code except when using `return_exceptions=True` (for backwards compatibility reasons)
* Adds `wrap_returned_exceptions: bool=True` flag to mapping operations that adds the legacy UserCodeException wrapping. Having this flag be true (the default at the moment) should yield a deprecation warning. Setting it to False will no longer wrap the exception and instead return the original user exception.

Seemingly unrelated (but necessary) change to client side retries:
* `.remote()` retries no longer deserializes (or blob-fetches) the output exception before retrying a failed input attempt. It's only deserializes once it needs to raise or return the payload (small optimization). This was the behavior for .map already and is arguably slightly better.

Unrelated change for consistency between .remote() and .map():
* `.map()` family client side retries will no longer attempt to retry TERMINATED (i.e. cancelled) inputs. Those were previously submitted using FunctionAttemptRetry but then errored with a GRPC error. They will now error with a `RemoteError("input was cancelled")` similar to to how `.remote()` when an input is cancelled, without attempting a retry.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

* Added a `wrap_returned_exceptions: bool = True` argument to the `Function.map` family of functions. Setting this to False will instead return the original exception type, which will be the default behavior in a future version of Modal.

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -342,7 +342,7 @@ async def _map_invocation(
         except Exception as e:
             if return_exceptions:
                 if wrap_returned_exceptions:
-                    # Prior to client 1.0.5 there was a bug where return_exceptions would wrap
+                    # Prior to client 1.0.4 there was a bug where return_exceptions would wrap
                     # any returned exceptions in a synchronicity.UserCodeException. This adds
                     # deprecated non-breaking compatibility bandaid for migrating away from that:
                     output = modal.exception.UserCodeException(e)


modal/_functions.py
```diff
@@ -284,7 +297,7 @@ async def run_function(self) -> Any:
         while True:
             try:
                 return await self._get_single_output(ctx.input_jwt)
-            except (UserCodeException, FunctionTimeoutError) as exc:
+            except (_WrappedUserException, FunctionTimeoutError) as exc:
```
freider: @rculbertson Do you have some context here if we actually need to look at the exception type here or if we have enough information in the `FunctionGetOutputsItem` to determine if we should retry?

It seems to me like we roughly have the current mapping here:
* `api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE` -> `InternalFailure` -> Retry immediately
* `api_pb2.GenericResult.GENERIC_STATUS_TIMEOUT` -> `FunctionTimeoutError` -> Retry after delay
Then there is the remaining case of:
`result.status != api_pb2.GenericResult.GENERIC_STATUS_SUCCESS`
This branches depending on if the output item has an attached `.data` attribute and can get deserialized.
* Can be deserialized to an exception -> `UserCodeException` (now `_WrappedUserException`) -> retry after delay
* Otherwise - don't retry at all - usually because of `api_pb2.GenericResult.GENERIC_STATUS_TERMINATED` (cancellation) but could also be deserialization errors etc.

freider: On the (sync) map side we seem to have a simpler logic - if status is `api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE` we retry immediately (same as here), otherwise we retry after a delay? 🤔 Not sure how this interacts with cancellations (those would cancel the full map anyways, so maybe it doesn't matter?)

freider: Just tested this and cancellations (`GENERIC_STATUS_TERMINATED`) are actually attempted to be retried in case of maps, but then receive a `FAILED_PRECONDITION` when trying to send those inputs

modal/parallel_map.py
```diff
@@ -742,6 +773,7 @@ async def handle_get_outputs_response(
             return _OutputType.FAILED_COMPLETION
 
         self.state = _MapItemState.WAITING_TO_RETRY
+        print("RETRYING", item.idx, item.retry_count, item.result.status)
```
mwaskom: 👀

modal/parallel_map.py
```diff
@@ -339,7 +341,13 @@ async def fetch_output(item: api_pb2.FunctionGetOutputsItem) -> tuple[int, Any]:
             output = await _process_result(item.result, item.data_format, client.stub, client)
         except Exception as e:
             if return_exceptions:
-                output = e
+                if wrap_returned_exceptions:
+                    # Prior to client 1.0.5 there was a bug where return_exceptions would wrap
```
mwaskom: Nit: this change will go out in 1.0.4

mwaskom (approved):
👍 Handling of the `wrap_returned_exceptions` flag looks right (pending later addition of the deprecation warning). I'm less familiar with the nuances of client-side retries though.

modal/parallel_map.py
```diff
@@ -411,6 +419,7 @@ async def _map_helper(
     kwargs={},  # any extra keyword arguments for the function
     order_outputs: bool = True,  # return outputs in order
     return_exceptions: bool = False,  # propagate exceptions (False) or aggregate them in the results list (True)
+    wrap_returned_exceptions: bool = True,
```
mwaskom, line 422: Should we explain what this is for?

mwaskom: Oh maybe we do that below?

modal/_functions.py
rculbertson: We don't need the actual exception type to determine the retry, the status code is enough, so you optimization makes sense!

> Just tested this and cancellations (GENERIC_STATUS_TERMINATED) are actually attempted to be retried in case of maps, but then receive a FAILED_PRECONDITION when trying to send those inputs

!!! Good find!

PR #3227, opened 2025-06-11
Fix `Cls.with_options` stacking

mwaskom:
## Describe your changes

- Fixes CLI-422

Pretty dumb bug on my end originally ... not sure how I deluded myself into thinking that `dataclass.replace` ignored default field values.

This is pretty finicky. Because we expose the individual members of the `Resources` proto in the public API, we need to special-case how we do that merge so that you can do e.g. `C.with_options(cpu=16).with_options(gpu="A100")`. It also relies on default values in the dataclass having untruthiness so we can tell that they are "unset". Perhaps it would be better to use an Unset singleton or something.

I decided *not* to merge the Volume / Secret lists. Feels like you could expect it either way, and not doing it is simpler?

Also while working on this I remembered another gotcha which is that you might have something like

```python
@app.cls(gpu="T4")
class C: ...

C.with_options(gpu=None)
```

and expect that the instance would not have a GPU. That's tricky because we would need to distinguish between (a) None and "unset" in the `.with_options` argument list, and (b) default-but-set vs. unset values in the protobuf message fields. I think it's technically possible but would be rather complicated so for now I'm just going to document that as not working.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- When `Cls.with_options` is called multiple times on the same `Cls` instance, the overrides will now be merged. For example, the following will use an H100 GPU with 8 CPUS:
  ```python
  Model.with_options(cpu=8, gpu="A100").with_options(gpu="H100")
  ```

### DIFF
--- a/test/cls_test.py
+++ b/test/cls_test.py
@@ -256,7 +256,7 @@ def test_service_options_defaults_untruthiness():
     # In the future we may change the implementation to use an "Unset" sentinel default, in
     # which case we wouldn't need this assertion.
     default_options = _ServiceOptions()
-    for value in dataclasses.asdict(default_options).values():
+    for value in dataclasses.asdict(default_options).values():  # type: ignore  # synchronicity type stubs
         assert not value
 
 


modal/cls.py
```diff
@@ -75,6 +75,7 @@ def _get_class_constructor_signature(user_cls: type) -> inspect.Signature:
 
 @dataclasses.dataclass()
 class _ServiceOptions:
+    # Note that default values should always be "untruthy" so we can detect when they are not set
```
thomasjpfan, line 78: Can we add a test to ensure this is always the case?

modal/cls.py
```diff
@@ -664,15 +684,30 @@ def with_options(
         container_idle_timeout: Optional[int] = None,  # Now called `scaledown_window`
         allow_concurrent_inputs: Optional[int] = None,  # See `.with_concurrency`
     ) -> "_Cls":
-        """Create an instance of the Cls with configuration options overridden with new values.
+        """Override the static Function configuration at runtime.
+
+        This method will return a new instance of the cls that will autoscale independently of the
+        original instance. Note that options cannot be "unset" with this method (i.e., if a GPU
```
thomasjpfan, line 690: Not being able to unset is a bit unfortunate. It does not feel like a common enough use case tho.

modal/cls.py
```diff
@@ -664,15 +684,30 @@ def with_options(
...
+        You can use this method after looking up the Cls from a deployed App or if you have a
+        direct reference to a Cls from another Function or local entrypoint on its App:
+
         ```python notest
         Model = modal.Cls.from_name("my_app", "Model")
         ModelUsingGPU = Model.with_options(gpu="A100")
         ModelUsingGPU().generate.remote(42)  # will run with an A100 GPU
         ```
+
+        The method can be called multiple times to "stack" updates, but note that container
```
thomasjpfan: Can the note about "Volume and secrets will not be merged" come after the example? At first glance, I thought that the example would showcase the behavior of merging.

modal/cls.py
mwaskom: 👍 good idea!

modal/cls.py
mwaskom: Yeah. If a sufficiently compelling need arises it's something we probably _could_ add in the future, but it would be pretty complicated to do, and I think the mindset of "you can dynamically add parameterization but not take it away" is reasonably easy to understand.

thomasjpfan (approved):

PR #3228, opened 2025-06-11
Added verbose to sandboxes in client

akchang13:
Added verbose to sandboxes in client.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- When `verbose=True` is set in `Sandbox.create()`, execs and file system operations are logged in the sandbox logs. For example, 
```
p = sb.exec("python", "-c", "print('hello')")
```
logs
image

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -105,6 +105,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         proxy: Optional[_Proxy] = None,
         _experimental_scheduler_placement: Optional[SchedulerPlacement] = None,
         enable_snapshot: bool = False,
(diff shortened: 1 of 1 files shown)


modal/sandbox.py
```diff
@@ -259,6 +261,7 @@ async def create(
             SchedulerPlacement
         ] = None,  # Experimental controls over fine-grained scheduling (alpha).
         client: Optional[_Client] = None,
+        verbose: bool = False,
```
pawalt: This should go above the experimental options, and it should have a comment noting how it's used.

akchang13: Done

PR #3230, opened 2025-06-11
Adds uv_pip_install

thomasjpfan:
## Describe your changes

Simplified version of https://github.com/modal-labs/modal-client/pull/3216

This PR copies the signature of `pip_install` to make it very easy to migrate to `uv_pip_install`.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Adds `uv_pip_install` to install packages with `uv`, which can improve build times by 50% compared to `pip_install`. This feature is in beta as we improve it based on feedback. During the beta period, updating `modal` may cause image rebuilds.

```python
image = modal.Image.debian_slim().uv_pip_install("torch==2.7.1", "numpy")
```


modal/image.py
```diff
@@ -1221,6 +1221,76 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             gpu_config=parse_gpu_config(gpu),
         )
 
+    def uv_pip_install(
```
mwaskom, line 1225: WDYT about just naming this `uv_install`? The `uv pip install` nomenclature feels more oriented towards CLI namespacing, which is somewhat less relevant here, and it would save modal users some typing (and reduce potential confusion if you're reading the code quickly).

Not something I feel super strongly about though if you think it's important to have image methods literally match the underlying command.

modal/image.py
```diff
@@ -1221,6 +1221,76 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+        elif not _validate_packages(pkgs):
+            raise InvalidError(
+                "Package list for `Image.uv_pip_install` cannot contain other arguments;"
+                " try the `extra_options` parameter instead."
+            )
+
+        def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
+            commands = ["FROM base"]
+            if uv_version is None:
+                commands.append("COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv")
```
mwaskom: Just flagging this as we should be consistent with the pending decision about `uv sync` about where we put the `uv` binary (i.e. maybe it should go in `/.uv` as mentioned in that PR.)

modal/image.py
```diff
@@ -1221,6 +1221,76 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+            else:
+                commands.append(f"COPY --from=ghcr.io/astral-sh/uv:{uv_version} /uv /usr/local/bin/uv")
+
+            # TODO: Assumes python is on the PATH and uv is installing into the first python in the path
+            uv_pip_args = ["--python $(which python)"]
```
mwaskom: I'm a little worried about the portability of this shell substitution ... will this always work?

Just for some context ... the syntax we use for Image building has a somewhat higher bar for stability than other code, since if we change it in a subsequent release, it would force image rebuilds on all users of that method (even if the method was previously working correctly for them). So I'm a little paranoid about getting wedged with an approach that is broken for some slice of our users.

We could also discuss documenting `uv` support as "beta" or some similar terminology that gives us more leeway for potential followup changes to the syntax, although surely not all users read the documentation closely and it will cause pain for someone :)

mwaskom (commented):
Whoops had a couple pending review comments stuck in the buffer, sorry!

modal/image.py
thomasjpfan: I think I'll still go with `uv_pip_install` because I am confident that `uv` will keep the `uv pip` namespace.

If `uv` introduces `uv install`, then Modal's `uv_install` would be confusing. (This was the motivation for having the `uv pip` namespace in the first place).

mwaskom: Fair enough — personally I think it's more about "This is the `uv` namespace where the CLI spellings are pip-ish" which is less relevant for a wrapper API.

modal/image.py
thomasjpfan: I updated the shell substitution to use backticks to also support `/bin/sh` and documented the assumptions this makes:

1. Python is on the PATH
2. The shell supports backticks
3. `which` is a command on the PATH.

2 and 3 are satisfied by most Unix-like systems. 1 has the same assumption as `pip_install` (The only gotcha is when Python is set in the entrypoint)

I'm okay with marking it as beta and not guarantee image stability for a few months just to see if we run into any issues.

modal/image.py
```diff
@@ -1222,6 +1222,115 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+                    local_path = os.path.expanduser(req)
+
+                    return {
+                        "basename": basename,
+                        "local_path": local_path,
+                        "context_path": f"./{basename}",
+                        "dest_path": f"{UV_ROOT}/{basename}",
+                    }
+
+                requirement_paths = [_generate_paths(req) for req in requirements]
```
mwaskom: Given that `requirements="requirements.txt"` would be an easy mistake and iteration over strings is valid (but would cause a confusing error) should we proactively catch that and fail nicely? (I guess I'd also be more open to `requirements` having a `str | sequence[str]` type since it's not grammatically confused 🤓).

modal/image.py
```diff
@@ -1222,6 +1222,115 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
...
+                requirements_cli = " ".join(f"--requirements {req['dest_path']}" for req in requirement_paths)
+                uv_pip_args.append(requirements_cli)
+
+                commands.extend([f"COPY {req['context_path']} {req['dest_path']}" for req in requirement_paths])
+                context_files.update({req["context_path"]: req["local_path"] for req in requirement_paths})
+
+            uv_pip_args.extend(shlex.quote(p) for p in sorted(pkgs))
+            uv_pip_args_joined = " ".join(uv_pip_args)
+
+            commands.append(f"RUN /.uv/uv pip install {uv_pip_args_joined}")
```
mwaskom: Should we be passing `--compile-bytecode`?

modal/image.py
mwaskom: oh wait i think that happens in `_normalize_items` actually?

mwaskom: Oh I guess that happens in `_normalize_items` actually? Nvm.

mwaskom (approved):

PR #3233, opened 2025-06-12
Add Warning for Ignored Modal Shell Flags

joyliu-q:
## Describe your changes

Follow-up to #3232. Discussion on https://linear.app/modal-labs/issue/WRK-1048/add-secret-to-modal-shell#comment-8e13cc64.

We should warn for ignored flags in `modal shell`.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -597,6 +597,20 @@ def shell(
     modal shell hello_world.py -c 'uv pip list' > env.txt
     ```
     """
+
+    if container_or_function and len(sys.argv) > 3:
+        supported_opts = {"--cmd"}
+        ignored_or_unrecognized = [
+            arg
+            for arg in sys.argv[3:]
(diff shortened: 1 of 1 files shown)


modal/cli/run.py
```diff
@@ -597,6 +597,20 @@ def shell(
     modal shell hello_world.py -c 'uv pip list' > env.txt
     ```
     """
+
+    if container_or_function and len(sys.argv) > 3:
+        supported_opts = {"--cmd"}
```
mwaskom, line 602: I think this is maybe insufficiently robust since you can also use `-c` for the cmd argument. I'm pretty sure that `--pty` (and `--no-pty`, not that there's much reason to pass that) and `--env`/`-e` are usable with or without a REF too.

It might be better to do this check w/r/t the Python variable values, rather than the command line arguments, since it avoids some of the aliasing complexity. Which will probably be a little annoying :(

mwaskom (changes requested):

PR #3236, opened 2025-06-13
Cancel tasks if asyncio.wait is cancelled

freider:
Unlike `asyncio.wait_for`, `asyncio.wait` doesn't cancel the underlying tasks if the wait is cancelled, we need to clean up after a potential cancellation to prevent downstream ClientClosed raises during event loop shut down

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,39 +6,23 @@ This changelog documents user-facing updates (features, enhancements, fixes, and
 
 <!-- NEW CONTENT GENERATED BELOW. PLEASE PRESERVE THIS COMMENT. -->
 
-#### 1.0.4.dev12 (2025-06-13)
+### 1.0.4 (2025-06-13)
 
-- When `Cls.with_options` is called multiple times on the same `Cls` instance, the overrides will now be merged. For example, the following will use an H100 GPU with 8 CPUS:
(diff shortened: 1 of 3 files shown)


modal/parallel_map.py
```diff
@@ -278,17 +278,19 @@ async def get_all_outputs():
...
             map_done_task = asyncio.create_task(map_done_event.wait())
-            done, pending = await asyncio.wait([get_response_task, map_done_task], return_when=FIRST_COMPLETED)
-            if get_response_task in done:
-                map_done_task.cancel()
-                response = get_response_task.result()
-            else:
-                assert map_done_event.is_set()
-                # map is done, cancel the pending call
+            try:
+                done, pending = await asyncio.wait([get_response_task, map_done_task], return_when=FIRST_COMPLETED)
```
mwaskom, line 282: 😵‍💫 relationship between return value names and task names 😁

modal/parallel_map.py
```diff
@@ -278,17 +278,19 @@ async def get_all_outputs():
...
+                if get_response_task in done:
+                    map_done_task.cancel()
+                    response = get_response_task.result()
+                else:
+                    assert map_done_event.is_set()
```
mwaskom, line 287: Should this ever send an assertion error back to the user?

PR #3238, opened 2025-06-13
[WRK-1002] Added sandbox user in api.proto and sandbox.py

akchang13:
Added sandbox user in api.proto and sandbox.py (for supporting non-root users in sandboxes using Sandbox.create())

WRK-1002

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- When `Sanbox.create()` is called with with a specific `user` parameter, the Sandbox process will run with the specified UID and GID


pawalt:
btw we won't merge this - using this to keep the work handy

pawalt:
we're not going to do this for now

PR #3241, opened 2025-06-16
Relax grpclib constraint

aek-gubra:
Based on our internal testing the client still works fine with the next version of `grpclib`


aek-gubra:
> Originally the reasoning was that as `grpclib` was pure Python it had fewer installation headaches across our userbase. It's possible that situation has changed. We've discussed a little bit internally whether we want to re-evaluate this but haven't made any firm decisions.

I can confirm that we've had issues with `grpclib` but I did not dig too deep into the issue.

PR #3242, opened 2025-06-16
Deprecate vestigial Object.clone method

mwaskom:
Deprecates the `modal.Object.clone` method which is basically dead code at this point. We used to call it from `modal.Cls.with_options` but now that uses `modal.Object._initialize_from_other` directly (`.clone` is basically "`_initialize_from_other` after asserting that other is hydrated").

This is technically public API, although it's hidden from the docs. AFAIK it's not used in any examples or documentation and I don't think it's really relevant to any user workflows. But I think it's good practice post 1.0 to treat all not-explicitly-private API as public and deprecate it gracefully. The cost here is low since this is just a small standalone method.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_object.py
+++ b/modal/_object.py
@@ -10,6 +10,7 @@ from typing_extensions import Self
 
 from ._resolver import Resolver
 from ._utils.async_utils import aclosing
+from ._utils.deprecation import deprecation_warning
 from .client import _Client
 from .config import config, logger
 from .exception import ExecutionError, InvalidError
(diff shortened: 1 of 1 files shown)


modal/image.py
```diff
@@ -416,7 +416,6 @@ def _initialize_from_empty(self):
         self.force_build = False
 
     def _initialize_from_other(self, other: "_Image"):
-        # used by .clone()
```
mwaskom, line 419: AFAICT `Image.clone()` is not used anywhere...

mwaskom:
@ehdr has convinced me that it's preferable for us to default to graceful deprecations even when the chances of actually breaking someone are very low. Could imagine making the decision the other way if an abrupt break let us remove a lot complexity, but that's not the case here as it's just a standalone simple method.

PR #3243, opened 2025-06-16
Relax grpclib constraint to allow 0.4.8

mwaskom:
Given that grpclib is critical to nearly all client behaviors and that they've had breaking chances in patch releases before ([e.g.](https://grpclib.readthedocs.io/en/latest/changelog/#id2)) we'll err on the side of caution and keep a strict upper bound pin on known-to-work versions.


mwaskom:
Gonna merge so we can start exercising this in integration tests and see if anything pops up @prbot approve

PR #3244, opened 2025-06-16
Include base class modal.parameters in class schema [CLI-450]

freider:
Some users making use of class inheritance ran into issues when migrating from explicit constructors to `modal.parameter` based parameterization, since base class parameters weren't included in the schema of the class.
By switching to `get_type_hints` instead of `.__dict__["__annotations__"]` we automatically get base class resolution.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

* When an `@app.cls()`-decorated class inherits from other classes and those classes have `modal.parameter()` definitions, the base class parameters will be included in the parameter set for the modal Cls.

### DIFF
--- a/test/cls_test.py
+++ b/test/cls_test.py
@@ -1307,3 +1307,36 @@ def test_concurrent_decorator_stacked_with_method_decorator():
             @modal.concurrent(max_inputs=10)
             def method(self):
                 pass
+
+
+def test_parameter_inheritance(client):
+    app = modal.App("inherit-params")
+
+    class Base:
+        a: int = modal.parameter()  # parameter in base class
+
+    @app.cls(serialized=True)
+    class ConcatenatingParams(Base):
+        b: str = modal.parameter()  # add additional parameter
+
+    @app.cls(serialized=True)
+    class RepeatingParams(Base):
+        # In versions prior to ~1.0.5, base class parameters were not
+        # included, so subclasses would always have to repeat parameters
+        # from the base class.
+        # We allow this as long as the definitions are the same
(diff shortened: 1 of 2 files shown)


test/cls_test.py
```diff
@@ -1307,3 +1307,36 @@ class UsesMethodAndConcurrentDecorators:
...
+        b: str = modal.parameter()
+
+    @app.cls(serialized=True)
+    class ChangingParameterDefinitions(Base):
+        # change type of base class parameter - not allowed
+        # this is similar to how type checkers normally don't let you do this
+        a: str = modal.parameter()
+
+    with app.run(client=client):
+        ConcatenatingParams(a=10, b="hello").update_autoscaler()  # stupid way of hydrating the class
```
mwaskom: Bruh

freider:
@prbot approve

test/cls_test.py
```diff
@@ -1307,3 +1307,37 @@ class UsesMethodAndConcurrentDecorators:
...
+        # In versions prior to ~1.0.5, base class parameters were not
+        # included, so subclasses would always have to repeat parameters
+        # from the base class.
+        # We allow this as long as the definitions are the same
+        a: int = modal.parameter()  # redefine base class parameter
+        b: str = modal.parameter()
+
+    @app.cls(serialized=True)
+    class ChangingParameterDefinitions(Base):
+        # change type of base class parameter - not allowed
```
mwaskom: Confused by the comment — aren't we allowing the type override?

freider: Ah, old comment before I changed my mind 🤦

PR #3248, opened 2025-06-16
Improve repr for Dict

thomasjpfan:
## Describe your changes

Related to https://linear.app/modal-labs/issue/CLI-428/python-client-api-for-object-management

For object management, I think it's important to have a nice repr. With this PR, `Dict` gets an improved repr along with a `name` attribute.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Improves `repr` to `Dict` and adds `name` to `Dict`. You can now access the Dict's name:

```python
d = Dict.from_name("custom-dict")
print(d.name)
```

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -158,7 +158,7 @@ class _Dict(_Object, type_prefix="di"):
         if environment_name is None:
             obj_repr = f"Dict.from_name('{name}')"
         else:
-            obj_repr = f"Dict.fromt_name('{name}', environment_name='{environment_name}')"
+            obj_repr = f"Dict.from_name('{name}', environment_name='{environment_name}')"
 
         obj = _Dict._from_loader(_load, obj_repr, is_another_app=True, hydrate_lazily=True)
         return obj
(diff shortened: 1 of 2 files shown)


modal/dict.py
```diff
@@ -155,7 +165,15 @@ async def _load(self: _Dict, resolver: Resolver, existing_object_id: Optional[st
             logger.debug(f"Created dict with id {response.dict_id}")
             self._hydrate(response.dict_id, resolver.client, None)
 
-        return _Dict._from_loader(_load, "Dict()", is_another_app=True, hydrate_lazily=True)
+        obj = _Dict._from_loader(_load, "Dict()", is_another_app=True, hydrate_lazily=True)
```
freider: `_Object` has a repr implementation that returns `self._repr`, and the second argument to `_from_loader` takes this attribute in order to set good reprs, so it's better to just pass `f"Dict.from_name('{name}')"` as the second argument here instead of overriding __repr__

freider (commented):
There is another way, see inline comment

mwaskom:
Does this also need to include the `environment_name`?

mwaskom:
Superseded by #3432, thanks for getting this work started!

PR #3251, opened 2025-06-16
Add Sandbox.tags API for listed sandboxes

ekzhang:
## Describe your changes

Resolves CLI-457

This depends on https://github.com/modal-labs/modal/pull/23678 to be merged first. Also it doesn't work for Sandbox.from_id yet, I saw this comment from @freider but I think I'm getting in too deep now to fix it :')

```
# TODO: send SandboxInfo as metadata to _new_hydrated?
```

## Changelog

- Added `Sandbox.tags` property to objects returned by `Sandbox.list()`, which is a `dict[str, str]` that contains tags for each of the listed Sandboxes.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -74,6 +74,7 @@ class _Sandbox(_Object, type_prefix="sb"):
     """
 
     _result: Optional[api_pb2.GenericResult]
+    _tags: dict[str, str]
     _stdout: _StreamReader[str]
     _stderr: _StreamReader[str]
     _stdin: _StreamWriter
@@ -764,6 +765,15 @@ class _Sandbox(_Object, type_prefix="sb"):
             yield event
 
     @property
+    def tags(self) -> dict[str, str]:
+        """Tags (key-value pairs) associated with the Sandbox.
+
+        Tags can be used to filter results in `Sandbox.list`, and they are updated with the
+        `set_tags()` function.
+        """
+        return dict(self._tags)
+
+    @property
     def stdout(self) -> _StreamReader[str]:
         """
         [`StreamReader`](https://modal.com/docs/reference/modal.io_streams#modalio_streamsstreamreader) for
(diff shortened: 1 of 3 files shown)


mwaskom:
It does feel like this should work with a sandbox you look up via `.from_id` not just `.list`.

ekzhang:
I did take a look at that when writing this PR! I think we could do that but it would be automatically through resolving this existing TODO comment in the code:

```
# TODO: send SandboxInfo as metadata to _new_hydrated?
```

Not super familiar with object hydration. Though if this PR doesn't make the current situation worse (there's already planned future TODO to fix populating the info in `_new_hydrated`, which is called from `from_id`) then maybe it's fine to just add it to list() initially?

freider:
Resolving the TODO is more about the code structure, making sure that any hydration path will need to provide the correct metdata and not having to have each constructor set the state fields individually. That would be resolved by moving the 
```py
obj._result = sandbox_info.task_info.result  # TODO: send SandboxInfo as metadata to _new_hydrated?
obj._tags = {tag.tag_name: tag.tag_value for tag in sandbox_info.tags}
```
part  into `def hydrate_metadata(...)` and passing the SandboxInfo as the third argument to `_new_hydrated`. 

However, to do that for `from_id()` (and .create) as well, we'd have to add the required `SandboxInfo` metadata to the responses for those rpcs (`SandboxCreate` and `SandboxWait` in this case), which is probably a good idea. 

It also seems like we have a `SandboxHandleMetadata` proto which is returned by AppGetObjects, and is completely unused by the client, so there is plenty of cleanup to do in this area :(

modal/sandbox.py
```diff
@@ -763,6 +764,15 @@ async def watch(
...
             yield event
 
+    @property
+    def tags(self) -> dict[str, str]:
+        """Tags (key-value pairs) associated with the Sandbox.
+
+        Tags can be used to filter results in `Sandbox.list`, and they are updated with the
+        `set_tags()` function.
+        """
+        return dict(self._tags)
```
freider, line 774: There is no default value for _tags, so you'd get an AttributeError when trying ot do this on a Sandbox that wasn't constructed via the `.list` method.

Unless we add the relevant metadata to the other hydration paths we'd have to keep some state and warn to the user why this field isn't available for those Sandboxes (adding another "color" of sandboxes) which is quite ugly

freider (commented):
I'm not a huge fan of having this be a selectively working on a subset of hydrated sandboxes, so I'd prefer if we could add the needed metadata to the other hydration paths as well (`.from_id` and `.create`) before merging this new functionality.

It's not so much about my old TODO, and more about having the backend returning the SandboxInfo on those paths.

Unrelated - but wouldn't it make sense to be able to set the tags of a sandbox at creation time as well (instead set_tags()) to make this more atomic?

PR #3252, opened 2025-06-17
[prototype] modal torchrun wrapper

pawalt:
this prototypes a new decorator, `@modal.experimental.torchrun` for multinode jobs that allows the user to specify in a more modal-esque way the function they want to run. we handle the torchrunning behind the scenes.

this is 100% ai coded so the implementation is a bit jank, but i really like the interface here! it seems like a natural way to extend the modal way of doing things to multi-node.

i tested this out by migrating our [benchmark code](https://github.com/modal-labs/multinode-training-guide/blob/14beb9bd50ee659faeb2edc473210988c74074a1/benchmark/modal_train.py) to this script. it looks like this now:

```python
"""
Improved multi-node benchmark using @modal.experimental.torchrun decorator.

This version eliminates the need for a separate train.py file and manual torchrun setup.
All the benchmark logic is contained within a single decorated function.
"""

import os
import modal
import modal.experimental

cuda_version = "12.4.0"  # should be no greater than host CUDA version
flavor = "devel"  # includes full CUDA toolkit
operating_sys = "ubuntu22.04"
tag = f"{cuda_version}-{flavor}-{operating_sys}"

# Benchmark configuration
N_NODES = 2
N_PROC_PER_NODE = 8

# Benchmark parameters (from EleutherAI cookbook)
WARMUP_ITERS, TRIALS = 5, 50
# These emulate the payload which will become a M * N * 4-sized tensor
N = 500000
M = 2000

# Create the image with necessary dependencies
image = (
    modal.Image.from_registry(f"nvidia/cuda:{tag}", add_python="3.12")
    .apt_install(
        "libibverbs-dev",
        "libibverbs1",
    )
    .pip_install(
        "torch==2.6.0",
        "numpy",
        "importlib-metadata",
    )
)

with image.imports():
    import torch
    import torch.distributed as dist

app = modal.App("multinode-benchmark-improved")


@app.function(
    gpu="H100:8",
    cloud="oci",
    image=image,
    timeout=60 * 10,  # 10 minutes
)
@modal.experimental.clustered(size=N_NODES, rdma=True)
@modal.experimental.torchrun(nproc_per_node=N_PROC_PER_NODE)
def run_benchmark():
    """
    Run a simple benchmark that measures all-reduce bandwidth across nodes.

    This function runs in "data mode" - the torchrun orchestration is handled
    automatically by the decorator. No separate train.py file needed!
    """
    # Get local rank from environment
    local_rank = int(os.environ["LOCAL_RANK"])
    global_rank = int(os.environ["RANK"])
    world_size = int(os.environ["WORLD_SIZE"])

    # Initialize the process group
    torch.cuda.set_device(local_rank)
    dist.init_process_group(
        backend="nccl", device_id=torch.device(f"cuda:{local_rank}")
    )

    if global_rank == 0:
        print(
            f"Starting benchmark with {world_size} total processes across {N_NODES} nodes"
        )
        print(f"Each node has {N_PROC_PER_NODE} processes")
        print(f"Total payload size: {M * N * 4 / 1e9:.2f} GB")

    # Helper functions for synchronization
    def sync_all():
        torch.cuda.synchronize()
        dist.barrier()

    def timed_allreduce(mat, start_event, end_event, warmup_iters, iters):
        sync_all()

        # Warmup iterations
        for _ in range(warmup_iters):
            dist.all_reduce(mat)
        sync_all()

        # Timed iterations
        start_event.record()
        for _ in range(iters):
            dist.all_reduce(mat)
        end_event.record()

        sync_all()

        # Calculate bandwidth
        duration = start_event.elapsed_time(end_event) / 1000  # Convert to seconds
        avg_duration = duration / iters

        size = M * N * 4  # 4 bytes per float32
        # Following the same math as NVIDIA/nccl-tests
        algbw = torch.tensor([size / avg_duration]).cuda(local_rank)

        # Calculate mean across all ranks
        dist.reduce(algbw, dst=0, op=dist.ReduceOp.SUM)
        algbw /= world_size

        return algbw.item()

    # Create the test tensor
    mat = torch.rand(N, M, dtype=torch.float32).cuda(local_rank)

    # Create CUDA events for timing
    start_event = torch.cuda.Event(enable_timing=True)
    end_event = torch.cuda.Event(enable_timing=True)

    # Run the benchmark
    algbw = timed_allreduce(
        mat, start_event, end_event, warmup_iters=WARMUP_ITERS, iters=TRIALS
    )

    # Calculate bus bandwidth
    # The 2*(n-1)/n correction factor for all-reduce is explained here:
    # https://github.com/NVIDIA/nccl-tests/blob/master/doc/PERFORMANCE.md#allreduce
    busbw = algbw * (2 * (world_size - 1) / world_size)

    # Report results from rank 0
    if global_rank == 0:
        print("\n" + "=" * 60)
        print(f"Benchmark Results:")
        print(f"Payload size: {M * N * 4 / 1e9:.2f} GB")
        print(f"Number of trials: {TRIALS}")
        print(
            f"Total ranks: {world_size} ({N_NODES} nodes × {N_PROC_PER_NODE} processes/node)"
        )
        print("-" * 60)
        print(
            f"Algorithm bandwidth: {algbw / 1e9:.3f} GB/s ({algbw * 8 / 1e9:.1f} Gb/s)"
        )
        print(
            f"Bus bandwidth:       {busbw / 1e9:.3f} GB/s ({busbw * 8 / 1e9:.1f} Gb/s)"
        )
        print("=" * 60 + "\n")

    # Clean up
    sync_all()
    dist.destroy_process_group()

    if global_rank == 0:
        print("Benchmark completed successfully!")


@app.local_entrypoint()
def main():
    """
    Entry point for running the benchmark.

    Notice how simple this is now - just call .remote() on the decorated function!
    All the torchrun orchestration happens automatically.
    """
    print("Starting multi-node benchmark...")
    print(f"Configuration: {N_NODES} nodes, {N_PROC_PER_NODE} GPUs per node")
    print(f"This will launch {N_NODES * N_PROC_PER_NODE} total processes")

    run_benchmark.remote()

    print("Benchmark submission completed!")


if __name__ == "__main__":
    main()
```

Shockingly, it works!

```
(modal) ➜  benchmark git:(3cefdab) ✗ MODAL_SYNC_ENTRYPOINT=1 modal run improved_bench.py
✓ Initialized. View run at https://modal.com/apps/modal-labs/peyton-dev/ap-6DjFo0FPcFIStDondn8O16
✓ Created objects.
├── 🔨 Created mount /home/ec2-user/projects/multinode-training-guide/benchmark/improved_bench.py
├── 🔨 Created mount /home/ec2-user/projects/client/modal, /home/ec2-user/projects/client/modal_proto, 
│   /home/ec2-user/projects/client/modal_version, 
│   /home/ec2-user/projects/modal/venv/lib64/python3.11/site-packages/synchronicity
└── 🔨 Created function run_benchmark.
Starting multi-node benchmark...
Configuration: 2 nodes, 8 GPUs per node
This will launch 16 total processes
Acknowledged cluster creation request for modal.Function fu-j5QAcIGElHtTCFsf9TTbOn. Expected p95 wait time: 3.0 minutes.
Made reservation of 2 GPUType.H100 workers for clustered modal.Function fu-j5QAcIGElHtTCFsf9TTbOn. Waiting for workers to be ready...

.......

Container rank 1: Executing torchrun
Command: /usr/local/bin/python -m torch.distributed.run --nnodes=2 --nproc-per-node=8 --node-rank=1 --master-addr=fdaa:5137:3ebf:b64:80f:84a:a192:7732 --master-port=29500 /tmp/tmpdk7e4u6s_torchrun_wrapper.py
Container rank 0: Executing torchrun
Command: /usr/local/bin/python -m torch.distributed.run --nnodes=2 --nproc-per-node=8 --node-rank=0 --master-addr=fdaa:5137:3ebf:b64:80f:84a:a192:7732 --master-port=29500 /tmp/tmpyonshjlx_torchrun_wrapper.py
Made reservation of 2 GPUType.H100 workers for clustered modal.Function fu-j5QAcIGElHtTCFsf9TTbOn. Waiting for workers to be ready...
Starting benchmark with 16 total processes across 2 nodes
Each node has 8 processes
Total payload size: 4.00 GB

============================================================
Benchmark Results:
Payload size: 4.00 GB
Number of trials: 50
Total ranks: 16 (2 nodes × 8 processes/node)
------------------------------------------------------------
Algorithm bandwidth: 245.881 GB/s (1967.0 Gb/s)
Bus bandwidth:       461.027 GB/s (3688.2 Gb/s)
============================================================

Benchmark completed successfully!
```

### DIFF
--- a/modal/_partial_function.py
+++ b/modal/_partial_function.py
@@ -922,3 +922,174 @@ def _clustered(size: int, broadcast: bool = True, rdma: bool = False):
         return pf
 
     return wrapper
+
+
+# NOTE: torchrun is currently exposed through modal.experimental, not the top-level namespace
+def _torchrun(
+    _warn_parentheses_missing=None,
+    *,
+    nproc_per_node: int = 1,
+    master_port: int = 29500,
+) -> Callable[
+    [Union[_PartialFunction[P, ReturnType, ReturnType], Callable[P, ReturnType]]],
+    _PartialFunction[P, ReturnType, ReturnType],
+]:
+    """Decorator that enables torchrun orchestration for distributed training.
+
+    When used with @modal.experimental.clustered, this decorator automatically handles
+    the torchrun orchestration, allowing you to write your training logic directly
+    in the decorated function without needing a separate script file.
+
+    The function will be called in two modes:
+    - Orchestration mode: Sets up and launches torchrun
+    - Data mode: Executes the actual training logic (called by torchrun)
+
+    Parameters:
+    nproc_per_node: int = 1
+        Number of processes to launch per node
+    master_port: int = 29500
+        Port for inter-process communication
+
+    Example:
+    ```python
+    @app.function(gpu="H100:8")
+    @modal.experimental.clustered(4, rdma=True)
+    @modal.experimental.torchrun(nproc_per_node=8)
+    def train():
+        # This code runs in data mode
+        import torch.distributed as dist
+        dist.init_process_group(backend="nccl")
+
+        # Your training logic here
+        ...
+    ```
+    """
+    if _warn_parentheses_missing is not None:
+        raise InvalidError(
+            "Positional arguments are not allowed. Did you forget parentheses? Suggestion: `@modal.experimental.torchrun()`."
+        )
+
+    def wrapper(
+        obj: Union[_PartialFunction[P, ReturnType, ReturnType], Callable[P, ReturnType]],
+    ) -> _PartialFunction[P, ReturnType, ReturnType]:
+        import functools
+        import os
+
+        # Get the original function
+        if isinstance(obj, _PartialFunction):
+            if obj.raw_f is None:
+                raise InvalidError("Cannot apply @modal.experimental.torchrun to a class-level decorator")
+            original_func = obj.raw_f
+        else:
+            original_func = obj
+
+        @functools.wraps(original_func)
+        def torchrun_wrapper(*args, **kwargs):
+            # Check if we're in data mode (torchrun has set these env vars)
+            if all(var in os.environ for var in ["RANK", "LOCAL_RANK", "WORLD_SIZE"]):
+                # We're in data mode - run the original function
+                return original_func(*args, **kwargs)
+            else:
+                # We're in orchestration mode
+                import sys
+                import subprocess
+                import tempfile
+                import json
+                import modal
+
+                # Get cluster info
+                cluster_info = modal.experimental.get_cluster_info()
+                container_rank = cluster_info.rank
+                main_ip_addr = cluster_info.container_ips[0]
+                n_nodes = len(cluster_info.container_ips)
+
+                # Create a Python script that torchrun will execute
+                # This script needs to reconstruct the function call
+                script_template = """
+import sys
+import os
+import json
+
+# Set up the environment to find the user's code
+sys.path.insert(0, '{working_dir}')
+
+# Import the function
+from {module_name} import {func_name}
+
+# Load arguments if provided
+args_file = sys.argv[1] if len(sys.argv) > 1 else None
+if args_file and os.path.exists(args_file):
+    with open(args_file, 'r') as f:
+        call_info = json.load(f)
+        args = call_info.get('args', [])
+        kwargs = call_info.get('kwargs', {{}})
+else:
+    args = []
+    kwargs = {{}}
+
+# Call the function
+{func_name}.local(*args, **kwargs)
+"""
+
+                # Get function details
+                module_name = original_func.__module__
+                func_name = original_func.__name__
+                working_dir = os.getcwd()
+
+                # Format the script
+                script_content = script_template.format(
+                    working_dir=working_dir, module_name=module_name, func_name=func_name
+                )
+
+                # Write the script to a temporary file
+                with tempfile.NamedTemporaryFile(mode="w", suffix="_torchrun_wrapper.py", delete=False) as f:
+                    f.write(script_content)
+                    script_path = f.name
+
+                # Save arguments to a temporary file (if any)
+                args_file = None
+                if args or kwargs:
+                    # Convert args/kwargs to JSON-serializable format
+                    # Note: This is a simplification and may not work for all argument types
+                    with tempfile.NamedTemporaryFile(mode="w", suffix="_args.json", delete=False) as f:
+                        json.dump({"args": list(args), "kwargs": kwargs}, f)
+                        args_file = f.name
+
+                try:
+                    # Build the torchrun command
+                    cmd = [
+                        sys.executable,
+                        "-m",
+                        "torch.distributed.run",
+                        f"--nnodes={n_nodes}",
+                        f"--nproc-per-node={nproc_per_node}",
+                        f"--node-rank={container_rank}",
+                        f"--master-addr={main_ip_addr}",
+                        f"--master-port={master_port}",
+                        script_path,
+                    ]
+
+                    if args_file:
+                        cmd.append(args_file)
+
+                    print(f"Container rank {container_rank}: Executing torchrun")
+                    print(f"Command: {' '.join(cmd)}")
+
+                    # Run torchrun
+                    result = subprocess.run(cmd, check=True)
+
+                finally:
+                    # Clean up temporary files
+                    os.unlink(script_path)
+                    if args_file:
+                        os.unlink(args_file)
+
+        # Wrap the function
+        if isinstance(obj, _PartialFunction):
+            # Replace the raw function with our wrapper
+            obj.raw_f = torchrun_wrapper
+            return obj
+        else:
+            return _PartialFunction(torchrun_wrapper, _PartialFunctionFlags(0), _PartialFunctionParams())
+
+    return wrapper
--- a/modal/experimental/__init__.py
+++ b/modal/experimental/__init__.py
@@ -9,7 +9,7 @@ from modal_proto import api_pb2
 from .._clustered_functions import ClusterInfo, get_cluster_info as _get_cluster_info
 from .._functions import _Function
 from .._object import _get_environment_name
-from .._partial_function import _clustered
+from .._partial_function import _clustered, _torchrun
 from .._runtime.container_io_manager import _ContainerIOManager
 from .._utils.async_utils import synchronize_api, synchronizer
 from .._utils.deprecation import deprecation_warning
@@ -46,6 +46,7 @@ def get_cluster_info() -> ClusterInfo:
 
 
 clustered = synchronize_api(_clustered, target_module=__name__)
+torchrun = synchronize_api(_torchrun, target_module=__name__)
 
 
 @dataclass


modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+    *,
+    nproc_per_node: int = 1,
+    master_port: int = 29500,
+) -> Callable[
+    [Union[_PartialFunction[P, ReturnType, ReturnType], Callable[P, ReturnType]]],
+    _PartialFunction[P, ReturnType, ReturnType],
+]:
+    """Decorator that enables torchrun orchestration for distributed training.
+
+    When used with @modal.experimental.clustered, this decorator automatically handles
```
mwaskom, line 939: j/w but would this ever be useful for single-node training?

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+    the torchrun orchestration, allowing you to write your training logic directly
+    in the decorated function without needing a separate script file.
+
+    The function will be called in two modes:
+    - Orchestration mode: Sets up and launches torchrun
+    - Data mode: Executes the actual training logic (called by torchrun)
```
mwaskom, line 945: AFAICT this is not super-useful user-facing information right? This is part of the complexity we're trying to prevent them from having to think about? (i.e. I _think_ this is a comment about _this function_ not the user's function it's wrapping, but it's a little ambiguous).

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+
+    Parameters:
+    nproc_per_node: int = 1
+        Number of processes to launch per node
+    master_port: int = 29500
+        Port for inter-process communication
```
mwaskom, line 951: Our convention would be to document these as comments in the signature (which I don't love, but at least it does cut down on some duplicated information about types / defaults).

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+            else:
+                # We're in orchestration mode
+                import sys
+                import subprocess
+                import tempfile
+                import json
+                import modal
+
+                # Get cluster info
+                cluster_info = modal.experimental.get_cluster_info()
```
mwaskom, line 1001: Once nice thing about this is we wouldn't necessarily need to expose this stuff in a public API (though we probably still would unless we think we'll cover all viable multi-node training frameworks with integrations I guess).

modal/_partial_function.py
mwaskom: If not, another option would be combining this somehow with the `@modal.clustered` decorator. Since I think we want to leave the door open for using `@modal.clustered` open for multinode inference or other non-training usecases I don't think we necessarily want to commit to something like `@modal.clustered(framework="torchrun")` though, so maybe that's a bad idea.

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+    nproc_per_node: int = 1
+        Number of processes to launch per node
+    master_port: int = 29500
+        Port for inter-process communication
+
+    Example:
+    ```python
+    @app.function(gpu="H100:8")
+    @modal.experimental.clustered(4, rdma=True)
+    @modal.experimental.torchrun(nproc_per_node=8)
```
mwaskom, line 957: The `nproc_per_node` is redundant with the GPU specification in the function decorator right? I wonder if we could implicitly make that connection behind the scenes.

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
 
     return wrapper
+
+
+# NOTE: torchrun is currently exposed through modal.experimental, not the top-level namespace
+def _torchrun(
+    _warn_parentheses_missing=None,
+    *,
+    nproc_per_node: int = 1,
+    master_port: int = 29500,
```
mwaskom, line 932: Would users ever need to change this?

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+                    print(f"Command: {' '.join(cmd)}")
+
+                    # Run torchrun
+                    result = subprocess.run(cmd, check=True)
+
+                finally:
+                    # Clean up temporary files
+                    os.unlink(script_path)
+                    if args_file:
+                        os.unlink(args_file)
```
mwaskom, line 1085: Is this needed since it will be in the ephemeral FS anyway?

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+                    script_path = f.name
+
+                # Save arguments to a temporary file (if any)
+                args_file = None
+                if args or kwargs:
+                    # Convert args/kwargs to JSON-serializable format
+                    # Note: This is a simplification and may not work for all argument types
+                    with tempfile.NamedTemporaryFile(mode="w", suffix="_args.json", delete=False) as f:
+                        json.dump({"args": list(args), "kwargs": kwargs}, f)
+                        args_file = f.name
```
mwaskom, line 1056: j/w but do we need to do this thing where we pass the extra args / kwargs into the subprocess via the filesystem? Could we not pass them through in the cli args?

mwaskom: Also if you `modal run` against a function using this decorator, do we still successfully propagate its parameters out to the `modal run` CLI? That would be neat!

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+                finally:
+                    # Clean up temporary files
+                    os.unlink(script_path)
+                    if args_file:
+                        os.unlink(args_file)
+
+        # Wrap the function
+        if isinstance(obj, _PartialFunction):
+            # Replace the raw function with our wrapper
+            obj.raw_f = torchrun_wrapper
```
mwaskom, line 1090: This part is a little unconventional right? I wonder if it's going to wedge us on a specific implementation for our modal decorators.

modal/_partial_function.py
```diff
@@ -922,3 +922,174 @@ def wrapper(
...
+                    ]
+
+                    if args_file:
+                        cmd.append(args_file)
+
+                    print(f"Container rank {container_rank}: Executing torchrun")
+                    print(f"Command: {' '.join(cmd)}")
+
+                    # Run torchrun
+                    result = subprocess.run(cmd, check=True)
```
mwaskom, line 1079: Checking my understanding about the control flow here, say you have an App script with a `train` function using this decorator. Then when you do `modal run app.py::train` it will

- Spins up a container for the `train` Function
- Execute this wrapper function, which will be in orchestration mode, which
  - Writes a script to the ephemeral FS and
  - Submits the training script to torchrun in a subprocess, which
    - Makes a local call to the decorated function, which
      - Mostly delegates to the underlying user function

whew! One question is, what are the tracebacks gonna look like when you have an error?

mwaskom (commented):
This is really cool, asked a bunch of questions mostly for my own understanding!

modal/_partial_function.py
thomasjpfan: `torchrun` can be used with single node training with multiple gpus. I think it's still common practice for `DistributedDataParallel`.

mwaskom:
Mentioning something we talked about in person, but the subprocess script won’t have a hydrated App which could cause some surprising failures if eg your training function tries to make remote calls to other functions defined on the App.

thomasjpfan:
I wanted to avoid manually subprocessing out, but it does not work:

https://github.com/modal-labs/modal-examples/blob/thomasjpfan/fabric_single_node/misc/train_fabric_does_not_work.py#L23-L27

(Although above still subprocess out, but it's through a Python SDK)

`torchrun` also has their own `elastic_launch` that has similar semantics: https://github.com/pytorch/pytorch/blob/53057fc16af2d381a61fe891b52ac8731ee9dfae/torch/distributed/launcher/api.py#L187-L191

PR #3253, opened 2025-06-17
Deprecate namespace parameter

ehdr:
## Describe your changes

Part of a larger project (CLI-335) to remove `DeploymentNamespace` from RPCs.

---

## Compatibility checklist

</details>

## Changelog

- Deprecated the `namespace` parameter on `Secret`, `Function`, `Cls`, `Dict`, `Queue`, `Volume`, `NetworkFileSystem`, and `deploy_app`.

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1245,7 +1245,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         cls,
         app_name: str,
         name: str,
-        namespace=_ARGUMENT_NOT_PASSED,
+        namespace=_ARGUMENT_NOT_PASSED,  # mdmd:line-hidden
         environment_name: Optional[str] = None,
     ):
         # internal function lookup implementation that allows lookup of class "service functions"
@@ -1282,7 +1282,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         app_name: str,
         name: str,
         *,
-        namespace=_ARGUMENT_NOT_PASSED,
+        namespace=_ARGUMENT_NOT_PASSED,  # mdmd:line-hidden
         environment_name: Optional[str] = None,
     ) -> "_Function":
         """Reference a Function from a deployed App by its name.
@@ -1313,7 +1313,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
     async def lookup(
         app_name: str,
         name: str,
-        namespace=_ARGUMENT_NOT_PASSED,
+        namespace=_ARGUMENT_NOT_PASSED,  # mdmd:line-hidden
         client: Optional[_Client] = None,
         environment_name: Optional[str] = None,
     ) -> "_Function":
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1255,7 +1255,6 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
             request = api_pb2.FunctionGetRequest(
                 app_name=app_name,
                 object_tag=name,
-                namespace=api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
                 environment_name=_get_environment_name(environment_name, resolver) or "",
             )
             try:
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -40,7 +40,7 @@ from ._utils.async_utils import (
     synchronizer,
     warn_if_generator_is_not_consumed,
 )
-from ._utils.deprecation import deprecation_warning
+from ._utils.deprecation import _ARGUMENT_NOT_PASSED, deprecation_warning, warn_if_passing_namespace
 from ._utils.function_utils import (
     ATTEMPT_TIMEOUT_GRACE_PERIOD,
     OUTPUTS_TIMEOUT,
@@ -1241,7 +1241,13 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         await self.update_autoscaler(min_containers=warm_pool_size)
 
     @classmethod
-    def _from_name(cls, app_name: str, name: str, namespace, environment_name: Optional[str]):
+    def _from_name(
+        cls,
+        app_name: str,
+        name: str,
+        namespace=_ARGUMENT_NOT_PASSED,
+        environment_name: Optional[str] = None,
+    ):
         # internal function lookup implementation that allows lookup of class "service functions"
         # in addition to non-class functions
         async def _load_remote(self: _Function, resolver: Resolver, existing_object_id: Optional[str]):
@@ -1249,7 +1255,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
             request = api_pb2.FunctionGetRequest(
                 app_name=app_name,
                 object_tag=name,
-                namespace=namespace,
+                namespace=namespace,  # type: ignore
                 environment_name=_get_environment_name(environment_name, resolver) or "",
             )
             try:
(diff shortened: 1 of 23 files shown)


modal/_utils/deprecation.py
```diff
@@ -122,3 +124,40 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
 
     return wrapper
+
+
+# Utilities for deprecating the namespace parameter across Modal resources
+class _ArgumentNotPassedType:
+    """Sentinel object to detect when an argument is explicitly passed vs using default."""
+
+    def __repr__(self) -> str:
+        return f"{__name__}._ARGUMENT_NOT_PASSED"
```
mwaskom: Do we need this? Wouldn't `None` work fine?

modal/_utils/deprecation.py
```diff
@@ -122,3 +124,40 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+
+
+_ARGUMENT_NOT_PASSED = _ArgumentNotPassedType()
+
+
+def warn_if_passing_namespace(
+    namespace,
+    resource_name: str = "Modal resource",
```
mwaskom: ??

modal/_utils/deprecation.py
```diff
@@ -122,3 +124,40 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+
+    Args:
+        namespace: The namespace parameter value (may be sentinel or actual value)
+        resource_name: Name of the resource type for the warning message
+
+    Returns:
+        The appropriate namespace value to use
+    """
+    if namespace is _ARGUMENT_NOT_PASSED:
+        return api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE
```
mwaskom: Do we need to actually be returning something here or should we just be warning when the user is passing something to `namespace=` in the Python interface? AFAIK for the object types you're touching here you can't actually accomplish anything productive by passing `namespace=`, so I don't think we actually need go out of our way to _use_ the value they passed.

The situation is indeed a little different for `modal.Image`, although even there I don't think there's a well-defined use for users passing some `namespace`.

modal/_functions.py
```diff
@@ -1292,13 +1298,14 @@ def from_name(
                 f"instance.{method_name}.remote(...)\n",
             )
 
-        return cls._from_name(app_name, name, namespace, environment_name)
+        warn_if_passing_namespace(namespace, "modal.Function")
+        return cls._from_name(app_name, name, api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE, environment_name)
```
mwaskom: Could we push the namespace default down into `_from_name`? I think generally speaking we should only need it at the level of the actual `*Get{OrCreate}Request` constructors.

modal/_functions.py
```diff
@@ -1292,13 +1298,14 @@ def from_name(
                 f"instance.{method_name}.remote(...)\n",
             )
 
-        return cls._from_name(app_name, name, namespace, environment_name)
+        warn_if_passing_namespace(namespace, "modal.Function")
```
mwaskom: Nit, but based on the way we interpolate the "resource" should we pass `modal.Function.from_name` here? Alternatively might be able to write `warn_if_passing_namepsace` as a decorator, like `warn_on_renamed_autoscaler_settings` and get the name from the wrapped function.

modal/_functions.py
ehdr: Yeah good call, done!

modal/_functions.py
ehdr: I updated the names now. I might take a stab at making a more generic deprecation decorator (like `renamed_parameter` but `parameter_will_be_removed` or something), but suggest doing it in a separate PR.

modal/_utils/deprecation.py
```diff
@@ -122,3 +122,23 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+) -> None:
+    """Issue deprecation warning for namespace parameter if non-None value is passed.
+
+    Args:
+        namespace: The namespace parameter value (may be None or actual value)
+        resource_name: Name of the resource type for the warning message
+    """
+    if namespace is not None:
+        deprecation_warning(
+            (2025, 6, 17),
```
mwaskom: Not super important but this is technically a bit out of date at this point.

modal/runner.py
```diff
@@ -102,13 +103,12 @@ async def _init_local_app_new(
 async def _init_local_app_from_name(
     client: _Client,
     name: str,
-    namespace: Any,
+    namespace: Any = None,
```
mwaskom: I think this function is fully private so could just remove the parameter here.

modal_docs/mdmd/mdmd.py
```diff
@@ -5,27 +5,30 @@
...
 
 
-def format_docstring(docstring: str):
+def format_docstring(docstring: Optional[str]) -> str:
     if docstring is None:
         docstring = ""
     else:
         docstring = inspect.cleandoc(docstring)
 
+    docstring = "\n".join(l for l in docstring.split("\n") if "mdmd:line-hidden" not in l)
```
mwaskom, line 21: What would it look like to use `mdmd:line-hidden` inside a docstring itself?

test/cls_test.py
```diff
@@ -1346,3 +1346,21 @@ class ChangingParameterDefinitions(Base):
...
+
+    # Test that from_name without namespace parameter doesn't warn about namespace
+    import warnings
+
+    with warnings.catch_warnings(record=True) as record:
+        warnings.simplefilter("always")
+        Cls.from_name("test-app", "test-cls")
+    # Filter out any unrelated warnings
+    namespace_warnings = [w for w in record if "namespace" in str(w.message).lower()]
+    assert len(namespace_warnings) == 0
```
mwaskom, line 1364: Could you be using `with pytest.warns(DeprecationError, match="namespace")` here?

modal/_utils/deprecation.py
```diff
@@ -122,3 +122,23 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+            f"The `namespace` parameter for `{resource_name}` is deprecated and will be"
+            " removed in a future release. It is no longer needed, so can be removed"
+            " from your code.",
+            pending=True,
```
mwaskom: Why `pending`?

ehdr: No good reason, will remove!

ehdr: `pending=True` is only used in [one single location](https://github.com/modal-labs/modal-client/blob/4f83474c06256deefa5c36f5c5ea2a9c2918ea81/modal/_functions.py#L1448) - could it even be removed entirely? Not super clear to me what the difference between a pending deprecation and a deprecation is?

mwaskom: The difference is that any unhandled `DeprecationError` warnings will cause integration tests to fail, while `PendingDeprecationError` won't. So it's used when we are putting in some change that requires requires the following sequence:

- Add new spelling in the client, pending deprecate the old spelling
- Update integration tests etc. to use the new spelling
- Drop pending from the warning

It's potentially less necessary now that we're not releasing every client commit (i.e. we could live with an intermediate state where there are "two ways to do it" because users won't be exposed to it.

modal_docs/mdmd/mdmd.py
ehdr: Like this:
image
image

But I get your point - we might as well just delete the line at that point... Should I remove it?

modal/_utils/deprecation.py
ehdr: Makes sense, thanks!

test/cls_test.py
mwaskom: nvm dumb comment!

PR #3254, opened 2025-06-17
Add read-only to volumes

thomasjpfan:
## Describe your changes

Closes https://linear.app/modal-labs/issue/CLI-273/support-read-only-volumes

This PR adds `Volume.with_options` to be consistent with `Cls.with_options`. Note that if a `Volume` is mounted as read only any writes to the volume would raise an OSError:

```python
import modal

app = modal.App("read_only")
volume = modal.Volume.from_name("polars-volume")


@app.function(volumes={"/mnt/polars": volume.with_options(read_only=True)})
def main():
    with open("/mnt/polars/modal_here.txt", "w") as f:
        f.write("this is some text")
```

The above will raise `OSError: [Errno 30] Read-only file system: '/mnt/polars/modal_here.txt'`.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Add `read_only` to Volume, which makes the volume read only.
 
```python
volume = modal.Volume.from_name("my-data-volume", read_only=True)
```

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -142,9 +142,16 @@ class _Volume(_Object, type_prefix="vo"):
     def read_only(self) -> bool:
         return self._read_only
 
-    def with_read_only(self) -> "_Volume":
+    def with_options(
+        self,
+        *,
+        read_only: Optional[bool] = None,  # Configure Volume to be read-only.
+    ) -> "_Volume":
+        """Configure Volume to mount as read-only"""
+        if read_only is None:
+            return self
         volume = deepcopy(self)
-        volume._read_only = True
+        volume._read_only = read_only
         return volume
 
     async def _get_lock(self):
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -137,7 +137,7 @@ class _Volume(_Object, type_prefix="vo"):
     _metadata: "typing.Optional[api_pb2.VolumeMetadata]"
     _read_only: bool = False
 
-    def as_read_only(self) -> "_Volume":
+    def read_only(self) -> "_Volume":
         """Configure Volume to mount as read-only.
 
         **Example**
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -155,9 +155,6 @@ class _Volume(_Object, type_prefix="vo"):
 
         The Volume is mounted as a read-only volume in a function. Any file system write operation into the
         mounted volume will result in an error.
-
-        Volume methods such as `batch_upload`, `copy_files`, or `remove_file` will still work for
-        writing data into the volume.
         """
 
         async def _load(new_volume: _Volume, resolver: Resolver, existing_object_id: Optional[str]):
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -164,7 +164,7 @@ class _Volume(_Object, type_prefix="vo"):
             new_volume._initialize_from_other(self)
             new_volume._read_only = True
 
-        obj = _Volume._from_loader(_load, "Volume.read_only()", hydrate_lazily=True, deps=lambda: [self])
+        obj = _Volume._from_loader(_load, "Volume()", hydrate_lazily=True, deps=lambda: [self])
         return obj
 
     async def _get_lock(self):
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -10,7 +10,6 @@ import re
 import time
 import typing
 from collections.abc import AsyncGenerator, AsyncIterator, Generator, Sequence
-from copy import deepcopy
 from dataclasses import dataclass
 from io import BytesIO
 from pathlib import Path, PurePosixPath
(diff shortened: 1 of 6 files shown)


modal/volume.py
```diff
@@ -135,6 +136,23 @@ def g():
...
+    _read_only: bool = False
+
+    @property
+    def read_only(self) -> bool:
+        return self._read_only
+
+    def with_options(
+        self,
+        *,
+        read_only: Optional[bool] = None,  # Configure Volume to be read-only.
```
mwaskom: Nit: we should be super clear in the documentation that the Volume itself isn't `read_only` (you can write to it from elsewhere), but it will be read-only as mounted to the function.

See next comment but this being a little bit nuanced is one reason to prefer a standalone method over it being one parameter in a "group of options" method; more natural space for verbose documentation.

modal/volume.py
```diff
@@ -135,6 +136,23 @@ def g():
...
+    ) -> "_Volume":
```
mwaskom: See some previous discussion of the API options [here](https://modal-com.slack.com/archives/C07S93UH343/p1743598212777729?thread_ts=1743517057.511459&cid=C07S93UH343) (took me a while to dig this up sorry!) where we went back-and-forth on the potential for additional options that would make sense to group.

One tricky thing here is that other options we eventually come up with (although we've basically blanked so far) may not strictly-speaking be properties of the Volume *mount* the way `read_only` is. Naming this overly generically could conceivably cause problems in the future.

Also, this is pretty nuanced, but `Cls.with_options` actually creates a distinct server-side instance, which is somewhat different from what's happening here.

OTOH I generally speaking like the approach here as being more extensible in the future. But it's tricky, which is part of why we've been holding off so far.

modal/volume.py
```diff
@@ -135,6 +136,23 @@ def g():
...
+        """Configure Volume to mount as read-only"""
+        if read_only is None:
+            return self
+        volume = deepcopy(self)
```
mwaskom: We don't use `deepcopy()` on other objects and frankly I'm not sure if there's a reason for that or not but I could imagine it potentially causing problems! There is existing infrastructure for effectively cloning an object that we use in `Cls.with_options` (note, not the public `.clone`, which we're likely to remove) and I'd probably advocate for using it here despite being a little bit awkward.

modal/volume.py
```diff
@@ -135,6 +136,23 @@ def g():
 
     _lock: Optional[asyncio.Lock] = None
     _metadata: "typing.Optional[api_pb2.VolumeMetadata]"
+    _read_only: bool = False
+
+    @property
+    def read_only(self) -> bool:
+        return self._read_only
```
mwaskom: Are users likely to need to read this property?

modal/volume.py
```diff
@@ -135,6 +136,23 @@ def g():
 
     _lock: Optional[asyncio.Lock] = None
     _metadata: "typing.Optional[api_pb2.VolumeMetadata]"
+    _read_only: bool = False
```
mwaskom, line 138: It's definitely within precedent to store random attributes on the class like this but it's sort of messy. We can change it later though, e.g. if we come up with other options that we want to bundle together.

mwaskom:
Note don't forget to update the changelog when we converge on a spelling.

mwaskom:
Another discussion question here: should it be possible to use `.batch_upload` (or, generally speaking, write methods) on a Volume with `read_only` enabled? On the one hand the read-onliness is a property of the mount point, not "the Volume", so that should be fine, but I could also imagine it being confusing to users. The risk is probably low since in most cases where you want read only volumes (e.g., if you're executing untrusted code while a Volume is mounted), there wouldn't be any way for the untrusted execution to use the Volume API. But whichever way we go, we should document it clearly.

modal/volume.py
```diff
@@ -138,21 +137,35 @@ def g():
...
+
+        Volume methods such as `batch_upload`, `copy_files`, or `remove_file` will still work for
+        writing data into the volume.
+        """
+
+        async def _load(new_volume: _Volume, resolver: Resolver, existing_object_id: Optional[str]):
+            new_volume._initialize_from_other(self)
+            new_volume._read_only = True
+
+        obj = _Volume._from_loader(_load, "Volume.read_only()", hydrate_lazily=True, deps=lambda: [self])
```
mwaskom: Is `Volume.read_only()` a "valid" repr?

thomasjpfan: I'll switch it back to `"Volume()"` to not tie this PR with defining a good repr.

In the long term, I'll like `Volume.from_name("my-volume").read_only()` to keep the semantics for "You can copy and paste this to create the instance".

mwaskom: 👍 that plan makes sense

test/volume_test.py
```diff
@@ -50,6 +58,31 @@ def test_volume_bad_paths():
...
+    with servicer.intercept():
+        with app.run(client=client):
+            with pytest.raises(InvalidError):
+                app.function()(batch_upload)
+
+            with pytest.raises(InvalidError):
+                app.function()(remove_file)
+
+            with pytest.raises(InvalidError):
+                app.function()(copy_files)
```
mwaskom: Not totally obvious to me why you're wrapping these in a modal function; can't you exercise the API methods outside of a modal context?

thomasjpfan: I wanted a context where the volume was hydrated. I updated to just hydrate the volume directly with the client: [`14b858d` (#3254)](https://github.com/modal-labs/modal-client/pull/3254/commits/14b858d012318a3be5e42c58fe68f75c608ba084)

PR #3255, opened 2025-06-17
Make env part optional in cell magic

ekzhang:
Notebooks have environments on their state now, so this is well-defined:

```
%modal from eric-gleb-jun-17-temp import f
```


gleb-modal:
lgtm

PR #3256, opened 2025-06-17
Experimental Modal Flash

thecodingwizard:
Experimental client implementation of Modal Flash.

### DIFF
--- a/modal/experimental/__init__.py
+++ b/modal/experimental/__init__.py
@@ -269,6 +269,12 @@ FlashManager = synchronize_api(_FlashManager)
 
 @synchronizer.create_blocking
 async def flash_forward(port: int) -> _FlashManager:
+    """
+    Forward a port to the Modal Flash service, exposing that port as a stable web endpoint.
+
+    This is a highly experimental method that can break or be removed at any time without warning.
+    Do not use this method unless explicitly instructed to do so by Modal support.
+    """
     client = await _Client.from_env()
 
     manager = _FlashManager(client, port)


modal/experimental/__init__.py
```diff
@@ -209,3 +212,69 @@ async def update_autoscaler(
...
+                print(f"[Modal Flash] Heartbeat failed: {e}")
+
+            try:
+                await asyncio.sleep(1)
+            except asyncio.CancelledError:
+                print("[Modal Flash] Shutting down...")
+                break
+
+    async def stop(self):
+        self.alive = False
```
gongy: not used yet right

modal/experimental/__init__.py
```diff
@@ -209,3 +212,69 @@ async def update_autoscaler(
...
+                    api_pb2.FlashContainerRegisterRequest(
+                        priority=self.priority,
+                        weight=self.weight,
+                        host=host,
+                        port=port,
+                    ),
+                    timeout=1,
+                )
+                if first_registration:
+                    print(f"[Modal Flash] Listening at {resp.url}")
```
gongy: tbh `logger.info` more appropriate

gongy (approved):

modal/experimental/__init__.py
thecodingwizard: oops, dead code I forgot to remove. thanks for catching

modal/experimental/__init__.py
thecodingwizard: updated (and also cleaned up some imports)

thecodingwizard: hmm I had to use `logger.warning` for it to show up (by default I think loglevel is warning).

I think `logger.warning` > `print` > `logger.info` but let me know if you think differently. Again this is just for v0; medium/long-term we should figure out something better.

CleanShot 2025-06-18 at 00 36 24@2x

modal/experimental/__init__.py
```diff
@@ -209,3 +213,64 @@ async def update_autoscaler(
...
+            api_pb2.FlashContainerDeregisterRequest(),
+        )
+        await self.tunnel_manager.__aexit__(*sys.exc_info())
+
+
+FlashManager = synchronize_api(_FlashManager)
+
+
+@synchronizer.create_blocking
+async def flash_forward(port: int) -> _FlashManager:
```
mwaskom, line 271: Something I've been trying to do with all the `modal.experimental` functions I've added is to set clear expectations in the docstring about what's experimental. Is this functionality we know we're going to add, but we're not sure about the client API? Or is it experimental _functionality_ and could be dropped entirely, etc. What is the timeline for graduating out of experimental (if known), etc. Basically we want users to understand if this is something they should rely on or not.

PR #3260, opened 2025-06-18
Client: Handle Input Plane Region Metadata

bokeefe123:
- SVC-601

- Include input-plane-region in requests

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -198,7 +198,7 @@ async def retry_transient_errors(
     total_timeout: Optional[float] = None,  # timeout for the entire function call
     attempt_timeout_floor=2.0,  # always have at least this much timeout (only for total_timeout)
     retry_warning_message: Optional[RetryWarningMessage] = None,
-    base_metadata: list[tuple[str, str]] = [],
+    metadata: list[tuple[str, str]] = [],
 ) -> ResponseType:
     """Retry on transient gRPC failures with back-off until max_retries is reached.
     If max_retries is None, retry forever."""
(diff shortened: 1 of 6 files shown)


modal/_utils/grpc_utils.py
```diff
@@ -198,6 +198,7 @@ async def retry_transient_errors(
     total_timeout: Optional[float] = None,  # timeout for the entire function call
     attempt_timeout_floor=2.0,  # always have at least this much timeout (only for total_timeout)
     retry_warning_message: Optional[RetryWarningMessage] = None,
+    input_plane_region: Optional[str] = None,
```
thecodingwizard: Could we make this more general, like `metadata: list[tuple[str, str]] = []` or something?

This `retry_transient_errors` function is used in a lot of different places, so it would be nice if the arguments aren't specific to the input plane.

modal_proto/api.proto
```diff
@@ -710,6 +710,7 @@ message ClassGetResponse {
   repeated Warning server_warnings = 3;
 }
 
+//TODO:(ben-okeefe) handle hydrating region
```
```suggestion
```

test/input_plane_test.py
bokeefe123, line 1: Check input_plane_url as well (empty string)

modal/_utils/grpc_utils.py
```diff
@@ -198,6 +198,7 @@ async def retry_transient_errors(
     total_timeout: Optional[float] = None,  # timeout for the entire function call
     attempt_timeout_floor=2.0,  # always have at least this much timeout (only for total_timeout)
     retry_warning_message: Optional[RetryWarningMessage] = None,
+    base_metadata: list[tuple[str, str]] = [],
```
thecodingwizard:
```suggestion
    metadata: list[tuple[str, str]] = [],
```

really minor nit, but what do you think about calling it `metadata` instead of `base_metadata`? From the caller's perspective I think they might not understand what `base_` means (`base_metadata` seems to suggest that there should also be a `metadata` parameter somewhere).

thecodingwizard (approved):
nice! (you might need `Optional[str]` for python 3.9)

PR #3262, opened 2025-06-19
Draft: Clean tracebacks (on Python 3.11+)

freider:
## Describe your changes

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -8,13 +8,10 @@ from asyncio import Future
 from collections.abc import Hashable
 from typing import TYPE_CHECKING, Optional
 
-from grpclib import GRPCError, Status
-
 from modal_proto import api_pb2
 
 from ._utils.async_utils import TaskContext
 from .client import _Client
-from .exception import NotFoundError
 
 if TYPE_CHECKING:
     from rich.tree import Tree
@@ -144,12 +141,8 @@ class Resolver:
                 # Load the object itself
                 if not obj._load:
                     raise Exception(f"Object {obj} has no loader function")
-                try:
-                    await obj._load(obj, self, existing_object_id)
-                except GRPCError as exc:
-                    if exc.status == Status.NOT_FOUND:
-                        raise NotFoundError(exc.message)
-                    raise
+
+                await obj._load(obj, self, existing_object_id)
 
                 # Check that the id of functions didn't change
                 # Persisted refs are ignored because their life cycle is managed independently.
@@ -170,7 +163,6 @@ class Resolver:
             if deduplication_key is not None:
                 self._deduplication_cache[deduplication_key] = cached_future
 
-        # TODO(elias): print original exception/trace rather than the Resolver-internal trace
         return await cached_future
 
     def objects(self) -> list["modal._object._Object"]:
--- a/modal/_resolver.py
+++ b/modal/_resolver.py
@@ -8,6 +8,7 @@ from asyncio import Future
 from collections.abc import Hashable
 from typing import TYPE_CHECKING, Optional
 
+from modal._traceback import suppress_tb_frames
 from modal_proto import api_pb2
 
 from ._utils.async_utils import TaskContext
@@ -162,8 +163,9 @@ class Resolver:
             self._local_uuid_to_future[obj.local_uuid] = cached_future
             if deduplication_key is not None:
                 self._deduplication_cache[deduplication_key] = cached_future
-
-        return await cached_future
+        with suppress_tb_frames(2):
+            # skip current frame + `loader()` closure frame from above
+            return await cached_future
 
     def objects(self) -> list["modal._object._Object"]:
         unique_objects: dict[str, "modal._object._Object"] = {}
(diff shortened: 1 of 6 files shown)


modal/_traceback.py
```diff
@@ -127,3 +129,42 @@ def print_server_warnings(server_warnings: Iterable[api_pb2.Warning]):
...
         warnings.warn_explicit(warning.message, ServerWarning, "<modal-server>", 0)
+
+
+# for some reason, the traceback cleanup here can't be moved into a context manager :(
+traceback_suppression_note = (
+    "Internal Modal traceback frames are suppressed for readability. Use MODAL_TRACEBACK=1 to show a full traceback."
+)
+
+
+class suppress_tb_frames:
```
thomasjpfan, line 140: I think we can simplify this:

```python
from contextlib import contextmanager


@contextmanager
def suppress_tb_frames(n: int):
    try:
        yield
    except Exception as exc:
        is_base_exception = not isinstance(exc, Exception)
        if is_base_exception:
            return

        tb = exc.__traceback__

        # modify traceback on exception object
        try:
            final_tb = tb
            for _ in range(n):
                final_tb = final_tb.tb_next
        except AttributeError:
            print(f"Failed to suppress {n} traceback frames from {str(exc)}")
            raise

        exc.with_traceback(final_tb)
        notes = getattr(exc, "__notes__", [])
        if traceback_suppression_note not in notes:
            # .add_note was added in Python 3.11
            notes.append(traceback_suppression_note)
        raise exc
```

Same for `grpc_error_converter`, but that one is less complex.

modal/_resolver.py
```diff
@@ -169,9 +163,9 @@ async def loader():
             self._local_uuid_to_future[obj.local_uuid] = cached_future
             if deduplication_key is not None:
                 self._deduplication_cache[deduplication_key] = cached_future
-
-        # TODO(elias): print original exception/trace rather than the Resolver-internal trace
-        return await cached_future
+        with suppress_tb_frames(2):
```
thomasjpfan, line 166: I'm unsure about skipping two frames. If we may refactor the closure, then we may need to update the number of frames.

thomasjpfan (commented):
From someone that goes through many long Python stack traces, I feel like the current stack trace is not too complex.

With my Python hat one, usually the pattern is to reraise from:

```python
except GRPCError as exc:
	raise SomeException(...) from exc
```

But I'm okay with hiding the `GRPCError`.

mwaskom:
> From someone that goes through many long Python stack traces, I feel like the current stack trace is not too complex.

I'm quite in favor of this effort. Modal is a fairly odd case where we are executing the user's code on behalf of them but trying to make that as transparent as possible. Add to that some of our features like `synchronicty` wrapping, and we end up with tracebacks that have very poor user-relevant information density. Additionally, without deep knowledge of Modal, users may not know that various stacktrace frames aren't relevant to them. We should try to make it as easy for them as possible to understand what's wrong with the code that they wrote, and i don't think that we have any specific examples where that requires tracking the particular path that flow took through the various modal internals.

modal/_resolver.py
mwaskom: This is a good catch, that hardcoded magic number is brittle.

Though I think it's something we could fix in a followup (since it's just about robustness to future hypothetical changes) right?

thomasjpfan: Yea, let's fix it in a follow up.

modal/_utils/async_utils.py
```diff
@@ -31,6 +32,10 @@
 P = ParamSpec("P")
 V = TypeVar("V")
 
+if sys.platform == "win32":
+    # quick workaround for deadlocks on shutdown - need to investigate further
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
thomasjpfan, line 37: Changing the event loop policy feels like a major change.

test/app_test.py
```diff
@@ -415,7 +416,8 @@ def test_app_create_bad_environment_name_error(client):
         ):  # TODO: why isn't environment_name an argument to app.run?
             pass
 
-    assert len(asyncio.all_tasks(synchronizer._loop)) == 1  # no trailing tasks, except the `loop_inner` ever-task
+    if sys.platform != "win32":  # proactor event loop has long lived helper tasks
```
thomasjpfan, line 419: Is this still needed if we set `WindowsSelectorEventLoopPolicy` as the event loop policy?

thomasjpfan (approved):

modal/_utils/async_utils.py
freider: This inverts a synchronicity change in the other direction. Want to use the default one but it broke some test fixtures in modal client so reverted it "modal side" until we have time to look at that

test/app_test.py
freider: Nope, but I'd like to keep it since I intend to revert the event loop policy change for windows and this is required for that (since the test still runs on other platforms and windows is documented to leak networking tasks I'd say it's fine)

thomasjpfan (approved):

PR #3266, opened 2025-06-22
 Simplify FlashManager.stop() 

gongy:
Instead of specifying a grace period in `.stop()`, skip `TunnelStop` which is superfluous in this use case.


gongy:
@prbot approve

PR #3268, opened 2025-06-22
Add logging for flash tunnel URL

gongy:


modal/experimental/__init__.py
```diff
@@ -223,9 +223,9 @@ def __init__(self, client: _Client, port: int):
         self.stopped = False
 
     async def _start(self):
-        tunnel = await self.tunnel_manager.__aenter__()
+        self.tunnel = await self.tunnel_manager.__aenter__()
 
-        hostname = tunnel.url.split("://")[1]
+        hostname = self.tunnel.url.split("://")[1]
```
luiscape, line 228: Wouldn't it be better to use `urlparse`? 

```python
hostname = urlparse(self.tunnel.url).hostname
```

thecodingwizard: Nice, I'll fix this in a follow-up PR. Thanks!

PR #3271, opened 2025-06-24
Remove some mentions of "lookup" from docs

ehdr:
## Describe your changes

The goal is to reduce traces of obsolete code.


mwaskom (approved):
Great thank you!

PR #3275, opened 2025-06-25
Increase concurrency of blob uploads/downloads

rculbertson:
This increases the concurrency with which we upload and download blobs from 10 to 20. 

This is in response to a user issue where the overall runtime of a map operation was limited by the speed of blob downloads. The server will reject requests if it has more than 100 outputs waiting to be consumed. In this case, the rate at which inputs were sent and processed was much higher than the rate the large outputs could be downloaded. The quicker we download outputs, the quicker we can send more inputs.

Trying out the change on my dev box, the total runtime of [this starmap script](https://gist.github.com/rculbertson/3b0e9e18ffd80f4881ddccf4f49a38db) went from ~4 min to ~2 min. 

This will have a positive impact only if you have enough bandwidth of course. Running the same script on my laptop, I saw no performance bump because I'm already at my max throughput apparently.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -46,7 +46,7 @@ MAX_ASYNC_OBJECT_SIZE_BYTES = 8 * 1024  # 8 KiB
 LARGE_FILE_LIMIT = 4 * 1024 * 1024  # 4 MiB
 
 # Max parallelism during map calls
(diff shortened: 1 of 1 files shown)


rculbertson:
@prbot approve

gongy:
:+1:

PR #3276, opened 2025-06-26
Report R2 Blob Upload Failure To The Server

CyrilSharma:
This adds fields to the FunctionPutItems protobuf which states which report whether an upload to R2 failed and what the latency when uploading to R2 was.

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -188,16 +188,23 @@ def get_content_length(data: BinaryIO) -> int:
     return content_length - pos
 
 
-async def _blob_upload_with_fallback(items, blob_ids, callback):
+async def _blob_upload_with_fallback(items, blob_ids: list[str], callback) -> tuple[str, bool, int]:
+    r2_latency_ms = 0
+    r2_failed = False
     for idx, (item, blob_id) in enumerate(zip(items, blob_ids)):
         # We want to default to R2 95% of the time and S3 5% of the time.
         # To ensure the failure path is continuously exercised.
         if idx == 0 and len(items) > 1 and random.random() > HEALTHY_R2_UPLOAD_PERCENTAGE:
             continue
         try:
+            init_time = time.monotonic_ns()
             await callback(item)
-            return blob_id
+            if blob_id.endswith(":r2"):
+                r2_latency_ms = (time.monotonic_ns() - init_time) // 1_000_000
+            return blob_id, r2_failed, r2_latency_ms
         except Exception as _:
+            if blob_id.endswith(":r2"):
+                r2_failed = True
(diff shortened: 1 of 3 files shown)


modal/_utils/blob_utils.py
```diff
@@ -188,25 +188,32 @@ def get_content_length(data: BinaryIO) -> int:
...
+async def _blob_upload_with_fallback(items, blob_ids: list[str], callback) -> tuple[str, bool, int]:
+    r2_latency_ns = 0
+    r2_failed = False
     for idx, (item, blob_id) in enumerate(zip(items, blob_ids)):
         # We want to default to R2 95% of the time and S3 5% of the time.
         # To ensure the failure path is continuously exercised.
         if idx == 0 and len(items) > 1 and random.random() > HEALTHY_R2_UPLOAD_PERCENTAGE:
             continue
         try:
+            init_time = time.time_ns()
```
CyrilSharma: This doesn't account for exponential backoff.

devennavani: Is there a reason we're using `ns` instead of `ms`?

modal/_utils/blob_utils.py
```diff
@@ -188,25 +188,32 @@ def get_content_length(data: BinaryIO) -> int:
...
             await callback(item)
-            return blob_id
+            if blob_id.endswith(":r2"):
+                r2_latency_ns = time.time_ns() - init_time
+            return blob_id, r2_failed, r2_latency_ns
         except Exception as _:
             # Ignore all errors except the last one, since we're out of fallback options.
```
devennavani: this comment should be directly above the `if idx == len(items) - 1:` line

modal/_utils/blob_utils.py
```diff
@@ -188,16 +188,23 @@ def get_content_length(data: BinaryIO) -> int:
...
+async def _blob_upload_with_fallback(items, blob_ids: list[str], callback) -> tuple[str, bool, int]:
+    r2_latency_ms = 0
+    r2_failed = False
     for idx, (item, blob_id) in enumerate(zip(items, blob_ids)):
         # We want to default to R2 95% of the time and S3 5% of the time.
         # To ensure the failure path is continuously exercised.
         if idx == 0 and len(items) > 1 and random.random() > HEALTHY_R2_UPLOAD_PERCENTAGE:
             continue
         try:
+            init_time = time.time_ns()
```
thomasjpfan: I tend to prefer [time.monotonic_ns](https://docs.python.org/3/library/time.html#time.monotonic_ns) or [time.perf_counter_ns](https://docs.python.org/3/library/time.html#time.perf_counter_ns) for this use case.

PR #3277, opened 2025-06-27
Adds quick workaround for pickle -> proto parameter upgrade glitches

freider:
Also added the reverse in case of downgrades...

I'd also like to investigate a "backend" fix that validates parameters, since this patch only deals with decode errors - not the less common parameter addition/rename class of container crashloops (e.g. if a new required parameter is added but there are still instances trying to bind that class without that parameter - this ends up in a crash loop instead of a FunctionBindParams error)

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

* Added more gracefull handling for upgrading class constructor based parameters to `modal.parameter()` annotation based parameters.

### DIFF
--- a/modal/_serialization.py
+++ b/modal/_serialization.py
@@ -474,20 +474,28 @@ def deserialize_params(serialized_params: bytes, function_def: api_pb2.Function,
         # legacy serialization format - pickle of `(args, kwargs)` w/ support for modal object arguments
         try:
             param_args, param_kwargs = deserialize(serialized_params, _client)
-        except DeserializationError:
-            # Fallback in case of proto -> pickle downgrades, where a pickle function could end
-            # up getting bound to proto argument during the downgrade transition.
+        except DeserializationError as original_exc:
+            # Fallback in case of proto -> pickle downgrades of a parameter serialization format
+            # I.e. FunctionBindParams binding proto serialized params to a function defintion
+            # that now assumes pickled data according to class_parameter_info
             param_args = ()
-            param_kwargs = deserialize_proto_params(serialized_params)
+            try:
+                param_kwargs = deserialize_proto_params(serialized_params)
+            except Exception:
+                raise original_exc
 
     elif function_def.class_parameter_info.format == api_pb2.ClassParameterInfo.PARAM_SERIALIZATION_FORMAT_PROTO:
         param_args = ()  # we use kwargs only for our implicit constructors
         try:
             param_kwargs = deserialize_proto_params(serialized_params)
-        except google.protobuf.message.DecodeError:
-            # Fallback in case of pickle -> proto upgrades, where a proto function could end
-            # up getting bound to pickle argument during the upgrade transition.
-            param_args, param_kwargs = deserialize(serialized_params, _client)
+        except google.protobuf.message.DecodeError as original_exc:
+            # Fallback in case of pickle -> proto upgrades of a parameter serialization format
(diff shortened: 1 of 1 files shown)


modal/_serialization.py
```diff
@@ -470,10 +472,23 @@ def deserialize_params(serialized_params: bytes, function_def: api_pb2.Function,
         api_pb2.ClassParameterInfo.PARAM_SERIALIZATION_FORMAT_PICKLE,
     ):
         # legacy serialization format - pickle of `(args, kwargs)` w/ support for modal object arguments
-        param_args, param_kwargs = deserialize(serialized_params, _client)
+        try:
+            param_args, param_kwargs = deserialize(serialized_params, _client)
+        except DeserializationError:
```
mwaskom: Is this case we're expecting the only way we can get a `DeserializeError` here?

modal/_serialization.py
```diff
@@ -470,10 +472,23 @@ def deserialize_params(serialized_params: bytes, function_def: api_pb2.Function,
...
+            # Fallback in case of proto -> pickle downgrades, where a pickle function could end
+            # up getting bound to proto argument during the downgrade transition.
```
mwaskom: what does it mean for a "pickle function [to get] bound to proto argument"?

modal/_serialization.py
```diff
@@ -470,10 +472,23 @@ def deserialize_params(serialized_params: bytes, function_def: api_pb2.Function,
...
+            param_args = ()
+            param_kwargs = deserialize_proto_params(serialized_params)
+
     elif function_def.class_parameter_info.format == api_pb2.ClassParameterInfo.PARAM_SERIALIZATION_FORMAT_PROTO:
         param_args = ()  # we use kwargs only for our implicit constructors
-        param_kwargs = deserialize_proto_params(serialized_params)
+        try:
+            param_kwargs = deserialize_proto_params(serialized_params)
+        except google.protobuf.message.DecodeError:
```
mwaskom: Similar question above, is there any other reason we would get this error?

modal/_serialization.py
freider: There may be other ways of getting this, but in those cases the proto decode would also fail and result in container crashes afaik

modal/_serialization.py
freider: a function that has parameter_info.serialization_format set to pickle getting a `serialized_params` payload in FunctionBindParams that's proto encoded

modal/_serialization.py
freider: I could imagine something like us adding a new parameter enum type which isn't supported, or messing with the proto in forbidden ways, but in that case it would still error on pickle decode and result in a similarly crashlooping container as today

modal/_serialization.py
mwaskom: Mostly just asking for clearer comments so we understand the point of this in a few months.

modal/_serialization.py
mwaskom: Yeah my concern is that the error message / traceback in that case will be misleading.

freider: I can't come up with any cases where this would be something users should have to care about, but maybe if they are pickling some custom type parameter that they don't have available in the container? Feels very niche (and also only applicable for the deprecated explicit constructor style classes)

mwaskom:
I think that makes sense.

PR #3278, opened 2025-06-27
Adds deprecation warning when using wrap_returned_exceptions=True

freider:
## Describe your changes

Closes CLI-419

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2024
 import asyncio
 import enum
+import inspect
 import time
 import typing
 from asyncio import FIRST_COMPLETED
@@ -410,7 +411,7 @@ async def _map_invocation(
         async_merge(drain_input_generator(), pump_inputs(), poll_outputs(), retry_inputs())
     ) as streamer:
         async for response in streamer:
-            if response is not None:
+            if response is not None:  # type: ignore[unreachable]
                 yield response.value
     log_debug_stats_task.cancel()
     await log_debug_stats_task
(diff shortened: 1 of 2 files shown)


modal/parallel_map.py
```diff
@@ -434,6 +435,17 @@ async def _map_helper(
...
+    if return_exceptions and wrap_returned_exceptions:
+        deprecation_warning(
+            (2025, 6, 27),
+            (
+                ".map and .starmap with return_exceptions=True will change behavior in a future version of Modal. "
+                + "The returned exceptions will no longer be wrapped in a modal.exceptions.UserCodeException.\n"
+                + "To get the future behavior now and silence this warning, use the `wrap_returned_exceptions=False` "
+                + "argument.\nE.g. `func.map(..., return_exceptions=True, wrap_returned_exceptions=False)"
+            ),
+            show_source=False,  # this is several frames down the stack so wouldn't help much
```
mwaskom: I thought our `deprecation_warning` machinery already attributed the warning to the last non-modal frame in the stack, no?

modal/parallel_map.py
```diff
@@ -434,6 +435,17 @@ async def _map_helper(
     We could make this explicit as an improvement or even let users decide what they
     prefer: throughput (prioritize queueing inputs) or latency (prioritize yielding results)
     """
+    if return_exceptions and wrap_returned_exceptions:
+        deprecation_warning(
+            (2025, 6, 27),
+            (
+                ".map and .starmap with return_exceptions=True will change behavior in a future version of Modal. "
```
mwaskom: I sort of  think we should just be very direct in the message that this is an amelioration + graceful transition for a bug, not an "API change". Thoughts? Basically I want to manage expectations about whether we're likely to continue introducing a lot of deprecations post 1.0.

modal/parallel_map.py
```diff
@@ -434,6 +435,17 @@ async def _map_helper(
...
+                + "The returned exceptions will no longer be wrapped in a modal.exceptions.UserCodeException.\n"
```
mwaskom: Nit but I don't think you need explicit string concatenation here (and i think having it will suppress the earning message from `inv show-deprecations` but maybe we added handling for that).

mwaskom (approved):

modal/parallel_map.py
freider: I think it ended up being some asyncio internal frame for me when I tested this

mwaskom: Yeah I'm seeing that too, annoying.

mwaskom: It works if we warn from the "implementation functions" so i'm gonna refactor a bit and accept some duplication to get the warning attribution.

PR #3286, opened 2025-07-01
modal run --detach should use .spawn

gongy:
## Describe your changes

Critical for training workflows where `local_entrypoint` is not specified and `--detach` is used.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Fixed handling of retries when using `modal run --detach` directly against a remote Function.


mwaskom (approved):
Nice find

PR #3287, opened 2025-07-01
Client Cancelation Changes

bokeefe123:
Necessary RPC changes to support client cancelation

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1471,7 +1471,7 @@ message FunctionCallCallGraphInfo {
 message FunctionCallCancelRequest {
   string function_call_id = 1;
   bool terminate_containers = 2;
-  optional string function_id = 3; // Only used for sync cancellation on the input plane. Async cancellation does not require this.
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1471,6 +1471,7 @@ message FunctionCallCallGraphInfo {
 message FunctionCallCancelRequest {
   string function_call_id = 1;
   bool terminate_containers = 2;
+  optional string function_id = 3; // Only used for sync cancellation on the input plane. Async cancellation does not require this.
```
```suggestion
  optional string function_id = 3; // Only provided for sync input cancellation on the input plane. Async input cancellation does not provide this field this.
```

thecodingwizard (approved):

modal_proto/api.proto
```diff
@@ -1471,6 +1471,7 @@ message FunctionCallCallGraphInfo {
 message FunctionCallCancelRequest {
   string function_call_id = 1;
   bool terminate_containers = 2;
```
thecodingwizard, line 1473: oops can we also add a comment that sync cancellations on the input plane do not support this field?

PR #3288, opened 2025-07-01
Update micromamba base image tag for next builder version

mwaskom:
The [micromamba docker images](https://hub.docker.com/r/mambaorg/micromamba) appear to have changed their naming scheme since the last image builder update. Previously they used Debian "codenames" but more recent tags have numeric debian versions. That means we need to change the way we store versioned micromamba base images.

Closes CLI-459


mwaskom:
old image builder versions pass the hash tests and the new one won't affect anyone yet

thomasjpfan (commented):
✅

PR #3289, opened 2025-07-01
Refresh input plane auth token

rculbertson:
Five minutes before the input plane auth token expires, request a new one as part of AppHeartbeat.

## Compatibility checklist

---

## Release checklist

---

</details>


rculbertson:
@prbot approve

rculbertson:
@thecodingwizard pointed out that this approach might not work if we do a function lookup, rather than running as an app. So we'll need to hook in somewhere else to check if we need to refresh.

rculbertson:
I changed the approach so we don't use AppHeartbeat. Instead we check if the token needs to be refreshed before each call to AttemptStart and AttemptAwait.
For now it calls ClientHello, but that function has no authentication. So we need to call a different endpoint, or maybe add a new one.

rculbertson:
Ok, pushed another commit, and updated the description at the top.

requirements.dev.txt
```diff
@@ -9,6 +9,7 @@ grpcio-tools==1.66.2;python_version>='3.13'
 grpclib==0.4.7
 httpx~=0.23.0
 invoke~=2.2
+pyjwt==2.10.1
```
rculbertson, line 12: This is used only in the tests to create tokens.

modal_proto/api.proto
```diff
@@ -3253,6 +3260,9 @@ service ModalClient {
   rpc AttemptRetry(AttemptRetryRequest) returns (AttemptRetryResponse);
   rpc AttemptStart(AttemptStartRequest) returns (AttemptStartResponse);
 
+  // Auth Token
+  rpc AuthTokenGet(AuthTokenGetRequest) returns (AuthTokenGetResponse);
+
```
rculbertson: I will break the proto changes out into a separate PR.

rculbertson: Done

modal/_functions.py
```diff
@@ -419,9 +417,7 @@ async def run_function(self) -> Any:
                 timeout_secs=OUTPUTS_TIMEOUT,
                 requested_at=time.time(),
             )
-            metadata: list[tuple[str, str]] = []
-            if self.input_plane_region and self.input_plane_region != "":
-                metadata.append(("x-modal-input-plane-region", self.input_plane_region))
+            metadata = await self._get_metadata(self.input_plane_region, self.client)
```
mwaskom, line 420: j/w/ but is there a reason we need to pass the `self.input_plane_region` and `self.client` into this call? Feels like the `metadata` becomes a fixed property once those are known. (Can they change?)

modal/_utils/grpc_utils.py
```diff
@@ -149,21 +149,7 @@ async def send_request(event: grpclib.events.SendRequest) -> None:
 
         logger.debug(f"Sending request to {event.method_name}")
 
-    async def recv_initial_metadata(initial_metadata: grpclib.events.RecvInitialMetadata) -> None:
-        # If we receive an auth token from the server, include it in all future requests.
-        # TODO(nathan): This isn't perfect because the metadata isn't propagated when the
-        # process is forked and a new channel is created. This is OK for now since this
-        # token is only used by the experimental input plane
```
mwaskom, line 156: Is it correct that this PR sufficiently changes the approach to getting the auth-token that this TODO is no longer relevant?

modal/client.py
```diff
@@ -78,6 +79,7 @@ class _Client:
     _cancellation_context: TaskContext
     _cancellation_context_event_loop: asyncio.AbstractEventLoop = None
     _stub: Optional[api_grpc.ModalClientStub]
+    _auth_token_manager: _AuthTokenManager = None
```
mwaskom, line 82: Persnickety, but should use explicit `Optional[]` here.

test/auth_token_manager_test.py
```diff
@@ -0,0 +1,299 @@
...
+    exp_time = int(time.time()) + 240  # 4 minutes from now
+    payload = {"exp": exp_time, "type": "near_expiry"}
+    return jwt.encode(payload, "my-secret-key", algorithm="HS256")
+
+
+@pytest.mark.asyncio
+async def test_get_token_initial_fetch(auth_token_manager, valid_jwt_token, client, servicer):
+    """Test getting token when no token exists."""
+
+    @synchronize_api
```
mwaskom, line 64: Is there a reason you can't just `await` the underlying coroutine? Is this testing some nasty synchronicity wrapping issue?

test/auth_token_manager_test.py
```diff
@@ -0,0 +1,299 @@
...
+    token = await wrapped_get_token.aio()
+    assert token == valid_jwt_token
+
+
+@pytest.mark.asyncio
+async def test_get_token_expired(auth_token_manager, expired_jwt_token, valid_jwt_token, servicer):
+    """Test that expired token triggers refresh."""
+    # Set up expired token
+    auth_token_manager._token = expired_jwt_token
+    auth_token_manager._expiry = time.time() - 100  # Expired
```
mwaskom: j/w but it looks like the `expired_jwt_token` payload already encodes a (different) expiry time. Is there an obvious reason the expiry is not always read from it directly to maintain a single source of truth?

modal/_utils/auth_token_manager.py
```diff
@@ -0,0 +1,107 @@
...
+        async with lock:
+            # Double check inside lock: maybe another coroutine refreshed already
+            if self._token and not self._needs_refresh():
+                return
+            resp: api_pb2.AuthTokenGetResponse = await retry_transient_errors(
+                self._stub.AuthTokenGet, api_pb2.AuthTokenGetRequest()
+            )
+            if not resp.token:
+                # Not expected
+                raise RemoteError("Did not receive auth token from server")
```
mwaskom: This feels like the wrong error type — although I'll grant that it's consistent with the way that type is documented! Seems like (based on this but also extant usage) that we're sort of overloading the "remote" concept to mean both remote execution (of user code) and remote in the sense of the server being remote to the client.

Given that the `.remote()` concept is so important to Modal that's a little unfortunate (although we do mostly propagate the _original_ error type back from remote execution and only raise `RemoteError` in cases where we can't, e.g. if the type isn't defined locally or if the remote failure is a non-python image build issue.

Also, is this error message going to be meaningful to a user who sees it?

modal/_utils/auth_token_manager.py
```diff
@@ -0,0 +1,107 @@
...
+                # This should never happen.
+                logger.warning("x-modal-auth-token does not contain exp field")
+                # We'll use the token, and set the expiry to 20 min from now.
+                self._expiry = time.time() + self.DEFAULT_EXPIRY
+
+    async def _get_lock(self) -> asyncio.Lock:
+        # Note: this function runs no async code but is marked as async to ensure it's
+        # being run inside the synchronicity event loop and binds the lock to the
+        # correct event loop on Python 3.9 which eagerly assigns event loops on
+        # constructions of locks
```
mwaskom, line 91: Glad you got to learn about this "fun" aspect of synchronicity.

modal/_utils/auth_token_manager.py
```diff
@@ -0,0 +1,107 @@
...
+    @staticmethod
+    def _decode_jwt(token: str) -> dict[str, Any]:
+        """Decodes a JWT into a dict without verifying signature."""
+        try:
+            payload = token.split(".")[1]
+            padding = "=" * (-len(payload) % 4)
+            decoded_bytes = base64.urlsafe_b64decode(payload + padding)
+            return json.loads(decoded_bytes)
+        except Exception as e:
+            raise ValueError("Cannot parse auth token") from e
```
mwaskom: Same question as above about what we're communicating to the user on a failure here.

modal/_utils/auth_token_manager.py
```diff
@@ -0,0 +1,107 @@
...
+from .logger import logger
+
+
+class _AuthTokenManager:
+    """ Handles fetching and refreshing of the input plane auth token. """
+
+    # Start refreshing this many seconds before the token expires
+    REFRESH_WINDOW = 5 * 60
+    # If the token doesn't have an expiry use this default (not expected)
+    DEFAULT_EXPIRY = 20 * 60
```
mwaskom: Super nit but since `expiry` is otherwise an absolute time (I think?) should this be `DEFAULT_EXPIRY_OFFSET` or something?

modal/_utils/auth_token_manager.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        When called, the AuthTokenManager can be in one of three states:
+        1. Has a valid cached token. It is returned to the caller.
+        2. Has no cached token, or the token is expired. We fetch a new one and cache it. If `get_token` is called
+        concurrently by multiple coroutines, all requests will block until the token has been fetched. But only one
+        coroutine will actually make a request to the control plane to fetch the new token. This ensures we do not hit
+        the control plane with more requests than needed.
+        3. Has a valid cached token, but it is going to expire in the next 5 minutes. In this case we fetch a new token
+        and cache it. If `get_token` is called concurrently, only one request will fetch the new token, and the others
+        will be given the old (but still valid) token - i.e. they will not block.
```
mwaskom, line 40: Nice

modal/_utils/auth_token_manager.py
```diff
@@ -0,0 +1,107 @@
...
+        return self._token
+
+    async def _refresh_token(self):
+        """
+        Fetch a new token from the control plane. If called concurrently, only one coroutine will make a request for a
+        new token. The others will block on a lock, until the first coroutine has fetched the new token.
+        """
+        lock = await self._get_lock()
+        async with lock:
+            # Double check inside lock: maybe another coroutine refreshed already
```
mwaskom: Is there a specific race condition that could lead to this?

modal/_functions.py
rculbertson: `self.input_plane_region` and `self.client` won't change, but the value returned by `get_metadata` can change. `get_metadata` fetches an auth token and inserts it into the metadata, so that can be different from call to call.

modal/_utils/auth_token_manager.py
rculbertson: Oh, that's a fair point. Will update.

modal/_utils/auth_token_manager.py
rculbertson: Yea, this would happen the very first time, when we have no token at all. All but one coroutine will block on the lock, waiting for the one coroutine to fetch it.

Updated the comment to this:
```
# Double check inside lock - maybe another coroutine refreshed already. This happens the first time we fetch
# the token. The first coroutine will fetch the token, while the others block on the lock, waiting for the
# new token. Once we have a new token, the other coroutines will unblock and return from here.
```

modal/_utils/auth_token_manager.py
rculbertson: Yea, I wasn't sure what would be a good error type here. How about `ExecutionError`? That seems like it's used in other places like this one when something unexpected happens:

https://github.com/modal-labs/modal-client/blob/110957c51054356bfce37307365ec1bbce0969a6/modal/_runtime/asgi.py#L498

Open to suggestions on how to improve the error message. This one is tricky because this should never happen - so something's gone very wrong if this happens - and there's nothing the user can do.

modal/_utils/auth_token_manager.py
rculbertson: I'm just glad elias hit this issue already and had a solution I could copy!

modal/_utils/auth_token_manager.py
rculbertson: hmmm...I could do something like this:

https://github.com/modal-labs/modal-client/blob/d5d7a8334e59cb315036657ea8f3ddc6b79ba66b/modal/_runtime/user_code_imports.py#L339

Calling out this is an internal error, and something they have no control over.

test/auth_token_manager_test.py
rculbertson: If I don't wrap it in this `@synchronize_api` method, the test just hangs. I'm not sure why, but I copied this technique from other tests which are doing the same thing. For example, this one:

https://github.com/modal-labs/modal-client/blob/79eb3852a7f4f945a8d4291f37ccace1b3ee262a/test/grpc_utils_test.py#L71

I should add a comment about that though.

modal/_utils/auth_token_manager.py
mwaskom: Yeah I think it's helpful to make it very clear when a fatal exception is arising from an unexpected internal state that users can't affect.

test/auth_token_manager_test.py
mwaskom: 😵‍💫

test/auth_token_manager_test.py
rculbertson: Yea that's a good point. Updated test.

modal/_utils/auth_token_manager.py
rculbertson: I updated it to `Internal error: Did not receive auth token from server. Please contact Modal support.`, using this line as an example:

https://github.com/modal-labs/modal-client/blob/d5d7a8334e59cb315036657ea8f3ddc6b79ba66b/modal/_runtime/user_code_imports.py#L338-L340

modal/_utils/auth_token_manager.py
rculbertson: Sounds good. Updated.

modal/_utils/grpc_utils.py
rculbertson: Oh, good eye, I forgot about this comment. But yes, with the new system if the token isn't present for some reason, we will just fetch a new one. So this case will be handled correctly.

modal/client.py
rculbertson: oops, thanks for catching. Fixed.

PR #3290, opened 2025-07-02
volume list --show-id

danielnorberg:
## Describe your changes

Sometimes it's very useful to know the id of a volume, e.g. for troubleshooting, but it can be tricky to find out.

---

## Compatibility checklist

---

## Release checklist

---

</details>


freider:
I wouldn't be opposed to always including it - wdyt @mwaskom ?

danielnorberg:
Then I'd be in favor of always showing the id.

mwaskom:
Are Volume IDs something users every need to think about?

danielnorberg:
Users usually do not need to think about volume ids.

PR #3292, opened 2025-07-03
Let server configure max object size 

rculbertson:
Part of SVC-756.

This adds `max_object_size_bytes` to FunctionHandleMetadata, so that the server can tell the client what value to use. 

This is being added because we want to allow some users to send larger inputs. To do this we will add a new experimental option `large_inputs_enabled`. If this is set, the server will set a larger max object size, otherwise it will set the current default. If this field isn't set at all, the client uses the current default.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/test/function_test.py
+++ b/test/function_test.py
@@ -1445,6 +1445,10 @@ def test_function_namespace_deprecated(servicer, client):
     namespace_warnings = [w for w in record if "namespace" in str(w.message).lower()]
     assert len(namespace_warnings) == 0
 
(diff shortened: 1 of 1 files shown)


thecodingwizard (approved):

thecodingwizard:
Should we leave a comment somewhere that this parameter is deliberately only respected for the input plane?

thecodingwizard (approved):
lgtm!

PR #3296, opened 2025-07-04
Prevent generator task leakage CLI-254 CLI-495

freider:
When exceptions (including cancellations) occurred with web endpoints, we previoulsy just exited the input context and left behind the task responsible for sending generator outputs indefinitely polling for additional generator outputs (on the synchronicity event loop).

This refactors the generator output task to be wrapped in an async context manager to always ensure it's cleaned up on exit (sending any queued outputs before exiting). A bonus is that it gets rid of the last synchronicity `_future=True` use in the client code :) 

**This tentatively resolves the long standing "Task was destroyed but is pending" spam with web endpoints**

This also has the potential to fix correctness edge cases with container shutdown where queued outputs wouldn't have been sent before in case of exceptions, or where the exception/failure status would have been sent before the queued generator outputs that led to the exception.

TODO: add regression test that reproduces the old task leakage

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

* Fixed containers printing "Task was destroyed but it is pending" on exit after exceptions in generators or web endpoints

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -451,8 +451,10 @@ class _InputPlaneInvocation:
                         self.attempt_token = retry_response.attempt_token
                         continue
 
+                control_plane_stub = self.client.stub
+                # Note: Blob download is done on the control plane stub, not the input plane stub!
                 return await _process_result(
-                    await_response.output.result, await_response.output.data_format, self.stub, self.client
+                    await_response.output.result, await_response.output.data_format, control_plane_stub, self.client
                 )
 
 
--- a/modal_version/__init__.py
+++ b/modal_version/__init__.py
@@ -1,4 +1,4 @@
 # Copyright Modal Labs 2025
 """Supplies the current version of the modal client library."""
 
-__version__ = "1.0.6.dev13"
+__version__ = "1.0.6.dev14"


modal/_runtime/container_io_manager.py
```diff
@@ -508,33 +507,47 @@ async def put_data_out(
...
-                if message is self._GENERATOR_STOP_SENTINEL:
-                    received_sentinel = True
+    @asynccontextmanager
+    async def generator_output_sender(
+        self, function_call_id: str, data_format: int, message_rx: asyncio.Queue
+    ) -> AsyncGenerator[None, None]:
+        """Runs background task that feeds generator outputs into a function call's `data_out` stream."""
+        GENERATOR_STOP_SENTINEL = Sentinel()
+
+        async def generator_output_task():
```
freider, line 517: This is just the old implementation untouched and indented

modal/_runtime/container_io_manager.py
```diff
@@ -508,33 +507,47 @@ async def put_data_out(
...
+                        total_size += len(serialized_messages[-1]) + 512  # 512 bytes for estimated framing overhead
+                await self.put_data_out(function_call_id, index, data_format, serialized_messages)
+                index += len(serialized_messages)
+
+        task = asyncio.create_task(generator_output_task())
+        try:
+            yield
+        finally:
+            # gracefully stop the task after all current inputs have been sent
+            await message_rx.put(GENERATOR_STOP_SENTINEL)
```
freider, line 549: Moved the sentinel injection here, further cleaning up the separation of concerns of the container entrypoint vs implementation details in the generator output sender

PR #3299, opened 2025-07-07
Actually report latency

CyrilSharma:
Fix a mistake where we measured total upload time instead of request latency.


thecodingwizard (approved):
lgtm!

PR #3302, opened 2025-07-08
Fix with_options when arguments are pre-hydrated

mwaskom:
## Describe your changes

- Fixes CLI-493

</details>

## Changelog

- Fixed a bug where `Cls.with_options` would fail when provided with a Secret object that was already hydrated.

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -98,6 +98,7 @@ class _ServiceOptions:
 
         This mostly exists to support "stacking" of `.with_options()` calls.
         """
+        # Don't use dataclasses.asdict() because it does a deepcopy(), which chokes on a hydrated object
         new_options_dict = {k.name: getattr(new_options, k.name) for k in dataclasses.fields(new_options)}
 
         # Resources needs special merge handling because individual fields are parameters in the public API


mwaskom:
@prbot approve

modal/_functions.py
```diff
@@ -1156,7 +1156,8 @@ async def _load(param_bound_func: _Function, resolver: Resolver, existing_object
 
         def _deps():
             if options:
-                return [v for _, v in options.validated_volumes] + list(options.secrets)
+                all_deps = [v for _, v in options.validated_volumes] + list(options.secrets)
```
thomasjpfan, line 1159: Does passing in un-hydrated deps break any assumptions around `_Object.deps`?

https://github.com/modal-labs/modal-client/blob/ddc68d72f0cdee23699e9c7dcd0a5a45678b8842/modal/_object.py#L239-L246

I think the only caller is here:

https://github.com/modal-labs/modal-client/blob/ddc68d72f0cdee23699e9c7dcd0a5a45678b8842/modal/_resolver.py#L142

mwaskom: Do you mean passing _only_ unhydrated deps, or unhydrated deps at all? If the latter, the purpose of `load()` is to hydrate the object (confusing terminology). If the former, I'm not 100% sure. As previously discussed, we'd like for hydration to be a strictly binary state so these questions are easier to reason about. But i'm not completely certain we're not currently violating that in some way.

thomasjpfan: > Do you mean passing only unhydrated deps

I mean this one.

Looking over other `_deps` implementation, I think it's okay to just pass in only unhydrated deps:
https://github.com/modal-labs/modal-client/blob/cb8c6fa61d95a0838f96b62dd9a3d19949c54211/modal/cls.py#L140-L148

PR #3305, opened 2025-07-09
Measure Throughput Instead

CyrilSharma:
This provides a way for the client to signal R2's health.


erikbern:
Should we make the client agnostic of the storage backend we're using?

devennavani (approved):
LGTM

PR #3301, opened 2025-07-07
Define the 2025.06 Image Builder version

mwaskom:
## Describe your changes

Potentially subject to change if issues are discovered during further testing, but otherwise not expected to.

- Closes CLI-488
- Closes CLI-471

<details> <summary>Checklists</summary>

---

## Compatibility checklist

Check these boxes or delete any item (or this section) if not relevant for this PR.

- [ ] Client+Server: this change is compatible with old servers
- [ ] Client forward compatibility: this change ensures client can accept data intended for later versions of itself

Note on protobuf: protobuf message changes in one place may have impact to
multiple entities (client, server, worker, database). See points above.

---

## Release checklist

If you intend for this commit to trigger a full release to PyPI, please ensure that the following steps have been taken:

- [ ] Version file (`modal_version/__init__.py`) has been updated with the next logical version
- [ ] Changelog has been cleaned up and given an appropriate subhead

---

</details>

## Changelog

- TK Changelog note introducing the 2025.06 Image Builder Version

PR #3306, opened 2025-07-09
Add 2025.06 requirements file

mwaskom:
Didn't notice that this was missing from https://github.com/modal-labs/modal-client/pull/3301; our gitignore has `**/*.txt` so these aren't picked up by `git commit -a` and need to be force added, which I forgot 😞.

Edit: also caught a couple places in the CI workflow where we needed to do things on 2025.06 too.


mwaskom:
@prbot approve

ehdr (commented):
Oh, yeah maybe we should remove that `**/*.txt` and replace it with more specific ignores? But LGTM!

PR #3307, opened 2025-07-09
Don't error while handling dependency mount creation failure

mwaskom:
Fixes a minor bug where we trip over an AttributeError while trying to report dependency mount creation failures. I think the async version of `Process` may not have a `.args` attribute.


matslina (commented):
LGTM

PR #3309, opened 2025-07-09
[SpawnMapV2](1) Introduce Experimental Spawn Map To The API

CyrilSharma:
This PR introduces `experimental_spawn_map` which is equivalent to `spawn_map` except in the backend it is a single function call invocation. It also return a `FunctionCall` handler instead of `None`. Follow-up PRs will add more functionality to the `FunctionCall` handler such as `.get(index)` and an iterator.

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -71,6 +71,8 @@ from .mount import _get_client_mount, _Mount
 from .network_file_system import _NetworkFileSystem, network_file_system_mount_protos
 from .output import _get_output_manager
 from .parallel_map import (
+    _experimental_spawn_map_async,
+    _experimental_spawn_map_sync,
     _for_each_async,
     _for_each_sync,
     _map_async,
@@ -78,6 +80,7 @@ from .parallel_map import (
     _map_invocation_inputplane,
     _map_sync,
     _spawn_map_async,
+    _spawn_map_invocation,
     _spawn_map_sync,
     _starmap_async,
     _starmap_sync,
@@ -1543,6 +1546,19 @@ Use the `Function.get_web_url()` method instead.
                 async for item in stream:
                     yield item
 
+    @live_method
+    async def _spawn_map(self, input_queue: _SynchronizedQueue) -> "_FunctionCall":
+        self._check_no_web_url("map")
+        if self._is_generator:
+            raise InvalidError("A generator function cannot be called with `.spawn_map(...)`.")
+
+        assert self._function_name
+        return await _spawn_map_invocation(
+            self,
+            input_queue,
+            self.client,
+        )
+
     async def _call_function(self, args, kwargs) -> ReturnType:
         invocation: Union[_Invocation, _InputPlaneInvocation]
         if self._input_plane_url:
@@ -1789,6 +1805,7 @@ Use the `Function.get_web_url()` method instead.
     starmap = MethodWithAio(_starmap_sync, _starmap_async, synchronizer)
     for_each = MethodWithAio(_for_each_sync, _for_each_async, synchronizer)
     spawn_map = MethodWithAio(_spawn_map_sync, _spawn_map_async, synchronizer)
+    experimental_spawn_map = MethodWithAio(_experimental_spawn_map_sync, _experimental_spawn_map_async, synchronizer)
 
 
 class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1547,17 +1547,19 @@ Use the `Function.get_web_url()` method instead.
                     yield item
 
     @live_method
-    async def _spawn_map(self, input_queue: _SynchronizedQueue) -> "_FunctionCall":
+    async def _spawn_map(self, input_queue: _SynchronizedQueue) -> "_FunctionCall[ReturnType]":
         self._check_no_web_url("spawn_map")
         if self._is_generator:
             raise InvalidError("A generator function cannot be called with `.spawn_map(...)`.")
 
         assert self._function_name
-        return await _spawn_map_invocation(
+        function_call_id = await _spawn_map_invocation(
             self,
             input_queue,
             self.client,
         )
+        fc: _FunctionCall[ReturnType] = _FunctionCall._new_hydrated(function_call_id, self.client, None)
+        return fc
 
     async def _call_function(self, args, kwargs) -> ReturnType:
         invocation: Union[_Invocation, _InputPlaneInvocation]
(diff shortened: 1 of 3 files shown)


modal/_functions.py
```diff
@@ -1502,6 +1505,26 @@ async def _map(
             async for item in stream:
                 yield item
 
+    @live_method
+    async def _experimental_spawn_map(self, input_queue: _SynchronizedQueue) -> None:
+        self._check_no_web_url("map")
+        if self._is_generator:
+            raise InvalidError("A generator function cannot be called with `.map(...)`.")
```
mwaskom: Should these consider the current method to be called `spawn_map`?

modal/_functions.py
```diff
@@ -1502,6 +1505,26 @@ async def _map(
...
+            count_update_callback = output_mgr.function_progress_callback(self._function_name, total=None)
+        else:
+            count_update_callback = None
+
+        await _experimental_spawn_map_invocation(
+            self,
+            input_queue,
+            self.client,
+            count_update_callback,
+            api_pb2.FUNCTION_CALL_INVOCATION_TYPE_SYNC,
```
mwaskom: Doesn't `spawn_map` send inputs on the `ASYNC` path?

modal/parallel_map.py
```diff
@@ -642,6 +858,33 @@ def _call_with_args(args):
...
+    Example:
+    ```python
+    @app.function()
+    def my_func(a):
+        return a ** 2
+
+
+    @app.local_entrypoint()
+    def main():
+        my_func.experimental_spawn_map([1, 2, 3, 4])
```
mwaskom: AFAICT this is being exposed as _experimental_spawn_map (with a leading underscore) but the documentation and error messages below are omitting the leading underscore.

I would probably just put `spawn_map` in all eventually-user-facing messaging. Nobody will get confused before you ship, and it reduces the chance that you miss some cleanup and confuse people after you do.

modal/parallel_map.py
```diff
@@ -520,6 +717,25 @@ async def _map_async(
         yield output
 
 
+async def _experimental_spawn_map_async(self, *input_iterators, kwargs={}) -> None:
```
mwaskom: Do internal functions need the `_experimental` namespacing?

mwaskom (commented):
> It's a lot of code but the logic is very similar to .map().

I'm a little sad about this because (a) the existing `.map()` implementation is already quite hard to follow and (b) it looks like there's a lot of straight-up code duplication here for things that aren't (obviously) different between the sync and async versions. Is there any opportunity for refactoring to cut down on how much we need to be copy/pasting?

CyrilSharma:
TODO: An extra call when the iterator terminates which just uploads the length.

modal/_functions.py
devennavani: @CyrilSharma `_experimental_spawn_map_invocation` shouldn't need to even take in a function call invocation type, since we know it's always `ASYNC`

modal/parallel_map.py
```diff
@@ -86,6 +86,203 @@ class _OutputValue:
...
+        parent_input_id=current_input_id() or "",
+        function_call_type=api_pb2.FUNCTION_CALL_TYPE_MAP,
+        function_call_invocation_type=function_call_invocation_type,
+    )
+    response: api_pb2.FunctionMapResponse = await retry_transient_errors(client.stub.FunctionMap, request)
+
+    # Initialize state
+    retry_policy = response.retry_policy
+    retry_queue = TimestampPriorityQueue()
+    sync_client_retries_enabled = response.sync_client_retries_enabled
```
devennavani: Why are `retry_queue` and `sync_client_retries_enabled` needed here?

modal/parallel_map.py
```diff
@@ -86,6 +86,203 @@ class _OutputValue:
...
+        function_call_id=response.function_call_id,
+        map_items_manager=_MapItemsManager(
+            retry_policy,
+            function_call_invocation_type,
+            retry_queue,
+            sync_client_retries_enabled,
+            max_inputs_outstanding,
+        ),
+    )
+
```
devennavani: From here onwards we can be using the `InputPreprocessor` and `InputPumper` code from the previous PR yes?

modal/_functions.py
```diff
@@ -1502,6 +1505,26 @@ async def _map(
             async for item in stream:
                 yield item
 
+    @live_method
+    async def _experimental_spawn_map(self, input_queue: _SynchronizedQueue) -> None:
```
devennavani: Return type should be same as `spawn` yes? `"_FunctionCall[ReturnType]"`?

CyrilSharma: I originally was not going to add the handler in this proposal.

modal/_functions.py
```diff
@@ -1543,6 +1546,19 @@ async def _map(
                 async for item in stream:
                     yield item
 
+    @live_method
+    async def _spawn_map(self, input_queue: _SynchronizedQueue) -> "_FunctionCall":
+        self._check_no_web_url("map")
```
devennavani: `map` -> `spawn_map`?

test/function_test.py
```diff
@@ -1199,6 +1199,17 @@ def test_spawn_map_sync(client, servicer):
...
+    dummy_function = app.function()(dummy)
+    with servicer.intercept() as ctx:
+        with app.run(client=client):
+            dummy_function.experimental_spawn_map([1, 2, 3])
+
+        # Verify the correct invocation type was used
+        function_put_inputs = ctx.pop_request("FunctionPutInputs")
+        assert function_put_inputs is not None
+
+
```
devennavani: update test

modal/_functions.py
```diff
@@ -1543,6 +1546,19 @@ async def _map(
                 async for item in stream:
                     yield item
 
+    @live_method
+    async def _spawn_map(self, input_queue: _SynchronizedQueue) -> "_FunctionCall":
```
devennavani: return type should match `spawn`

modal/_functions.py
```diff
@@ -1543,6 +1546,19 @@ async def _map(
...
+        if self._is_generator:
+            raise InvalidError("A generator function cannot be called with `.spawn_map(...)`.")
+
+        assert self._function_name
+        return await _spawn_map_invocation(
+            self,
+            input_queue,
+            self.client,
+        )
+
```
devennavani: construct FunctionCall from function_call_id returned by `_spawn_map_invocation`, using `_FunctionCall._new_hydrated(`

modal/parallel_map.py
```diff
@@ -260,6 +260,89 @@ async def retry_inputs(self):
...
+        client: "modal.client._Client",
+        *,
+        input_queue: asyncio.Queue,
+        function: "modal.functions._Function",
+        function_call_id: str,
+    ):
+        super().__init__(client, input_queue=input_queue, function=function, function_call_id=function_call_id)
+
+
+async def _spawn_map_invocation(
```
devennavani, line 275: this should return a str function_call_id

devennavani (approved):
LGTM

PR #3310, opened 2025-07-09
[SpawnMapV2](2) Implement `.get(i)` for FunctionCalls.

CyrilSharma:
Adds the `.get(i)` method to the FunctionCall class. The purpose of the index is for batch async jobs (spawn_map).

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1141,6 +1141,8 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
                     target_concurrent_inputs=options.target_concurrent_inputs,
                     batch_max_size=options.batch_max_size,
                     batch_linger_ms=options.batch_wait_ms,
+                    scheduler_placement=options.scheduler_placement,
+                    cloud_provider_str=options.cloud,
                 )
             else:
                 options_pb = None
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -12,3 +12,4 @@
 1. Compile protobuf files: `inv protoc`
 1. Install the repo in editable mode: `pip install -e .`
 1. Build type Python stubs and check types: `inv type-check`
+1. Install pre-commit: `pre-commit install`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ This changelog documents user-facing updates (features, enhancements, fixes, and
 
 <!-- NEW CONTENT GENERATED BELOW. PLEASE PRESERVE THIS COMMENT. -->
 
+#### 1.1.2.dev31 (2025-08-11)
+
+- Allow multi-node Class-based functions to use RDMA
+
+
+#### 1.1.2.dev29 (2025-08-11)
+
+- Added `.objects.create()` methods
+
+
+#### 1.1.2.dev23 (2025-08-07)
+
+- `Cls.with_options` now supports `region` and `cloud` keyword arguments to override `region` or `cloud` during runtime.
+
+
 #### 1.1.2.dev16 (2025-08-06)
 
 - Added `.objects.delete()` methods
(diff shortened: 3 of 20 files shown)


modal/_functions.py
```diff
@@ -1906,6 +1905,111 @@ async def gather(*function_calls: "_FunctionCall[T]") -> typing.Sequence[T]:
             raise exc
 
 
+class _BatchFunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
```
mwaskom: Did we discuss having multiple distinct object types that share a type prefix? I'm not sure this is a good idea.

CyrilSharma: We haven't discussed it yet. My opinion is we shouldn't change the type prefix as all backend logic tries to treat batch function calls as similar to regular function calls as possible (only one table for outputs, inputs, and metadata keyed by function-call id, identical validation logic, etc.).

Would a better API be to create an "is_batch_function_call" attribute on the FunctionCall object, move all the BatchFunctionCall methods to FunctionCall, and throw errors if users attempt to call, `.iter()`, `.status()`, or `.get()` on something that wasn't a batch call?

mwaskom: Not sure I understand the proposal but it sounds pretty clunky to me.

modal/_functions.py
```diff
@@ -1779,16 +1786,19 @@ class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
     def _invocation(self):
         return _Invocation(self.client.stub, self.object_id, self.client)
 
-    async def get(self, timeout: Optional[float] = None) -> ReturnType:
+    async def get(self, timeout: Optional[float] = None, index: int = 0) -> ReturnType:
```
mwaskom: Unfortunate that we weren't already requiring keyword arguments here, but can we add it before `index`?

modal/_functions.py
```diff
@@ -1779,16 +1786,19 @@ class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
...
         """Get the result of the function call.
 
+        Calling `get` without an `index` will return the first output of the function call.
```
mwaskom: I think this is implied by the signature and the next line.

modal/_functions.py
```diff
@@ -1779,16 +1786,19 @@ class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
...
+        Calling `get` with an `index` will return the index-th output of the function call.
```
mwaskom: You may want to expand this documentation here a little bit. The key thing a user needs to understand is that, when they've used `spawn_map`, this method retrieves the result for an individual input not the entire batch result.

One downside to shipping this separately from the changes to `Function.spawn_map` is that we can't demonstrate it with an example, which would be the best way to make the point.

modal/_functions.py
```diff
@@ -1506,7 +1513,7 @@ async def _map(
                 yield item
 
     @live_method
-    async def _experimental_spawn_map(self, input_queue: _SynchronizedQueue) -> None:
+    async def _experimental_spawn_map(self, input_queue: _SynchronizedQueue) -> str:
```
mwaskom: What is the `str` that's returned here? Shouldn't this return a `modal.FunctionCall` like `Function.spawn`?

mwaskom:
> Adds the .get(i) method to the FunctionCall class

This would get the first index with a timeout of `i` right?

modal/_functions.py
```diff
@@ -1824,16 +1832,19 @@ class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
     def _invocation(self):
         return _Invocation(self.client.stub, self.object_id, self.client)
 
-    async def get(self, timeout: Optional[float] = None) -> ReturnType:
+    async def get(self, timeout: Optional[float] = None, *, index: int = 0) -> ReturnType:
         """Get the result of the function call.
```
devennavani: `Get the result of the function call` -> `Get the result of the index-th input of the function call`?

devennavani (approved):
LGTM

PR #3311, opened 2025-07-09
Added container exec timeout logic

mattnappo:
Introduce a client-side timer to ensure that the Sandbox.exec timeout parameter is respected. The timer begins right after `ContainerExecRequest` is sent to the server, and is checked within Sandbox `.wait` and `.poll` right before the `ContainerExecWaitRequest` RPC.

Context:
* User was experiencing sandbox exec hanging indefinitely ([context](https://modal-com.slack.com/archives/C07JNJMJPQC/p1751923161013099))
* See design discussion [here](https://modal-com.slack.com/archives/C07JNJMJPQC/p1752077999250329)

## Changelog
* Fixes a bug where the specified timeout for a `Sandbox.exec` is not respected by `wait()` or `poll()`.

### DIFF
--- a/modal/container_process.py
+++ b/modal/container_process.py
@@ -112,7 +112,6 @@ class _ContainerProcess(Generic[T]):
         return None
 
     async def _wait_for_completion(self) -> int:
-        """Wait for the container process to finish running."""
         while True:
             req = api_pb2.ContainerExecWaitRequest(exec_id=self._process_id, timeout=10)
             resp: api_pb2.ContainerExecWaitResponse = await retry_transient_errors(
@@ -126,18 +125,15 @@ class _ContainerProcess(Generic[T]):
         if self._returncode is not None:
             return self._returncode
 
-        if self._exec_deadline:
-            try:
-                remaining = self._exec_deadline - time.monotonic()
-                if remaining <= 0:
+        try:
+            timeout = None
+            if self._exec_deadline:
+                timeout = self._exec_deadline - time.monotonic()
+                if timeout <= 0:
                     raise TimeoutError()
-                self._returncode = await asyncio.wait_for(self._wait_for_completion(), timeout=remaining)
-            except (asyncio.TimeoutError, TimeoutError):
-                self._returncode = -1
-                return self._returncode
-        else:
-            self._returncode = await self._wait_for_completion()
-
+                self._returncode = await asyncio.wait_for(self._wait_for_completion(), timeout=timeout)
+        except (asyncio.TimeoutError, TimeoutError):
+            self._returncode = -1
         return self._returncode
 
     async def attach(self):
--- a/modal/container_process.py
+++ b/modal/container_process.py
@@ -96,15 +96,15 @@ class _ContainerProcess(Generic[T]):
         """
         if self._returncode is not None:
             return self._returncode
-
         if self._exec_deadline >= time.monotonic():
             self._returncode = -1
-            raise TimeoutError("container timed out while running exec")
+            return self._returncode
 
         req = api_pb2.ContainerExecWaitRequest(exec_id=self._process_id, timeout=0)
         resp: api_pb2.ContainerExecWaitResponse = await retry_transient_errors(self._client.stub.ContainerExecWait, req)
 
         if resp.completed:
+            # TODO(matt): In the future, it would be nice to raise a ContainerExecTimeoutError
             self._returncode = resp.exit_code
             return self._returncode
 
@@ -136,9 +136,9 @@ class _ContainerProcess(Generic[T]):
 
             return self._returncode
 
-        except asyncio.TimeoutError:
+        except asyncio.TimeoutError | TimeoutError:
             self._returncode = -1
-            raise TimeoutError("container timed out while running exec")
+            return self._returncode
 
     async def attach(self):
         """mdmd:hidden"""
--- a/modal/container_process.py
+++ b/modal/container_process.py
@@ -39,7 +39,7 @@ class _ContainerProcess(Generic[T]):
     ) -> None:
         self._process_id = process_id
         self._client = client
-        self._exec_deadline = self._exec_deadline
+        self._exec_deadline = exec_deadline
         self._text = text
         self._by_line = by_line
         self._stdout = _StreamReader[T](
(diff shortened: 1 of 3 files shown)


modal/container_process.py
```diff
@@ -31,6 +34,8 @@ def __init__(
         client: _Client,
         stdout: StreamType = StreamType.PIPE,
         stderr: StreamType = StreamType.PIPE,
+        timeout_secs: Optional[int] = None,
+        exec_start: Optional[datetime] = None,
```
thundergolfer: Could these be combined into one field, `exec_deadline` which is expressed as a timestamp in the future?

modal/container_process.py
```diff
@@ -93,6 +101,10 @@ async def poll(self) -> Optional[int]:
         if self._returncode is not None:
             return self._returncode
 
+        if self._timeout and datetime.now() >= self._exec_start + self._timeout:
```
thundergolfer: Using `datetime `is likely not valid, because this clock can go backwards. I think you want https://docs.python.org/3/library/time.html#time.monotonic

modal/container_process.py
```diff
@@ -93,6 +101,10 @@ async def poll(self) -> Optional[int]:
...
+            self._returncode = -1
+            raise TimeoutError("container timed out while running exec")
```
thundergolfer: Is this the same behavior as previous?

thundergolfer (commented):
We need some tests here. Also I'd venture that TDD could help with checking that the behavior isn't modified from the status quo (with only worker-side timeouts) to this diff (worker-side plus client-side timeout).

thundergolfer:
@mattnappo seeing this error in CI

```
>       except asyncio.TimeoutError | TimeoutError:
E       TypeError: catching classes that do not inherit from BaseException is not allowed
```

modal/sandbox.py
```diff
@@ -655,7 +656,16 @@ async def exec(
         )
         resp = await retry_transient_errors(self._client.stub.ContainerExec, req)
         by_line = bufsize == 1
-        return _ContainerProcess(resp.exec_id, self._client, stdout=stdout, stderr=stderr, text=text, by_line=by_line)
+        exec_deadline = time.monotonic() + timeout if timeout else None
```
thundergolfer: As discussed we need to add a tolerance here for latency between the client and the worker to avoid missing messages sent before the specificed timeout on the worker side but which are received after the specified timeout on the client side.

modal/sandbox.py
```diff
@@ -49,6 +50,12 @@
 # e.g. 'runsc exec ...'. So we use 2**16 as the limit.
 ARG_MAX_BYTES = 2**16
 
+
+# A user-supplied timeout of n seconds on a ContainerExec will be extended by this buffer.
+# This is introduced to minimize the effects of a race condition between the Sandbox.exec timeout
+# handler on the client side and the receiving of server RPCs carrying status codes or IO data.
+CONTAINER_EXEC_TIMEOUT_BUFFER = 5
+
```
mattnappo:
```suggestion

# This buffer extends the user-supplied timeout on ContainerExec-related RPCs. This was introduced to
# (hopefully) allow any in-flight status codes/IO data to reach the client before the stream is closed.
CONTAINER_EXEC_TIMEOUT_BUFFER = 5
```

modal/sandbox.py
```diff
@@ -49,6 +50,10 @@
 # e.g. 'runsc exec ...'. So we use 2**16 as the limit.
 ARG_MAX_BYTES = 2**16
 
+
+# This buffer extends the user-supplied timeout on ContainerExec-related RPCs. This was introduced to
+# (hopefully) allow any in-flight status codes/IO data to reach the client before the stream is closed.
```
mattnappo:
```suggestion
# give any in-flight status codes/IO data more time to reach the client before the stream is closed.
```

modal/sandbox.py
```diff
@@ -49,6 +50,10 @@
 # e.g. 'runsc exec ...'. So we use 2**16 as the limit.
 ARG_MAX_BYTES = 2**16
 
+
```
mattnappo, line 53: Thoughts on this description and name?

modal/sandbox.py
```diff
@@ -655,7 +660,16 @@ async def exec(
         )
         resp = await retry_transient_errors(self._client.stub.ContainerExec, req)
         by_line = bufsize == 1
-        return _ContainerProcess(resp.exec_id, self._client, stdout=stdout, stderr=stderr, text=text, by_line=by_line)
+        exec_deadline = time.monotonic() + timeout + CONTAINER_EXEC_TIMEOUT_BUFFER if timeout else None
```
thundergolfer: Do we need to `int(timeout)` here to ensure both the server/worker and client are using the same value?

modal/container_process.py
```diff
@@ -102,22 +111,37 @@ async def poll(self) -> Optional[int]:
...
         return None
 
-    async def wait(self) -> int:
-        """Wait for the container process to finish running. Returns the exit code."""
-
-        if self._returncode is not None:
-            return self._returncode
-
+    async def _wait_for_completion(self) -> int:
+        """Wait for the container process to finish running."""
```
thundergolfer:
```suggestion
```

modal/container_process.py
```diff
@@ -102,22 +111,37 @@ async def poll(self) -> Optional[int]:
...
+        """Wait for the container process to finish running. Returns the exit code."""
+        if self._returncode is not None:
+            return self._returncode
+
+        if self._exec_deadline:
+            try:
+                remaining = self._exec_deadline - time.monotonic()
+                if remaining <= 0:
+                    raise TimeoutError()
+                self._returncode = await asyncio.wait_for(self._wait_for_completion(), timeout=remaining)
```
thundergolfer: I think you can simplify the code with this behavior when `None` is passed: 

> timeout can either be None or a float or int number of seconds to wait for. If timeout is None, block until the future completes.

mattnappo: do you mean something like
```python
        if not self._exec_deadline:
            self._returncode = await self._wait_for_completion()
            return self._returncode
        
        try:
            remaining = self._exec_deadline - time.monotonic()
            if remaining <= 0:
                raise TimeoutError()
            self._returncode = await asyncio.wait_for(self._wait_for_completion(), timeout=remaining)
        except (asyncio.TimeoutError, TimeoutError):
            self._returncode = -1
            return self._returncode
```

thundergolfer (approved):

PR #3312, opened 2025-07-09
reload_volumes() accepts volume_ids

AmitPr:
## Describe your changes
Adds optional `volume_ids` to `modal.Sandbox.reload_volumes`
- PRD-957

---

## Compatibility checklist

---

</details>

## Changelog

* `modal.Sandbox.reload_volumes` accepts `volume_ids` to reload specific volumes. All volumes are reloaded if not specified.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -538,13 +538,17 @@ class _Sandbox(_Object, type_prefix="sb"):
 
         return self._tunnels
 
-    async def reload_volumes(self) -> None:
-        """Reload all Volumes mounted in the Sandbox."""
+    async def reload_volumes(self, volume_ids: list[str] | None = None) -> None:
+        """Reloads volumes mounted in the Sandbox.
+
+        If `volume_ids` is provided, only reloads the specified volumes, otherwise reloads all volumes.
+        """
         task_id = await self._get_task_id()
         await retry_transient_errors(
             self._client.stub.ContainerReloadVolumes,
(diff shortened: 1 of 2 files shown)


mwaskom:
I guess that could work, not sure if it would also make sense to parameterize it in terms of the mount points that you want to reload. If you go in that direction it would be necessary for this method to have internal access to the mount point -> volume dict which may not currently be held in an accessible location.

mwaskom:
Given that `.reload()` is a method on the Volume and that's how you use it inside a Function I think passing the Volume to a `Sandbox.reload_volumes()` would make sense.

modal/sandbox.py
```diff
@@ -538,12 +538,13 @@ async def tunnels(self, timeout: int = 50) -> dict[int, Tunnel]:
...
 
-    async def reload_volumes(self, volume_ids: list[str] | None = None) -> None:
+    async def reload_volumes(self, volumes: list[_Volume] | None = None) -> None:
         """Reloads volumes mounted in the Sandbox.
 
-        If `volume_ids` is provided, only reloads the specified volumes, otherwise reloads all volumes.
+        If `volumes` is provided, only reloads the specified volumes, otherwise reloads all volumes.
         """
         task_id = await self._get_task_id()
+        volume_ids = [volume.object_id for volume in volumes] if volumes else None
```
mwaskom, line 547: Something to consider ... will Volume objects always be hydrated when they're passed to this method?

PR #3315, opened 2025-07-10
Add basic experimental public interface for getting App layout

mwaskom:
This is a useful feature to expose, but there are some complicated reasons why we can't implement the desired stable API right now, so this adds it as an experimental function with a provisional API:

```python
>>> modal.experimental.get_app_objects("my-app")
{"MyClass": modal.Cls.from_name("my-app", "MyClass"), "f": modal.Function.from_name("my-app", "f")}
```

Closes CLI-501

### DIFF
--- a/modal/experimental/__init__.py
+++ b/modal/experimental/__init__.py
@@ -17,8 +17,9 @@ from .._tunnel import _forward as _forward_tunnel
 from .._utils.async_utils import synchronize_api, synchronizer
 from .._utils.deprecation import deprecation_warning
 from .._utils.grpc_utils import retry_transient_errors
+from ..app import _App
 from ..client import _Client
-from ..cls import _Obj
+from ..cls import _Cls, _Obj
 from ..config import logger
 from ..exception import InvalidError
 from ..image import DockerfileSpec, ImageBuilderVersion, _Image, _ImageRegistryConfig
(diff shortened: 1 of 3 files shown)


modal/experimental/__init__.py
```diff
@@ -88,6 +89,53 @@ async def list_deployed_apps(environment_name: str = "", client: Optional[_Clien
...
+    app_name: str, *, environment_name: Optional[str] = None, client: Optional[_Client] = None
+) -> dict[str, Union[_Function, _Cls]]:
+    """Experimental interface for retrieving a dictionary of the Functions / Clses in an App.
+
+    The return value is a dictionary mapping names to Function or Cls objects.
+
+    We plan to support this functionality through a stable API in the future. It's likely that
+    the stable API will look different (it will probably be a method on the App object itself).
+    Additionally, note that the returned objects are currently not hydrated, but a future stable
+    API would likely return hydrated objects instead.
```
mwaskom: I guess we could run the handles through parallel explicit `.hydrate()` calls here , at the cost of another (longest in the set) round trip in this function (currently there are already 2 so it will be rather slow).

mwaskom: Actually! It doesn't really make sense to pre-hydrate a Cls, since you don't know the parameters in this context. Bah.

PR #3316, opened 2025-07-10
Make Function.from_name repr have string arguments

mwaskom:
Noticed this minor blemish:

```python
$ python from_name_oops.py
Cls.from_name('synthetic-traffic', 'Server')
Function.from_name(synthetic-traffic, trigger)
```


mwaskom:
@prbot approve

PR #3317, opened 2025-07-10
Implement AppGetOrCreate in mock server

mwaskom:
This behavior is slightly closer to what the actual server does.


mwaskom:
@prbot approve

PR #3319, opened 2025-07-10
Client Changes: map basic input plane

bokeefe123:


modal/_functions.py
```diff
@@ -401,7 +402,9 @@ async def create(
             parent_input_id=current_input_id() or "",
             input=input_item,
         )
-        metadata = await _InputPlaneInvocation._get_metadata(input_plane_region, client)
+
+        token = await client._auth_token_manager.get_token()
+        metadata = [("x-modal-input-plane-region", input_plane_region), ("x-modal-auth-token", token)]
```
thecodingwizard: hmm ... I don't particularly love how the metadata is duplicated so many times in this codebase, but I'm not super sure what else to do. Maybe it's OK for now?

(cc @rculbertson)

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
 
+async def _map_invocation_inputplane(
+    function: "modal.functions._Function",
+    raw_input_queue: _SynchronizedQueue,
+    client: "modal.client._Client",
+    order_outputs: bool,
+    return_exceptions: bool,
+    wrap_returned_exceptions: bool,
+    count_update_callback: Optional[Callable[[int, int], None]],
+    function_call_invocation_type: "api_pb2.FunctionCallInvocationType.ValueType",
```
thecodingwizard: We don't need `function_call_invocation_type` anymore right?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+    `FunctionMap` / `FunctionPutInputs` / `FunctionGetOutputs` RPCs it speaks
+    the input-plane protocol consisting of `MapStartOrContinue` and `MapAwait`.
+
+    The implementation purposefully ignores retry handling for now – a stub is
+    left in place so that a future change can add support for the retry path
+    without re-structuring the surrounding code.
+    """
+
+    assert function._input_plane_url, "_map_invocation_inputplane should only be used for input-plane backed functions"
+    assert client.stub, "Client must be hydrated with a stub"
```
thecodingwizard:
```suggestion
```

input plane shouldn't use client stub right?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+
+    input_plane_stub = await client.get_stub(function._input_plane_url)
+    input_id_to_idx: dict[str, int] = {}
```
thecodingwizard: what's the purpose of this map?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+
+    # ------------------------------------------------------------
+    # Helper structures
+    # ------------------------------------------------------------
+
+    @dataclass
+    class _QueueItem:
```
thecodingwizard: Long-term it's probably best to move this out of this function, but I think it's fine for now (input plane is still experimental anyway)

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+        """Wrapper for items that should be sent in the next MapStartOrContinue.
+
+        If `attempt_token` is `None` this is a *new* input, otherwise it
+        represents a retry of a previous attempt.
+        """
+
+        put_inputs_item: api_pb2.FunctionPutInputsItem
+        # idx: int
```
thecodingwizard: idx here isn't needed... right? it can be included in `attempt_token`

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+        attempt_token: str | None = None  # populated for retries
+
+        def to_start_or_continue_item(self) -> api_pb2.MapStartOrContinueItem:
```
thecodingwizard: idea: `to_proto` instead? I have no strong preference, you can pick

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+                    output_val = modal.exception.UserCodeException(exc)
+                else:
+                    output_val = exc
+            else:
+                raise exc
+
+        return (item.idx, output_val)
+
+    async def poll_outputs():
+        received: dict[int, Any] = {}
```
thecodingwizard: could we add a brief comment explaining what the key and value is?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+        next_idx = 1  # 1-indexed map call idx
+        async with aclosing(
+            async_map_ordered(get_all_outputs_and_clean_up(), fetch_output, concurrency=BLOB_MAX_PARALLELISM)
```
thecodingwizard, line 658: hmm doesn't `async_map_ordered` already return an ordered list of function outputs?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+    # ------------------------------------------------------------
+    # Coroutine: convert FunctionGetOutputsItem → actual result value
+    # ------------------------------------------------------------
+
+    async def fetch_output(item: api_pb2.FunctionGetOutputsItem) -> tuple[int, Any]:
+        try:
+            output_val = await _process_result(item.result, item.data_format, input_plane_stub, client)
+        except Exception as exc:
+            if return_exceptions:
+                if wrap_returned_exceptions:
```
thecodingwizard: The comment in the existing client code says

```
                    # Prior to client 1.0.4 there was a bug where return_exceptions would wrap
                    # any returned exceptions in a synchronicity.UserCodeException. This adds
                    # deprecated non-breaking compatibility bandaid for migrating away from that:
```

given that this code is only used by new clients, is this still necessary?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+            pass
+
+    # ------------------------------------------------------------
+    # Coroutine: stream outputs via MapAwait
+    # ------------------------------------------------------------
+
+    async def get_all_outputs():
+        """Continuously fetch outputs until the map is complete."""
+
+        assert client.stub
```
thecodingwizard:
```suggestion
```

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+        last_entry_id = "0-0"
+        while not map_done_event.is_set():
+            if function_call_id is None:
+                # We haven't received the first MapStartOrContinueResponse yet.
+                await asyncio.sleep(0.05)
```
thecodingwizard: hmm ... maybe we can use another event or some asyncio queue or something other than sleeping in a while loop?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+                token = await client._auth_token_manager.get_token()
+                metadata = [("x-modal-input-plane-region", function._input_plane_region), ("x-modal-auth-token", token)]
+                response: api_pb2.MapAwaitResponse = await retry_transient_errors(
+                    input_plane_stub.MapAwait,
+                    request,
+                    max_retries=20,
+                    attempt_timeout=OUTPUTS_TIMEOUT + ATTEMPT_TIMEOUT_GRACE_PERIOD,
+                    metadata=metadata,
+                )
+                # print(f"MapAwaitResponse: {response}")
```
thecodingwizard:
```suggestion
```

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+            except Exception:
+                # If the RPC failed we retry in next loop iteration.
+                continue
```
thecodingwizard: is this logic present in the Python codebase? Do we want to retry next loop iteration or should we just fail here (we are retrying 20 times above already)?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+    # ------------------------------------------------------------
+    # Coroutine: **stub** – retry handling will be added in the future
+    # ------------------------------------------------------------
+
+    async def retry_inputs():
+        """Simple ticker that prints 'hello' every second while the map is running."""
+
+        try:
+            while not map_done_event.is_set():
+                # print("hello")
```
thecodingwizard:
```suggestion
```

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+                attempt_tokens[item.idx] = item.attempt_token
+
+        yield
+
+    # ------------------------------------------------------------
+    # Coroutine: **stub** – retry handling will be added in the future
+    # ------------------------------------------------------------
+
+    async def retry_inputs():
+        """Simple ticker that prints 'hello' every second while the map is running."""
```
thecodingwizard:
```suggestion
        """Temporary stub for retrying inputs. Retry handling will be added in the future."""
```

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+
+        try:
+            while not map_done_event.is_set():
+                # print("hello")
+                await asyncio.sleep(1)
+                # Keep generator semantics for async_merge; value is ignored.
+                yield
```
thecodingwizard: just for my understanding, why does this generator need to yield once a second? what happens if this never yields until the generator exits?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+    # ------------------------------------------------------------
+    # Coroutine: send queued items to the input-plane server
+    # ------------------------------------------------------------
+
+    async def pump_inputs():
+        nonlocal function_call_id, max_inputs_outstanding
+
+        async for batch in queue_batch_iterator(queue, max_batch_size=MAP_INVOCATION_CHUNK_SIZE):
+            # Convert the queued items into the proto format expected by the RPC.
+            # input_id -> idx
```
thecodingwizard: what exactly is input id -> idx?

modal/parallel_map.py
```diff
@@ -424,6 +424,333 @@ def log_stats():
...
+            request_items: list[api_pb2.MapStartOrContinueItem] = [qi.to_start_or_continue_item() for qi in batch]
+            # Build request
+            request = api_pb2.MapStartOrContinueRequest(
+                function_id=function.object_id,
+                function_call_id=function_call_id if function_call_id else None,
```
thecodingwizard:
```suggestion
                function_call_id=function_call_id,
```

modal/parallel_map.py
```diff
@@ -424,6 +424,331 @@ def log_stats():
...
+            map_done_event.set()
+
+    async def create_input(argskwargs):
+        idx = inputs_created + 1  # 1-indexed map call idx
+        update_counters(created_delta=1)
+        (args, kwargs) = argskwargs
+        put_item: api_pb2.FunctionPutInputsItem = await _create_input(
+            args,
+            kwargs,
+            input_plane_stub,
```
thecodingwizard: `_create_input` should be called against client stub, not input plane stub 😰 (we should probably clearly document this in `_create_input`; I've made this mistake twice

modal/parallel_map.py
```diff
@@ -424,6 +424,331 @@ def log_stats():
...
+    # ------------------------------------------------------------
+    # Coroutine: drain user input iterator, upload blobs, enqueue for sending
+    # ------------------------------------------------------------
+
+    async def input_iter():
+        while True:
+            raw = await raw_input_queue.get()
+            if raw is None:
+                break
+            yield raw
```
thecodingwizard: If you copied this from the existing python code, could you also preserve the comments (about what raw is, and what raw_input being None means)?

modal_proto/api.proto
```diff
@@ -611,7 +611,65 @@ message AttemptStartRequest {
...
+  FunctionPutInputsItem input = 1;
+  string attempt_token = 2;
+}
+
+message MapStartOrContinueItem {
+  oneof input_oneof {
+    FunctionPutInputsItem input = 1;
+    MapContinueRetryItem retry_item = 2;
+  }
+}
```
thecodingwizard: Maybe we can simplify this and remove the oneof, just keeping "input" and an optional retry_attempt_token?

modal_proto/api.proto
```diff
@@ -611,7 +611,65 @@ message AttemptStartRequest {
...
+  string parent_input_id = 2;
+  optional string function_call_id = 3;
+  repeated MapStartOrContinueItem items = 4;
+}
+
+message MapStartOrContinueResponseItem {
+    int32 idx = 1;
+    string input_id = 2;
+    string attempt_token = 3;
+  }
```
thecodingwizard: the indentation looks off

modal_proto/api.proto
```diff
@@ -611,7 +611,65 @@ message AttemptStartRequest {
...
+  repeated FunctionGetOutputsItem outputs = 1;
+  string last_entry_id = 2;
+}
+
+message MapCheckInputsRequest {
+  string function_id = 1;
+  message MapCheckInputsRequestItem {
+    string input_id = 1;
+    string attempt_token = 2;
+  }
```
thecodingwizard: doesn't the attempt token include the input ID?

modal_proto/api.proto
```diff
@@ -611,7 +611,65 @@ message AttemptStartRequest {
...
+  repeated MapCheckInputsRequestItem items = 2;
+}
+
+message MapCheckInputsResponse {
+  repeated string lost_attempt_tokens = 1;
 }
```
thecodingwizard: I think we can maybe simplify things if we return a list of booleans corresponding 1:1 with the given attempt tokens. True if the input is still running and False if the input is no longer running or something.

Feel free to also move the MapCheckInputsRequest/Response into a separate PR if that's simpler.

modal_proto/api.proto
```diff
@@ -1484,6 +1542,17 @@ message FunctionCallCancelRequest {
...
+message FunctionCallCancelInputPlaneRequest {
+  string function_id = 1;
+  string function_call_id = 2;
+}
+
+message FunctionCallCancelMaybeInputPlaneRequest {
+  string function_id = 1;
+  string function_call_id = 2;
+  bool terminate_containers = 3;
+}
```
thecodingwizard: Can we move this to a separate PR / is this still needed?

modal_proto/api.proto
```diff
@@ -1841,6 +1911,8 @@ message FunctionPutOutputsItem {
   double output_created_at = 4;
   DataFormat data_format = 7; // for result.data_oneof
   uint32 retry_count = 8;
+  string function_call_id = 9; // injected by the worker 
+  optional int32 function_map_idx = 10;
```
thecodingwizard: function_map_idx is also injected by the worker right?

modal_proto/api.proto
```diff
@@ -1651,6 +1720,7 @@ message FunctionGetInputsItem {
   string function_call_id = 5;
   FunctionCallInvocationType function_call_invocation_type = 6;
   uint32 retry_count = 7;
+  optional int32 function_map_idx = 8;
```
thecodingwizard: can we add a comment saying this is intercepted and only used by the worker? (assuming this is indeed the case)

modal_proto/api.proto
```diff
@@ -3358,6 +3435,7 @@ service ModalClient {
   rpc FunctionAsyncInvoke(FunctionAsyncInvokeRequest) returns (FunctionAsyncInvokeResponse);
   rpc FunctionBindParams(FunctionBindParamsRequest) returns (FunctionBindParamsResponse);
   rpc FunctionCallCancel(FunctionCallCancelRequest) returns (google.protobuf.Empty);
+  rpc FunctionCallCancelMaybeInputPlane(FunctionCallCancelMaybeInputPlaneRequest) returns (google.protobuf.Empty);
```
thecodingwizard: not needed?

modal/parallel_map.py
bokeefe123: It causes an AttributeError without the yield. I'm not certain on the semantics. However, I'll introduce a false path to resolve the error:
```
async def retry_inputs():
  """Temporary stub for retrying inputs. Retry handling will be added in the future."""
  
  try:
      while not map_done_event.is_set():
          await asyncio.sleep(1)
          if False:
              yield
  except asyncio.CancelledError:
      pass
```

modal/parallel_map.py
bokeefe123: Should I retain this due to the _create_input need for client.stub?

modal_proto/api.proto
bokeefe123: Sounds reasonable. I'll remove from this PR for now.

modal/_functions.py
```diff
@@ -401,7 +402,8 @@ async def create(
             parent_input_id=current_input_id() or "",
             input=input_item,
         )
-        metadata = await _InputPlaneInvocation._get_metadata(input_plane_region, client)
+
+        metadata = await client.get_input_plane_metadata(input_plane_region)
```
thecodingwizard, line 404: nice!

modal/parallel_map.py
thecodingwizard: Oops, good catch. Yes, let's retain this, thanks!

modal/parallel_map.py
```diff
@@ -364,32 +364,18 @@ async def fetch_output(item: api_pb2.FunctionGetOutputsItem) -> tuple[int, Any]:
...
+                async_map_ordered(get_all_outputs_and_clean_up(), fetch_output, concurrency=BLOB_MAX_PARALLELISM)
+            ) as streamer:
+                async for _, output in streamer:
+                    yield _OutputValue(output)
+        else:
+            async with aclosing(
+                async_map(get_all_outputs_and_clean_up(), fetch_output, concurrency=BLOB_MAX_PARALLELISM)
+            ) as streamer:
+                async for _, output in streamer:
                     yield _OutputValue(output)
```
thecodingwizard: hmm, can we remove this change from this PR?

I think this is correct, but:

1) would be good to check with the client team
2) needs to be included in the changelog (since this impacts normal `.map`, which is not experimental)
3) this change is distinct from the overall PR, making it harder to `git blame`, view git history, and revert

modal/parallel_map.py
```diff
@@ -424,6 +410,282 @@ def log_stats():
...
+    The implementation purposefully ignores retry handling for now - a stub is
+    left in place so that a future change can add support for the retry path
+    without re-structuring the surrounding code.
+    """
+
+    assert function._input_plane_url, "_map_invocation_inputplane should only be used for input-plane backed functions"
+
+    input_plane_stub = await client.get_stub(function._input_plane_url)
+
+    assert input_plane_stub, "Client must be hydrated with an input-plane stub"
```
thecodingwizard: We probably don't need this -- get_stub always returns a non-None object right?

modal/parallel_map.py
```diff
@@ -424,6 +410,282 @@ def log_stats():
...
+    assert client.stub, "Client must be hydrated with a stub for _map_invocation_inputplane"
+
+    # ------------------------------------------------------------
+    # Invocation-wide state
+    # ------------------------------------------------------------
+
+    have_all_inputs = False
```
thecodingwizard, line 457:
```suggestion
    submitted_all_inputs = False
```

Do you think this is better? Will let you decide.

modal/parallel_map.py
```diff
@@ -424,6 +410,282 @@ def log_stats():
...
+        except asyncio.CancelledError:
+            pass
+
+    # ------------------------------------------------------------
+    # Coroutine: stream outputs via MapAwait
+    # ------------------------------------------------------------
+
+    async def get_all_outputs():
+        """Continuously fetch outputs until the map is complete."""
+        last_entry_id = "0-0"
```
thecodingwizard: Oops sorry for backtracking on this -- I think it's not great to hardcode this in the client in case we change our backend implementation to not use Redis someday. Maybe let's make `last_entry_id` either an empty string or `None`, and hardcode `0-0` in the server? (I know this contradicts what I said in the other review -- on second thought I think hardcoding `0-0` in the client is worse than hardcoding it in the server since the client is harder to change than the server)

modal_proto/api.proto
```diff
@@ -611,7 +611,48 @@ message AttemptStartRequest {
...
 message AttemptStartResponse {
   string attempt_token = 1;
-  FunctionRetryPolicy retry_policy = 2;
+  FunctionRetryPolicy retry_policy = 2; // TODO(ben-okeefe) TODO(nathan): Not currently used
+}
+
+message MapContinueRetryItem {
+  FunctionPutInputsItem input = 1;
+  string attempt_token = 2;
+}
```
thecodingwizard: this isn't needed anymore right?

modal_proto/api.proto
```diff
@@ -611,7 +611,48 @@ message AttemptStartRequest {
...
+
+message MapStartOrContinueItem {
+  FunctionPutInputsItem input = 1;
+  optional string attempt_token = 2; // None if this is a fresh input, otherwise it is the attempt token for a retry.
+}
+
+message MapStartOrContinueRequest {
+  string function_id = 1;
+  string parent_input_id = 2;
+  optional string function_call_id = 3;
```
thecodingwizard: can we add a comment that this is given iff it is a continue request?

modal_proto/api.proto
```diff
@@ -3302,6 +3346,10 @@ service ModalClient {
   rpc AttemptRetry(AttemptRetryRequest) returns (AttemptRetryResponse);
   rpc AttemptStart(AttemptStartRequest) returns (AttemptStartResponse);
 
+  // Map
+  rpc MapStartOrContinue(MapStartOrContinueRequest) returns (MapStartOrContinueResponse);
+  rpc MapAwait(MapAwaitRequest) returns (MapAwaitResponse);
```
```suggestion
  // Input Plane Map
  rpc MapStartOrContinue(MapStartOrContinueRequest) returns (MapStartOrContinueResponse);
  rpc MapAwait(MapAwaitRequest) returns (MapAwaitResponse);
```

thecodingwizard (commented):
Nice!

modal/parallel_map.py
bokeefe123: I think have_all_inputs might be more correct because it's not when all the inputs have been submitted but rather when the input generator has been drained.

modal/parallel_map.py
bokeefe123: Sounds good adding a test to correspond to this change as well in MapAwait and comments on both trim and getMap about caller responsibility.

thecodingwizard (approved):

PR #3320, opened 2025-07-10
Adds `enable_gpu_snapshot` to proto definition

luiscape:
We are moving GPU memory snapshots from experimental to alpha. I am renaming the flag from `_experimental_enable_gpu_snapshot` to `enable_gpu_snapshot` to promote the interface. I am first making a proto change to add backend support and will update the client interface after.


mwaskom (approved):
FWIW

PR #3321, opened 2025-07-10
Document limitations of Sandbox FS APIs

ekzhang:
See also, [here](https://linear.app/modal-labs/issue/WRK-1081/provide-clear-caveats-in-filesystem-api-guide-and-reference-about) and [here](https://github.com/modal-labs/modal/pull/24426/files)

### DIFF
--- a/modal/file_io.py
+++ b/modal/file_io.py
@@ -125,7 +125,7 @@ class _FileIO(Generic[T]):
     may be limited in size to 100 MiB, and the throughput of requests is
     restricted in the current implementation. For our recommendations on large file transfers
     see the Sandbox [filesystem access guide](https://modal.com/docs/guide/sandbox-files).
-    
+
     **Usage**
 
     ```python


modal/file_io.py
```diff
@@ -117,10 +117,15 @@ class FileWatchEvent:
...
-    """FileIO handle, used in the Sandbox filesystem API.
+    """[Alpha] FileIO handle, used in the Sandbox filesystem API.
 
     The API is designed to mimic Python's io.FileIO.
 
+    Currently this API is in Alpha and is subject to change. File I/O operations
+    may be limited in size to 100 MiB, and the throughput of requests is
+    restricted in the current implementation. We recommend using
+    [Tunnels](https://modal.com/docs/guide/tunnels) for large file transfers.
+
```
saltzm:
```suggestion
    restricted in the current implementation. For our recommendations on large file transfers 
    see the Sandbox [filesystem access guide](https://modal.com/docs/guide/sandbox-files).
    
```

I think now that I'm pushing changes to the guide we can forward people to that - I think Volumes/Image APIs are still preferable to tunnels for a lot of use cases

PR #3323, opened 2025-07-11
Allow Modal Cls functions to get i6pn addresses

pawalt:
## Describe your changes

This extends i6pn support to Cls so users with lifecycle hooks can use i6pn.

---

## Compatibility checklist

---

## Release checklist

- [x] Version file (`modal_version/__init__.py`) has been updated with the next logical version
- [x] Changelog has been cleaned up and given an appropriate subhead

---

</details>

## Changelog

- Allows i6pn configuration for Modal Classes

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -863,6 +863,7 @@ class _App:
         # Limits the number of inputs a container handles before shutting down.
         # Use `max_inputs = 1` for single-use containers.
         max_inputs: Optional[int] = None,
+        i6pn: Optional[bool] = None,  # Whether to enable IPv6 container networking within the region.
         include_source: Optional[bool] = None,  # When `False`, don't automatically add the App source to the container.
         experimental_options: Optional[dict[str, Any]] = None,
         # Parameters below here are experimental. Use with caution!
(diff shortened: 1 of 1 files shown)


modal/app.py
```diff
@@ -863,6 +863,7 @@ def cls(
         # Limits the number of inputs a container handles before shutting down.
         # Use `max_inputs = 1` for single-use containers.
         max_inputs: Optional[int] = None,
+        i6pn: Optional[bool] = None,  # Whether to enable IPv6 container networking within the region.
```
luiscape, line 866: Can we not put the default value `False`?

mwaskom: For `@app.function` we have a null default because using `@modal.clustered` implies True:

https://github.com/modal-labs/modal-client/blob/279d01fac87cbd64a5c8422089c74e8306d1e787/modal/app.py#L712

pawalt: We can - I don't think it makes much of a difference either way. I was matching the pattern on Functions

mwaskom: Which I guess the `None` communicates _slightly_ better (even though `i6pn=False` is not respected for clustered functions 🤷‍♂️)

mwaskom (approved):

PR #3325, opened 2025-07-11
support Sandbox Names

rolyatmax:
This PR adds support for Sandbox names by adding a `name` keyword arg to `Sandbox.create()` and a `Sandbox.from_name()` static method. (Note: protobuf defs were updated in a [previous PR](https://github.com/modal-labs/modal-client/pull/3332).)

## Usage

```
app = App.lookup("my-app", create_if_missing=True)
sb = Sandbox.create(app=app, name="my-sandbox")
sb2 = Sandbox.create(app=app, name="my-sandbox") # raises an Error, cannot have more than one running sandbox with the same name in the same environment

# from_name() returns the _running_ sandbox with the given name (in the same environment)
sb2 = Sandbox.from_name("my-app", "my-sandbox")
assert sb.object_id == sb2.object_id

sb.terminate()
sb.wait(raise_on_termination=False)

Sandbox.from_name("my-app", "my-sandbox-name") # raises an Error because no running sandbox has the given name
```

---

## Compatibility checklist

---

</details>

## Changelog

- Added a `name` parameter to `Sandbox.create()`
- Added a `Sandbox.from_name()` static method.
- Added a `name` parameter to `Sandbox._experimental_from_snapshot()`

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -73,13 +73,12 @@ def _validate_exec_args(entrypoint_args: Sequence[str]) -> None:
         )
 
 
-"""
-This is a singleton class that represents the default sandbox name override.
-It is used to indicate that the sandbox name should not be overridden.
-"""
-
-
 class DefaultSandboxNameOverride:
+    """
+    This is a singleton class that represents the default sandbox name override.
+    It is used to indicate that the sandbox name should not be overridden.
+    """
+
     def __repr__(self) -> str:
         return "DefaultSandboxNameOverride"
 
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -74,8 +74,8 @@ def _validate_exec_args(entrypoint_args: Sequence[str]) -> None:
 
 
 class DefaultSandboxNameOverride:
-    """
-    This is a singleton class that represents the default sandbox name override.
+    """A singleton class that represents the default sandbox name override.
+
     It is used to indicate that the sandbox name should not be overridden.
     """
 
@@ -472,10 +472,10 @@ class _Sandbox(_Object, type_prefix="sb"):
         environment_name: Optional[str] = None,
         client: Optional[_Client] = None,
     ) -> "_Sandbox":
-        """Return the running Sandbox in the given app with the given name. Raises an error if no running
-        sandbox is found with the given name.
+        """Get a running Sandbox by name from the given app.
 
-        A Sandbox's name is the `name` argument passed to `Sandbox.create`.
+        Raises an error if no running sandbox is found with the given name. A Sandbox's name
+        is the `name` argument passed to `Sandbox.create`.
         """
         if client is None:
             client = await _Client.from_env()
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -93,6 +93,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         entrypoint_args: Sequence[str],
         image: _Image,
         secrets: Sequence[_Secret],
+        name: Optional[str] = None,
         timeout: Optional[int] = None,
         workdir: Optional[str] = None,
         gpu: GPU_T = None,
@@ -216,6 +217,7 @@ class _Sandbox(_Object, type_prefix="sb"):
                 proxy_id=(proxy.object_id if proxy else None),
                 enable_snapshot=enable_snapshot,
                 verbose=verbose,
+                name=name,
                 experimental_options=experimental_options,
             )
 
@@ -231,6 +233,7 @@ class _Sandbox(_Object, type_prefix="sb"):
     async def create(
         *entrypoint_args: str,
         app: Optional["modal.app._App"] = None,  # Optionally associate the sandbox with an app
+        name: Optional[str] = None,  # Optionally give the sandbox a name. Unique within an app.
         image: Optional[_Image] = None,  # The image to run as the container for the sandbox.
         secrets: Sequence[_Secret] = (),  # Environment variables to inject into the sandbox.
         network_file_systems: dict[Union[str, os.PathLike], _NetworkFileSystem] = {},
@@ -295,6 +298,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         return await _Sandbox._create(
             *entrypoint_args,
             app=app,
+            name=name,
             image=image,
             secrets=secrets,
             network_file_systems=network_file_systems,
@@ -324,6 +328,7 @@ class _Sandbox(_Object, type_prefix="sb"):
     async def _create(
         *entrypoint_args: str,
         app: Optional["modal.app._App"] = None,  # Optionally associate the sandbox with an app
+        name: Optional[str] = None,  # Optionally give the sandbox a name. Unique within an app.
         image: Optional[_Image] = None,  # The image to run as the container for the sandbox.
         secrets: Sequence[_Secret] = (),  # Environment variables to inject into the sandbox.
         mounts: Sequence[_Mount] = (),
@@ -376,6 +381,7 @@ class _Sandbox(_Object, type_prefix="sb"):
             entrypoint_args,
             image=image or _default_image,
             secrets=secrets,
+            name=name,
             timeout=timeout,
             workdir=workdir,
             gpu=gpu,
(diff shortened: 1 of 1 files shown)


modal/sandbox.py
```diff
@@ -435,6 +442,28 @@ def _hydrate_metadata(self, handle_metadata: Optional[Message]):
...
+        environment_name: Optional[str] = None,
+        client: Optional[_Client] = None,
+    ) -> Optional["_Sandbox"]:
+        """Return the running Sandbox with the given name if one exists. Otherwise, return None.
+
+        A Sandbox's name is the `name` argument passed to `Sandbox.create`.
+        """
+        if client is None:
+            client = await _Client.from_env()
+        env_name = ensure_env(environment_name)
```
mwaskom: It's confusing but I think you want `_get_environment_name` here not `ensure_env`. The latter is just used from the CLI and also _mutates_ the internal config, while the former just implements the fallbacks from an explicit interface to the config defaults.

rolyatmax: Ahh good catch! Thank you!

modal/sandbox.py
```diff
@@ -435,6 +441,28 @@ def _hydrate_metadata(self, handle_metadata: Optional[Message]):
         self._stdin = StreamWriter(self.object_id, "sandbox", self._client)
         self._result = None
 
+    @staticmethod
+    async def from_name(
+        sandbox_name: str,
```
thomasjpfan: If we want to be consistent with most of the other `from_name` signatures:

```suggestion
        name: str,
```

modal/sandbox.py
```diff
@@ -435,6 +441,28 @@ def _hydrate_metadata(self, handle_metadata: Optional[Message]):
...
+        *,
+        environment_name: Optional[str] = None,
+        client: Optional[_Client] = None,
+    ) -> Optional["_Sandbox"]:
+        """Return the running Sandbox with the given name if one exists. Otherwise, return None.
```
thomasjpfan: With my Python hat on, I prefer to raise an error if the sandbox does not exist.

@mwaskom WDYT?

rolyatmax: Since `from_name()` only returns a _running_ sandbox (if one exists), I think not finding a sandbox will be normal / expected. I feel like returning `None` makes more sense in these cases instead of asking users to wrap these calls in `try/except`s. Thoughts?

modal/sandbox.py
rolyatmax: Ah yes, I see that now. Good call!

modal/sandbox.py
mwaskom: Could see the argument either way. Using an optional return type means that users will need to do all interaction under an `if sb is not None:` conditional, although that's probably (marginally) simpler than a try/except construct. But given that we have the precedent of other `.from_name` methods raising `NotFound` I'd say there's value in consistency / predictability.

rolyatmax: Yeah, the consistency argument makes sense. I'll switch this to raising an error.

modal/sandbox.py
```diff
@@ -230,6 +232,7 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
     async def create(
         *entrypoint_args: str,
         app: Optional["modal.app._App"] = None,  # Optionally associate the sandbox with an app
```
mwaskom: Oops, not in your diff obviously but I think "Optionally" is no longer accurate here!

mwaskom: (Maybe nuanced: I think the current state is that a Sandbox must be associated with an App but if you create it from a modal Function we will implicitly associate it with the Function's App).

rolyatmax: Is `app` required when creating a sandbox from a container? https://github.com/modal-labs/modal-client/blob/dc755f8c09fbfe4159b1ed601b19c7c4281ac1ad/modal/sandbox.py#L409-L411

rolyatmax: @mwaskom gotcha gotcha - so leave this as `Optional`, yeah?

mwaskom: The parameter type is definitely still correct, not sure if we want to update the _description_ to note that it's required if you're not in a Function ... but doesn't need to be solved here!

rolyatmax: Updated!

modal/sandbox.py
```diff
@@ -437,6 +445,27 @@ def _hydrate_metadata(self, handle_metadata: Optional[Message]):
...
 
+    @staticmethod
+    async def from_name(
+        app_name: str,
+        name: str,
+        *,
+        environment_name: Optional[str] = None,
+        client: Optional[_Client] = None,
+    ) -> "_Sandbox":
+        """Return the running Sandbox in the given app with the given name. Raises an error if no running
```
mwaskom: Nit: Per PEP257 the docstring should start with a single-line summary; additional context can be provided in subsequent paragraphs.

modal/sandbox.py
```diff
@@ -710,10 +753,31 @@ async def _load(self: _SandboxSnapshot, resolver: Resolver, existing_object_id:
         return obj
 
     @staticmethod
-    async def _experimental_from_snapshot(snapshot: _SandboxSnapshot, client: Optional[_Client] = None):
+    async def _experimental_from_snapshot(
+        snapshot: _SandboxSnapshot,
+        client: Optional[_Client] = None,
+        *,
+        sandbox_name: Optional[Union[str, type[DefaultSandboxNameOverride]]] = _DEFAULT_SANDBOX_NAME_OVERRIDE,
```
rolyatmax: I'm torn between calling this `sandbox_name` or simply `name`

modal/sandbox.py
```diff
@@ -710,10 +752,31 @@ async def _load(self: _SandboxSnapshot, resolver: Resolver, existing_object_id:
...
-        restore_req = api_pb2.SandboxRestoreRequest(snapshot_id=snapshot.object_id)
+        if sandbox_name is _DEFAULT_SANDBOX_NAME_OVERRIDE:
+            restore_req = api_pb2.SandboxRestoreRequest(
+                snapshot_id=snapshot.object_id,
+                sandbox_name_override_type=api_pb2.SandboxRestoreRequest.SANDBOX_NAME_OVERRIDE_TYPE_UNSPECIFIED,
+            )
+        elif sandbox_name is None:
+            restore_req = api_pb2.SandboxRestoreRequest(
+                snapshot_id=snapshot.object_id,
+                sandbox_name_override_type=api_pb2.SandboxRestoreRequest.SANDBOX_NAME_OVERRIDE_TYPE_NONE,
```
thomasjpfan, line 781: What's the difference between "None" and "Unspecified"?

modal/sandbox.py
thomasjpfan: I slightly like name more. Are you concerned with `name` being confused with a "snapshot name" (which does not exist)?

modal/sandbox.py
rolyatmax: Unspecified means to use the default behavior (i.e. use the name from the original sandbox). And `None` is to explicitly run the new sandbox without a name.

modal/sandbox.py
rolyatmax: Yeah, that was my thinking in choosing `sandbox_name`. `name` better matches the `Sandbox.create()` API though.

rolyatmax: Gonna switch to `name`

modal/sandbox.py
```diff
@@ -710,10 +752,31 @@ async def _load(self: _SandboxSnapshot, resolver: Resolver, existing_object_id:
         return obj
 
     @staticmethod
-    async def _experimental_from_snapshot(snapshot: _SandboxSnapshot, client: Optional[_Client] = None):
+    async def _experimental_from_snapshot(
+        snapshot: _SandboxSnapshot,
+        client: Optional[_Client] = None,
+        *,
+        name: Optional[Union[str, type[DefaultSandboxNameOverride]]] = _DEFAULT_SANDBOX_NAME_OVERRIDE,
```
mwaskom: I wonder if making your `DefaultSandboxNameOverride` type subclass `str` would let you avoid this ugly annotation.

mwaskom (approved):
Two nits but LGTM in general.

modal/sandbox.py
rolyatmax: Ah - I loved this idea and was so hoping it would work. But alas, it doesn't seem to :-(

mwaskom: What doesn't work about it? I tried something I thought was similar and it seemed to work, but not the full example 😞

rolyatmax: I am Very Bad™️ at Python, so I might be doing something wrong, but I made [these changes](https://github.com/modal-labs/modal-client/commit/5a6aad6a5d3edbc6fceb37e202deaf40cb305f2a) then ran `inv type-check` and I get (among other pre-existing issues) this:

```
modal/sandbox.pyi:191: error: Incompatible default for argument "name" (default has type "type[DefaultSandboxNameOverride]", argument has type "str | None")  [assignment]
modal/sandbox.pyi:568: error: Incompatible default for argument "name" (default has type "type[DefaultSandboxNameOverride]", argument has type "str | None")  [assignment]
modal/sandbox.pyi:576: error: Incompatible default for argument "name" (default has type "type[DefaultSandboxNameOverride]", argument has type "str | None")  [assignment]
```

rolyatmax: **Update for posterity:** we finally figured out what the issue was. See the last commit on this PR for more info.

PR #3326, opened 2025-07-11
Enforce timeouts on sandbox `StreamReader` operations

mattnappo:
Enforce a timeout (`deadline`) on `StreamReader` consumption. This is introduced since `.read()` and `async for` consumption can currently hang indefinitely. The buffer (`CONTAINER_EXEC_TIMEOUT_BUFFER`) is used so that in-flight `ContainerExecGetOutput` RPCs are given a bit more time to reach the client after a timeout.

Impl detail: in `_container_process_logs_iterator`, I create the stream once and then break the `async for` into a `while` with manual `.__next__()` consumption of the iterator. This makes it easier to ensure that the timeout is not exceeded while stepping through the iterator.

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2022
 import asyncio
 import time
+from asyncio import timeout as asyncio_timeout
 from collections.abc import AsyncGenerator, AsyncIterator
 from typing import (
     TYPE_CHECKING,
@@ -22,11 +23,6 @@ from ._utils.async_utils import synchronize_api
 from ._utils.grpc_utils import RETRYABLE_GRPC_STATUS_CODES, retry_transient_errors
 from .client import _Client
 from .stream_type import StreamType
-
-# This buffer extends the user-supplied timeout on ContainerExec-related RPCs. This was introduced to
-# give any in-flight status codes/IO data more time to reach the client before the stream is closed.
-CONTAINER_EXEC_TIMEOUT_BUFFER = 5
-
 
 if TYPE_CHECKING:
     pass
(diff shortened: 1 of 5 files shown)


test/sandbox_test.py
```diff
@@ -332,6 +334,19 @@ def test_sandbox_exec_poll_timeout(app, servicer):
...
+@mock.patch("modal.io_streams.CONTAINER_EXEC_TIMEOUT_BUFFER", 0)
+@mock.patch("modal.sandbox.CONTAINER_EXEC_TIMEOUT_BUFFER", 0)
+@skip_non_subprocess
+def test_sandbox_exec_output_timeout(app, servicer):
+    sb = Sandbox.create("sleep", "infinity", app=app)
+
+    cp = sb.exec("sh", "-c", "echo hi; sleep 999", timeout=1)
+    t1 = time.monotonic()
+    assert cp.stdout.read() == "hi\n"
+    assert 1 < time.monotonic() - t1 < 1.2
```
saltzm:
```suggestion
    assert 1 < time.monotonic() - t1 < 2
```
Maybe I'm being overly paranoid but I might bump this to 2 just in case of a really slow machine or something. I think it'd still be testing the behavior we want closely enough

modal/io_streams.py
```diff
@@ -58,11 +65,30 @@ async def _container_process_logs_iterator(
         get_raw_bytes=True,
         last_batch_index=last_index,
     )
-    async for batch in client.stub.ContainerExecGetOutput.unary_stream(req):
+    # Create the stream once and then pull batches manually so we can wrap
+    # each receive in `asyncio.wait_for`, guaranteeing the deadline.
+    stream = client.stub.ContainerExecGetOutput.unary_stream(req)
+    original_deadline = deadline - CONTAINER_EXEC_TIMEOUT_BUFFER if deadline else None
+    while True:
```
saltzm, line 65: I think you could maybe use the `timeout()` [context manager](https://docs.python.org/3/library/asyncio-task.html#timeouts) here to simplify things, like

```py
async with asyncio.timeout(deadline - time.monotonic()):
    async for batch in client.stub.ContainerExecGetOutput.unary_stream(req):
        ...
```

modal/io_streams.py
```diff
@@ -58,11 +65,30 @@ async def _container_process_logs_iterator(
...
+            yield None, -1
+            break
+        except StopAsyncIteration:
+            break
+
         if batch.HasField("exit_code"):
             yield None, batch.batch_index
             break
         for item in batch.items:
+            # Extend the buffer if a message is received during the buffer window
```
saltzm: What's the motivation for this? It seems like in the worst case this could lead to a timeout being indefinitely extended

test/io_streams_test.py
```diff
@@ -282,3 +284,53 @@ async def container_exec_get_output(servicer, stream):
...
+                (0, 1.5),  # message 0: between 0 and 1.5s
+                (1, 2.0),  # message 1: between 1 and 2
+            ]
+            async for line in stdout:
+                elapsed = time.monotonic() - time_first_send
+                min_time, max_time = expected_times[len(output)]
+                assert min_time < elapsed < max_time
+                output.append(line)
+
+            # message 3 should not be received, due to the timeout
```
saltzm, line 334: nice!

saltzm (approved):

PR #3327, opened 2025-07-12
reliability: _get_task_id not retried

thundergolfer:
## Describe your changes

Over the past week about 3.3% of our Sandbox HFM errors are: `error_message="StreamTerminatedError('Stream reset by remote party, error_code: 0')"`. Looking at the code executed and the relevant RPCs, I think this is the cause. 

Being optimistic that a different error wouldn't have happened, fixing this gives us an `100 - 99.883 * 0.0337 = 0.00394%` boost 😄. 

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -578,7 +578,9 @@ class _Sandbox(_Object, type_prefix="sb"):
 
     async def _get_task_id(self) -> str:
         while not self._task_id:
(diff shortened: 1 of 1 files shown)


erikbern:
@freider wdyt about moving the retry logic into the proto stub generator? feels like this keeps coming up

thundergolfer:
It keeps coming up Rust-side too, but Rust's ownership model makes it much harder to implement default gRPC request retries.

PR #3329, opened 2025-07-14
Adjust scaledown window docs

thomasjpfan:
## Describe your changes

Adjust `scaledown_window` to reflect it's actual behavior.

---

## Compatibility checklist

N/A Docs only

---

## Release checklist

---

</details>

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -638,7 +638,9 @@ class _App:
         min_containers: Optional[int] = None,  # Minimum number of containers to keep warm, even when Function is idle.
         max_containers: Optional[int] = None,  # Limit on the number of containers that can be concurrently running.
         buffer_containers: Optional[int] = None,  # Number of additional idle containers to maintain under active load.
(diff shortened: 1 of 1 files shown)


mwaskom:
The existing text is intentional, if not entirely explicatory. The key distinction is that the scale up / scale down decision is not made on the basis of individual containers; it's the Function's container pool that is scaling up or down. So when scaling down (i.e. when the Function is over-provisioned), idle containers can be killed even if the entire scaledown period has not elapsed.

Also since the raw signature is interpolated in the docs, I strongly prefer keeping each parameter on one line if possible, since it's harder to read otherwise.

thomasjpfan:
>  I strongly prefer keeping each parameter on one line if possible, since it's harder to read otherwise.

The pre-commit hook did the auto formatting.

Reading through https://modal.com/docs/guide/cold-start#run-more-warm-containers, it feels more like a `max_scaledown_window` or `max_idle_time`.

PR #3330, opened 2025-07-14
Flip MODAL_FUNCTION_SCHEMAS to True

gleb-modal:
Flip the flag to True so that we can ship UIs for functions later.

## Compatibility checklist

### DIFF
--- a/modal/config.py
+++ b/modal/config.py
@@ -238,7 +238,7 @@ _SETTINGS = {
     "strict_parameters": _Setting(False, transform=_to_boolean),  # For internal/experimental use
     "snapshot_debug": _Setting(False, transform=_to_boolean),
     "cuda_checkpoint_path": _Setting("/__modal/.bin/cuda-checkpoint"),  # Used for snapshotting GPU memory.
-    "function_schemas": _Setting(False, transform=_to_boolean),
+    "function_schemas": _Setting(True, transform=_to_boolean),
     "build_validation": _Setting("error", transform=_check_value(["error", "warn", "ignore"])),
 }
 
--- a/modal/config.py
+++ b/modal/config.py
@@ -238,7 +238,6 @@ _SETTINGS = {
     "strict_parameters": _Setting(False, transform=_to_boolean),  # For internal/experimental use
     "snapshot_debug": _Setting(False, transform=_to_boolean),
     "cuda_checkpoint_path": _Setting("/__modal/.bin/cuda-checkpoint"),  # Used for snapshotting GPU memory.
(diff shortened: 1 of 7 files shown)


modal/config.py
```diff
@@ -238,7 +238,7 @@ class _Setting(typing.NamedTuple):
     "strict_parameters": _Setting(False, transform=_to_boolean),  # For internal/experimental use
     "snapshot_debug": _Setting(False, transform=_to_boolean),
     "cuda_checkpoint_path": _Setting("/__modal/.bin/cuda-checkpoint"),  # Used for snapshotting GPU memory.
-    "function_schemas": _Setting(False, transform=_to_boolean),
+    "function_schemas": _Setting(True, transform=_to_boolean),
```
mwaskom: I can't remember if we only intended this to be a development feature flag, in which case it _should_ be fine to remove it altogether. Should make those explanatory comments mandatory 🥲

thomasjpfan (commented):
@mwaskom From reading the code, I think the risk is from `inspect.signature` failing:

https://github.com/modal-labs/modal-client/blob/9f9c5884abbf4c3879df72a649239387932acfb8/modal/_serialization.py#L562

If we want to be extra safe thing to do is to catch exceptions from `inspect.signature` write a debug message and `return None`.

gleb-modal:
> @mwaskom From reading the code, I think the risk is from `inspect.signature` failing:
> 
> https://github.com/modal-labs/modal-client/blob/9f9c5884abbf4c3879df72a649239387932acfb8/modal/_serialization.py#L562
> 
> If we want to be extra safe thing to do is to catch exceptions from `inspect.signature` write a debug message and `return None`.

FWIW seems like we call inspect.signature for batched functions already: https://github.com/modal-labs/modal-client/blob/6028af80ee6014a1cc58c249737055836dd936db/modal/_functions.py#L724-L730

mwaskom:
My question (since it’s been a few months since we were thinking about this) is more how tolerant *we* can be to just not collecting a schema for the Function. I.e. is anything downstream going to expect that one exists and not be able to handle that case gracefully?

gleb-modal:
> My question (since it’s been a few months since we were thinking about this) is more how tolerant _we_ can be to just not collecting a schema for the Function. I.e. is anything downstream going to expect that one exists and not be able to handle that case gracefully?

No hard dependency, it's ok to just log something if it goes wrong and still deploy!

mwaskom:
Great, can we also put a comment to that effect in the proto file? I don't want someone a little down the line to write some server code assuming that these will 100% always exist.

PR #3333, opened 2025-07-14
Overide Image.hydrate to hide from docs and error better

mwaskom:
`modal.Image` inherits `.hydrate()` from `modal.Object`, which [shows up](https://modal.com/docs/reference/modal.Image#hydrate) (first!) in the docs.

But Images cannot be hydrated on demand. Worse, the current error message is circular and very unclear:

```python-traceback
In [6]: modal.Image.debian_slim().hydrate()

...

File ~/client/build/__editable__.modal-1.0.6.dev35-py3-none-any/modal/_object.py:133, in _Object._validate_is_hydrated(self)
    130 else:
    131     # Technically possible, but with an ambiguous cause.
    132     reason = ""
--> 133 raise ExecutionError(
    134     f"{object_type} has not been hydrated with the metadata it needs to run on Modal{reason}."
    135 )

ExecutionError: Image has not been hydrated with the metadata it needs to run on Modal.
```

This PR overrides the `hydrate` method so we can suppress it from the docs and raise a better error.

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -2300,9 +2300,11 @@ class _Image(_Object, type_prefix="im"):
         """mdmd:hidden"""
         # Image inherits hydrate() from Object but can't be hydrated on demand
         # Overriding the method lets us hide it from the docs and raise a better error message
-        raise ExecutionError(
-            "Images cannot currently be hydrated on demand; build or retreive the Image by running an App that uses it."
-        )
+        if not self.is_hydrated:
+            raise ExecutionError(
+                "Images cannot currently be hydrated on demand; you can build an Image by running an App that uses it."
+            )
+        return self
 
 
 Image = synchronize_api(_Image)


thomasjpfan (approved):
LGTM

modal/image.py
```diff
@@ -2389,5 +2390,15 @@ async def _logs(self) -> typing.AsyncGenerator[str, None]:
...
 
+    async def hydrate(self, client: Optional[_Client] = None) -> Self:
+        """mdmd:hidden"""
+        # Image inherits hydrate() from Object but can't be hydrated on demand
+        # Overriding the method lets us hide it from the docs and raise a better error message
+        if not self.is_hydrated:
+            raise ExecutionError(
+                "Images cannot currently be hydrated on demand; you can build an Image by running an App that uses it."
+            )
+        return self
```
thomasjpfan, line 2307: For internal use cases, we'll need to make sure we use `_hydrate` first and then `.hydrate` becomes a noop.

Is it safer to call `super().hydrate`, catch the exception and then re-raise with a more specific exception?

mwaskom: > we'll need to make sure we use _hydrate first and then .hydrate becomes a noop

I'm not sure we're changing much relative to the status quo, currently calling `.hydrate()` on an Image will always fail (if you haven't previously used `_hydrate` with an ID and metadata), right? Or are you seeing somewhere that isn't the case?

thomasjpfan: Yea it's the status quote, I think all use cases calls `_hydrate` properly.

The only weird case is `Image._log`:

https://github.com/modal-labs/modal-client/blob/c9c4ee845d0df1c19ac808b39b930f09af454c70/modal/image.py#L2278-L2283

An App needs to hydrate the `image` object before `_logs` can be used, but that already breaks today.

thomasjpfan (approved):
Merge conflict, otherwise. LGTM

PR #3334, opened 2025-07-14
Pass put_data_out through input-plane from containers

CyrilSharma:
This uses the same mechanism a `.remote()` request on an input-plane container uses, namely it doesn't write into the Unix Domain Socket and completely bypasses the worker.

This isn't the only approach (ideally our requests should get routed to the worker, like most of our gRPCs), but I think it's fine to do that in a follow-up if necessary.

### DIFF
--- a/modal/_runtime/container_io_manager.py
+++ b/modal/_runtime/container_io_manager.py
@@ -511,7 +511,12 @@ class _ContainerIOManager:
                 chunk.data = message_bytes
             data_chunks.append(chunk)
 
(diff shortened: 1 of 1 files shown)


danielshaar (approved):
LGTM, slight concern about performance if an app has a bunch of objects, but good enough for now. If someone from client team is super concerned, we can revise.

mwaskom:
One thing I'm a little confused about is the status of the input plane support. A lot of the original code that Nathan put in was heavily guarded by "experimental, can be removed" comments etc. Now it looks like we're integrating it much more deeply. Should we update the existing comments to reflect that shifting state?

PR #3340, opened 2025-07-15
Renames `_experimental_enable_gpu_snapshot` parameter to `enable_gpu_snapshot`

luiscape:
This renames the parameter `_experimental_enable_gpu_snapshot` to `enable_gpu_snapshot` as a promotion to alpha from experimental.


luiscape:
Leaving this in draft until backend supports changes.

modal/app.py
```diff
@@ -874,6 +874,7 @@ def cls(
         cloud: Optional[str] = None,  # Cloud provider to run the function on. Possible values are aws, gcp, oci, auto.
         region: Optional[Union[str, Sequence[str]]] = None,  # Region or regions to run the function on.
         enable_memory_snapshot: bool = False,  # Enable memory checkpointing for faster cold starts.
+        enable_gpu_snapshot: bool = False,  # Enable GPU memory snapshots (alpha).
```
mwaskom, line 877: What's the relationship between `enable_memory_snapshot` and `enable_gpu_snapshot`? Do you have to pass both? Or are they mutually exclusive? Or is there different behavior if you just set one vs setting both?

luiscape: `enable_gpu_snapshot` is a subset of `enable_memory_snapshot`. `enable_gpu_snapshot` doesn't work if `enable_memory_snapshot=False`.

We can merge both flags and deprecate `enable_gpu_snapshot` when that's GA.

mwaskom: Is there a reason not to keep it as `_experimental` until then? I'd prefer to minimize the amount of non-explicitly-experimental parameters that we need to deprecate going forward.

luiscape: Two reasons:

1. It'll likely take a long time to merge both pieces of functionality (6 months+)
2. I would like to ask users to try this as an alpha feature. `_experimental` gives the wrong idea.

Let me know if that makes sense.

mwaskom: > I would like to ask users to try this as an alpha feature. _experimental gives the wrong idea.

We have plenty of precedent for giving users features that have an `_experimental` parameter name or that live in the `modal.experimental` namepsace. Although, going forward, we should use `experimental_options` for experimental parameters instead.

We're at the point where we have to commit to long support windows for any API surface that isn't explicitly experimental so I'd like to avoid unnecessary churn if possible.

mwaskom: > enable_gpu_snapshot is a subset of enable_memory_snapshot. enable_gpu_snapshot doesn't work if enable_memory_snapshot=False.

FWIW  this seems pretty cumbersome. Do we catch and error if you set `enable_gpu_snapshot` but not `enable_memory_snapshot`?

EzequielAlejandroLastra:
Do you have ETA for this feature? I have a modal function with Surya-OCR and I think that this feature solve me a problem with cold start. How would it work? Would it work with @enter?

luiscape:
Hi @EzequielAlejandroLastra, we will make an announcement soon about the release of this feature in `alpha`. Stay tunned.

luiscape:
@mwaskom I am closing this PR to follow your recommendation to use `experimental_options` instead. Thank you!

PR #3341, opened 2025-07-15
deprecate environment_name from Sandbox.create()

rolyatmax:
## Describe your changes

Linear issue: [CLI-504](https://linear.app/modal-labs/issue/CLI-504/sandbox-environment-membership-is-ambiguous)

Consider the following code:

```py
app = modal.App("sandbox-wtf")
with modal.enable_output(), app.run(environment_name="main"):
    sb = modal.Sandbox.create(app=app, environment_name="dev")
```

It's unclear which environment the sandbox should belong to. As it turns out, the sandbox will be created in the `dev` environment, but it will be grouped with the sandboxes under the `main` environment in the UI. Other App dependents (Function, Cls, etc) don't have this problem as they have no independent way to specify an environment. Let's deprecate the `environment_name` arg from `Sandbox.create()`.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -228,7 +228,6 @@ class _Sandbox(_Object, type_prefix="sb"):
     async def create(
         *entrypoint_args: str,
         app: Optional["modal.app._App"] = None,  # Optionally associate the sandbox with an app
-        environment_name: Optional[str] = None,  # *DEPRECATED* Optionally override the default environment
         image: Optional[_Image] = None,  # The image to run as the container for the sandbox.
         secrets: Sequence[_Secret] = (),  # Environment variables to inject into the sandbox.
         network_file_systems: dict[Union[str, os.PathLike], _NetworkFileSystem] = {},
@@ -267,6 +266,7 @@ class _Sandbox(_Object, type_prefix="sb"):
             SchedulerPlacement
         ] = None,  # Experimental controls over fine-grained scheduling (alpha).
         client: Optional[_Client] = None,
+        environment_name: Optional[str] = None,  # *DEPRECATED* Optionally override the default environment
     ) -> "_Sandbox":
         """
         Create a new Sandbox to run untrusted, arbitrary code. The Sandbox's corresponding container
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -365,7 +365,7 @@ class _Sandbox(_Object, type_prefix="sb"):
             deprecation_warning(
                 (2025, 7, 16),
                 "Passing `environment_name` to `Sandbox.create` is deprecated and will be removed in a future release.",
-                "A sandbox's environment is now determined by the app it is associated with.",
+                "A sandbox's environment will then be determined by the app it is associated with.",
             )
 
         environment_name = _get_environment_name(environment_name)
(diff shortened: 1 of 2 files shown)


modal/sandbox.py
```diff
@@ -360,6 +361,13 @@ async def _create(
         # sandbox that runs the shell session
         from .app import _App
 
+        if environment_name is not None:
+            deprecation_warning(
+                (2025, 7, 16),
+                "Passing `environment_name` to `Sandbox.create` is deprecated and will be removed in a future release.",
+                "A sandbox's environment is now determined by the app it is associated with.",
```
mwaskom: You know I'm not actually sure this is accurate ... we might need to change the server for it to be.

rolyatmax: Yeah, when we change the functionality here, I imagine we'd stop sending `environment_name` from the client altogether. There are a few changes we'd need to make in the `SandboxCreate` handler too, like requiring an app.

I've gone ahead and tweaked the deprecation wording to indicate that these changes will come in a future update. Just curious: how long do we leave up notices like these before making functionality changes?

mwaskom: I'd actually argue that the current behavior is broken since 

```python
import modal

if __name__ == "__main__":
    app = modal.App("sandbox-wtf")
    with modal.enable_output(), app.run(environment_name="dev"):
        print(sb.object_id)
```

will associate the sandbox with the `main` environment (my current implicit default) not `dev` as you'd expect if Sandboxes are "managed" by an App.

But unless I'm missing something, this does not yet have any overt user impact, since (A) you can't do a name-based lookup of sandboxes (yet), (B) we haven't shipped environment features like budgeting / RBAC, and (C) the Sandbox actually appears in the UI under the App's environment. Based on that, I think it would make sense to just switch the backend associations abruptly and deprecate the parameter, though there may be a complication that I'm missing.

rolyatmax: I see. Okay, I'll open a PR on the modal server for this.

rolyatmax: Updated this PR to no longer send `environment_name` in the `SandboxCreate` request.

modal/sandbox.py
```diff
@@ -230,7 +228,7 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
     async def create(
         *entrypoint_args: str,
         app: Optional["modal.app._App"] = None,  # Optionally associate the sandbox with an app
-        environment_name: Optional[str] = None,  # Optionally override the default environment
+        environment_name: Optional[str] = None,  # *DEPRECATED* Optionally override the default environment
```
mwaskom: Can we move this to the bottom on the arg list? (It's safe since `*entrypoint_args` being variadic  means that all other arguments require keywords.)

modal_proto/api.proto
```diff
@@ -2524,7 +2524,7 @@ message Sandbox {
 message SandboxCreateRequest {
   string app_id = 1 [ (modal.options.audit_target_attr) = true ];
   Sandbox definition = 2;
-  string environment_name = 3;
+  string environment_name = 3; // *DEPRECATED* 7/16/2025
```
mwaskom, line 2527: This date may confuse our poor Swedish friends 😀

mwaskom (approved):
Thank you for dealing with this!

PR #3342, opened 2025-07-15
tweak sandbox-name-related protobuf defs

rolyatmax:
After more discussion on the Sandbox Names project, we've decided to slightly change the `Sandbox.from_name()` function to take an `app_name`. We've also decided to allow sandboxes restored from snapshots to take an optional `name` override. This PR updates those two definitions.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2534,6 +2534,7 @@ message SandboxCreateResponse {
 message SandboxGetFromNameRequest {
   string sandbox_name = 1;
   string environment_name = 2;
+  string app_name = 3;
 }
 
 message SandboxGetFromNameResponse {
@@ -2608,6 +2609,7 @@ message SandboxListResponse {
 
 message SandboxRestoreRequest {
   string snapshot_id = 1;
+  string sandbox_name = 2;
 }
 
 message SandboxRestoreResponse {


modal_proto/api.proto
```diff
@@ -2608,6 +2609,7 @@ message SandboxListResponse {
 
 message SandboxRestoreRequest {
   string snapshot_id = 1;
+  string sandbox_name = 2;
```
mwaskom, line 2612: These are not pointing to the same type of object; how would you use them?

rolyatmax: The `sandbox_name` isn't used as a query. It is the name you want to _give_ to the sandbox that's created from the sandbox restore. The plan right now is that if no name is given, it will default to using the name of the sandbox that the snapshot was created from.

mwaskom: Interesting! Feels like a lot of potential for complexity when thinking about the relationship between snapshots and names. Hope it doesn't invalidate any of our previous discussions 😁

rolyatmax: After some discussions, it seems clear that giving different names when restoring sandboxes is important for certain sandbox-forking use cases. The default behavior - to just use the name from the original sandbox - is a pretty safe path, I think. The name override would be more of an escape hatch for a user who is intentionally wanting to start several sandboxes with different names from the same snapshot.

mwaskom: Dumb question, but does taking a snapshot terminate the sandbox? Or can you snapshot to fork a live sandbox (and then potentially have `.from_id()` fail with a conflict in its default parameterization?

rolyatmax: I believe taking a snapshot does stop the sandbox. Then restoring from that snapshot creates a _new_ sandbox (with a new ID) based on the snapshot.

PR #3343, opened 2025-07-15
Adds an Index Field FunctionGetOutputs Protobuf

CyrilSharma:
This field is needed for supporting queries like `.get()` on a SpawnMap Handler.


devennavani (approved):
LGTM

PR #3348, opened 2025-07-16
Update get_data_in to use input plane.

danielshaar:
## Describe your changes

Mirrored https://github.com/modal-labs/modal-client/pull/3334. Fix for generators will go in as part of other work.

SVC-606

### DIFF
--- a/modal/_runtime/container_io_manager.py
+++ b/modal/_runtime/container_io_manager.py
@@ -486,7 +486,12 @@ class _ContainerIOManager:
 
     async def get_data_in(self, function_call_id: str) -> AsyncIterator[Any]:
         """Read from the `data_in` stream of a function call."""
-        async for data in _stream_function_call_data(self._client, function_call_id, "data_in"):
+        stub = (
+            await self._client.get_stub(self.input_plane_server_url)
+            if self.input_plane_server_url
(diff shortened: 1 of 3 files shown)


danielshaar:
@prbot approve

modal/_runtime/container_io_manager.py
```diff
@@ -486,7 +486,11 @@ async def format_blob_data(self, data: bytes) -> dict[str, Any]:
 
     async def get_data_in(self, function_call_id: str) -> AsyncIterator[Any]:
         """Read from the `data_in` stream of a function call."""
-        async for data in _stream_function_call_data(self._client, function_call_id, "data_in"):
+        stub = self._client.stub
+        if self.input_plane_server_url:
+            stub = await self._client.get_stub(self.input_plane_server_url)
```
mwaskom, line 491: Would it make sense for `self._client.get_stub` to own the "fall back to `self.stub` when `input_plane_server_url` is None? No idea if that is a good idea or not just seems plausible.

rohansingh: The trivial patch for this is not possible because of this line in `_Client._open`:
```
self._stub = await self.get_stub(self.server_url)
```

PR #3349, opened 2025-07-16
update SandboxRestoreRequest protobuf to support a null value for the sandbox name

rolyatmax:
While working on the [Sandbox Names](https://www.notion.so/modal-com/Sandbox-Names-n-e-Keys-Locks-21b1e7f16949802b86a4eec4cfa9f204) project, we've had to decide on the desired behavior when restoring from a sandbox snapshot. So far we've landed on having the new sandboxes use the same name as the original snapshotted sandbox by default, but allowing a user to override that configuration for the new sandbox if desired. That means that there are three possible configurations here: (1) use the original sandbox's name value (either string or null), (2) don't give the new sandbox a name at all (set its name to null), or (3) use the provided string as the sandbox name.

I'll wait to implement this with the other logic changes in a separate PR, but I imagine I'd do something like this to differentiate between these three states on the Python side:

```py
_NO_SANDBOX_OVERRIDE = NoSandboxOverride()

def from_snapshot(snapshot, sandbox_name: Optional[str] = _NO_SANDBOX_OVERRIDE):
    if sandbox_name is _NO_SANDBOX_OVERRIDE:
        # use default behavior
    elif sandbox_name is None:
        # do not give the new sandbox a name
    else:
        # give the sandbox the provided name
```

---

## Compatibility checklist

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2609,8 +2609,15 @@ message SandboxListResponse {
 }
 
 message SandboxRestoreRequest {
+  enum SandboxNameOverrideType {
+    SANDBOX_NAME_OVERRIDE_TYPE_UNSPECIFIED = 0;
+    SANDBOX_NAME_OVERRIDE_TYPE_NONE = 1;
+    SANDBOX_NAME_OVERRIDE_TYPE_STRING = 2;
+  }
+
   string snapshot_id = 1;
-  string sandbox_name = 2;
+  string sandbox_name_override = 2;
+  SandboxNameOverrideType sandbox_name_override_type = 3;
 }
 
 message SandboxRestoreResponse {


modal_proto/api.proto
```diff
@@ -2609,8 +2609,15 @@ message SandboxListResponse {
 }
 
 message SandboxRestoreRequest {
+  enum SandboxNameOverrideType {
+    SANDBOX_NAME_OVERRIDE_TYPE_UNSPECIFIED = 0;
+    SANDBOX_NAME_OVERRIDE_TYPE_NONE = 1;
+    SANDBOX_NAME_OVERRIDE_TYPE_STRING = 2;
```
mwaskom, line 2615: Do we need a fourth type to represent the default "use previous name" behavior or is it sufficient to interpret UNSPECIFIED that way? I guess there won't be any clients that support `.from_name` but don't set this field?

rolyatmax: I'm thinking we just interpret `sandbox_name_override_type is None` and `sandbox_name_override_type == UNSPECIFIED` the same - which is that the original sandbox's name should be used. ([implementation here](https://github.com/modal-labs/modal/pull/24484/files#diff-a781648aff9677e0f644ed5cc871e72cc203c197838d11a52bee3a363c81f413R1313-R1318))

PR #3352, opened 2025-07-17
Update `include_source` for 1.0+ and improve some documentation

mwaskom:
(This PR experienced some scope creep during review).

- Tried to improve the parameter descriptions for `include_source` (which is a very hard thing to describe; we lack some key vocabulary)
- Tried to improve the documentation of the App()-level configuration so it's clear that `volumes=` and `secrets=` will always be used (merged with any function-level configuration)
- Cleaned up some complexity around `include_source=` that dates to having it as an optional feature flag pre-1.0

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -2095,23 +2095,24 @@ class _Image(_Object, type_prefix="im"):
         raw_f: Callable[..., Any],
         *,
         secrets: Sequence[_Secret] = (),  # Optional Modal Secret objects with environment variables for the container
-        gpu: Union[GPU_T, list[GPU_T]] = None,  # Requested GPU or or list of acceptable GPUs( e.g. ["A10", "A100"])
         volumes: dict[Union[str, PurePosixPath], Union[_Volume, _CloudBucketMount]] = {},  # Volume mount paths
         network_file_systems: dict[Union[str, PurePosixPath], _NetworkFileSystem] = {},  # NFS mount paths
+        gpu: Union[GPU_T, list[GPU_T]] = None,  # Requested GPU or or list of acceptable GPUs( e.g. ["A10", "A100"])
         cpu: Optional[float] = None,  # How many CPU cores to request. This is a soft limit.
         memory: Optional[int] = None,  # How much memory to request, in MiB. This is a soft limit.
         timeout: Optional[int] = 60 * 60,  # Maximum execution time of the function in seconds.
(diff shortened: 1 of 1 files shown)


modal/app.py
```diff
@@ -171,10 +171,10 @@ def __init__(
...
         name: Optional[str] = None,
         *,
-        image: Optional[_Image] = None,  # default image for all functions (default is `modal.Image.debian_slim()`)
-        secrets: Sequence[_Secret] = [],  # default secrets for all functions
-        volumes: dict[Union[str, PurePosixPath], _Volume] = {},  # default volumes for all functions
-        include_source: Optional[bool] = None,
+        image: Optional[_Image] = None,  # Default Image for the App (otherwise default to `modal.Image.debian_slim()`)
+        secrets: Sequence[_Secret] = [],  # Secrets to add for all Functions in the App
+        volumes: dict[Union[str, PurePosixPath], _Volume] = {},  # Volume mounts to use for all Functions
+        include_source: Optional[bool] = None,  # default for auto-adding Function source file(s) to the container
```
thomasjpfan: `default=None` is hard to understand without looking at the code. I think it maps to `INCLUDE_MAIN_PACKAGE` (True). 

Looking over the code, can `App.include_source` can be set to `True` now?

```suggestion
        include_source: bool = True,  # default for auto-adding Function source file(s) to the container
```

(`App.function(include_source)` still needs to be `Optional`)

mwaskom: Wondered that myself, but I think that if `include_source` were set to `True` in the `App.function` signature, we wouldn't fall back to the `App()`-level configuration when the user omits it, which is the intended behavior.

thomasjpfan: I think we are in agreement here. Explicitly, I meant this:

```python
App(..., include_source: bool = True)

# Use `include_source` if defined, otherwise use `App.include_source` 
App.function(..., include_source: Optional[bool] = None)
```

mwaskom: Doh, sorry didn't read carefully enough on where you were commenting.

App() probably could have a `True` default now. It was `None` prior to 1.0 because using it opted into the post-1.0 automounting behavior _in toto_ (not mounting local dependencies, etc.).

mwaskom: I don't really know if there would be any surprising consequences to changing the default. I hope not!

thomasjpfan (approved):

PR #3356, opened 2025-07-18
Address some flaky tests

mwaskom:
I'm pretty sure that one of these started failing as a result of https://github.com/modal-labs/modal-client/pull/3180 breaking the assumption that the Sandbox actually will be terminated when `sb.terminate()` is called. Fixes CLI-517

The other is pretty complicated and recently added so I'm going to delegate a fix to @mattnappo , but I marked it as flaky for now so it should crash fewer CI builds.


mwaskom:
@prbot approve

PR #3359, opened 2025-07-18
Input Plane: Map Retry

bokeefe123:
Input plane map retries

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -456,6 +456,7 @@ async def _map_invocation_inputplane(
 
     have_all_inputs = False
     map_done_event = asyncio.Event()
+    retry_event = asyncio.Event()
 
     inputs_created = 0
     outputs_completed = 0
@@ -549,6 +550,7 @@ async def _map_invocation_inputplane(
             async_map_ordered(input_iter(), create_input, concurrency=BLOB_MAX_PARALLELISM)
         ) as streamer:
             async for q_item in streamer:
+                # TODO(ben-okeefe): Handle retrying inputs.
                 await queue.put(time.time(), q_item)
 
         # All inputs have been read.
@@ -622,7 +624,7 @@ async def _map_invocation_inputplane(
                     function_id=function.object_id,
                     function_call_id=function_call_id,
                     last_entry_id=last_entry_id,
-                    timeout=OUTPUTS_TIMEOUT,
+                    timeout=0,  # Non-blocking read
                     items=check_inputs,
                 )
                 token = await client._auth_token_manager.get_token()
@@ -1217,6 +1219,7 @@ class _MapItemContext:
 
         self.state = _MapItemState.WAITING_TO_RETRY
 
+        # TODO(ben-okeefe): Handle retrying inputs.
         await retry_queue.put(now_seconds + (delay_ms / 1000), item.idx)
 
         return _OutputType.RETRYING
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -429,7 +429,6 @@ async def _map_invocation_inputplane(
     client: "modal.client._Client",
     order_outputs: bool,
     return_exceptions: bool,
-    wrap_returned_exceptions: bool,
     count_update_callback: Optional[Callable[[int, int], None]],
 ) -> typing.AsyncGenerator[Any, None]:
     """Input-plane implementation of a function map invocation.
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -129,7 +129,7 @@ async def _map_invocation(
     no_context_duplicates = 0
 
     retry_queue = TimestampPriorityQueue()
-    completed_outputs: set[str] = set()  # Set of input_ids whose outputs are complete (expecting no more values)
+    # completed_outputs: set[str] = set()  # Set of input_ids whose outputs are complete (expecting no more values)
     input_queue: asyncio.Queue[api_pb2.FunctionPutInputsItem | None] = asyncio.Queue()
     map_items_manager = _MapItemsManager(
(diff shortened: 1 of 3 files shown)


modal/_utils/async_utils.py
```diff
@@ -279,7 +279,8 @@ class TimestampPriorityQueue(Generic[T]):
 
     def __init__(self, maxsize: int = 0):
         self.condition = asyncio.Condition()
-        self._queue: asyncio.PriorityQueue[tuple[float, Union[T, None]]] = asyncio.PriorityQueue(maxsize=maxsize)
+        self._queue: asyncio.PriorityQueue[tuple[float, int, Union[T, None]]] = asyncio.PriorityQueue(maxsize=maxsize)
+        self._counter = itertools.count()
```
rculbertson, line 284: Maybe add a comment why this counter is needed, just as a reminder to our future selves?

modal/parallel_map.py
```diff
@@ -1226,6 +1355,16 @@ def handle_retry_response(self, input_jwts: list[str]):
...
                 ctx.handle_retry_response(input_jwt)
 
+    async def handle_check_inputs_response(self, response: list[tuple[int, bool]]):
+        for idx, lost in response:
+            ctx = self._item_context.get(idx, None)
+            if ctx is not None:
+                if lost:
+                    ctx.state = _MapItemState.WAITING_TO_RETRY
+                    retry_item = await ctx.create_map_start_or_continue_item(idx)
+                    delay_ms = ctx.retry_manager.get_delay_ms()
```
rculbertson: In the control-plane version, the logic is to call `retry_manager.get_delay_ms()`, but then ignore the returned value, and set it to 0. This is because we want to increment the retry count (which happens when we call `get_delay_ms`), but we want to retry immediately. We should probably do the same thing here to keep parity.

modal/parallel_map.py
```diff
@@ -1045,12 +1098,19 @@ class _MapItemContext:
...
     previous_input_jwt: Optional[str]
     _event_loop: asyncio.AbstractEventLoop
 
-    def __init__(self, input: api_pb2.FunctionInput, retry_manager: RetryManager, sync_client_retries_enabled: bool):
+    def __init__(
+        self,
+        input: api_pb2.FunctionInput,
+        retry_manager: RetryManager,
+        sync_client_retries_enabled: bool,
+        input_plane_instance: bool = False,
```
rculbertson: nit: what would you think about calling this variable `is_input_plane_instance`, to indicate this is a bool?

modal/parallel_map.py
```diff
@@ -477,11 +479,23 @@ async def _map_invocation_inputplane(
...
-    # ------------------------------------------------------------
-    # Helper functions
-    # ------------------------------------------------------------
+    # Input plane does not yet return a retry policy. So we set a default.
+    retry_policy = api_pb2.FunctionRetryPolicy(
+        retries=INPUTS_RETRY_MAX_RETRIES,
+        initial_delay_ms=1000,
+        max_delay_ms=1000,
+        backoff_coefficient=1.0,
+    )
```
rculbertson, line 487: This effectively makes it so that every function has a retry policy right? Most users don't have a retry policy, so this might be unexpected. But I guess we're not giving this to users yet right? Might want to hold off until we have the retry policy sorted out.

modal/parallel_map.py
```diff
@@ -477,11 +479,23 @@ async def _map_invocation_inputplane(
...
+        retry_policy=retry_policy,
+        function_call_invocation_type=api_pb2.FUNCTION_CALL_INVOCATION_TYPE_SYNC,
+        retry_queue=queue,
+        sync_client_retries_enabled=True,
+        max_inputs_outstanding=MAX_INPUTS_OUTSTANDING_DEFAULT,
+        input_plane_instance=True,
+    )
 
-    def update_counters(created_delta: int = 0, completed_delta: int = 0, set_have_all_inputs: bool | None = None):
+    def update_counters(created_delta: int = 0, completed_delta: int = 0, set_have_all_inputs=None):
```
rculbertson: was removing the type annotation intentional?

modal/parallel_map.py
```diff
@@ -527,25 +537,23 @@ async def drain_input_generator():
             async_map_ordered(input_iter(), create_input, concurrency=BLOB_MAX_PARALLELISM)
         ) as streamer:
             async for q_item in streamer:
+                # TODO(ben-okeefe): Handle retrying inputs.
```
rculbertson: Could you expand the comment to explain why the input-plane version has this TODO, but the control-plane doesn't?

test/conftest.py
```diff
@@ -218,6 +218,7 @@ def __init__(self, blob_host, blobs, blocks, files_sha2data, credentials, port):
         self.files_sha2data = files_sha2data
         self.function_id_for_function_call = {}
         self.function_call_inputs = {}
+        self.inputplane_function_call_inputs = {}
```
rculbertson: is this not used?

modal/_utils/async_utils.py
bokeefe123: Sounds great, thanks! For reference:
```
# Used to tiebreak items with the same timestamp that are not comparable. (eg. protos)
```

modal/parallel_map.py
bokeefe123: Yeah I'm not sure how it was passing before but this was failing on the tests without removing them.

rculbertson: Just curious what was the failure? Was it a test failure or mypy error? If it was a mypy error maybe we can figure out how to fix it?

modal/parallel_map.py
rculbertson: Was this TODO not needed?

rculbertson (approved):
nice job!

PR #3366, opened 2025-07-22
[modal flash autoscaler] fix blue green deployments

thecodingwizard:
when redeploying, we may temporarily burst to 2x number of target containers


thecodingwizard:
@prbot approve

gongy:
:+1:

PR #3368, opened 2025-07-22
[SpawnMapV2](0) Refactor Map

CyrilSharma:
This is a relatively small refactor which just groups related functionality into classes. This refactor is necessary because
1. The Map Code is a mess
2. The map logic should be re-usable, especially with the soon-to-arrive introduction of `Spawn-Map` v2.

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -151,7 +151,6 @@ class InputPreprocessor:
 class InputPumper:
     """
     Reads inputs from a queue of FunctionPutInputsItems, and sends them to the server.
-    Also reads inputs from a queue of FunctionRetryInputsItems, and sends them to the server.
     """
 
     def __init__(
@@ -159,27 +158,23 @@ class InputPumper:
         client: "modal.client._Client",
         *,
         input_queue: asyncio.Queue,
-        retry_queue: TimestampPriorityQueue,
         function: "modal.functions._Function",
-        map_items_manager: "_MapItemsManager",
-        function_call_jwt: str,
         function_call_id: str,
+        map_items_manager: Optional["_MapItemsManager"] = None,
     ):
         self.client = client
         self.function = function
         self.map_items_manager = map_items_manager
-        self.retry_queue = retry_queue
         self.input_queue = input_queue
         self.inputs_sent = 0
-        self.inputs_retried = 0
-        self.function_call_jwt = function_call_jwt
         self.function_call_id = function_call_id
 
     async def pump_inputs(self):
         assert self.client.stub
         async for items in queue_batch_iterator(self.input_queue, max_batch_size=MAP_INVOCATION_CHUNK_SIZE):
             # Add items to the manager. Their state will be SENDING.
-            await self.map_items_manager.add_items(items)
+            if self.map_items_manager is not None:
+                await self.map_items_manager.add_items(items)
             request = api_pb2.FunctionPutInputsRequest(
                 function_id=self.function.object_id,
                 inputs=items,
@@ -193,30 +188,12 @@ class InputPumper:
             resp = await self._send_inputs(self.client.stub.FunctionPutInputs, request)
             self.inputs_sent += len(items)
             # Change item state to WAITING_FOR_OUTPUT, and set the input_id and input_jwt which are in the response.
-            self.map_items_manager.handle_put_inputs_response(resp.inputs)
+            if self.map_items_manager is not None:
+                self.map_items_manager.handle_put_inputs_response(resp.inputs)
             logger.debug(
                 f"Successfully pushed {len(items)} inputs to server. "
                 f"Num queued inputs awaiting push is {self.input_queue.qsize()}."
             )
-        yield
-
-    async def retry_inputs(self):
-        async for retriable_idxs in queue_batch_iterator(self.retry_queue, max_batch_size=MAP_INVOCATION_CHUNK_SIZE):
-            # For each index, use the context in the manager to create a FunctionRetryInputsItem.
-            # This will also update the context state to RETRYING.
-            inputs: list[api_pb2.FunctionRetryInputsItem] = await self.map_items_manager.prepare_items_for_retry(
-                retriable_idxs
-            )
-            request = api_pb2.FunctionRetryInputsRequest(
-                function_call_jwt=self.function_call_jwt,
-                inputs=inputs,
-            )
-            resp = await self._send_inputs(self.client.stub.FunctionRetryInputs, request)
-            # Update the state to WAITING_FOR_OUTPUT, and update the input_jwt in the context
-            # to the new value in the response.
-            self.map_items_manager.handle_retry_response(resp.input_jwts)
-            logger.debug(f"Successfully pushed retry for {len(inputs)} to server.")
-            self.inputs_retried += len(inputs)
         yield
 
     async def _send_inputs(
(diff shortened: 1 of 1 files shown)


modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
     import modal.functions
 
 
+class InputPreprocessor:
+    """
+    Preprocesses inputs to be sent to the server.
+    """
```
devennavani: Could we add some more color to this docstring?

modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
...
+        self.processed_all_inputs = True
+        if self.done_callback is not None:
+            self.done_callback(self.inputs_created)
+        yield
+
+
+class InputPumper:
+    """
+    Pumps inputs to the server.
+    """
```
devennavani: Likewise, add more color to this docstring.

modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
...
+    Preprocesses inputs to be sent to the server.
+    """
+
+    def __init__(
+        self,
+        client: "modal.client._Client",
+        *,
+        raw_input_queue: _SynchronizedQueue,
+        processed_input_queue: asyncio.Queue,
```
devennavani, line 99: Can't the type annotation for `processed_input_queue` be more specific here?

`asyncio.Queue[api_pb2.FunctionPutInputsItem | None]`

modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
...
+        self.done_callback = done_callback
+
+    async def input_iter(self):
+        while 1:
+            raw_input = await self.raw_input_queue.get()
+            if raw_input is None:  # end of input sentinel
+                break
+            yield raw_input  # args, kwargs
+
+    def create_input(self):
```
devennavani: Might be a better name for this function, since it's returning a function that creates the input, not actually creating the input itself

modal/parallel_map.py
```diff
@@ -395,8 +470,11 @@ async def log_debug_stats():
         def log_stats():
             logger.debug(
                 f"Map stats: sync_client_retries_enabled={sync_client_retries_enabled} "
-                f"have_all_inputs={have_all_inputs} inputs_created={inputs_created} input_sent={inputs_sent} "
-                f"inputs_retried={inputs_retried} outputs_received={outputs_received} "
+                f"have_all_inputs={input_preprocessor.processed_all_inputs} "
```
devennavani: Can you not use `have_all_inputs` instead of `input_preprocessor.processed_all_inputs` here?

modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
...
+        function: "modal.functions._Function",
+        created_callback: Optional[Callable[[int], None]],
+        done_callback: Optional[Callable[[int], None]],
+    ):
+        self.client = client
+        self.function = function
+        self.inputs_created = 0
+        self.raw_input_queue = raw_input_queue
+        self.processed_input_queue = processed_input_queue
+        self.processed_all_inputs = False
```
devennavani: Why is `processed_all_inputs` needed?

modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
...
+            async_map_ordered(self.input_iter(), self.create_input(), concurrency=BLOB_MAX_PARALLELISM)
+        ) as streamer:
+            async for item in streamer:
+                await self.processed_input_queue.put(item)
+
+        # close queue iterator
+        await self.processed_input_queue.put(None)
+        self.processed_all_inputs = True
+        if self.done_callback is not None:
+            self.done_callback(self.inputs_created)
```
devennavani: I thought the `done_callback` doesn't take any args? And is `self.done_callback` ever actually `None`?

modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
...
+
+    def __init__(
+        self,
+        client: "modal.client._Client",
+        *,
+        raw_input_queue: _SynchronizedQueue,
+        processed_input_queue: asyncio.Queue,
+        function: "modal.functions._Function",
+        created_callback: Optional[Callable[[int], None]],
+        done_callback: Optional[Callable[[int], None]],
```
devennavani: Unclear to me why these are `Optional`

modal/parallel_map.py
```diff
@@ -135,25 +289,24 @@ async def _map_invocation(
...
+        client=client,
+        input_queue=input_queue,
+        retry_queue=retry_queue,
+        function=function,
+        map_items_manager=map_items_manager,
+        function_call_jwt=function_call_jwt,
+        function_call_id=function_call_id,
+    )
 
     def update_state(set_have_all_inputs=None, set_inputs_created=None, set_outputs_completed=None):
```
devennavani, line 329: Can we not clean up `update_state` now that the callbacks are being used? (it only needs to handle `set_outputs_completed` I think?)

modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
...
+            )
+            resp = await self.send_inputs(self.client.stub.FunctionRetryInputs, request)
+            # Update the state to WAITING_FOR_OUTPUT, and update the input_jwt in the context
+            # to the new value in the response.
+            self.map_items_manager.handle_retry_response(resp.input_jwts)
+            logger.debug(f"Successfully pushed retry for {len(inputs)} to server.")
+            self.inputs_retried += len(inputs)
+        yield
+
+    async def send_inputs(
```
devennavani: Can we make it `_send_inputs` if it's only being used internally by `pump_inputs` and `retry_inputs`?

modal/parallel_map.py
```diff
@@ -86,6 +86,162 @@ class _OutputValue:
...
+class InputPumper:
+    """
+    Pumps inputs to the server.
+    """
+
+    def __init__(
+        self,
+        client: "modal.client._Client",
+        *,
+        input_queue: asyncio.Queue,
```
devennavani, line 159: Ditto, typing can be more specific here (`asyncio.Queue[api_pb2.FunctionPutInputsItem | None]`)

devennavani (commented):
Left some comments, please address. Also please verify that a `.map` is successful using these changes.

modal/parallel_map.py
CyrilSharma: I have tried this. It actually does not work for some reason, synchronicity starts screaming at me.

CyrilSharma: parallel_map:842
```
# Both these futures are strings. Omitting generic type because
# it causes an error when running `inv protoc type-stubs`.
```

Other people have run into the same error ^

modal/parallel_map.py
CyrilSharma: Actually it's a bit non-trivial. When any fragment of the state changes any of these if statements could activate. The net effect is factoring this code out into independent functions does not simplify it much.

modal/parallel_map.py
```diff
@@ -86,6 +86,161 @@ class _OutputValue:
     import modal.functions
 
 
+class InputPreprocessor:
+    """
+    Consumes unregistered, user provided inputs from a raw-input queue
+    Registers each of them with the control-plane, and puts them in a processed-input queue.
```
devennavani: Let's make this a bit more accurate -- InputPreprocessor is only constructing `FunctionPutInputsItem` objects and putting them into the client-side `processed_input_queue`

PR #3371, opened 2025-07-22
Add protos for communicating resource creation metadata

mwaskom:
## Describe your changes

Adds protos so that we can send creation metadata back from the server to the client for "resource objects" (Volume / Dict / Queue / Secret). This will be considered part of the general hydration metadata; I've also added type-specific Metadata messages for most of these objects (Volume already had one) and updated the `.from_name` loading to pass the response metadata through to `_hydrate`.

I'll follow up in a second PR (after the server-side changes land) to actually store the metadata on hydration and expose it.

- Towards CLI-428
 
---

## Compatibility checklist

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1145,7 +1145,8 @@ message DictListResponse {
 }
 
 message DictMetadata {
-  CreationInfo creation_info = 1;
+  string name = 1;
+  CreationInfo creation_info = 2;
 }
 
 message DictPopRequest {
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1136,6 +1144,10 @@ message DictListResponse {
   repeated DictInfo dicts = 1;
 }
 
+message DictMetadata {
+  CreationInfo creation_info = 1;
```
thomasjpfan: In the future, do you intend to have the dict's "deployment_name" to be a part of the metadata?

mwaskom: Discussed live but for posterity: good question and a bit undefined, I think we lean towards considering "names" metadata as it simplifies ensuring that we know the name of a hydrated object.

mwaskom: That's what I've done here.

PR #3372, opened 2025-07-22
Integrate container IPs into clusterinfo

pawalt:
## Describe your changes

- Integrates the `container_ipv4_ips` field into `ClusterInfo`.
- For multi-node clusters, `container_ipv4_ips` is populated from the backend's `TaskClusterHelloResponse`.
- For single-node scenarios, `container_ipv4_ips` is an empty list since the container already has a local V4 IP.

---

## Compatibility checklist

### DIFF
--- a/modal/_clustered_functions.py
+++ b/modal/_clustered_functions.py
@@ -16,6 +16,7 @@ class ClusterInfo:
     rank: int
     container_ips: list[str]
     container_ipv4_ips: list[str]
+    error: Optional[str] = None
 
 
 cluster_info: Optional[ClusterInfo] = None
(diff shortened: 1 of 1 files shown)


thundergolfer (approved):
LGTM

modal/_clustered_functions.py
```diff
@@ -15,6 +15,7 @@
 class ClusterInfo:
     rank: int
     container_ips: list[str]
```
mwaskom, line 17: are these effectively `container_ipv6_ips`?

pawalt: yeah. the naming is bad, but I didn't want to break this interface for now. When we graduate out of experimental I'd like to make this more class-ful. would appreciate your brain on thinking about how to do this

PR #3373, opened 2025-07-23
Refactor GPU memory snapshots

luiscape:
This refactors GPU memory snapshots to parallelize both checkpoint and restore operations. This is impactful when processes have multiple CUDA sessions with large memory footprints.

### DIFF
--- a/modal/_runtime/gpu_memory_snapshot.py
+++ b/modal/_runtime/gpu_memory_snapshot.py
@@ -1,17 +1,18 @@
 # Copyright Modal Labs 2022
 #
 # This module provides a simple interface for creating GPU memory snapshots,
-# provising a convenient interface to `cuda-checkpoint` [1]. This is intended
+# providing a convenient interface to `cuda-checkpoint` [1]. This is intended
 # to be used in conjunction with memory snapshots.
 #
 # [1] https://github.com/NVIDIA/cuda-checkpoint
 
 import subprocess
 import time
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
+from typing import List, Optional
 
 from modal.config import config, logger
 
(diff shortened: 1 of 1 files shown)


modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -39,16 +44,31 @@ class CudaCheckpointProcess:
...
         logger.debug(f"PID: {self.pid} Toggling CUDA checkpoint state to {target_state.value}")
 
         start_time = time.monotonic()
+        retry_count = 0
+        max_retries = 3
 
         while self._should_continue_toggle(target_state, start_time, timeout_secs):
-            self._execute_toggle_command()
-            time.sleep(0.1)
+            try:
```
cweld510, line 58: Why do we need exponential backoff here? Are there actually retryable errors?

cweld510 (approved):

modal/_runtime/gpu_memory_snapshot.py
luiscape: They are retryable. I added exponential backoff as a defense against contention between sessions. Admittedly, I am not sure exponential backoffs are necessary.

modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -125,75 +161,143 @@ def _get_cuda_pids(self) -> list[CudaCheckpointProcess]:
...
-
-                # If the command succeeds (return code 0), this PID has a CUDA session
-                if result.returncode == 0:
-                    state_str = result.stdout.strip().lower()
-                    state = CudaCheckpointState(state_str)
-
-                    cuda_checkpoint_process = CudaCheckpointProcess(pid=pid, state=state)
-                    cuda_pids.append(cuda_checkpoint_process)
+        # Get all numeric directories (PIDs) from /proc
+        pid_dirs = [entry for entry in proc_dir.iterdir() if entry.name.isdigit()]
```
rst0git, line 165: @luiscape Some PIDs can be CPU-only processes without GPU state. In the CUDA Plugin for CRIU we use the `--get-restore-tid` option to retrieve the cuda TIDs: https://github.com/checkpoint-restore/criu/blob/57251d8/plugins/cuda/cuda_plugin.c#L211

PR #3375, opened 2025-07-23
[WRK-718] Add SnapshotBehavior proto

mattnappo:
Renaming of #3370

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -240,6 +240,13 @@ enum SystemErrorCode {
   SYSTEM_ERROR_CODE_NOSPC = 28;         // ENOSPC: No space left on device
 }
 
+enum TaskSnapshotBehavior {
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -240,6 +240,13 @@ enum SystemErrorCode {
   SYSTEM_ERROR_CODE_NOSPC = 28;         // ENOSPC: No space left on device
 }
 
+enum TaskSnapshotBehavior {
```
luiscape, line 243: enums always start with `_UNSPECIFIED`.

luiscape (changes requested):

modal_proto/api.proto
mattnappo: Should I remove `_UNKNOWN` then?

luiscape: yes

cweld510 (approved):

PR #3376, opened 2025-07-23
Protos for async worker snapshots

paulgb:
## Describe your changes

Since filesystem snapshots can take arbitrarily long, we should use a pollable async API instead of a standard functional RPC.

This PR adds protos and an RPC. To maintain backwards compatibility, it creates new protos / RPC instead of modifying the existing one.

---

## Compatibility checklist

---

</details>


saltzm:
You could maybe just watch for [this response](https://github.com/modal-labs/modal/blob/a7bad9189e87c388d6ef63560fab7e928d228afa/server/modal_server/objects/sandbox.py#L1434-L1440)  in the client and then start sending the SandboxSnapshotFsAsyncGetRequest

thundergolfer (approved):

PR #3379, opened 2025-07-23
[Input-Plane Parameterized Webendpoints](9) Add Field To ContainerArguments

CyrilSharma:
Currently the container uses the input-plane-url to decide where to get and receive data from in FunctionPutDataIn and FunctionPutDataOut. We are able to infer the input-plane-url using the AppLayout struct, but unfortunately this will not work once we start supporting parameterized functions, since the AppLayout will not contain every instance of every parameterized function.

To that end, we allow ContainerArguments to contain an input-plane-url.

I confirmed every ContainerArguments constructor already uses ..Default::default() so this should be safe.

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -865,7 +865,7 @@ message ContainerArguments {  // This is used to pass data from the worker to th
   string environment_name = 13;
   optional string checkpoint_id = 14;
   AppLayout app_layout = 15;
-  optional string input_plane_server_url = 16;
+  string input_plane_server_url = 16;
 }
 
 message ContainerCheckpointRequest {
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -865,6 +865,7 @@ message ContainerArguments {  // This is used to pass data from the worker to th
   string environment_name = 13;
   optional string checkpoint_id = 14;
   AppLayout app_layout = 15;
+  optional string input_plane_server_url = 16;
```
devennavani: let's remove the `optional`

devennavani (approved):
LGTM w/ one comment

PR #3382, opened 2025-07-24
Relax pin on click

mwaskom:
Got a user report that `click>=8.2` is required by another framework that they want to use with Modal so testing whether any issues arise if we allow that.

### DIFF
--- a/test/cli_test.py
+++ b/test/cli_test.py
@@ -47,7 +47,12 @@ dummy_other_module_file = "x = 42"
 
 
 def _run(args: list[str], expected_exit_code: int = 0, expected_stderr: str = "", expected_error: str = ""):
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -16,7 +16,7 @@ authors = [
 dependencies = [
     "aiohttp",
     "certifi",
-    "click~=8.1.0",
+    "click~=8.1",
```
thomasjpfan, line 19: Should this be less string so we are more compatible with other libraries? (specifically `click>=8.2`)

```suggestion
    "click>=8.1",
```

thomasjpfan: Nevermind, I misread.

PR #3383, opened 2025-07-24
Update minimal version of wheel in development requirements

mwaskom:
Addresses https://github.com/modal-labs/modal-client/security/dependabot/53


thomasjpfan (commented):
✅

PR #3389, opened 2025-07-24
add support for key not found

molocule:
## Describe your changes
if the key is not found, the server side returns None (according to type definition)
add support for handling that case on client side

## Compatibility checklist

- [ x] Client+Server: this change is compatible with old servers

---

</details>

## Changelog
- Improve error handling for DictPop on nonexistent keys

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -309,7 +309,7 @@ class _Dict(_Object, type_prefix="di"):
         """Remove a key from the Dict, returning the value if it exists."""
         req = api_pb2.DictPopRequest(dict_id=self.object_id, key=serialize(key))
         resp = await retry_transient_errors(self._client.stub.DictPop, req)
-        if not resp.found:
+        if not resp or not resp.found:
             raise KeyError(f"{key} not in dict {self.object_id}")
         return deserialize(resp.value, self._client)
 


mwaskom:
I'm a little confused, it seems this is the existing behavior?

```python
In [1]: import modal

In [2]: d = modal.Dict.from_name("test-dict")

In [3]: d
Out[3]: Dict()

In [4]: d.pop("blah")
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[4], line 1
----> 1 d.pop("blah")

File ~/client/build/__editable__.modal-1.1.1.dev22-py3-none-any/modal/_object.py:285, in live_method.<locals>.wrapped(self, *args, **kwargs)
    282 @wraps(method)
    283 async def wrapped(self, *args, **kwargs):
    284     await self.hydrate()
--> 285     return await method(self, *args, **kwargs)

File ~/client/build/__editable__.modal-1.1.1.dev22-py3-none-any/modal/dict.py:313, in _Dict.pop(self, key)
    311 resp = await retry_transient_errors(self._client.stub.DictPop, req)
    312 if not resp.found:
--> 313     raise KeyError(f"{key} not in dict {self.object_id}")
    314 return deserialize(resp.value, self._client)

KeyError: 'blah not in dict di-RBcAeFRVLpblxTMlhknkJs'
```

Is there a bug report associated with this?

claudiallium:
Oh yes you're right! Made the relevant change on server side

luiscape (approved):

PR #3392, opened 2025-07-26
Set force_jupyter=False on rich Console objects

ekzhang:
Rich's autodetection of Jupyter is a bit aggressive and changes the way interactive CLI outputs are displayed in notebooks, causing problems while rendering. It appears that just using the terminal output mode is better, and the only way to do this seems to be by "disabling" Jupyter with `force_jupyter=False`.

## Describe your changes

Resolves PRD-1096

https://github.com/user-attachments/assets/92e3da73-5355-49cc-800e-f4d8ead68724

### DIFF
--- a/modal/_output.py
+++ b/modal/_output.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import functools
-import io
 import platform
 import re
 import socket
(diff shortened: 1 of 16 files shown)


modal/_output.py
```diff
@@ -147,12 +156,11 @@ class OutputManager:
     def __init__(
         self,
         *,
-        stdout: io.TextIOWrapper | None = None,
```
ekzhang, line 150: Checked that this wasn't used anywhere in the code, so removing it for now.

thomasjpfan (approved):

PR #3393, opened 2025-07-28
Expose resource metadata through `.name` and `.info`

mwaskom:
## Describe your changes

- Closes SDK-538, see also [ideation doc](https://www.notion.so/modal-com/Emergent-issues-around-object-metadata-21c1e7f1694980b38d6dd33744ec6aac)
- Depends on https://github.com/modal-labs/modal/pull/25030

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Added a `.name` property and a `.info()` method to `modal.Dict`, `modal.Queue`, `modal.Volume`, and `modal.Secret` objects.

### DIFF
--- a/modal/_object.py
+++ b/modal/_object.py
@@ -48,6 +48,10 @@ class _Object:
     _is_hydrated: bool
     _is_rehydrated: bool
 
+    # Not all object subclasses have a meaningful "name" concept
+    # So whether they expose this is a matter of having a name property
+    _name: Optional[str]
+
     @classmethod
     def __init_subclass__(cls, type_prefix: Optional[str] = None):
         super().__init_subclass__()
@@ -68,6 +72,7 @@ class _Object:
         hydrate_lazily: bool = False,
         deps: Optional[Callable[..., Sequence["_Object"]]] = None,
         deduplication_key: Optional[Callable[[], Awaitable[Hashable]]] = None,
+        name: Optional[str] = None,
     ):
         self._local_uuid = str(uuid.uuid4())
         self._load = load
@@ -82,6 +87,8 @@ class _Object:
         self._client = None
         self._is_hydrated = False
         self._is_rehydrated = False
+
+        self._name = name
 
         self._initialize_from_empty()
 
@@ -163,10 +170,11 @@ class _Object:
         hydrate_lazily: bool = False,
         deps: Optional[Callable[..., Sequence["_Object"]]] = None,
         deduplication_key: Optional[Callable[[], Awaitable[Hashable]]] = None,
+        name: Optional[str] = None,
     ):
         # TODO(erikbern): flip the order of the two first arguments
         obj = _Object.__new__(cls)
-        obj._init(rep, load, is_another_app, preload, hydrate_lazily, deps, deduplication_key)
+        obj._init(rep, load, is_another_app, preload, hydrate_lazily, deps, deduplication_key, name)
         return obj
 
     @staticmethod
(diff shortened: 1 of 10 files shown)


modal/_object.py
```diff
@@ -48,6 +48,10 @@ class _Object:
     _is_hydrated: bool
     _is_rehydrated: bool
 
+    # Not all object subclasses have a meaningful "name" concept
+    # So whether they expose this is a matter of having a name property
+    _name: Optional[str]
```
thomasjpfan, line 53: I understand why this is optional, but typing checking would be a little more annoying for objects that usually have a name:

```python
my_dict = modal.Dict.from_name("hello")
# do something with my_dict

if my_dict.name is not None:
    use_dict_name(my_dict.name)
```

I do not see a great solution to be more strict on the type tho, especially since ephemeral Dicts do not names.

modal/volume.py
```diff
@@ -167,6 +180,20 @@ async def _load(new_volume: _Volume, resolver: Resolver, existing_object_id: Opt
         obj = _Volume._from_loader(_load, "Volume()", hydrate_lazily=True, deps=lambda: [self])
         return obj
 
+    @property
+    def name(self) -> Optional[str]:
+        return self._name
+
+    def _hydrate_metadata(self, metadata: Optional[Message]):
+        if metadata and isinstance(metadata, api_pb2.VolumeMetadata):
```
mwaskom: Maybe we should do this kind of type checking consistently (this predated these changes since metadata was used for Volume version traking); not sure, it's a little annoying.

thomasjpfan: From scanning the code base on `main`, we already do the type checking consistently. I'm +1 on consistently checking.

Very nit: I like early existing more:

```python
def _hydrate_metadata(self, metadata: Optional[Message]):
    if not (metadata and isinstance(metadata, api_pb2.VolumeMetadata)):
        raise TypeError(...)

    self._metadata = metadata
    ...
```

modal/volume.py
```diff
@@ -92,6 +93,18 @@ def _from_proto(cls, proto: api_pb2.FileEntry) -> "FileEntry":
...
+@dataclass
+class VolumeInfo:
+    """Information about the Volume object."""
+
+    # This dataclass should be limited to information that is unchanging over the lifetime of the Volume,
+    # since it is transmitted from the server when the object is hydrated and could be stale when accessed.
+
+    name: Optional[str] = None
+    created_at: Optional[datetime] = None
+    created_by: Optional[str] = None
```
thomasjpfan: When is `created_at` or `created_by` `None`?

modal/_object.py
mwaskom: > especially since ephemeral Dicts do not names

Yeah this is the main and fairly unmitigatable reason

modal/volume.py
mwaskom: For `created_by` it' None of the object was created before we tracked the creator (i.e. last week).

I think `created_at` might actually always be available though.

modal/volume.py
mwaskom: yeah we do it inconsistently though, and usually with an assertion, which i think is more about satisfying static typing. it's not a very actionable user error message.

modal/volume.py
mwaskom: Oh right, I originally wasn't returning any creation metadata for ephemeral resources, so it would have been absent in that case. (In theory there should not be ambiguity from the perspective of the process that can access an ephemeral resource who created it). But I decided it was simpler on the client side to always have that info.

thomasjpfan (approved):
✅

PR #3395, opened 2025-07-28
New protos for list APIs

mwaskom:
## Describe your changes

Updates protos for resource list RPCs to (1) support pagination and (2) return metadata so a list API can produce hydrated objects.

- Part of SDK-428

---

## Compatibility checklist

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1140,8 +1140,7 @@ message DictLenResponse {
 
 message DictListRequest {
   string environment_name = 1;
-  int32 max_objects = 2;
-  double created_before = 3;
+  ListPagination pagination = 2;
 }
 
 message DictListResponse {
@@ -2098,6 +2097,11 @@ message InputInfo {
   bool task_first_input = 7;
 }
 
+message ListPagination {
+  int32 max_objects = 1;
+  double created_before = 2;
+}
+
 message MapAwaitRequest {
   string function_call_id = 1;
   string last_entry_id = 2;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1140,6 +1140,8 @@ message DictLenResponse {
 
 message DictListRequest {
   string environment_name = 1;
+  int32 max_objects = 2;
+  double created_before = 3;
```
thomasjpfan: Is `max_objects` the amount of objects returned? Specifically:

```python
for vol in modal.Volume.objects.list(max_objects=20, created_before=0):
    assert len(vol) == 20
```

mwaskom: Yes it's for pagination. Not sure I'll use those names exactly in the public API (also the `created_before` will be `datetime | str`).

modal_proto/api.proto
```diff
@@ -1140,15 +1140,20 @@ message DictLenResponse {
...
```
mwaskom: I kind of want to package these two into a `ListPagination` message.

thomasjpfan: +1 on `ListPagination`. In the future, do you see add more options to `ListPagination`?

mwaskom: It's hard to imagine what they would be, so it doesn't feel critical. Just a little cleaner.

mwaskom: Went with `ListPagination` seems nice.

mwaskom:
@prbot approve

thomasjpfan (approved):
✅

PR #3396, opened 2025-07-29
Sandboxes: renamed variadic 'entrypoint_args' to 'cmd'

thundergolfer:
## Describe your changes

Functionally setting this overrides `CMD` on the container. `cmd` is thus IMO a better name.

**Aside:** Kubernetes unhelpfully has different terms here: 

<kbd>
image
</kbd>

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -60,12 +60,12 @@ if TYPE_CHECKING:
     import modal.app
 
 
-def _validate_exec_args(cmd: Sequence[str]) -> None:
+def _validate_exec_args(args: Sequence[str]) -> None:
     # Entrypoint args must be strings.
-    if not all(isinstance(arg, str) for arg in cmd):
+    if not all(isinstance(arg, str) for arg in args):
         raise InvalidError("All entrypoint arguments must be strings")
     # Avoid "[Errno 7] Argument list too long" errors.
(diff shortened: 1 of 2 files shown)


modal/sandbox.py
```diff
@@ -250,7 +250,7 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
 
     @staticmethod
     async def create(
-        *entrypoint_args: str,
+        *cmd: str,  # Set the CMD of the Sandbox, overriding any CMD of the container image.
```
mwaskom: Does it feel like it's easier to miss that this is a variadic arglist and not a single parameter with the singular name?

mwaskom: It looks like `Sandbox.exec` uses `*cmds` which is better in being non-plural but also a bit ambiguous as to whether you can provide _multiple commands_ rather than a sequence of arguments constituting a single command.

thundergolfer: Yeah it probably does. But it trades off against the fact that this is setting the CMD. Could be persuaded to just go with the K8s `args` and comment that it's setting CMD.

mwaskom (approved):

PR #3398, opened 2025-07-29
Add Authentication String To FunctionBindParams

CyrilSharma:
This is only used by the input-plane, which unfortunately is forced to use a public facing API to communicate with the control plane.


devennavani (approved):
LGTM

PR #3401, opened 2025-07-30
Better error message in uv_sync

ekzhang:
Make this error message better:

image

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1397,7 +1397,7 @@ class _Image(_Object, type_prefix="im"):
                 dependencies = pyproject_toml_content["project"]["dependencies"]
             except KeyError as e:
                 raise InvalidError(
-                    f"Invalid pyproject.toml file: missing key {e!r} in {pyproject_toml}. "
+                    f"Invalid pyproject.toml file: missing key {e} in {pyproject_toml}. "
                     "See https://packaging.python.org/en/latest/guides/writing-pyproject-toml for guidelines."
                 )
 
--- a/modal/image.py
+++ b/modal/image.py
@@ -1393,22 +1393,28 @@ class _Image(_Object, type_prefix="im"):
                 # a requirement in `uv.lock`
                 return
 
-            dependencies = pyproject_toml_content["project"]["dependencies"]
+            try:
+                dependencies = pyproject_toml_content["project"]["dependencies"]
 
-            for group in groups:
-                if (
-                    "dependency-groups" in pyproject_toml_content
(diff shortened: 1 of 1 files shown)


modal/image.py
```diff
@@ -1393,22 +1393,28 @@ def _check_pyproject_toml(pyproject_toml: str, version: ImageBuilderVersion):
...
+                    if (
+                        "project" in pyproject_toml_content
+                        and "optional-dependencies" in pyproject_toml_content["project"]
+                        and extra in pyproject_toml_content["project"]["optional-dependencies"]
+                    ):
+                        dependencies += pyproject_toml_content["project"]["optional-dependencies"][extra]
+            except KeyError as e:
+                raise InvalidError(
+                    f"Invalid pyproject.toml file: missing key {e!r} in {pyproject_toml}. "
+                    "See https://packaging.python.org/en/latest/guides/writing-pyproject-toml for guidelines."
```
mwaskom, line 1401: This is a good change in general but is it a distinct failure mode if they have a valid `pyproject.toml` but pass the name of a `group` or `extra` that isn't defined as a parameter to `uv_sync`? Is it worth distinguishing between those cases to avoid sending users down the wrong debugging path?

thomasjpfan: I think we can restrict the `try, except KeyError` to just `dependencies = pyproject_toml_content["project"]["dependencies"]`. The rest of the code is defensive on checking for the key existing.

mwaskom: Won't they get a similar unhelpful (when truncated) error message if they misspell or misremember the name of the `group` or `extra` they want to pass? I like the idea of enriching the error message a lot, I just think it's not quite right that the pyproject.toml itself is invalid in that case.

thomasjpfan: Moving `try except` up would only catch the error around "project" being missing from `pyproject.toml`, which is what Eric ran into.

We'll need another PR to check if `groups` and `extras` are misspelled, etc. Currently we do not check in the client and do not raise an error. Although, there will be an error during image build time from `uv`, so it's not hard to fix.

ekzhang: Maybe better to just close this PR then — sounds like you have some ideas of where you want to take this refactor differently! I was wondering if a broad stroke would be better but it seems that being more precise would be good.

thomasjpfan: I push a commit to move the check up to only catch the error Eric ran into. I think it's already better than what we have in `main`.

thomasjpfan (approved):

PR #3402, opened 2025-07-30
Add `.objects.list()` method to resource object APIs

mwaskom:
## Describe your changes

Client side work that closes SDK-544

---

## Compatibility checklist

</details>

## Changelog

- Added a `.options` namespace to Modal resource object types (`modal.Dict`, `modal.Queue`, `modal.Volume`, and `modal.Secret`).

### DIFF
--- a/modal/cli/dict.py
+++ b/modal/cli/dict.py
@@ -38,16 +38,10 @@ async def create(name: str, *, env: Optional[str] = ENV_OPTION):
 async def list_(*, json: bool = False, env: Optional[str] = ENV_OPTION):
     """List all named Dicts."""
     env = ensure_env(env)
-    client = await _Client.from_env()
-    # request = api_pb2.DictListRequest(environment_name=env)
-    # response = await retry_transient_errors(client.stub.DictList, request)
-    # rows = [(d.name, timestamp_to_local(d.created_at, json)) for d in response.dicts]
-
-    dicts = await _Dict.objects.list(client=client, environment_name=env)  # TODO expose pagination
+    dicts = await _Dict.objects.list(environment_name=env)  # TODO expose pagination?
     rows = []
     for obj in dicts:
         info = await obj.info()
-        # TODO decide what to do about time zone localization in the info API
         rows.append((info.name, timestamp_to_localized_str(info.created_at.timestamp(), json), info.created_by))
 
     display_table(["Name", "Created at", "Created by"], rows, json)
--- a/modal/cli/dict.py
+++ b/modal/cli/dict.py
@@ -38,7 +38,7 @@ async def create(name: str, *, env: Optional[str] = ENV_OPTION):
 async def list_(*, json: bool = False, env: Optional[str] = ENV_OPTION):
     """List all named Dicts."""
     env = ensure_env(env)
-    dicts = await _Dict.objects.list(environment_name=env)  # TODO expose pagination in the request
+    dicts = await _Dict.objects.list(environment_name=env)
     rows = []
     for obj in dicts:
         info = await obj.info()
--- a/modal/cli/dict.py
+++ b/modal/cli/dict.py
@@ -38,7 +38,7 @@ async def create(name: str, *, env: Optional[str] = ENV_OPTION):
 async def list_(*, json: bool = False, env: Optional[str] = ENV_OPTION):
     """List all named Dicts."""
     env = ensure_env(env)
-    dicts = await _Dict.objects.list(environment_name=env)  # TODO expose pagination?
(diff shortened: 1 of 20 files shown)


modal/cli/dict.py
```diff
@@ -41,11 +39,18 @@ async def list_(*, json: bool = False, env: Optional[str] = ENV_OPTION):
...
-    display_table(["Name", "Created at"], rows, json)
+    # request = api_pb2.DictListRequest(environment_name=env)
+    # response = await retry_transient_errors(client.stub.DictList, request)
+    # rows = [(d.name, timestamp_to_local(d.created_at, json)) for d in response.dicts]
+
+    dicts = await _Dict.objects.list(client=client, environment_name=env)  # TODO expose pagination
+    rows = []
+    for obj in dicts:
+        info = await obj.info()
+        # TODO decide what to do about time zone localization in the info API
```
thomasjpfan: For me, I'll want the times shown here to be consistent with the times shown in the UI.

mwaskom:
Doctests are gonna fail until the server-side is merged since they hit prod

mwaskom:
Decided to make the APIs (and CLIs) return all objects by default, which seems easier than forcing users to use the awkward pagination pattern if they want everything.

modal/cli/profile.py
```diff
@@ -19,6 +19,7 @@
 @profile_cli.command(help="Change the active Modal profile.")
 def activate(profile: str = typer.Argument(..., help="Modal profile to activate.")):
     config_set_active_profile(profile)
+    typer.echo(f"Active profile: {_profile}")
```
ehdr: This is addressing another user request right?

test/cli_test.py
```diff
@@ -46,6 +47,13 @@
 dummy_other_module_file = "x = 42"
 
 
+def windows_sleep():
```
ehdr, line 50: Could this be moved into the test fixtures themselves somehow? I'm guessing not, but best to ask!

modal/cli/profile.py
mwaskom: whoops somehow pulled in a change i'd applied when testing https://github.com/modal-labs/modal-client/pull/3408; thanks for catching

test/cli_test.py
mwaskom: i don't think so; you need to do it between creating the object and listing them. i'm not 100% certain of this explanation but i think the issue is that since we return objects with `created_at < created_before`, and the windows clock resolution is so low, without the sleep some objects will be created at the "exact same time" as the list query originates and won't be included in the result.

modal/dict.py
```diff
@@ -39,6 +46,70 @@ class DictInfo:
...
+        number of results and to filter by creation date:
+
+        ```python
+        dicts = modal.Dict.objects.list(max_objects=10, created_before="2025-01-01")
+        ```
+
+        """
+        client = await _Client.from_env() if client is None else client
+
+        async def retrieve_page(page_size: int, created_before: float) -> list[api_pb2.DictListResponse.DictInfo]:
```
thomasjpfan: Nit:

```suggestion
        async def retrieve_page(page_size: Optional[int], created_before: float) -> list[api_pb2.DictListResponse.DictInfo]:
```

Interesting that `ListPagination` interpreters `None` as zero and the server converts 0 to `MAX_LIST_MAX_OBJECTS`.

For `max_objects=0`, I think we end up returning objects and then filtering them all out later.

modal/dict.py
```diff
@@ -39,6 +46,70 @@ class DictInfo:
...
+            return resp.dicts
+
+        items = await retrieve_page(max_objects, as_timestamp(created_before))
+        if not items:
+            return []
+
+        while True:
+            if max_objects is not None and len(items) >= max_objects:
+                break
+            new_items = await retrieve_page(100, items[-1].metadata.creation_info.created_at)
```
thomasjpfan: What do you think of being exact and keep track of the remaining items we need? Something like:

```python
page_size = 100 if max_objects is None else min(max_objects - len(items), 100)
```

test/cli_test.py
ehdr: Yeah makes sense. Always awkward to have to rely on timestamps, but I guess not much we can do without some monotonic ID?

modal/dict.py
mwaskom: This is just normal protobuf behavior ... I think we need to do that so that we can preserve what we return to older clients, right?

thomasjpfan: Yea, I do not think we need to change anything in the RPC.

PR #3403, opened 2025-07-30
Localize timestamps in object info

mwaskom:
Follow-on to #3393 making the `created_at` metadata a localized datetime, which matches what we currently do in the UI and CLI

### DIFF
--- a/modal/_utils/time_utils.py
+++ b/modal/_utils/time_utils.py
@@ -3,13 +3,17 @@ from datetime import datetime
 from typing import Optional
 
 
-def timestamp_to_local(ts: float, isotz: bool = True) -> Optional[str]:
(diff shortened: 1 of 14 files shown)


modal/_utils/time_utils.py
```diff
@@ -3,13 +3,17 @@
 from typing import Optional
 
 
-def timestamp_to_local(ts: float, isotz: bool = True) -> Optional[str]:
+def timestamp_to_localized_dt(ts: float) -> datetime:
```
mwaskom, line 6: I kinda think it's preferable to avoid making the return type Optional and forcing users to handle that possibility. Though there is some risk that eventually someone is looking at 1970-01-01 if we start reading dates from the server that haven't always existed.

thomasjpfan (approved):

PR #3405, opened 2025-07-31
Increase slack on sandbox exec timeout test

mattnappo:
https://modal-com.slack.com/archives/C056CGAANRM/p1753970330154559

### DIFF
--- a/test/sandbox_test.py
+++ b/test/sandbox_test.py
@@ -318,7 +318,7 @@ def test_sandbox_exec_wait_timeout(app, servicer):
     cp = sb.exec("sleep", "999", timeout=1)
     t0 = time.monotonic()
     assert cp.wait() == -1
-    assert 0.5 < time.monotonic() - t0 <= 1.5
+    assert 0.8 < time.monotonic() - t0 <= 1.2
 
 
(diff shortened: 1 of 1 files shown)


test/sandbox_test.py
```diff
@@ -318,7 +318,7 @@ def test_sandbox_exec_wait_timeout(app, servicer):
     cp = sb.exec("sleep", "999", timeout=1)
     t0 = time.monotonic()
     assert cp.wait() == -1
-    assert 1 < time.monotonic() - t0 <= 1.2
+    assert 0.5 < time.monotonic() - t0 <= 1.5
```
thomasjpfan: Is this too much slack for `timeout=1`? Do you think its enough to do `0.8 < T < 1.2`?

mattnappo: Sure, we can try 0.8 first and if it still flakes often, I'll bump it even further.

mwaskom: The question for me is what are we actually trying to assert about here? The specific timeout duration is mostly controlled by the behavior of the mock server, which we don't actually care about, right?

That new range seems good though.

mattnappo: That's correct, which I why I don't think the exact tolerance needs to be tight in these sorts of timing-based tests.

thomasjpfan (approved):

PR #3407, opened 2025-07-31
docs: update contributing guide

davidxia:
with correct version of Python used in CI/CD and mention installing pre-commit hooks.


mwaskom:
> with correct version of Python used in CI/CD 

Hi, what do you mean by this? We run CI/CD across all supported Python versions

mwaskom:
We use 3.11 in our internal dev setup

davidxia:
OK, just added a pre-commit step then

PR #3409, opened 2025-07-31
Improve docs for Volume.commit

thomasjpfan:
## Describe your changes

Improves docstring for `Volume.commit`

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -377,7 +377,7 @@ class _Volume(_Object, type_prefix="vo"):
 
     @live_method
     async def commit(self):
-        """Commit changes to the volume.
+        """Commit local changes to a mounted volume.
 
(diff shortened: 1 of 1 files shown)


modal/volume.py
```diff
@@ -377,7 +377,7 @@ async def _do_reload(self, lock=True):
 
     @live_method
     async def commit(self):
-        """Commit changes to the volume.
+        """Commit local changes to a mounted volume.
```
mwaskom: Is this confusing in that we usually differentiate "local" from "remote" where "local" means something like "on your personal computer rather than on Modal infra"?

thomasjpfan: Removed the word "local" and kept "mounted"

mwaskom (approved):

PR #3410, opened 2025-07-31
Improve docs on cls.timeout regarding @modal.enter

thomasjpfan:
## Describe your changes

Improve docs on cls.timeout regarding `@modal.enter`

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -869,7 +869,8 @@ class _App:
         scaledown_window: Optional[int] = None,  # Max time (in seconds) a container can remain idle while scaling down.
         proxy: Optional[_Proxy] = None,  # Reference to a Modal Proxy to use in front of this function.
         retries: Optional[Union[int, Retries]] = None,  # Number of times to retry each input in case of failure.
-        timeout: Optional[int] = None,  # Maximum execution time of the function in seconds.
+        # Maximum execution time of the function in seconds, including time spent in `@modal.enter`.
+        timeout: Optional[int] = None,
         cloud: Optional[str] = None,  # Cloud provider to run the function on. Possible values are aws, gcp, oci, auto.
         region: Optional[Union[str, Sequence[str]]] = None,  # Region or regions to run the function on.
         enable_memory_snapshot: bool = False,  # Enable memory checkpointing for faster cold starts.


modal/app.py
```diff
@@ -869,7 +869,8 @@ def cls(
         scaledown_window: Optional[int] = None,  # Max time (in seconds) a container can remain idle while scaling down.
         proxy: Optional[_Proxy] = None,  # Reference to a Modal Proxy to use in front of this function.
         retries: Optional[Union[int, Retries]] = None,  # Number of times to retry each input in case of failure.
-        timeout: Optional[int] = None,  # Maximum execution time of the function in seconds.
+        # Maximum execution time of the function in seconds, including time spent in `@modal.enter`.
```
mwaskom: I could be misreading it, but I think this is not quite phrased correctly. The `timeout` is applied to the startup phase _and_ to each input, but it's applied separately. The timekeeping for an input doesn't "include" the time you spent in `enter`. If you have a 60 s timeout and spend 55s starting the container and then 30s running an input, the input doesn't timeout. Clearly this is hard to explain though!

Also, it's applied to the whole startup phase, not just to running the `enter()` methods. So e.g.g if you have an import that's really slow and a short timeout, we could fail the container even if there's no `@modal.enter` method.

modal/app.py
```diff
@@ -869,7 +869,9 @@ def cls(
         scaledown_window: Optional[int] = None,  # Max time (in seconds) a container can remain idle while scaling down.
         proxy: Optional[_Proxy] = None,  # Reference to a Modal Proxy to use in front of this function.
         retries: Optional[Union[int, Retries]] = None,  # Number of times to retry each input in case of failure.
-        timeout: int = 300,  # Maximum execution time of the function in seconds.
+        # Maximum execution time of the function in seconds. This timeout applies independently to `@modal.enter` and
+        # functions.
```
mwaskom: We should keep the documentation synced between `app.function()` and `app.cls()` IMO. (Of course the need to do this is part of why I don't like Cls, but still...)

Also I think these inline parameter comments are a lot easier to read if they're all on the right, so unless you really need some verbose explanation, it's user-friendly to play with the language until you can make it fit. e.g.

> Maximum execution time of the function in seconds; applies independently to the startup phase and each input

PR #3411, opened 2025-07-31
Add gpu count to task info

gleb-modal:
## Describe your changes

Corresponding modal pr: https://github.com/modal-labs/modal/pull/25217

https://linear.app/modal-labs/issue/PRD-988/display-number-of-gpus-next-to-type-in-container-drawer

## Compatibility checklist

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -3008,7 +3008,7 @@ message TaskInfo {
   string gpu_type = 6;
   string sandbox_id = 7;
   TaskSnapshotBehavior snapshot_behavior = 8;
-  uint32 gpu_count = 9;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -3008,6 +3008,7 @@ message TaskInfo {
   string gpu_type = 6;
   string sandbox_id = 7;
   TaskSnapshotBehavior snapshot_behavior = 8;
+  uint32 gpu_count = 9;
```
mwaskom: j/w but is there a reason not to be using the `GPUConfig` message here?

gleb-modal: Good idea, updated.

mwaskom (approved):

PR #3419, opened 2025-08-01
MVP: `modal launch workspace`

ekzhang:
This is a starting point for a command that just launches an instance with a mounted volume and allows you to SSH into it.

It's meant to be a basic primitive with reasonably good developer experience. The SSH key is generated randomly / ephemerally.

It relies on the notebook base image, so users don't need to build the image from scratch on first usage.

Resolves SPR-142.


modal/experimental/__init__.py
```diff
@@ -212,6 +213,109 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
     )
 
 
+@synchronizer.create_blocking
+async def notebook_base_image(*, python_version: Optional[str] = None, force_build: bool = False) -> _Image:
```
freider, line 217: Would it make sense to have a non-notebook-associated version of this recipe that's used by "machines", expose that in the sdk under some more general name, and then just superimpose whatever packages are only required by notebooks as a separate layer?

Also, have we considered "versioning" this image at some point with pinned dependencies per version? It could make the workflow of going from notebook/machine experimentation -> deployed code a lot more streamlined IMO

Issue #3381, opened 2025-07-24
Error with image.add_local_dir using FilePatternMatcher.from_file() for ignore


staceymelville:
### Summary

When creating a `FilePatternMatcher` using the `from_file` class method the `patterns` attribute is unset until the instance is called when the `_delayed_init` method runs. In version `1.1.0` the method `can_prune_directories` was [added](https://github.com/modal-labs/modal-client/commit/fc24968d845e55b2352bee4a05a18b202ff0109b#diff-a684f3a5f9eed73782c46d20fac17fc3aa01490148f829b9f6ddfcd31dbf8d6f) and it can be called off of a `FilePatternMatcher` instance that [hasn't previously been called](https://github.com/modal-labs/modal-client/blob/9eddb0cca8f0e1a6dbf3d0ef44e61faeae1515ec/modal/mount.py#L173) which leads to the error `'FilePatternMatcher' object has no attribute 'patterns'` when you create an image with the following:
```python
ignore = modal.FilePatternMatcher.from_file(".gitignore")
image = (
    modal.Image.from_registry(
        "nvidia/cuda:12.8.0-cudnn-devel-ubuntu22.04", add_python="3.11"
    )
    .add_local_dir("foo", remote_path="/foo",ignore=ignore)
```

The simplest fix would be to add the `delayed_init` call for the FilePatternMatcher to the `can_prune_directories` [method](https://github.com/modal-labs/modal-client/blob/9eddb0cca8f0e1a6dbf3d0ef44e61faeae1515ec/modal/file_pattern_matcher.py#L205) as is done for the `__call__` method below.

### Version

1.1.0

PR #3420, opened 2025-08-04
Make FilePatternMatcher.can_prune_directories handle delayed init

ehdr:
This fixes issue #3381 / SDK-535.

---

## Compatibility checklist

---

</details>

## Changelog

- Fixed a bug that would cause e.g. some image builds to fail with `'FilePatternMatcher' object has no attribute 'patterns'` when using a `FilePatternMatcher.from_file` ignore pattern.


mwaskom (approved):
Sort of annoying that we need to remember to do this everywhere, seems like a design smell?

ehdr:
Yeah good callout @mwaskom ! What do you think about this, as a `@cached_property` instead?

mwaskom:
Nice!!

PR #3426, opened 2025-08-05
fix: remove unused `modal._runtime.asgi` import

davidxia:


mwaskom:
fyi Elias is out right now — we should only request followup reviews from people that can get to them in a timely manner

PR #3428, opened 2025-08-06
Add preload_path_prefixes to sandbox defn

AmitPr:
## Describe your changes

Used internally to address sandbox startup performance in special cases.

---

## Compatibility checklist


ekzhang (commented):
sgtm 👍

PR #3432, opened 2025-08-06
Improve most Object subclass reprs

mwaskom:
## Describe your changes

This improves the way we provide reprs for our Object types.

Currently we have a mix of things that are not [strictly valid reprs](https://docs.python.org/3/reference/datamodel.html#object.__repr__) in the sense that they can
> be used to recreate an object with the same value (given an appropriate environment)

Now we generally speaking will try to represent an Object with how it could be instantiated via its `.from_name()` constructor, including accounting for the environment name (but _not_ any `create_if_missing` value or any other configuration that's snuck into that type's `.from_name()`):

```python
In [1]: import modal

In [2]: modal.Dict.from_name("a-dict")
Out[2]: modal.Dict.from_name('a-dict')

In [3]: modal.Dict.from_name("a-dict", environment_name="dev")
Out[3]: modal.Dict.from_name('a-dict', environment_name='dev')

In [4]: modal.Volume.objects.list()
Out[4]:
[modal.Volume.from_name('hf-hub-cache'),
 modal.Volume.from_name('train-data'),
 modal.Volume.from_name('test-build-enter-volume'),
 ...
 modal.Volume.from_name('test-vol')]
```

I do wonder if we'd like to also have a "richer" str representation of objects that makes the result of printing `modal.Dict.object.list()` look closer to e.g. the `modal dict list` output (in terms of showing useful metadata, not in terms of the tabular formatting) but this seems to be what we were previously aiming for, but not quite achieving, with the default reprs.

Closes SDK-545
Supersedes #3248

### DIFF
--- a/modal/_object.py
+++ b/modal/_object.py
@@ -192,8 +192,19 @@ class _Object:
         return cls._get_type_from_id(object_id) == cls
 
     @classmethod
+    def _repr(cls, name: str, environment_name: Optional[str] = None) -> str:
+        public_cls = cls.__name__.strip("_")
+        environment_repr = f", environment_name={environment_name!r}" if environment_name else ""
+        return f"modal.{public_cls}.from_name({name!r}{environment_repr})"
+
+    @classmethod
     def _new_hydrated(
-        cls, object_id: str, client: _Client, handle_metadata: Optional[Message], is_another_app: bool = False
+        cls,
+        object_id: str,
+        client: _Client,
+        handle_metadata: Optional[Message],
+        is_another_app: bool = False,
+        rep: Optional[str] = None,
     ) -> Self:
         obj_cls: type[Self]
         if cls._type_prefix is not None:
@@ -210,7 +221,7 @@ class _Object:
 
         # Instantiate provider
         obj = _Object.__new__(obj_cls)
-        rep = f"Object({object_id})"  # TODO(erikbern): dumb
+        rep = rep or f"modal.{obj_cls.__name__.strip('_')}.from_id({object_id!r})"
         obj._init(rep, is_another_app=is_another_app)
         obj._hydrate(object_id, client, handle_metadata)
 
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1276,7 +1276,8 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
 
             self._hydrate(response.function_id, resolver.client, response.handle_metadata)
 
-        rep = f"Function.from_name('{app_name}', '{name}')"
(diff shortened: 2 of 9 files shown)


modal/_object.py
```diff
@@ -210,7 +221,7 @@ def _new_hydrated(
 
         # Instantiate provider
         obj = _Object.__new__(obj_cls)
-        rep = f"Object({object_id})"  # TODO(erikbern): dumb
+        rep = rep or f"modal.{obj_cls.__name__.strip('_')}.from_id({object_id!r})"
```
mwaskom, line 224: This is not actually valid for most object types! Maybe we should add it. But it actually is valid as a fallback for some types that don't have a `.from_name()` (e.g. it works out of the box for `modal.FunctionCall.from_id`). And since we allow methods that call into here to define their own `rep`, we hit the fallback less often. So it's less "dumb" imo 😁

modal/_object.py
```diff
@@ -191,9 +191,20 @@ def _get_type_from_id(object_id: str) -> type["_Object"]:
     def _is_id_type(cls, object_id) -> bool:
         return cls._get_type_from_id(object_id) == cls
 
+    @classmethod
+    def _repr(cls, name: str, environment_name: Optional[str] = None) -> str:
```
thomasjpfan, line 195: **No action required**: The pattern of passing the repr to the constructor and not having the subclasses define their own `__repr__` is a bit weird to me, but it's the most consistent with the codebase today.

modal/_object.py
thomasjpfan: Although it is incorrect, I'm okay with this because it's better than `Object({object_id})` which was wrong all the time.

**Hot take**: I would be okay with forcing all callers to pass in a repr.

modal/_object.py
mwaskom: I'd say it's a consequence of not having a "normal" constructor / one obvious way to instantiate the class. At this point we've mostly aligned on `.from_name()` in most cases, but not all objects types, or instances of any specific type, have a name. So that's why there needs to be some context-dependent definition of the repr.

modal/_object.py
mwaskom: That would probably be annoying in some edge cases, e.g. we call into `_new_hydrated` when unpicking a modal object (we have custom pickle hooks that use hydration to pass objects around — actually we should double check that this handles the new metadata ok!!) and don't do the work to figure out what kind of object it is until we're inside this method.

mwaskom: The hotter take I'd be more interested in is "why don't we support `.from_id()` everywhere?" 😁

thomasjpfan (approved):

PR #3434, opened 2025-08-07
Increase timeout on fork tests

mwaskom:
These tests are flaking often with a TimeoutError, especially on the MacOS-13 build.

I'm guessing that just doubling the timeout will cut down on flakes substantially but @thecodingwizard  would appreciate if you can give it a little thought whether there's a deeper issue here.

Closes SVC-808


thecodingwizard (commented):
nice. I took a look and am not sure why there would be a timeout error. However 2 seconds maybe seems a bit short to launch a process, import modal, connect to the server, fork the process, re-connect, etc, so hopefully 4 seconds will help here.

PR #3435, opened 2025-08-07
[SpawnMapV2](1) Proto Change

CyrilSharma:
This adds a new field to allow specifying if the client would like a single output or not.


devennavani (approved):
LGTM

PR #3438, opened 2025-08-07
[SpawnMapV2](?) Add An In-Order Iterator To Experimental Spawn Map

CyrilSharma:
Add An In-Order Iterator To Experimental Spawn Map


modal/_functions.py
```diff
@@ -360,6 +360,42 @@ async def run_generator(self):
...
+                self.stub.FunctionGetOutputs,
+                request,
+                attempt_timeout=ATTEMPT_TIMEOUT_GRACE_PERIOD,
+            )
+
+            outputs = list(response.outputs)
+            outputs.sort(key=lambda x: x.idx)
+            for output in outputs:
+                if output.idx != current_index:
+                    continue
```
devennavani: break

devennavani (commented):
fix nit and add usage to docstring

modal/_functions.py
```diff
@@ -1865,6 +1902,47 @@ async def get(self, timeout: Optional[float] = None, *, index: int = 0) -> Retur
         """
         return await self._invocation().poll_function(timeout=timeout, index=index)
 
+    @live_method_gen
+    async def iter(self, *args) -> AsyncIterator[Tuple[int, ReturnType]]:
```
mwaskom: FWIW my initialy reaction here was "WTF are the *args?". I think the `range()` signature is pretty fussy. Might prefer an explicit `start=0, end=None` parameterization, but don't feel super strongly.

modal/_functions.py
```diff
@@ -1865,6 +1902,47 @@ async def get(self, timeout: Optional[float] = None, *, index: int = 0) -> Retur
...
+        """Iterate in-order over (index, result) pairs of the function call.
+        Optionally, specify a range [start, end) to iterate over.
```
mwaskom, line 1909: Nit; the one-line summary should be a single line and should have a blank before any additional explanation.

modal/_functions.py
```diff
@@ -1865,6 +1902,47 @@ async def get(self, timeout: Optional[float] = None, *, index: int = 0) -> Retur
         """
         return await self._invocation().poll_function(timeout=timeout, index=index)
 
+    @live_method_gen
+    async def iter(self, *args) -> AsyncIterator[Tuple[int, ReturnType]]:
+        """Iterate in-order over (index, result) pairs of the function call.
```
mwaskom: I might expect a method that returns an (index, result) pair to be called something like "enumerate".

OTOH would some name that includes "get" help clarify what this is? IDK if `FunctionCall.iter_get()` is sort of akward though.

mwaskom: Another idea might be an opt-in flag to return the index but (but then you'd need an annoying overload type).

Is it possible for users to do `for i, result in enumerate(fc.iter()): ...`?

PR #3440, opened 2025-08-08
make the sandbox default timeout explicit

rolyatmax:
## Describe your changes

This PR makes the sandbox default timeout more explicit. Before now it was still set to 300 seconds, but this change has the docs more clearly representing that.

---

## Compatibility checklist

---

</details>


rolyatmax:
> our established pattern for things like this tends to be having the default actually set by the server
@mwaskom - To be sure, I think we'll still need to set defaults on the server (if for no other reason than to support old clients), but I see two issues with not putting those defaults in the client:

1. We use the SDK's function signatures to generate our docs. AFAICT, this is currently the only place in our docs where we actually talk about sandbox timeouts, and it implies that sandboxes don't have a timeout set by default - which is false.
2. Python tooling for IDEs show that the `Sandbox.create()` `timeout` arg's default is explicitly `None`, which is also misleading.

Also, there are several function args that _do_ have their defaults set in the client. Here are a few just from `Sandbox`:
```
sb.snapshot_filesystem(timeout=55)
sb.tunnels(timeout=50)
sb.open(..., mode="r")
sb.exec(..., stdout=PIPE, stderr=PIPE, text=True)
sb.wait(raise_on_termination=True)
```

mwaskom:
I'm not saying that defaults are _always_ set serverside; some of these wouldn't really make sense to have there because they only control client behavior. But the ones that just get passed through to the server and configure what it (or the worker it dispatches to) might.

Again, I totally see the argument for recoding the default in the client. It's a lot friendlier! But doing it inconsistently has costs IMO. Wouldn't it make a user more likely to assume that the `timeout: Optional[None]` in `App.function()` means that Functions don't have a default timeout?

I've sort of casually wondered in the past if we want to at least have some sort of `ServerDefault` sentinel to help users distinguish between "None means the behavior controlled by this behavior is not in effect by default" and "None means there is a default set by our servers". We could even potentially hook that up somehow in the docs build to show the (current at time of build) default on the website! That would be neat.

I don't know if we have a fully articulated rationale for why we tend to set these sort of defaults in the server. In theory it gives us more flexibility to change them, although in practice we'd probably see this sort of change as unacceptably disruptive.

rolyatmax:
@mwaskom 
> Wouldn't it make a user more likely to assume that the timeout: Optional[None] in App.function() means that Functions don't have a default timeout?
I actually don't think so. We already have lots of defaults set in the client. Wouldn't those be misleading users, too? I think the fact remains that `timeout=None` is misleading all by itself and _not_ because other parts of the docs happen to more accurately describe what the defaults are.

> I don't know if we have a fully articulated rationale for why we tend to set these sort of defaults in the server. In theory it gives us more flexibility to change them, although in practice we'd probably see this sort of change as unacceptably disruptive. It could just be that most of our devs spend most of their time in that repo and it makes it easy for them to be aware of what defaults are in effect. (Maybe that's a bad reason when weighed against our users devex!)
Completely agree! Also agree that changing defaults on users without them updating any code is unacceptably disruptive.

mwaskom:
We might have a lot of defaults set ... but [`timeout=`](https://github.com/modal-labs/modal-client/blob/35dc4acfdf8297047b19c3a9240bb0d387dbead0/modal/app.py#L644) isn't one of them. I'd prefer we change it in both places if we're going to change it here.

rolyatmax:
@mwaskom 

> We might have a lot of defaults set ... but [`timeout=`](https://github.com/modal-labs/modal-client/blob/35dc4acfdf8297047b19c3a9240bb0d387dbead0/modal/app.py#L644) isn't one of them. I'd prefer we change it in both places if we're going to change it here.

I think that's totally fair. I'll look through to find the other spots where we've got a default `timeout` and update those as well.

modal/image.py
```diff
@@ -2114,7 +2114,7 @@ def run_function(
         gpu: Union[GPU_T, list[GPU_T]] = None,  # Requested GPU or or list of acceptable GPUs( e.g. ["A10", "A100"])
         cpu: Optional[float] = None,  # How many CPU cores to request. This is a soft limit.
         memory: Optional[int] = None,  # How much memory to request, in MiB. This is a soft limit.
-        timeout: Optional[int] = 60 * 60,  # Maximum execution time of the function in seconds.
+        timeout: int = 60 * 60,  # Maximum execution time of the function in seconds.
```
rolyatmax, line 2117: Looks like the type on this one was wrong?

mwaskom: I think we would have accepted `None` here and passed through to `Function.from_local()` which does (did) accept None.

rolyatmax: Gotcha

mwaskom (approved):
Thanks!!

PR #3442, opened 2025-08-08
Add `.objects.create` methods

mwaskom:
## Describe your changes

- Closes SDK-549

---

## Compatibility checklist

---

</details>

## Changelog

- Added `.objects.create()` methods

### DIFF
--- a/modal/dict.py
+++ b/modal/dict.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from typing import Any, Optional, Union
 
 from google.protobuf.message import Message
-from grpclib import GRPCError
+from grpclib import GRPCError, Status
 from synchronicity import classproperty
 from synchronicity.async_wrap import asynccontextmanager
 
@@ -27,7 +27,7 @@ from ._utils.name_utils import check_object_name
 from ._utils.time_utils import as_timestamp, timestamp_to_localized_dt
 from .client import _Client
 from .config import logger
-from .exception import InvalidError, NotFoundError, RequestSizeError
+from .exception import AlreadyExistsError, InvalidError, NotFoundError, RequestSizeError
 
 
 def _serialize_dict(data):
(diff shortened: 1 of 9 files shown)


modal/dict.py
```diff
@@ -49,6 +49,59 @@ class DictInfo:
...
+            object_creation_type=object_creation_type,
+        )
+        try:
+            await retry_transient_errors(client.stub.DictGetOrCreate, req)
+        except GRPCError as exc:
+            if exc.status == Status.ALREADY_EXISTS:
+                if not allow_existing:
+                    raise AlreadyExistsError(exc.message)
+            else:
+                raise
```
thomasjpfan, line 102: Is it okay to raise the original `GRPCError` here? (I thought we were trying to hide them, but we already reraise in the rest of this file)

modal/dict.py
```diff
@@ -49,6 +49,59 @@ class DictInfo:
...
+            deployment_name=name,
+            environment_name=_get_environment_name(environment_name),
+            object_creation_type=object_creation_type,
+        )
+        try:
+            await retry_transient_errors(client.stub.DictGetOrCreate, req)
+        except GRPCError as exc:
+            if exc.status == Status.ALREADY_EXISTS:
+                if not allow_existing:
+                    raise AlreadyExistsError(exc.message)
```
thomasjpfan: Nit:

```suggestion
            if exc.status == Status.ALREADY_EXISTS and not allow_existing:
                raise AlreadyExistsError(exc.message)
```

Same for the other objects.

modal/volume.py
```diff
@@ -116,6 +116,58 @@ class VolumeInfo:
 class _VolumeManager:
     """Namespace with methods for managing named Volume objects."""
 
+    @staticmethod
+    async def create(
+        name: str,
```
thomasjpfan: This is missing comments to document the parameters:

```suggestion
        name: str,  # Name to use for the new Volume
```

modal/dict.py
mwaskom: Well I'd prefer not to have GRPCErrors escape but we can't handle every possible GRPCError at every RPC callsite — we need to spend some time adding some standardized GRPCError -> `modal.exception` type mapping to the internal gRPC utilities. But in the absence of that, we want to coerce the specific status that we're expecting here.

modal/secret.py
```diff
@@ -39,6 +39,63 @@ class SecretInfo:
 class _SecretManager:
     """Namespace with methods for managing named Secret objects."""
 
+    @staticmethod
+    async def create(
+        name: str,  # Name to use for the new Secret
+        env_dict: dict[str, str],  # Dictionary of environment variables to set in the Secret
```
ehdr: "environment variables" here first led me to think that it would read the values _from_ environment variables. Would it make sense to just say "dictionary of key-value pairs to set in the Secret" or something like that?

ehdr (approved):

PR #3446, opened 2025-08-11
volume: Strip trailing zeroes from each block when uploading

dflemstr:
## Describe your changes

We now internally store blocks in a more compact format by stripping trailing zeroes from each block. To reduce the number of bytes transferred over the network, and to aid in block de-duplication (which uses the content SHA256 checksum of each block), we can strip the trailing zeroes already client-side to speed up file transfers significantly.

The alternative to this would involve a lot of backend changes, since we would then instead need to track the stripped and un-stripped digest of every block when de-duplicating blocks.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_utils/blob_utils.py
+++ b/modal/_utils/blob_utils.py
@@ -445,13 +445,23 @@ _FileUploadSource2 = Callable[[], ContextManager[BinaryIO]]
 
 
 @dataclasses.dataclass
+class FileUploadBlock:
+    # The start (byte offset, inclusive) of the block within the file
+    start: int
+    # The end (byte offset, exclusive) of the block, after having removed any trailing zeroes
+    end: int
+    # Raw (unencoded 32 byte) SHA256 sum of the block, not including trailing zeroes
+    contents_sha256: bytes
+
+
+@dataclasses.dataclass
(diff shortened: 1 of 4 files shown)


dflemstr:
Re-based on latest `main`

modal/_utils/blob_utils.py
```diff
@@ -544,31 +554,77 @@ def ceildiv(a: int, b: int) -> int:
...
 
-            num_bytes_read = 0
-            while num_bytes_read < BLOCK_SIZE:
-                chunk = block_fp.read(BLOCK_SIZE - num_bytes_read)
 
-                if not chunk:
-                    break
+def _gather_block(source: _FileUploadSource2, block_idx: int) -> FileUploadBlock:
+    start = block_idx * BLOCK_SIZE
+    end = 1 + _find_last_nonzero_byte(source, start, start + BLOCK_SIZE)
```
danielnorberg: if `_find_last_nonzero_byte` returns `-1` here we might end up with `end == 0` and `start > 0`. Is that expected?

dflemstr: Good catch!

PR #3453, opened 2025-08-12
Avoid building PREVIEW image builder mounts

mwaskom:
We've closed a cycle of Image Builder Version development so we can disable building the PREVIEW mount until we start that up again.


ekzhang (approved):
Nice! Are you seeing that locally or in CI? — I think CI might be significantly slower so it may be good to double-check the runtime of the "setup" step.

PR #3455, opened 2025-08-12
Disable FileIO doctests

mwaskom:
I don't like doing this, but these flake too often, and the FileIO API is "in alpha".


mwaskom:
@prbot approve

PR #3459, opened 2025-08-12
Remove outdated `MapStartOrContinueResponse` comment

davidxia:
`ClientServer.MapStartOrContinue()` now sets
`MapStartOrContinueResponse.RetryPolicy`.


mwaskom:
It's good practice to clean up comments like these lest someone else be mislead and cause problems by changing or removing the field.

PR #3460, opened 2025-08-13
Expose version= in modal.Volume.objects.create

mwaskom:
Brings `modal.Volume.objects.create()` to parity with the `modal volume create` CLI.

Would sort of have preferred we didn't need to expose this semi-internal / transient configuration option in the public API (maybe we should have `experimental_options` everywhere?)


mwaskom:
@prbot approve

dflemstr (commented):
LGTM

PR #3461, opened 2025-08-13
SVC-602 Respect user's retry policy for `.map()`

davidxia:
for new input plane path.

---

## Compatibility checklist

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -673,9 +673,11 @@ async def _map_invocation_inputplane(
     # any reason).
     max_inputs_outstanding = MAX_INPUTS_OUTSTANDING_DEFAULT
 
-    # Input plane does not yet return a retry policy. So we currently disable retries.
+    # Set a default retry policy to construct an instance of _MapItemsManager.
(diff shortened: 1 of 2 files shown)


davidxia:
@bokeefe123 good call. Fixed some bugs on the server side. This now works with the below.

```
MODAL_PROFILE=local modal run test.py
...
Expected Total: 19900, Actual Total: 19900
```

`test.py` is

```python
import random

import modal

app = modal.App("dxia-test-flaky-map")
retries = modal.Retries(max_retries=10, backoff_coefficient=1.0, initial_delay=1.0)


@app.function(experimental_options={"input_plane_region": "us-east"}, retries=retries, max_containers=10)
@modal.concurrent(max_inputs=10)
async def flaky_map(i: int):
    if random.random() < 0.1:
        raise Exception(f"Randomly induced error for input {i}!")

    return i


@app.local_entrypoint()
def main():
    total = 0
    count = 200
    for result in flaky_map.map(range(count)):
        total += result
    print(f"Expected Total: {sum(range(count))}, Actual Total: {total}")
```

PR #3462, opened 2025-08-13
Large filesystem snapshot support

paulgb:
## Compatibility checklist

- [x] Client+Server: this change is compatible with old servers (server support shipped Aug 4 and the functionality is only invoked by an optional flag)

In theory the new code path could replace `snapshot_filesystem` entirely, but since it involves both new server and new client code, I'd rather initially gate it as experimental.

## Changelog

`sandbox.snapshot_filesystem()` now supports an optional `experimental_large_snapshot_support` argument. When enabled, it switches to an async snapshot API that supports arbitrarily large timeouts.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -556,7 +556,8 @@ class _Sandbox(_Object, type_prefix="sb"):
         return image
 
     async def experimental_snapshot_filesystem(self, timeout: Optional[int] = None) -> _Image:
-        """Snapshot the filesystem of the Sandbox, using the experimental async endpoint to support larger snapshots.
+        """mdmd:hidden
+        Snapshot the filesystem of the Sandbox, using the experimental async endpoint to support larger snapshots.
 
         Returns an [`Image`](https://modal.com/docs/reference/modal.Image) object which
         can be used to spawn a new Sandbox with the same filesystem.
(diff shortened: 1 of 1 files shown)


mwaskom:
Would it make sense to enable this via `Sandbox.create(..., experimental_options={...})`?

mwaskom:
I guess it's less clear in this case since the change involves client-side logic too 🤔

mwaskom:
There's different precedents here. E.g. the core-services folks are currently building a `Function.experimental_spawn_map` that shadows the existing `Function.spawn_map` that can then get rotated in as a drop-in replacement. That can also help avoid needing to clutter the signature of the stable method with the experimental parameter going forward.

paulgb:
> There's different precedents here. E.g. the core-services folks are currently building a `Function.experimental_spawn_map` that shadows the existing `Function.spawn_map` and can eventually get rotated in as a drop-in replacement. That can also help avoid needing to clutter the signature of the stable method with the experimental parameter going forward.

I'd be happy to call this `experimental_snapshot_filesystem` if that would be more consistent.

modal/sandbox.py
```diff
@@ -555,6 +555,52 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
 
         return image
 
+    async def experimental_snapshot_filesystem(self, timeout: Optional[int] = None) -> _Image:
+        """Snapshot the filesystem of the Sandbox, using the experimental async endpoint to support larger snapshots.
```
mwaskom: Can we use `mdmd:hidden` so this gets suppressed from the API reference docs? [Example](https://github.com/modal-labs/modal-client/pull/3466)

paulgb: Done!

PR #3464, opened 2025-08-13
Adds Image.build to eagerly build an image

thomasjpfan:
## Describe your changes

Although, eagerly building an image requires an app, I still think it's useful. In a future, where an app is not required, we can make the app kwarg optional.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Adds `image.build` to eagerly build an image:
 
```python
image = modal.Image.debian_slim().uv_pip_install("scipy", "numpy")
app = modal.App("build-image")
with app.run(), modal.enable_output():
    image.build(app)

# Use this object_id with `Image.from_id` to use the built image in a sandbox or function.
print(image.object_id)
```

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -880,6 +880,7 @@ class _Image(_Object, type_prefix="im"):
         with app.run(), modal.enable_output():
             image.build(app)
 
+        # Use this object_id with `Image.from_id` to use the built image in a sandbox or function.
         print(image.object_id)
         ```
         """
--- a/modal/image.py
+++ b/modal/image.py
@@ -868,6 +868,31 @@ class _Image(_Object, type_prefix="im"):
 
         return obj
 
+    async def build(self, app: "modal.app._App") -> "_Image":
+        """Eagerly build an image.
+
+        **Examples**
+
+        ```python
+        image = modal.Image.debian_slim().uv_pip_install("scipy", "numpy")
+
+        app = modal.App("build-image")
+        with app.run(), modal.enable_output():
+            image.build(app)
+
+        print(image.object_id)
+        ```
+        """
+        if app.app_id is None:
+            raise InvalidError("App has not been initialized yet. Run with `with app.run()`.")
+
+        app_id = app.app_id
+        app_client = app._client or await _Client.from_env()
+
+        resolver = Resolver(app_client, app_id=app_id)
+        await resolver.load(self)
+        return self
+
     def pip_install(
         self,
         *packages: Union[str, list[str]],  # A list of Python packages, eg. ["numpy", "matplotlib>=3.5.0"]
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -2267,3 +2267,14 @@ def test_raw_registry_image(servicer, client):
 def test_hydration_refusal():
     with pytest.raises(ExecutionError, match="Images cannot currently be hydrated on demand"):
         Image.debian_slim().hydrate()
+
+
+def test_build(servicer, client):
+    image = modal.Image.debian_slim().uv_pip_install("scipy", "numpy")
+
+    app = App()
+    with servicer.intercept() as ctx:
+        with app.run(client=client):
+            image.build(app)
+
+    assert len(ctx.get_requests("ImageGetOrCreate")) == 2


modal/image.py
```diff
@@ -868,6 +868,32 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
...
+        app = modal.App("build-image")
+        with app.run(), modal.enable_output():
+            image.build(app)
+
+        # Use this object_id with `Image.from_id` to use the built image in a sandbox or function.
+        print(image.object_id)
+        ```
+        """
+        if app.app_id is None:
+            raise InvalidError("App has not been initialized yet. Run with `with app.run()`.")
```
mwaskom: Regarding "Run with `with app.run()`, it would also be valid to use `App.lookup` and if this is oriented towards Sandbox usecases that's likely the more common pattern. Not sure we need to call out a specific suggestion though, although it's arguable a flaw that we don't have a single clear remediation to point to :/.

OTOH if you change the suggestion to "Run App with `with app.run()`" it will be a word-level palindrome and then I'd like It just for the luls 😁

modal/image.py
```diff
@@ -868,6 +868,32 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
 
         return obj
 
+    async def build(self, app: "modal.app._App") -> "_Image":
+        """Eagerly build an image.
```
mwaskom, line 872: Couple suggestions for the documentation:

First, can we make it clear somehow that you don't _normally_ need to do this when using Functions? i.e. we very much want to avoid people doing

```python
app = modal.App()

image = modal.Image.debian_slim()

with app.run():
    image.build(app)

@app.function()
def f():
    ...
```

Second, can we make it clear (again, somehow ... it's a little tricky) that "builds" do not actually always _build_ the Image? i.e. sometimes they just retrieve a cached version. I know Eric mentioned that the `docker build` prior art helps establish this pattern, otoh concepts in our API like `force_build=True` imply that there's a meaningful distinction between building and hydrating from the cache.

(Also: I wonder if this method should have a `force: bool` option 🤔).

modal/image.py
```diff
@@ -868,6 +868,32 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
...
+
+        ```python
+        image = modal.Image.debian_slim().uv_pip_install("scipy", "numpy")
+
+        app = modal.App("build-image")
+        with app.run(), modal.enable_output():
+            image.build(app)
+
+        # Use this object_id with `Image.from_id` to use the built image in a sandbox or function.
+        print(image.object_id)
```
mwaskom: I'm not sure this example is sufficiently clear w/r/t how you would do that.

mwaskom (commented):
One thing I don't super love here is that _nearly_ every other Image method is "recipe-oriented" (I think `.imports()` is the only other exception?). OTOH I'm not certain that `modal.build_image(image, app)` is that much better.

One thing I like about this is that we have a lot of internal uses of the pattern

```
app = modal.App()

@app.function(image=image)
def dummy():
    ...

with app.run():
    ...
```

As a way of building an Image on-demand and it will be nice to get rid of them!

modal/image.py
```diff
@@ -868,6 +868,32 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
...
+    async def build(self, app: "modal.app._App") -> "_Image":
+        """Eagerly build an image.
+
+        **Examples**
+
+        ```python
+        image = modal.Image.debian_slim().uv_pip_install("scipy", "numpy")
+
+        app = modal.App("build-image")
+        with app.run(), modal.enable_output():
```
mwaskom: Normally you need to do `modal.enable_output()` before you do `app.run()`

```python
with modal.enable_output(), app.run():
    ...
```

I guess with this way you'd see the Image build logs but not any of the other "App status" output?

modal/image.py
thomasjpfan: I'll add `App.lookup` in the error message as well.

For generic image building (independent of sandboxes) where you just want to build with and get an image id, I think an ephemeral app is a nicer (thus `app.run()`).

mwaskom: Yeah I agree that if you have some standalone `build_modal_image.py` script that makes sense.

In the Sandbox usecase there might be a pattern more like

```python
app = modal.App.lookup(...)

image = modal.Image.from_registry(...).build()

# Now we can start accepting requests or something else

sb = modal.Sandbox.create(app=app, image=image)

```

modal/image.py
thomasjpfan: I'm not sure if I like `force` here. One can already force with all "recipe-oriented" API. Do you see `build(force=True)` as a convenience for setting `force_build` on every layer?

mwaskom: Yeah I see it as a programmatic equivalent to using `MODAL_FORCE_BUILD=1 ...` but maybe that's enough. Force-rebuilding the entire Image is actually often regretful because it makes every other Image that uses the same base rebuild too. Now that I think about it a little more, ad hoc application of `MODAL_IGNORE_CACHE` is maybe more useful. But we can just drop it for now.

modal/image.py
```diff
@@ -884,7 +887,25 @@ async def build(self, app: "modal.app._App") -> "_Image":
...
+        For defining Modal functions, images are built automatically when deploying or running an App.
+        You do not need to built the image explicity:
+
+        ```python notest
+        app = modal.App()
+        image = modal.Image.debian_slim()
+
+        # No need to explicity build the image for defining a function.
+        # with app.run():
+        #     image.build(app)
```
mwaskom: This feels easily misread as giving them an example of the thing we _don't_ want them to do 😁

modal/image.py
```diff
@@ -868,6 +868,67 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
 
         return obj
 
+    async def build(self, app: "modal.app._App") -> "_Image":
+        """Eagerly build an image.
+
+        If your image was previously built, then this method will be a no-op and
```
mwaskom, line 874: Worried this is a little confusing since if you have an unhydrated Image object, we'll do an RPC and it's not _technically_  a  "no-op" it just doesn't rebuild.

modal/image.py
```diff
@@ -868,6 +868,67 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
...
+        your built image is returned.
```
mwaskom, line 875: s/built/cached/? Not sure

modal/image.py
```diff
@@ -868,6 +868,67 @@ async def _load(self: _Image, resolver: Resolver, existing_object_id: Optional[s
...
+            image.build(app)
+
+        # Save the image id
+        my_image_id = image.object_id
+
+        # Reference the image with the id or uses it another context.
+        built_image = modal.Image.from_id(my_image_id)
+        ```
+
+        Alternatively, you can re-built a image and use it in a sandbox.
```
mwaskom, line 893: Typo: re-built

mwaskom: (Also I'm not sure what "re-build" would mean here. Do you mean pre-build?)

modal/image.py
mwaskom: Thinking a little more I kinda like having `ignore_cache: bool` here, though not critical. Could be nice in some Sandbox usecases where you want the process to always use some "fresh" version of some external image. Thought might require some work to wire it up since we don't currently have "ignore cache" in the API anywhere.

PR #3467, opened 2025-08-14
Make deprecation warning assertion stricter to avoid flakes

mwaskom:
Fixes CLI-573 (I hope)

### DIFF
--- a/test/cli_test.py
+++ b/test/cli_test.py
@@ -236,7 +236,7 @@ def test_run_warns_without_module_flag(
 ):
     monkeypatch.chdir(supports_dir)
     _run(["run", "-m", f"{app_module}::foo"])
-    deprecation_warnings = [w for w in recwarn if issubclass(w.category, DeprecationError)]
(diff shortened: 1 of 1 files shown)


test/cli_test.py
```diff
@@ -236,7 +236,8 @@ def test_run_warns_without_module_flag(
 ):
     monkeypatch.chdir(supports_dir)
     _run(["run", "-m", f"{app_module}::foo"])
-    assert not len(recwarn)
+    deprecation_warnings = [w for w in recwarn if issubclass(w.category, DeprecationError)]
+    assert not deprecation_warnings
```
thomasjpfan: It's hard to tell if this fixes the flakiness without seeing the warning. Can we change the assert to the following to see what the warning is?

```python
assert not [w.message for w in recwarn]
```

mwaskom: I think it make sense on first principles to restrict this assertion to the type that we're "not expecting". But I agree that having a failure message provide more context about what warning we caught makes sense! Kind of annoying that pytest doesn't do that by default.

thomasjpfan (approved):

PR #3469, opened 2025-08-15
volume read_file: add block-level download retries

gimaker:
This adds block level retries for downloads - lack thereof seems like an oversight.

We think we might be getting some intermittent 400s (see FS-270), so we indiscriminately retry everything.

### DIFF
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -696,9 +696,6 @@ class _Volume(_Object, type_prefix="vo"):
         @retry(n_attempts=5, base_delay=0.1, timeout=None)
         async def read_block(block_url: str) -> bytes:
             async with ClientSessionRegistry.get_session().get(block_url) as get_response:
-                if get_response.status in [503, 429]:
-                    logger.warning("Received SlowDown signal, sleeping for 1 second before retrying.")
-                    await asyncio.sleep(1)
                 return await get_response.content.read()
 
         async def iter_urls() -> AsyncGenerator[str]:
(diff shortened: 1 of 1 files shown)


dflemstr (approved):

dflemstr:
I guess we should also gracefully handle when we do reach a timeout... do we want to raise an exception then or communicate the failure in some more user-friendly way? We don't do that in other places like this, so maybe it's fine...

gimaker:
> I guess we should also gracefully handle when we do reach a timeout... do we want to raise an exception then or communicate the failure in some more user-friendly way? We don't do that in other places like this, so maybe it's fine...

Yeah perhaps - the output isn't great right now (example below). But this doesn't change the error reporting in any way so seems like a separate problem.

```
$ modal volume get staffan-v2 sharegpt.jsonl --force
⠧ Downloading file(s) to local...
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [Errno 104] Connection reset by peer                                                                                                                                                 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

modal/volume.py
```diff
@@ -693,8 +693,12 @@ async def read_file(self, path: str) -> AsyncIterator[bytes]:
         except modal.exception.NotFoundError as exc:
             raise FileNotFoundError(exc.args[0])
 
+        @retry(n_attempts=5, base_delay=0.1, timeout=None)
         async def read_block(block_url: str) -> bytes:
             async with ClientSessionRegistry.get_session().get(block_url) as get_response:
+                if get_response.status in [503, 429]:
+                    logger.warning("Received SlowDown signal, sleeping for 1 second before retrying.")
```
freider: Is this expected to happen during normal operation? Not very actionable for users

gimaker: I'm just cargo-culting this tbh:

https://github.com/modal-labs/modal-client/blob/e0757d94cef8d3c40e54302eda1450cd465471ff/modal/_utils/blob_utils.py#L87-L89

Happy to remove both the `warning` and the extra `sleep`.

PR #3472, opened 2025-08-15
add an idle_timeout param for Sandbox.create()

rolyatmax:
## Describe your changes

Adds an `idle_timeout` param to `Sandbox.create()`. See design doc for this feature [here](https://www.notion.so/modal-com/Sandbox-Idle-Timeouts-2491e7f1694980208ccde60da0e8622e?source=copy_link).

---

## Compatibility checklist

---

</details>

## Changelog

- Added an `idle_timeout` param to `Sandbox.create()` which, when provided, will have the sandbox terminate after `idle_timeout` seconds of idleness.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -259,7 +259,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         image: Optional[_Image] = None,  # The image to run as the container for the sandbox.
         secrets: Sequence[_Secret] = (),  # Environment variables to inject into the sandbox.
         network_file_systems: dict[Union[str, os.PathLike], _NetworkFileSystem] = {},
-        timeout: int = 300,  # Maximum execution time of the sandbox in seconds.
+        timeout: int = 300,  # Maximum lifetime of the sandbox in seconds.
         # The amount of time in seconds that a sandbox can be idle before being terminated.
         idle_timeout: Optional[int] = None,
         workdir: Optional[str] = None,  # Working directory of the sandbox.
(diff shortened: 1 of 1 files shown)


modal/sandbox.py
```diff
@@ -234,6 +235,8 @@ async def _load(self: _Sandbox, resolver: Resolver, _existing_object_id: Optiona
                 name=name,
                 experimental_options=experimental_options,
             )
+            if idle_timeout is not None:
```
mwaskom: J/w but is there a reason you're not passing this directly to `api_pb2.Sandbox`?

modal/sandbox.py
```diff
@@ -258,6 +261,8 @@ async def create(
         secrets: Sequence[_Secret] = (),  # Environment variables to inject into the sandbox.
         network_file_systems: dict[Union[str, os.PathLike], _NetworkFileSystem] = {},
         timeout: int = 300,  # Maximum execution time of the sandbox in seconds.
```
mwaskom: Not in your diff but "Maximum execution time" is maybe ambiguous if we're going to be distinguishing between "doing something" and "idle". Might want to rephrase the comment as like "Maximum lifetime"?

modal/sandbox.py
rolyatmax: Hmmm, not really, I don't think. Maybe I saw this somewhere else in the codebase? I'll update to just pass this directly to `api_pb2.Sandbox()`

modal/sandbox.py
rolyatmax: Ah yes - that's better - will update!

mwaskom (approved):

PR #3474, opened 2025-08-18
Cbor2 serialization for function calling [SDK-584]

freider:
Adds opt-in support for cbor as a payload serialization format instead of pickle.

* If the input format is cbor, the corresponding output will also be encoded as such by the runtime
* Requires `cbor2` to be installed in the container environment to work. This happens automatically for image builder 2025.06 and newer (if users absolutely need to use older image builder they can manually add cbor2 as a dependency to their image to make this work)
* The *python* SDK only uses cbor for input serialization when the MODAL_PAYLOAD_FORMAT=cbor env var is set
* libmodal beta is intended to remove the pickle serde for function calling altogether, relying on remote functions being deployed with cbor support to be called
* Functions can be optionally marked as restricted to cbor output (`_experimental_restrict_output=True` for now). This makes output cbor regardless if the input is pickle (good for security in the receiver, or if you know the consumer will be on libmodal even if the input creator is in Python using .spawn)

## TODO
- [x] Respect `supported_input_formats` in the client by *sending* cbor if pickle isn't supported by the Function definition. There is currently no syntactic way to restrict the *input* to a function when deploying it, but if we were to add this in the future it's good if the client respects it already for forwards-compatibility.
- [x] Add `_experimental_restrict_output` to either the `@app.cls` or `@modal.method` decorators + add test that it's respected
- [x] ~Better error message if encode can't encode a datatype (linking to a docs page about how serialization works w/ Modal and libmodal etc.?)~ Good enough for now until we have a docs page for this
- [x] ~Error message if something can't be decoded?~ Skipping this for now. When accessing the undecodable data you get a CBORTag type on the python side which can be inspected or passed back as output. Good enough for now and we'll have to add support as we go if any unforeseen input types pop up

## Not in this patch (yet?):
* Tracebacks and python representations of exceptions are only provided for pickle calls - custom serialization of tracebacks could be a future improvement. libmodal only looks at the `exception` string repr of errors anyways at the moment. **Will not do for this initial version - opting to leave these fields empty in the response payload for cbor inputs.**
* We don't enforce worker-side that outputs for a function conform to the supported_output_format. For untrusted functions it would be good to enforce this in order to not allow "unpickle attacks" by the functions. 

---

## Compatibility checklist

---

</details>

## Changelog

* [Internal] Adds support for calling into deployed functions using a new cbor based serialization format used by beta versions of libmodal-ts and libmodal-go

### DIFF
--- a/modal/_runtime/container_io_manager.py
+++ b/modal/_runtime/container_io_manager.py
@@ -292,7 +292,7 @@ class IOContext:
     def _determine_output_format(self, input_format: "api_pb2.DataFormat.ValueType") -> "api_pb2.DataFormat.ValueType":
         if input_format in self.finalized_function.supported_output_formats:
             return input_format
-        else:
+        elif self.finalized_function.supported_output_formats:
             # This branch would normally be hit when calling a restricted_output function with Pickle input
             # but we enforce cbor output at function definition level. In the future we might send the intended
             # output format along with the input to make this disitinction in the calling client instead
@@ -301,6 +301,11 @@ class IOContext:
                 f" using formats {self.finalized_function.supported_output_formats}"
             )
             return self.finalized_function.supported_output_formats[0]
+        else:
+            # This should never happen since self.finalized_function.supported_output_formats should be
+            # populated with defaults in case it's empty, log a warning
+            logger.warning(f"Got an input with format {input_format}, but the function has no defined output formats")
+            return api_pb2.DATA_FORMAT_PICKLE
 
     async def output_items_exception(
         self, started_at: float, task_id: str, exc: BaseException


modal/_serialization.py
```diff
@@ -6,6 +6,14 @@
 from inspect import Parameter
 from typing import Any
 
+from modal._traceback import extract_traceback
+from modal.config import config
+
+try:
+    import cbor2  # type: ignore
+except Exception:  # pragma: no cover - optional dependency
```
ehdr:
```suggestion
except ImportError:  # pragma: no cover - optional dependency
```
?

modal/_runtime/container_io_manager.py
```diff
@@ -191,25 +203,206 @@ def _args_and_kwargs(self) -> tuple[tuple[Any, ...], dict[str, list[Any]]]:
...
+        ]
+
+    async def output_items(self, started_at: float, data: list[Any]) -> list[api_pb2.FunctionPutOutputsItem]:
+        output_created_at = time.time()
+
+        # Process all items concurrently and create output items directly
+        async def package_output(
+            item: Any, function_input: api_pb2.FunctionInput, input_id: str, retry_count: int
+        ) -> api_pb2.FunctionPutOutputsItem:
+            assert function_input.data_format  # TODO: Remove this - just for debugging
```
ehdr: Debugging leftovers?

test/container_test.py
```diff
@@ -1907,8 +1929,10 @@ def test_cancellation_aborts_current_input_on_match(
         api_pb2.ContainerHeartbeatResponse(cancel_input_event=api_pb2.CancelInputEvent(input_ids=cancelled_input_ids))
     )
     stdout, stderr = container_process.communicate()
-    assert stderr.decode().count("Successfully canceled input") == live_cancellations
-    assert "Traceback" not in stderr.decode()
+    stderr_str = stderr.decode()
+    print(stderr_str)
```
ehdr: Debug leftovers?

modal/_serialization.py
```diff
@@ -355,6 +369,10 @@ def serialize_data_format(obj: Any, data_format: int) -> bytes:
     elif data_format == api_pb2.DATA_FORMAT_GENERATOR_DONE:
         assert isinstance(obj, api_pb2.GeneratorDone)
         return obj.SerializeToString(deterministic=True)
+    elif data_format == api_pb2.DATA_FORMAT_CBOR:
+        if cbor2 is None:
+            raise InvalidError("CBOR support requires the 'cbor2' package to be installed.")
+        return cbor2.dumps(obj)
```
ehdr: Maybe we should catch `CBOREncodeTypeError`s here and give more context to the error message?

ehdr (commented):
Look good to me! Just minor comments on implementation, though also pending the larger discussion we started yesterday I guess.

modal/_runtime/container_io_manager.py
freider: Yup

modal/_serialization.py
freider: Thanks, fixed

test/container_test.py
freider: yp

modal/_serialization.py
freider: Good idea, I'll add a test for that too

test/container_test.py
```diff
@@ -2530,6 +2559,111 @@ def test_deserialization_error_returns_exception(servicer, client):
     assert int(deserialize(ret.items[1].result.data, ret.client)) == 4
 
 
+@skip_github_non_linux
+def test_cbor_input_payload_simple_function(servicer):
```
freider: TODO @freider : Add test for cbor generators

test/container_test.py
```diff
@@ -2530,6 +2559,111 @@ def test_deserialization_error_returns_exception(servicer, client):
...
+    assert len(ret.items) == 1
+    item = ret.items[0]
+    assert item.result.status == api_pb2.GenericResult.GENERIC_STATUS_SUCCESS
+    assert item.data_format == api_pb2.DATA_FORMAT_CBOR  # expect cbor output
+    value = deserialize_data_format(item.result.data, item.data_format, ret.client)
+    assert int(value) == 4
+
+
+@skip_github_non_linux
+def test_cbor_incompatible_output(servicer):
```
freider, line 2777: TODO @freider add test for general cbor general exceptions

modal/_runtime/user_code_imports.py
```diff
@@ -108,14 +108,18 @@ def get_finalized_functions(
...
         if not webhook_config.type:
             # for non-webhooks, the runnable is straight forward:
             return {
                 "": FinalizedFunction(
                     callable=self._user_defined_callable,
                     is_async=is_async,
                     is_generator=is_generator,
-                    data_format=api_pb2.DATA_FORMAT_PICKLE,
+                    supported_output_formats=fun_def.supported_output_formats
+                    # FIXME (elias): the following `or [api_pb2.DATA_FORMAT_PICKLE, api_pb2.DATA_FORMAT_CBOR]` is only
```
mwaskom, line 120: What's the plan for the fix?

modal/_utils/function_utils.py
```diff
@@ -549,27 +555,35 @@ async def _create_input(
...
+
     if idx is None:
         idx = 0
-    if method_name is None:
-        method_name = ""  # proto compatible
 
-    args_serialized = serialize((args, kwargs))
+    data_format = get_preferred_payload_format()
+    if not function._metadata:
+        raise ExecutionError("Attempted to call function that has not been hydrated with metadata")
```
mwaskom, line 573: Did we enhance this error message with the Function name or other helpful information in the other places we raise it?

modal/_utils/function_utils.py
```diff
@@ -549,27 +555,35 @@ async def _create_input(
...
+
+    supported_input_formats = function._metadata.supported_input_formats or [api_pb2.DATA_FORMAT_PICKLE]
+    if data_format not in supported_input_formats:
+        data_format = supported_input_formats[0]
```
mwaskom, line 577: What's the rationale behind silently ignoring non-compatible `supported_input_formats`?

modal/_partial_function.py
```diff
@@ -287,6 +288,7 @@ def _method(
     # Set this to True if it's a non-generator function returning
     # a [sync/async] generator object
     is_generator: Optional[bool] = None,
+    _experimental_restrict_output: bool = False,  # Prevent pickle as serializion (client 1.2+ and libmodal beta+)
```
mwaskom: Could we maybe just keep this at the Cls level?

mwaskom: Aside from my general woes regarding multi-method classes, if you have an untrusted Function don't you need the whole container to be rejecting malicious injection of Pickled outputs?

modal/_serialization.py
```diff
@@ -346,6 +354,12 @@ def unflatten_headers(obj):
         return None
 
 
+def get_preferred_payload_format() -> "api_pb2.DataFormat.ValueType":
+    payload_format = (config.get("payload_format") or "pickle").lower()
+    data_format = api_pb2.DATA_FORMAT_CBOR if payload_format == "cbor" else api_pb2.DATA_FORMAT_PICKLE
```
mwaskom, line 359: Safe because we use `_check_value` in the config parser.

modal/_serialization.py
```diff
@@ -579,3 +612,26 @@ def get_callable_schema(
         arguments=arguments,
         return_type=return_type_proto,
     )
+
+
```
mwaskom, line 616: Not exactly within the scope of this PR but I don't think I understand all the rules for how we pass around (serialized) exceptions or what we're hoping to accomplish by doing so.

modal/_partial_function.py
mwaskom: I guess part of the answer is that because you can have a web endpoint decorator on an individual method, we need to be able to track the data format accepted / produced by each method.

I might still advocate for scoping `restrict_output` at the Cls level (and have it inherited by all methods). Though I'm not sure if it makes sense to have an untrusted function with a web endpoint method 🤔

test/function_test.py
```diff
@@ -1539,6 +1551,171 @@ def webbie(query_param: int): ...
     assert schema.schema_type == api_pb2.FunctionSchema.FUNCTION_SCHEMA_UNSPECIFIED
 
 
+@pytest.mark.usefixtures("set_env_client")
+def test_cbor_output_complex_data_types(client, servicer):
+    """Test that a received cbor payload is decoded as such, even if the submitted input is pickle"""
+    app = App("app")
+
+    @app.function(serialized=True)
```
mwaskom, line 1760: Serialized how 🙃

test/function_test.py
```diff
@@ -1539,6 +1551,171 @@ def webbie(query_param: int): ...
...
+            "tags": ("test", "cbor", "serialization"),
+            "config": {"enabled": True, "count": 42},
+        },
+    }
+    expected_decoded_output = {
+        "processed": True,
+        "data": [1, 2.5, "string", {"nested": True, "values": [None, False, True]}],
+        "metadata": {
+            "timestamp": 1234567890,
+            "tags": ["test", "cbor", "serialization"],  # same but tuple converted to list since cbor doesnt distinguish
```
mwaskom, line 1781: Down the road, but is this something we'll need to think about if we want to support exporting type stubs for Modal functions?

modal/_container_entrypoint.py
```diff
@@ -204,33 +200,24 @@ async def run_input_async(io_context: IOContext) -> None:
...
                     current_attempt_token,
-                    io_context.finalized_function.data_format,
+                    io_context._generator_output_format(),
                     generator_queue,
                 ):
                     item_count = 0
-                    async for value in res:
-                        await container_io_manager._queue_put.aio(generator_queue, value)
-                        item_count += 1
+                    async with aclosing(io_context.call_generator_async()) as gen:
```
mwaskom, line 207: Couldn't figure out if this is directly related or a drive-by improvement.

modal/_runtime/container_io_manager.py
```diff
@@ -191,25 +200,220 @@ def _args_and_kwargs(self) -> tuple[tuple[Any, ...], dict[str, list[Any]]]:
...
         args, kwargs = self._args_and_kwargs()
-        res = self.finalized_function.callable(*args, **kwargs)
+        expected_value_or_values = self.finalized_function.callable(*args, **kwargs)
+        if (
+            inspect.iscoroutine(expected_value_or_values)
+            or inspect.isgenerator(expected_value_or_values)
+            or inspect.isasyncgen(expected_value_or_values)
+        ):
+            raise InvalidError(
+                f"Sync (non-generator) function return value of type {type(expected_value_or_values)}."
```
mwaskom, line 232: I found this sentence hard to parse.

mwaskom: Maybe `s/return/returned`, though even then I'm not quite sure what the parenthetical is communicating.

modal/_runtime/container_io_manager.py
```diff
@@ -191,25 +200,220 @@ def _args_and_kwargs(self) -> tuple[tuple[Any, ...], dict[str, list[Any]]]:
...
+            yield result
+        logger.debug(f"Finished generator input {self.input_ids}")
+
+    async def call_generator_async(self) -> AsyncGenerator[Any, None]:
+        assert not self._is_batched
+        logger.debug(f"Starting generator input {self.input_ids}")
+        args, kwargs = self._args_and_kwargs()
+        expected_async_gen = self.finalized_function.callable(*args, **kwargs)
+        if not inspect.isasyncgen(expected_async_gen):
+            raise InvalidError(f"Async generator function returned value of type {type(expected_async_gen)}")
```
mwaskom, line 273: Stray comment but I wonder if we're overloading `InvalidError` here and more specific error type(s) might be useful.

modal/_container_entrypoint.py
freider: It's *mostly* a drive-by improvement, but I'm pretty sure there was some structural reason I needed it in this case. There were a bunch of generator + batching invariants to keep track of and this made it less easy to make mistakes there

modal/_partial_function.py
freider: Yeah the web endpoints being on individual methods was the reason I put this on individual methods too. I agree we *could* put the flag on the class decorator instead (I had it like that initially), but as you say it doesn't have a meaning for web endpoint methods and I felt this better reflects the underlying data model (as complex as it is)

The enforcing part I think would be doable in the worker regardless if the flag is on the method definition or the service class itself (the worker has the full definition and knows which inputs belong to each method)

modal/_runtime/user_code_imports.py
freider: I have a followup PR that refactors pretty much all container_test tests to auto-infer the api_pb2.Function definitions based on the actual app/function definitions in code instead of having the test "duplicate" all this logic. With that patch, these fields are guaranteed to be set so we wouldn't need the fallback here (could even assert that they are non empty)

modal/_serialization.py
freider: I think so? Or I mean, it would always be "safe" by falling back to PICKLE if it's not set to ~"cbor"

modal/_serialization.py
freider: We reraise the deserialized exception in the caller, people can try: except with the exception type as raised by the called function. But naturally that only applies to python -> python calls (and only when the exception *type* actually exists in both the parent and child call)

modal/_utils/function_utils.py
freider: Not that I can find. Would also be tricky since we currently set the name *from* the hydration metadata... You could imagine setting it separately using the from_name and/or from_local function information though, in which case that would be doable

modal/_utils/function_utils.py
freider: At this point it's mostly a hypothetical since there is nothing that restricts the supported *input* formats to functions (apart from the function being old and assumed to only support pickle if the value is empty).

This was mostly put in place to be forwards compatible with an eventual future where a function might be deployed to only support cbor *inputs*, in which case we don't want to preclude an old (i.e. current) client from calling into that function (by switching to cbor encoded inputs)

test/function_test.py
freider: cbor-serialized functions? 🤡

test/function_test.py
freider: You mean the down/up-grading of tuples into lists?

iirc, the "schema" we store for functions today only support `list` type annotations, not tuples.
If we were to add tuple support in annotations, which feels reasonable, that wouldn't preclude us from using a cbor list as the cbor encoding of that tuple and then do client side casting of it into the relevant tuple type in each language.

One thing we *could* do today is add a *decoder* support for a custom cbor tag for tuples that enforces this typing at the data transfer layer (making the result a tuple in python automatically, even for untyped functions).
This would make at least the python usage of cbor a data transfer protocol more intuitive, and let it work even for untyped dicts/queues, at the expense of having to implement support for a custom "tuple" tag in every language.
Not sure what I prefer tbh...

freider:
> I think my only design-level question is about whether we should / need to support `restrict_output` at the level of individual methods. I understand why we need to track serialization format at the method level to support web interfaces on methods. But I'd prefer we not add new method-level functionality in general, and I think that we may actually need to make the restriction container level to avoid the pickle vulnerabilities around untrusted functions?

I don't think we need the restriction to be on a container level since the worker should be able to inspect the method definitions for each output (and for consistency w/ web endpoints I'd like to keep the values on the method definition), but I agree it would be nice to not have the option on method level from an SDK perspective if we're moving away from multiple-method classes eventually 🤔

modal/_utils/function_utils.py
mwaskom: Makes sense. Seen this mentioned before by users as a pain point, FWIW.

modal/_serialization.py
mwaskom: Right that's what I mean sorry. Originally thought this was risky if you e.g. typo `cbor` but it's 👍

modal/_runtime/container_io_manager.py
freider: Haha yeah, this was actually just copied from the old validation in container_entrypoint, so I'm also not sure. Not even sure if this branch can ever get triggered or if we have other validation before it that prevents this

modal/_runtime/container_io_manager.py
freider: Yeah. I think this is more of a ValueError, but this just moves existing logic from the container entrypoint into the IOContext (we might consider moving IOContext into its own module and call it InputBatch or something...)

freider:
@prbot approve

freider:
@prbot approve @mwaskom

PR #3477, opened 2025-08-19
add GENERIC_STATUS_IDLE_TIMEOUT variant for GenericStatus

rolyatmax:
## Describe your changes

We want to differentiate between sandboxes that hit their lifetime limit (i.e. `timeout` param), those that were explicitly terminated by the user, and those that were terminated due to hitting an idle timeout. This new `GenericStatus` variant makes that possible.

---

## Compatibility checklist

There are a few places on the server where we may need to update some code for this new variant. There's a PR for those changes [here](https://github.com/modal-labs/modal/pull/25917).

Note that running servers should not encounter this new variant until we actually release the `idle_timeout` client, but `cargo build`s may fail until we add support for the new variant.
---

</details>


mwaskom:
SGTM!

mwaskom (approved):

PR #3445, opened 2025-08-08
Add `.num_inputs()` to the FunctionCall API

CyrilSharma:
Stacked PRs:
 * __->__#3445

--- --- ---

PR #3479, opened 2025-08-19
Fix FunctionCall.from_id not setting client object

devennavani:
## Describe your changes

https://github.com/modal-labs/modal-client/pull/3445 changed the logic of `FunctionCall.from_id` to return an un-hydrated `FunctionCall` object, because there is now server-side object metadata (`num_inputs`) that can be set on the `FunctionCall` handle. This `num_inputs` information is only used by the newly introduced `FunctionCall.iter` (and `FunctionCall.num_inputs`). Previously, `FunctionCall.from_id` would use `_FunctionCall._new_hydrated` so it would return a hydrated `FunctionCall` object with `_client` set. Methods like `.get()`, `.get_call_graph()`, etc. depend on `_client` being set. With the change, we forgot to manually set `_client` on the `FunctionCall` in `from_id`.

Discussed this fix with @mwaskom. We are getting this bug fix in quickly for now but want to discuss further and hopefully move towards a cleaner state where an Object is only either hydrated (ID+server side metadata present) or not.

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

Fixes a bug introduced in `1.1.2` which causes invocation of `FunctionCall.get`, `FunctionCall.get_call_graph`, `FunctionCall.cancel`, and `FunctionCall.gather` to fail when the `FunctionCall` object is retrieved via `FunctionCall.from_id`.

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -2015,6 +2015,7 @@ class _FunctionCall(typing.Generic[ReturnType], _Object, type_prefix="fc"):
         fc: _FunctionCall[Any] = _FunctionCall._from_loader(_load, rep, hydrate_lazily=True)
         # We already know the object ID, so we can set it directly
         fc._object_id = function_call_id
+        fc._client = client
         return fc
 
     @staticmethod
--- a/test/function_test.py
+++ b/test/function_test.py
@@ -781,18 +781,23 @@ def test_from_id(client, servicer):
 
     @app.function(serialized=True)
     def foo():
-        pass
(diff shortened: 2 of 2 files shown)


mwaskom (approved):
Thanks for the fix

> causes invocation of FunctionCall.get, FunctionCall.get_call_graph, FunctionCall.cancel, and FunctionCall.gather

Do we need to test `.from_id().get_call_graph()` too?

PR #3485, opened 2025-08-20
Add startup_timeout to cls and functions

thomasjpfan:
## Describe your changes

- Towards https://linear.app/modal-labs/issue/SDK-417/support-a-distinct-startup-timeout

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- `app.function` and `app.cls` now support `startup_timeout` which configures the timeout for the startup phase


thomasjpfan:
> Why not @app.function too?

Are you thinking of supporting startup timeouts for the web servers? i.e `wsgi_app`, `asgi_app` and friends.

mwaskom:
I do think we need an answer to what to do when there are conflicts between those parameterizations. There's a long Slack thread about this that you might have seen.

But more broadly, Functions have a startup period too and the `timeout` will be applied to it. It's _generally_ a lot faster than if you have a Cls with a `@modal.enter()` method that loads a lot of data. But we've seen users have Functions unable to startup in the past because they set a really low timeout but all of their imports take a few seconds to run.

modal/_functions.py
```diff
@@ -966,6 +967,7 @@ async def _load(self: _Function, resolver: Resolver, existing_object_id: Optiona
                     proxy_id=(proxy.object_id if proxy else None),
                     retry_policy=retry_policy,
                     timeout_secs=timeout_secs or 0,
+                    startup_timeout_secs=startup_timeout or timeout_secs,
```
mwaskom, line 970: We should think about how to document this — ideally they'd be independent with their own defaults but for backwards compatibility we need to fallback to `timeout` for startup.

modal/app.py
```diff
@@ -870,6 +872,7 @@ def cls(
         proxy: Optional[_Proxy] = None,  # Reference to a Modal Proxy to use in front of this function.
         retries: Optional[Union[int, Retries]] = None,  # Number of times to retry each input in case of failure.
         timeout: int = 300,  # Maximum execution time in seconds; applies independently to startup and each input.
```
mwaskom: Should update the parameter explanation here.

modal/app.py
```diff
@@ -870,6 +872,7 @@ def cls(
...
+        startup_timeout: Optional[int] = None,  # Maximum startup execution time in seconds.
```
mwaskom: Is "startup execution time" a misnomer? i.e. it feels like "startup" and "execution" are two different phases.

mwaskom:
would be good to queue up an update to this section of the guide too: https://modal.com/docs/guide/timeouts#container-startup-timeout

PR #3488, opened 2025-08-21
Adds cbor2 dependency to PREVIEW image builder

freider:
I intend to manually publish a new version of the preview mounts once this is merged.

If subsequent testing shows it works, the intention is to patch the 2025.06 image builder to include cbor2 as well, since it's a relatively safe thing to do and would unlock cbor2 functionality on new clients without having to wait for/switch to a future image builder version.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/builder/PREVIEW.txt
+++ b/modal/builder/PREVIEW.txt
@@ -15,3 +15,4 @@ propcache==0.3.1
 protobuf==6.31.1
 typing_extensions==4.13.2
 yarl==1.20.0
+cbor2==5.7.0


modal/builder/PREVIEW.txt
```diff
@@ -15,3 +15,4 @@ propcache==0.3.1
 protobuf==6.31.1
 typing_extensions==4.13.2
 yarl==1.20.0
+cbor2==5.7.0
```
mwaskom: Could we keep the list alphabetical 🙏

freider: good point

PR #3490, opened 2025-08-21
[Old] FLASH-11 get_flash_url using RPC

emilyhanyf:
Closes FLASH-11

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -904,14 +904,13 @@ More information on class parameterization can be found here: https://modal.com/
 
     @synchronizer.create_blocking
     async def get_flash_url(self) -> Optional[str]:
-        """Get the stable Flash service URL for this class.
-
-        Returns a load-balanced Flash URL that route requests to all healthy
-        containers running this class. Returns None if Flash is not enabled for this class.
+        """
+        Get the stable Flash service URL for this class.
+        Returns None if this class is not a Flash service.
 
         Example:
-            Model = modal.Cls.from_name("my-app", "Model")
-            flash_url = Model.get_flash_url()
+        Model = modal.Cls.from_name("my-app", "Model")
+        flash_url = Model.get_flash_url()
         """
         await self.hydrate()
 
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -12,7 +12,7 @@ from grpclib import GRPCError, Status
 from modal_proto import api_pb2
 
 from ._functions import _Function, _parse_retries
-from ._object import _Object
+from ._object import _Object, live_method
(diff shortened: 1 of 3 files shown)


modal/cls.py
```diff
@@ -902,6 +902,28 @@ def __getattr__(self, k):
     def _is_local(self) -> bool:
         return self._user_cls is not None
 
+    @live_method
+    async def get_flash_url(self) -> Optional[str]:
```
thomasjpfan, line 906: This will be the first mention of `flash` in `cls.py` while flash is experimental.

@mwaskom Also, does it make sense to put the flash url in somewhere in `ClassGetResponse` as metadata and populated when it gets hydrated? Similar to `web_url`:

https://github.com/modal-labs/modal-client/blob/f38c82d712622fa9fc9bb16846f2f12e4bdc312f/modal/_functions.py#L1543-L1546

emilyhanyf: I'm working on a PR that adds flash_url to `FunctionHandleMetadata` (instead of creating a new RPC). Lmk if you think that makes more sense!

thomasjpfan: `FunctionHandleMetadata` looks okay to me. A few open questions:

1. Should we include flash into the public API for `cls`? Could we mark this method as experimental, so we are allowed to change things as flash evolves?
3. There is some inconsistency because functions have a `get_web_url` and now `cls` gets a `get_flash_url`. This makes sense today because `flash_forward` only really works with `cls`. Are we okay with this inconsistency?

---

Should it be plural, i.e. `flash_web_urls` since `fetch_flash_associated_urls` (in the `mono-repo`) can return multiple values?

mwaskom: Kneejerk reaction is that we don't need it in ClassGetResponse — the server-side Class object is not really used for anything.

mwaskom: However I do think it's kind of weird to have this as a public method on Cls when Flash stuff is otherwise segregated into an `experimental` namespace (i think?)

thomasjpfan: > However I do think it's kind of weird to have this as a public method on Cls when Flash stuff 

I'm okay with adding it to the object and signal that is it experimental. Options are:

1. `_get_flash_url`: Private method
2. `_experimental_get_flash_url`: Private method
3. `experimental_get_flash_url`: "namespaced" with `experimental`.

gongy: maybe #2 with plural? we aren't quite sure where this is going yet

thomasjpfan: Yup, I'm okay with 2 with plural.

PR #3493, opened 2025-08-21
clearer docs on how to look up a named sandbox

rolyatmax:
Updated the `Sandbox.from_name()` docstring to clarify that it only works with _deployed_ apps.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -479,7 +479,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         environment_name: Optional[str] = None,
         client: Optional[_Client] = None,
     ) -> "_Sandbox":
-        """Get a running Sandbox by name from the given app. Note that the app must be a deployed app.
+        """Get a running Sandbox by name from a deployed App.
 
         Raises a modal.exception.NotFoundError if no running sandbox is found with the given name.
         A Sandbox's name is the `name` argument passed to `Sandbox.create`.


modal/sandbox.py
```diff
@@ -479,7 +479,7 @@ async def from_name(
         environment_name: Optional[str] = None,
         client: Optional[_Client] = None,
     ) -> "_Sandbox":
-        """Get a running Sandbox by name from the given app.
+        """Get a running Sandbox by name from the given app. Note that the app must be a deployed app.
```
mwaskom: To be a little annoying, I think the first line of a docstring should always be a simple statement about what the function / class does and nuance or additional comments should go in the "body". Not a huge deal though. (the Function.from_name docstring summary does say `Reference a Function from a deployed App by its name.` though I think we probably also would benefit from explaining more obviously the implication that it _won't work_ for ephemeral Apps...

rolyatmax: That's fair. What do you think about this? https://github.com/modal-labs/modal-client/pull/3493/commits/919ed040a55661c8259b336179fc2a7d2fffd26a

mwaskom: Yeah that's great and matched the analagous methods on Function / Cls.

FWIW I wouldn't be opposed to belaboring the point in an extended description — perhaps pointing people towards development environments? Just like to keep the docstring summaries short and to the point. Thanks for bearing with me.

PR #3495, opened 2025-08-21
Add is_managed to EnvironmentListItem

RohanArepally:
We'll update `EnvironmentList` so that web calls it as well for the environment picker on the /settings page: https://github.com/modal-labs/modal/blob/77bc45ed4dfaa7391edfd34d1764ce34a40c582a/server/modal_server/objects/environment.py#L156

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1236,6 +1236,7 @@ message EnvironmentListItem {
   double created_at = 3;
   bool default = 4;
   bool is_managed = 5;
+  string environment_id = 6;
 }
 
 message EnvironmentListResponse {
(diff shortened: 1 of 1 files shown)


mwaskom (approved):

modal_proto/api.proto
```diff
@@ -1235,6 +1235,8 @@ message EnvironmentListItem {
   string webhook_suffix = 2;
   double created_at = 3;
   bool default = 4;
+  bool is_managed = 5;
+  string environment_id = 6;
```
gleb-modal, line 1239: Crazy we weren't passing this before.

PR #3496, opened 2025-08-21
Move `retry_transient_errors` into the stub generator

thomasjpfan:
## Describe your changes

This way we do not need to remember to call `retry_transient_errors` on the unary-unary rpcs. 

Defining a `_with_retires` was the simplest way I saw for getting types to work nicely.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -271,6 +271,40 @@ async def retry_transient_errors(
             delay = min(delay * delay_factor, max_delay)
 
 
+def _with_retries(fn: "modal.client.UnaryUnaryWrapper[RequestType, ResponseType]"):
+    """Wrap a UnaryUnaryWrapper with retry_transient_errors."""
+
+    async def wrapped(
+        req: RequestType,
+        base_delay: float = 0.1,
+        max_delay: float = 1,
+        delay_factor: float = 2,
+        max_retries: Optional[int] = 3,
+        additional_status_codes: list = [],
+        attempt_timeout: Optional[float] = None,
+        total_timeout: Optional[float] = None,
+        attempt_timeout_floor=2.0,
+        retry_warning_message: Optional[RetryWarningMessage] = None,
+        metadata: list[tuple[str, str]] = [],
+    ) -> ResponseType:
+        return await retry_transient_errors(
+            fn,
+            req,
+            base_delay=base_delay,
+            max_delay=max_delay,
+            delay_factor=delay_factor,
+            max_retries=max_retries,
+            additional_status_codes=additional_status_codes,
+            attempt_timeout=attempt_timeout,
+            total_timeout=total_timeout,
+            attempt_timeout_floor=attempt_timeout_floor,
+            retry_warning_message=retry_warning_message,
+            metadata=metadata,
+        )
+
+    return wrapped
+
+
 def find_free_port() -> int:
     """
     Find a free TCP port, useful for testing.
--- a/modal/client.py
+++ b/modal/client.py
@@ -29,7 +29,7 @@ from ._traceback import print_server_warnings, suppress_tb_frames
 from ._utils import async_utils
 from ._utils.async_utils import TaskContext, synchronize_api
 from ._utils.auth_token_manager import _AuthTokenManager
-from ._utils.grpc_utils import ConnectionManager, retry_transient_errors
+from ._utils.grpc_utils import ConnectionManager
 from .config import _check_config, _is_remote, config, logger
 from .exception import AuthError, ClientClosed, NotFoundError
 
@@ -159,7 +159,7 @@ class _Client:
     async def hello(self):
         """Connect to server and retrieve version information; raise appropriate error for various failures."""
         logger.debug(f"Client ({id(self)}): Starting")
-        resp = await retry_transient_errors(self.stub.ClientHello, empty_pb2.Empty())
+        resp = await self.stub.ClientHello(empty_pb2.Empty())
         print_server_warnings(resp.server_warnings)
 
     async def __aenter__(self):
--- a/modal/_clustered_functions.py
+++ b/modal/_clustered_functions.py
@@ -5,7 +5,6 @@ from dataclasses import dataclass
 from typing import Optional
 
 from modal._utils.async_utils import synchronize_api
-from modal._utils.grpc_utils import retry_transient_errors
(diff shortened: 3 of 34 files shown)


modal/_utils/grpc_utils.py
```diff
@@ -174,8 +171,11 @@ async def unary_stream(
         yield item
 
 
+R = TypeVar("R")
+
+
 async def retry_transient_errors(
-    fn: "modal.client.UnaryUnaryWrapper[RequestType, ResponseType]",
+    fn: Callable[..., Awaitable[R]],
```
thomasjpfan: I couldn't find a good way to get the typing to work with `UnaryUnaryWrapper` directly and the generics:

Screenshot 2025-09-05 at 4 31 34 PM

So I went with passing a callable and then pass the name directly in `fn_name`.

modal/client.py
```diff
@@ -412,6 +405,37 @@ def name(self) -> str:
...
+                additional_status_codes=additional_status_codes,
+                attempt_timeout=attempt_timeout,
+                total_timeout=total_timeout,
+                attempt_timeout_floor=attempt_timeout_floor,
+                retry_warning_message=retry_warning_message,
+                metadata=metadata,
+                fn_name=self.name,
+            )
+
+    async def direct(
```
thomasjpfan: You can still call the underlying wrapper with `direct`.

modal/client.py
```diff
@@ -412,6 +405,37 @@ def name(self) -> str:
         return self.wrapped_method.name
 
     async def __call__(
+        self,
+        req: RequestType,
+        base_delay: float = 0.1,
```
mwaskom: Should everything past `req` require kwargs?

modal/client.py
```diff
@@ -412,6 +405,37 @@ def name(self) -> str:
...
+        max_delay: float = 1,
+        delay_factor: float = 2,
+        max_retries: Optional[int] = 3,
+        additional_status_codes: list = [],
+        attempt_timeout: Optional[float] = None,
+        total_timeout: Optional[float] = None,
+        attempt_timeout_floor=2.0,
+        retry_warning_message: Optional[RetryWarningMessage] = None,
+        metadata: list[tuple[str, str]] = [],
```
mwaskom: This should probably go above the retry parameters (should you be able to pass it as a positional arg?)

modal/client.py
```diff
@@ -412,6 +405,37 @@ def name(self) -> str:
...
+        req: RequestType,
+        base_delay: float = 0.1,
+        max_delay: float = 1,
+        delay_factor: float = 2,
+        max_retries: Optional[int] = 3,
+        additional_status_codes: list = [],
+        attempt_timeout: Optional[float] = None,
+        total_timeout: Optional[float] = None,
+        attempt_timeout_floor=2.0,
+        retry_warning_message: Optional[RetryWarningMessage] = None,
```
mwaskom: Or alternatively, what would you think about packaging up the retry parameterization into a dataclass so the stub method signatures are more like `(request, metadata, retries)`? Feels like that could keep things organized a little better maybe?

modal/client.py
mwaskom: Would we never want to pass a `timeout` when calling with retries?

modal/client.py
mwaskom: I wonder if that would also let us spell the no-retry case something like `stub.MyMethod(request, retries=None)` which is maybe a little bit more obvious than the `.direct()` suffix (although I don't hate that).

modal/config.py
```diff
@@ -143,11 +143,12 @@ def _read_user_config():
 
 
 async def _lookup_workspace(server_url: str, token_id: str, token_secret: str) -> api_pb2.WorkspaceNameLookupResponse:
+    from ._utils.grpc_utils import NoRetry
     from .client import _Client
 
     credentials = (token_id, token_secret)
     async with _Client(server_url, api_pb2.CLIENT_TYPE_CLIENT, credentials) as client:
-        return await client.stub.WorkspaceNameLookup(Empty(), timeout=3)
+        return await client.stub.WorkspaceNameLookup(Empty(), retry=NoRetry(timeout=3))
```
mwaskom: `retry=NoRetry(...)` is a little 😵‍💫 IMO

modal/client.py
```diff
@@ -412,6 +405,22 @@ def name(self) -> str:
         return self.wrapped_method.name
 
     async def __call__(
+        self,
+        req: RequestType,
+        retry: Optional[Union[Retry, NoRetry]] = None,
```
mwaskom: I think I'm still not understanding the difference between `retry=NoRetry()` and `retry=None`, why do you need both? Also was there a reason you decided not to go with having an instance default?

modal/config.py
thomasjpfan: I updated to force it to be positional-only, which looks a little better:

```python
client.stub.WorkspaceNameLookup(Empty(), NoRetry(timeout=3))
```

Context: I originally tried to do `retry = Retry()` as the default, but `inv type-stubs` did not work correctly. The type stub did not import `Retry` from the correct module. Then I tried to do a sentinel value so I can use `retry=None` to mean "No retries", but `inv type-stubs` was still not happy.

thomasjpfan: Here are the options:

- `retry=None` means `Retry()`. (It's a sentinel)
- `retry=NoRetry()` actually means no retries.
- `retry=Retry(...)` is a custom `Retry`

mwaskom: Hm, yeah `inv type-stubs` can be pretty annoying. Do we need `UnaryUnaryWrapper` to be defined in module with synchronicity-wrapped (i.e. public) components? The playbook for type stub generation problems usually looks like
1. Move the annoying part into a private module (feels like it could go in `_grpc_utils`?)
2. Make the "implementation module" private and add a public module with the synchronicity wrapping (like `functions.py` / `_functions.py`)

We arguably should consistently do (2) everywhere but in this case (1) seems like it would not produce an unusual organization (though maybe I am missing something).

thomasjpfan:
With the most recent update, there is a new `_grpc_client.py` that have all the GRPC wrappers. As a developer, the API now is:

- Retry by default: `stub.MyRPCCall(req)`
- Retries with custom parameters: `stub.MyRPCCall(req, retry=Retry(...))`
- No retries: `stub.MyRPCCall(req, retry=None)`

PR #3497, opened 2025-08-21
Surface cluster ID for clustered functions

irfansharif:
## Describe your changes

We already pass it back down from the server. Include it in the client object. Useful now that we're looking at single-function, multiple clusters.

---

## Compatibility checklist

---

</details>

## Changelog

The type returned by `modal.experimental.get_cluster_info()` now also includes the cluster ID - shared across the set of tasks that spin up in tandem when using the `@clustered` decorator.


irfansharif:
@prbot approve

mwaskom:
✅

PR #3499, opened 2025-08-22
check sandbox name length and warn

rolyatmax:
## Describe your changes

This enforces Sandbox name length constraints on the client. Sandbox names must be between 1 and 128 characters.

---

## Compatibility checklist

---

</details>

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -73,6 +73,12 @@ def _validate_exec_args(args: Sequence[str]) -> None:
         )
 
 
+def _validate_sandbox_name(name: str) -> None:
+    message = f"Invalid Sandbox name: '{name}'.\n\nSandbox names must be between 1 and 128 characters long."
+    if len(name) == 0 or len(name) > 128:
+        raise InvalidError(message)
+
+
 class DefaultSandboxNameOverride(str):
     """A singleton class that represents the default sandbox name override.
 
@@ -404,6 +410,8 @@ class _Sandbox(_Object, type_prefix="sb"):
         from .app import _App
 
         _validate_exec_args(args)
+        if name is not None:
+            _validate_sandbox_name(name)
 
         # TODO(erikbern): Get rid of the `_new` method and create an already-hydrated object
         obj = _Sandbox._new(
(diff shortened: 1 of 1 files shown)


modal/sandbox.py
```diff
@@ -73,6 +73,12 @@ def _validate_exec_args(args: Sequence[str]) -> None:
         )
 
 
+def _validate_sandbox_name(name: str) -> None:
+    message = f"Invalid Sandbox name: '{name}'.\n\nSandbox names must be between 1 and 128 characters long."
+    if len(name) == 0 or len(name) > 128:
```
rolyatmax: Note: we're not using the `name_utils` validation functions because Sandbox names can (and often are) longer than 64 characters. This puts in place the same constraints we had at Jamsocket and that are currently set on the DB schema.

mwaskom: Interesting! Is there a reason Sandbox names need to be up to that length? All things being equal I'd say it's nice if we have consistent rules everywhere.

Also do we not want similar restrictions on the character set that you can use? I think one reason to enforce that would be if we ever want to have a sandbox name in a URL right? Which may be handy at some point...

rolyatmax:
@mwaskom That's a good point. I'm going to gather some more data over the next week or two, then we can land this with constraints that match our other standard naming constraints if all looks good.

rolyatmax:
> Any insights from the logs?

From the logs, it seems like we do have several users who regularly set sandbox names that will become invalid. Let's switch this to a warning first and then we can error in an upcoming release.

modal/sandbox.py
```diff
@@ -73,6 +74,16 @@ def _validate_exec_args(args: Sequence[str]) -> None:
...
 
 
+def _warn_if_invalid_name(name: str) -> None:
+    message = (
+        f"Warning: sandbox name '{name}' will be considered invalid in an upcoming release."
+        "\n\nNames may contain only alphanumeric characters, dashes, periods, and underscores,"
+        " must be shorter than 64 characters, and cannot conflict with App ID strings."
+    )
+    if not is_valid_object_name(name):
+        raise InvalidError(message)
```
mwaskom: Should this use `deprecation_warning` instead?

mwaskom (commented):
Would be great to have a quick test!

modal/sandbox.py
rolyatmax: yeah, you're right. updated

PR #3500, opened 2025-08-22
FLASH-11 get_flash_url

emilyhanyf:


modal_proto/api.proto
```diff
@@ -1791,6 +1791,7 @@ message FunctionHandleMetadata {
   optional string input_plane_region = 47;
   // Use optional to ensure unset values default to None instead of 0
   optional uint64 max_object_size_bytes = 48;
+  repeated string _experimental_flash_urls = 49;
```
thomasjpfan: Can you extract out this proto change into it's own PR?

They are usually done in their own PRs. For example: https://github.com/modal-labs/modal-client/pull/3435/files

modal/_functions.py
```diff
@@ -1477,6 +1480,11 @@ async def get_web_url(self) -> Optional[str]:
         """URL of a Function running as a web endpoint."""
         return self._web_url
 
+    @live_method
+    async def _experimental_get_flash_urls(self) -> Optional[list[str]]:
+        """URL of the flash service for the function."""
+        return self._experimental_flash_urls or None
```
thomasjpfan: For the type checker:

```suggestion
        return list(self._experimental_flash_urls) or None
```

thomasjpfan (commented):
✅

PR #3506, opened 2025-08-26
[flash] scale up more when containers can't hit /metrics

emilyhanyf:
Prometheus can't hit `/metrics` when containers are overloaded. We want to just mark containers that time out or error when hitting `/metrics` as "unhealthy" and scale up more based on that.

### DIFF
--- a/modal/experimental/flash.py
+++ b/modal/experimental/flash.py
@@ -275,7 +275,7 @@ class _FlashPrometheusAutoscaler:
 
         # Scale up / down conservatively: Any container that is missing the metric is assumed to be at the minimum
         # value of the metric when scaling up and the maximum value of the metric when scaling down.
-        scale_up_target_metric_value = sum_metric / current_replicas
+        scale_up_target_metric_value = sum_metric / (containers_with_metrics if containers_with_metrics > 0 else 1)
         scale_down_target_metric_value = (
             sum_metric + n_containers_missing_metric * target_metric_value
         ) / current_replicas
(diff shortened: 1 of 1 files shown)


modal/experimental/flash.py
```diff
@@ -275,7 +275,7 @@ async def _compute_target_containers(self, current_replicas: int) -> int:
 
         # Scale up / down conservatively: Any container that is missing the metric is assumed to be at the minimum
         # value of the metric when scaling up and the maximum value of the metric when scaling down.
-        scale_up_target_metric_value = sum_metric / current_replicas
+        scale_up_target_metric_value = sum_metric / (containers_with_metrics if containers_with_metrics > 0 else 1)
```
molocule: if the metrics endpoint fully times out, will sum_metric >0 ?

modal/experimental/flash.py
```diff
@@ -303,8 +303,11 @@ async def _get_metrics(self, url: str) -> Optional[dict[str, list[Any]]]:  # tec
 
         # Fetch the metrics from the endpoint
         try:
-            response = await self.http_client.get(url)
+            response = await asyncio.wait_for(self.http_client.get(url), timeout=5)
             response.raise_for_status()
+        except asyncio.TimeoutError:
```
molocule, line 308: might need to handle some of the other errors you saw as well! e.g. server disconnect in case the timeout doesn't catch

modal/experimental/flash.py
```diff
@@ -303,8 +303,11 @@ async def _get_metrics(self, url: str) -> Optional[dict[str, list[Any]]]:  # tec
 
         # Fetch the metrics from the endpoint
         try:
-            response = await self.http_client.get(url)
+            response = await asyncio.wait_for(self.http_client.get(url), timeout=5)
```
gongy: should use a timeout on the .get call I think?

modal/experimental/flash.py
gongy: hmm, if all containers are unhealthy, I'm not seeing how this scales up. I think we need to augment the numerator `sum_metric + constant * containers_without_metrics` maybe

PR #3509, opened 2025-08-27
Add min overprovision

molocule:
add min overprovision for flash autoscaling

### DIFF
--- a/modal/experimental/flash.py
+++ b/modal/experimental/flash.py
@@ -121,6 +121,7 @@ class _FlashPrometheusAutoscaler:
         target_metric_value: float,
         min_containers: Optional[int],
         max_containers: Optional[int],
+        min_overprovision_containers: Optional[int],
         scale_up_tolerance: float,
         scale_down_tolerance: float,
         scale_up_stabilization_window_seconds: int,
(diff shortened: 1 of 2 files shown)


mwaskom:
That already exists. Though I don't really have any context for what you're trying to do, maybe it's different.

modal/experimental/flash.py
```diff
@@ -418,29 +422,36 @@ async def _compute_target_containers_prometheus(self, current_replicas: int) ->
...
-        )
+        # number of total containers - buffer containers
+        # This is used in 1) scale ratio denominators 2) provisioning base.
+        # Max is used to handle case when buffer_containers are first initialized.
+        num_provisioned_containers = max(current_replicas - buffer_containers, 1)
+
+        # Scale up assuming that every unhealthy container is at (1 + scale_up_tolerance)x the target metric value.
+        # This way if all containers are unhealthy, we will increase our number of containers.
+        scale_up_target_metric_value = (
+            sum_metric + (1 + self.scale_up_tolerance) * n_containers_unhealthy * target_metric_value
```
molocule, line 438: we don't want cold start containers to be counted here since they aren't serving requests
iirc we might have tried this and it lead to infinite scaling also ;-;

emilyhanyf (approved):
Looks good!!

PR #3510, opened 2025-08-27
Add separate max map batch size for `experimental_spawn_map`

devennavani:
## Describe your changes

Add separate max map batch size for `experimental_spawn_map`

### DIFF
--- a/modal/parallel_map.py
+++ b/modal/parallel_map.py
@@ -81,6 +81,8 @@ MAX_INPUTS_OUTSTANDING_DEFAULT = 1000
 
 # maximum number of inputs to send to the server in a single request
 MAP_INVOCATION_CHUNK_SIZE = 49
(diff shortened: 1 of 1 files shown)


modal/parallel_map.py
```diff
@@ -79,8 +79,10 @@ class _OutputValue:
 
 MAX_INPUTS_OUTSTANDING_DEFAULT = 1000
 
-# maximum number of inputs to send to the server in a single request
+# Maximum number of inputs to send to the server per FunctionPutInputs request
 MAP_INVOCATION_CHUNK_SIZE = 49
+SPAWN_MAP_INVOCATION_CHUNK_SIZE = 512
```
danielshaar, line 84: This should be 500. 512 is just a mystic power of 2 and the backend batching logic is in groups of 25 anyway, so might as well be a multiple of that.

PR #3511, opened 2025-08-27
Change first input timeout response code (502 -> 408)

rohansingh:
If we have a first input timeout while waiting for an HTTP request body, this more closely resembles a request timeout rather than a gateway timeout.

Using a less-common status code will also help us distinguish between error responses being returned by our client code vs. user code.


ekzhang (approved):

ehdr:
LGTM!

PR #3516, opened 2025-08-29
[flash autoscaler] add flash_scaling_metrics

emilyhanyf:
Get metrics from containers for flash autoscaler

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1314,6 +1314,13 @@ message FlashContainerRegisterResponse {
   string url = 1;
 }
 
+message FlashScalingMetric {
+  double cpu_usage_percent = 1;
+  double memory_usage_percent = 2;
+  uint32 concurrent_requests = 3;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -1314,6 +1314,13 @@ message FlashContainerRegisterResponse {
   string url = 1;
 }
 
+message FlashScalingMetric {
+  double cpu_usage_percent = 1;
+  double memory_usage_percent = 2;
+  uint32 concurrent_requests = 3;
```
molocule: i think if we generalize the metric to metrics in general, we can keep this in, but this might be misleading for flash specifically since concurrent_requests (measured as num inputs) will be 0

molocule (approved):

modal_proto/api.proto
emilyhanyf: change all naming to `AutoscalingMetrics` instead!

PR #3520, opened 2025-09-02
Add ffmpeg to notebook base image

ekzhang:
Needed to run whisper and a couple other audio-related packages like kokoro and torchaudio, seems good to include!

Very minor change, going to apply the new image in prod


ekzhang:
@prbot approve @howard36

PR #3522, opened 2025-09-03
Check tunnel before registering

molocule:
The subprocess dies which causes the tunnel to only return 502s. However the heartbeat on the worker level is fine/still succeeds. This means the container is still alive even though it is “unhealthy”.

to fix:
- try hitting the tunnel url in the heartbeat 
- if this fails more than max_failures times, deregister the tunnel
- then another draining process will close the tunnel and stop the container

more details here: https://www.notion.so/modal-com/Flash-Container-Error-Death-Handling-23e1e7f16949809c81e6c2c0821e2162


modal/experimental/flash.py
```diff
@@ -19,28 +21,67 @@
...
+                await asyncio.sleep(1)
+            except asyncio.CancelledError:
+                logger.warning("[Modal Flash] Shutting down...")
+                return
 
     async def _run_heartbeat(self, host: str, port: int):
         first_registration = True
         while True:
             try:
+                resp = await self.http_client.get(self.tunnel.url, timeout=1)
```
gongy: I think we should probably check if the port is open since the server might not be HTTP -- [example](https://github.com/modal-labs/modal-client/blob/4fc33b78608a8a2484b5cb9a55f0ba42943a74ab/modal/cli/programs/vscode.py#L64-L74)

modal_proto/api.proto
```diff
@@ -1320,6 +1320,7 @@ message FlashContainerRegisterRequest {
 
 message FlashContainerRegisterResponse {
   string url = 1;
+  string task_id = 2;
```
gongy: `task_id` is `MODAL_TASK_ID`

modal/experimental/flash.py

modal_proto/api.proto
molocule: fixed

molocule:
baking now:
https://modal.com/apps/modal-labs/reducto/deployed/flash-upstream?start=1756928234.936&end=1756929134.936&live=true&activeTab=functions&functionId=fu-omBwqhpnqbxRrhzDj6igsD&functionSection=calls&limit=100&includeLogContext=false

image


better than before, also tracking metrics here:

image

modal/experimental/flash.py
```diff
@@ -19,28 +23,89 @@
...
+        """
+        while True:
+            try:
+                # Check if the container should be drained (e.g., too many failures)
+                if self.num_failures > MAX_FAILURES:
+                    logger.warning(
+                        f"[Modal Flash] Draining task {self.task_id} on {self.tunnel.url} due to too many failures."
+                    )
+                    await self.stop()
+                    await self.close()
```
gongy: we can probably let the close happen in the eventual `exit` right?

molocule: yes

PR #3533, opened 2025-09-04
fix port check return

molocule:
- made changes to port checking to return exception, but i never handled these issues
- fix tests to return correct values instead of raising exception
- actually handle the returned errors


molocule:
@prbot help a brother out

gongy:
:+1:

PR #3534, opened 2025-09-04
adjust flash manager timeout

molocule:
noticing that the 5 second timeout is pretty detrimental, we should just deregister quickly and allow more grace time for containers to recover

for example here is a container that was erroring out, but in the timeout leeway of 5s, it racked up ~400 errors

image


molocule:
@prbot approve @gongy

gongy:
:+1:

PR #3535, opened 2025-09-04
[flash autoscaler] apply buffer container logic to autoscaler_internal

emilyhanyf:


modal/experimental/flash.py
```diff
@@ -322,6 +322,66 @@ async def _run_autoscaler_loop(self):
                 logger.error(traceback.format_exc())
                 await asyncio.sleep(self.autoscaling_interval_seconds)
 
+    def _calculate_desired_replicas(
```
molocule, line 350: i feel like a lot of the code here is replicated between this function and the prometheus one, is there a way to combine them (like fetch metrics differently but handle the calculation in the same way) so we don't have two sets of logic to keep up to date

modal/experimental/flash.py
```diff
@@ -322,6 +322,66 @@ async def _run_autoscaler_loop(self):
...
+        # number of total containers - buffer containers
+        # This is used in 1) scale ratio denominators 2) provisioning base.
+        # Max is used to handle case when buffer_containers are first initialized.
+        num_provisioned_containers = max(current_replicas - buffer_containers, 1)
+
+        # Scale up assuming that every unhealthy container is at (1 + scale_up_tolerance)x the target metric value.
+        # This way if all containers are unhealthy, we will increase our number of containers.
+        scale_up_target_metric_value = (
+            sum_metric + (1 + self.scale_up_tolerance) * n_containers_unhealthy * target_metric_value
+        ) / (num_provisioned_containers)
```
molocule: does cpu, mem scale linearly across containers? a bit worried if there's some overhead it might affect calculations in a bad way

modal/experimental/flash.py
emilyhanyf: yep! working on that rn!

modal/experimental/flash.py
molocule: ah i guess i mean if we have 3 buffer containers, and we don't count in denominator, does cpu of each buffer container have evenly distributed load or does each buffer container have evenly distributed load + some baseline

modal/experimental/flash.py
```diff
@@ -337,95 +336,39 @@ async def _compute_target_containers_internal(self, current_replicas: int) -> in
...
 
-        # Gets metrics from prometheus
-        sum_metric = 0
-        containers_with_metrics = 0
+    def _calculate_desired_replicas(
+        self,
+        current_replicas: int,
+        sum_metric: float,
+        containers_with_metrics: int,
+        total_containers: int,
```
molocule: nit: i think total_containers might be ambiguous since we have current_replicas and len(containers), maybe specify what you mean here

modal/experimental/flash.py
```diff
@@ -468,6 +412,46 @@ async def _compute_target_containers_prometheus(self, current_replicas: int) ->
 
         return desired_replicas
 
+    async def _get_scaling_info_internal(self, containers) -> tuple[float, int]:
```
molocule: nit: if you're getting metrics maybe just use get_metrics... for better clarity

emilyhanyf: i think there's another `_get_metrics` that tries to hit the prometheus metrics endpoint though :,) can change that one's wording if that makes it clearer!

modal/experimental/flash.py
molocule: cause worried it will just cause infinite blow up (maybe we can test or add tests for this) or you can just try it w the sleeper app with buffer container metric

modal/experimental/flash.py
emilyhanyf: also i think we also want to get the `containers_with_metrics` from the helper function, so i thought maybe "scaling_info" is more accurate. lmk what u think!

modal/experimental/flash.py
emilyhanyf: renaming this to `n_live_containers` and adding a brief comment to make different variables clearer!

molocule: ok ty!!

molocule: feel free to name the rest of the variables in a way that makes more sense overall btw, i feel like we have all kinds of definitions of unhealthy, alive, etc.

modal/experimental/flash.py
molocule: ok makes sense! also i took a closer look at the two functions and i think you can probably combine then? cause they seem to be doing the same thing more or less : ). i think users shouldn't care too much about where the metric is being fetched from (i.e. internally vs prometheus), they should only care that they can use internal metrics

molocule (approved):
approved if load test goes well!

modal/experimental/flash.py
emilyhanyf: added locust load testing results in slack!

PR #3537, opened 2025-09-05
auth relay protobuf updates

rolyatmax:
This adds protobuf types for creating connect credentials for a sandbox.

---

## Compatibility checklist

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2671,6 +2671,16 @@ message SandboxCreateResponse {
   string sandbox_id = 1;
 }
 
+message SandboxCreateConnectCredentialsRequest {
+  string sandbox_id = 1;
+  string metadata = 2;
+}
+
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -2661,6 +2661,16 @@ message Sandbox {
   optional uint32 idle_timeout_secs = 33;
 }
 
+message SandboxCreateConnectCredentialsRequest {
+  string sandbox_id = 1;
+  string metadata = 2;
```
paulgb, line 2666: just to be sure I'm understanding, this will be JSON right? I wonder if calling it `metadata_json` would be clearer, but not deadset on that.

rolyatmax: This is just a string. If you provide a dict to the `metadata=` arg, we'll serialize it as JSON for you. And then in the container you receive it as a string (and have to parse it yourself if you passed in a dict).

PR #3538, opened 2025-09-05
auth relay connect_credentials() fn for sandboxes

rolyatmax:
Adds a `connect_credentials()` method for Sandboxes.

## Changelog

- Adds a `connect_credentials()` method for Sandboxes which may be used to generate credentials for making HTTP / Websocket requests to a server running in a Sandbox.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -1,9 +1,11 @@
 # Copyright Modal Labs 2022
 import asyncio
+import json
 import os
 import time
 from collections.abc import AsyncGenerator, Sequence
-from typing import TYPE_CHECKING, AsyncIterator, Literal, Optional, Union, overload
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, AsyncIterator, Literal, Optional, Union, overload
 
 from .config import config, logger
 
@@ -97,6 +99,14 @@ class DefaultSandboxNameOverride(str):
 
 
 _DEFAULT_SANDBOX_NAME_OVERRIDE = DefaultSandboxNameOverride()
+
+
+@dataclass(frozen=True)
+class SandboxConnectCredentials:
+    """Simple data structure storing credentials for making HTTP connections to a sandbox."""
+
+    url: str
+    token: str
 
 
 class _Sandbox(_Object, type_prefix="sb"):
@@ -625,6 +635,20 @@ class _Sandbox(_Object, type_prefix="sb"):
         }
 
         return self._tunnels
+
+    async def connect_credentials(
+        self, metadata: Optional[Union[str, dict[str, Any]]] = None
+    ) -> SandboxConnectCredentials:
+        """Create credentials for making HTTP connections to the sandbox."""
+        if metadata is not None and isinstance(metadata, dict):
+            try:
+                metadata = json.dumps(metadata)
+            except BaseException as e:
+                raise InvalidError(f"Failed to serialize metadata: {e}")
+
+        req = api_pb2.SandboxCreateConnectCredentialsRequest(sandbox_id=self.object_id, metadata=metadata)
+        resp = await retry_transient_errors(self._client.stub.SandboxCreateConnectCredentials, req)
+        return SandboxConnectCredentials(resp.url, resp.token)
 
     async def reload_volumes(self) -> None:
         """Reload all Volumes mounted in the Sandbox.


modal/sandbox.py
```diff
@@ -626,6 +636,20 @@ async def tunnels(self, timeout: int = 50) -> dict[int, Tunnel]:
 
         return self._tunnels
 
+    async def connect_credentials(
```
saltzm: [nit] `connect` is also a verb so this reads like "connect the credentials to something". I'd prefer `create_connect_credentials` or `create_http_credentials` or just `create_credentials`

mwaskom: I had the same reaction; based on the name I expected the credentials to be the argument, not the return value

rolyatmax: That's fair. Okay, maybe we'll go with `create_connect_credentials` then. Or maybe `create_connect_token`?

rolyatmax: Updated!

mwaskom: I guess I'm slightly irked by there being `Sandbox.create()` (a constructor) and `Sandbox.create_connect_token()` (an instance method) but I'm not sure any other verb would be better.

mwaskom:
```
# you can pass the token in an auth header or a query param
requests.get(
    creds.url,
    headers={"Authorization": f"Bearer {creds.token}"}
)
```

Is the `url` a url _for_ the creds?

modal/sandbox.py
rolyatmax: It seems like the original issue was that "connect" could be read as a verb, resulting in a confusing method name. If we could come up with a different adjective to describe "token", then maybe the method name could be simpler: `sb.adjective_token()`.

saltzm: >  slightly irked by there being Sandbox.create() (a constructor) and Sandbox.create_connect_token()

Why is this irksome?

saltzm (approved):

modal/sandbox.py
mwaskom: > If we could come up with a different adjective to describe "token",

I asked Claude for three adjectives and it suggested "bright", "curious", and "smooth". Do any of those work? Is this what people mean by vibe coding? 🤡 

> Why is this irksome?

Eh, it's a non-blocking, fairly low-grade amount of irk, but the methods have very different functions so ideally they would have suitably distinct names. I think `modal.Sandbox.create()` is probably the real offender here, it's basically a "normal" constructor and I think you'd typically just expect it to be an `__init__` method.

modal/sandbox.py
```diff
@@ -626,6 +636,20 @@ async def tunnels(self, timeout: int = 50) -> dict[int, Tunnel]:
 
         return self._tunnels
 
+    async def create_connect_token(
+        self, metadata: Optional[Union[str, dict[str, Any]]] = None
+    ) -> SandboxConnectCredentials:
+        """Create a token for making HTTP connections to the sandbox."""
```
mwaskom: Can we add some explanation of the `metadata` parameter?

modal/sandbox.py
```diff
@@ -626,6 +636,20 @@ async def tunnels(self, timeout: int = 50) -> dict[int, Tunnel]:
...
+        if metadata is not None and isinstance(metadata, dict):
+            try:
+                metadata = json.dumps(metadata)
+            except BaseException as e:
```
mwaskom: You very rarely want to catch `BaseException` in Python; control-related exceptions like `KeyboardInterrupt` derive from it but not from `Exception`. `except Exception as e:` is the recommended spelling when you want "any" exception that might be triggered by what you do in the `try` block.

modal/sandbox.py
mwaskom: Would, like, `sb.relay()` be a bad name?

modal/sandbox.py
rolyatmax: Updated!

modal/sandbox.py
rolyatmax: Updated!

modal/sandbox.py
rolyatmax: We could call them Sandbox Auth Tokens maybe? `sb.auth_token()`?
I don't think `sb.relay()` describes what it does. I also don't think anyone is going to confuse `Sandbox.create()` and `sb.create_connect_token()`.

modal/sandbox.py
mwaskom: Glad I could get a useful contribution in under the wire 😅

rolyatmax:
> the dataclass is not strictly holding _credentials_ then but I'm not sure what it is holding

@mwaskom yeah, i hear you. "Credentials" was the best term I could come up with.

modal/sandbox.py
mwaskom: It's not really about confusing them, it's more about making it harder to predict what `.create()` does (i.e. understand that it's a constructor). It's not a huge problem just inelegant.

PR #3539, opened 2025-09-05
FLASH-7 Proto changes to add target_containers

emilyhanyf:


mwaskom:
If this is a Flash specific thing I'm not sure it should go in the AutoscalerSettings message.

Would love to see a design doc on some of this stuff to better understand it.

emilyhanyf:
@mwaskom We are adding a target_container field for flash autoscaler, but we are actually going to make a new RPC instead of modifying this proto/RPC. Will revert the changes in this PR today!

PR #3541, opened 2025-09-08
[cbor support] Adds `supported_output_formats` and `supported_input_formats` to Function metadata

freider:
This is prep for the addition of actual cbor support in the client (see #3474 which is being reworked now to use this information)

* Adds proto fields `supported_input_formats` and `supported_output_formats`
* Propagate/set these correctly for existing functions and methods (disambiguating asgi and pickle at the moment)
* Adds DATA_FORMAT_CBOR, without actually implementing it in this PR (see #3474)

---

## Compatibility checklist

---

## Release checklist

---

</details>


modal_proto/api.proto
```diff
@@ -111,6 +111,7 @@ enum DataFormat {
   DATA_FORMAT_PICKLE = 1; // Cloudpickle
   DATA_FORMAT_ASGI = 2; // "Asgi" protobuf message
   DATA_FORMAT_GENERATOR_DONE = 3; // "GeneratorDone" protobuf message
+  DATA_FORMAT_CBOR = 4;
```
thomasjpfan, line 114: Is it okay to include proto changes in the same PR as the implementation changes or should proto changes be in their own PR?

modal/_functions.py
```diff
@@ -1529,6 +1554,8 @@ def _get_metadata(self):
             input_plane_region=self._input_plane_region,
             max_object_size_bytes=self._max_object_size_bytes,
             _experimental_flash_urls=self._experimental_flash_urls,
+            supported_input_formats=self._metadata.supported_input_formats if self._metadata else None,
+            supported_output_formats=self._metadata.supported_output_formats if self._metadata else [],
```
thomasjpfan: Why is the default input format `None` while the output format's default is `[]`?

test/function_test.py
```diff
@@ -1484,6 +1484,46 @@ def f(a: int) -> list[str]: ...
...
+    @app.function(serialized=True)
+    @modal.fastapi_endpoint()
+    def web_f(): ...
+
+    @app.cls(serialized=True)
+    class A:
+        @modal.method()
+        def f(self): ...
+
+        @modal.fastapi_endpoint()
```
thomasjpfan: Nit: Just to test a slightly different code path compared to the `fastapi_endpoint` above.

```suggestion
        @modal.web_server(port=8081)
```

modal/_functions.py
freider: Not intentional, thanks for spotting, fixing. I think it doesn't have any effect in practice though since python-proto will encode both the same.

modal_proto/api.proto
freider: I don't think it's necessary in this case, but normally good practice yeah :)

freider:
@prbot approve

PR #3542, opened 2025-09-08
[flash-autoscaler] fix edge case

molocule:
encountered a weird edge case where 9 containers are unhealthy and 1 container is cold starting (which was in the denominator, but not in numerator)

fix denominator
adding a test to make sure that when containers with metric = 0, we always scale up (scale ratio always positive)


emilyhanyf (commented):
thank you!

PR #3548, opened 2025-09-12
Accept `env` params in addition to `secrets`

ehdr:
## Describe your changes

Today you can pass environment variables via an ephemeral Secret using `secrets=[Secret.from_dict(dict(k="v"))]`, but it's not super discoverable, for a use case that is very common. This provides a simple convenience wrapper in the form of a separate `env` parameter.

Also changes the `secrets` params to `Collections` (because some already were, and we don't need them to be `Sequence`s), and removes the mutable default arguments.

---

## Compatibility checklist

</details>

## Changelog

- Added `env` parameters to several methods, as a more discoverable convenience method for passing non-secret environment variables to containers.

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -6,7 +6,7 @@ import textwrap
 import time
 import typing
 import warnings
-from collections.abc import AsyncGenerator, Sequence, Sized
+from collections.abc import AsyncGenerator, Collection, Sequence, Sized
 from dataclasses import dataclass
 from pathlib import PurePosixPath
 from typing import TYPE_CHECKING, Any, AsyncIterator, Callable, Optional, Union
@@ -597,7 +597,7 @@ class _FunctionSpec:
 
     image: Optional[_Image]
     mounts: Sequence[_Mount]
-    secrets: Sequence[_Secret]
+    secrets: Collection[_Secret]
     network_file_systems: dict[Union[str, PurePosixPath], _NetworkFileSystem]
     volumes: dict[Union[str, PurePosixPath], Union[_Volume, _CloudBucketMount]]
     # TODO(irfansharif): Somehow assert that it's the first kind, in sandboxes
@@ -661,7 +661,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         info: FunctionInfo,
         app,
         image: _Image,
-        secrets: Sequence[_Secret] = (),
+        secrets: Optional[Collection[_Secret]] = None,
         schedule: Optional[Schedule] = None,
         is_generator: bool = False,
         gpu: Union[GPU_T, list[GPU_T]] = None,
@@ -700,6 +700,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
         experimental_options: Optional[dict[str, str]] = None,
         _experimental_proxy_ip: Optional[str] = None,
         _experimental_custom_scaling_factor: Optional[float] = None,
+        env: Optional[dict[str, str]] = None,
     ) -> "_Function":
         """mdmd:hidden"""
         # Needed to avoid circular imports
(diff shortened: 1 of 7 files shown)


modal/_functions.py
```diff
@@ -700,6 +700,7 @@ def from_local(
         experimental_options: Optional[dict[str, str]] = None,
         _experimental_proxy_ip: Optional[str] = None,
         _experimental_custom_scaling_factor: Optional[float] = None,
+        env: Optional[dict[str, str]] = None,
```
mwaskom: Why all the way down here?

mwaskom: Also should we type this as a Mapping?

modal/app.py
```diff
@@ -920,6 +925,10 @@ def cls(
         if allow_cross_region_volumes is not None:
             deprecation_warning((2025, 4, 23), "The `allow_cross_region_volumes` parameter no longer has any effect.")
 
+        secrets = secrets or []
+        if env:
+            secrets = [*secrets, _Secret.from_dict(dict(**env))]
```
mwaskom: If `env` is a dict why do we need to unpack them into another dict?

modal/sandbox.py
```diff
@@ -708,10 +718,11 @@ async def exec(
         stderr: StreamType = StreamType.PIPE,
         timeout: Optional[int] = None,
         workdir: Optional[str] = None,
-        secrets: Sequence[_Secret] = (),
+        secrets: Optional[Collection[_Secret]] = None,
         text: Literal[True] = True,
         bufsize: Literal[-1, 1] = -1,
         _pty_info: Optional[api_pb2.PTYInfo] = None,
+        env: Optional[dict[str, Union[str, None]]] = None,
```
mwaskom: Why not put this next to `secrets` like you do elsewhere?

modal/sandbox.py
```diff
@@ -723,10 +734,11 @@ async def exec(
         stderr: StreamType = StreamType.PIPE,
         timeout: Optional[int] = None,
         workdir: Optional[str] = None,
-        secrets: Sequence[_Secret] = (),
+        secrets: Optional[Collection[_Secret]] = None,
         text: Literal[False] = False,
         bufsize: Literal[-1, 1] = -1,
         _pty_info: Optional[api_pb2.PTYInfo] = None,
+        env: Optional[dict[str, Union[str, None]]] = None,
```
mwaskom: Same q

modal/_functions.py
ehdr: Because some of these don't enforce kwargs, so changing the order of params is technically a breaking change... 😞

modal/app.py
ehdr: Because pyright says `Argument of type "dict[str, str]" cannot be assigned to parameter "env_dict" of type "dict[str, str | None]" in function "from_dict"`!

mwaskom: > Create a secret from a str-str dictionary. Values can also be `None`, which is ignored.

🤔 what does that even mean

mwaskom: Maybe it makes sense to make the values in `env` optional too? I don't know the reason for that behavior but maybe there is a good one.

ehdr: We just [filter out any pairs where the value is None](https://github.com/modal-labs/modal-client/blob/main/modal/secret.py#L257)!

modal/sandbox.py
mwaskom: Doesn't this one (and some of the others) require kwargs by virtue of accepting variadic `*args`?

modal/_functions.py
mwaskom: Hm, `Function.from_local` is internal though, although I guess it isn't named that way. It's undocumented anyway...

modal/app.py
mwaskom: My guess is that this makes the typing nicer for people defining the env with `os.environ.get()`

modal/sandbox.py
ehdr: Good catch - in fact all of them except `from_local`! Fixed now.

modal/app.py
ehdr: Yeah makes sense, changed it now

PR #3549, opened 2025-09-12
add wait_until_running() function [ART-5]

aadit-juneja:
to let user check if sandbox has been scheduled to a worker

**UPDATED: replacing `wait_for_scheduled` name with `wait_until_running`**
## Describe your changes

This PR provides a user facing API to check if their sandbox has been assigned onto a worker (Linear Issue ART-5 in agent runtime).
It exposes a `wait_until_running()` function that can be utilized to check if the sandbox has been assigned to a worker, raising an error if no worker is assigned and returning None if a worker gets assigned.

---

## Compatibility checklist

---
</details>

## Changelog

- Add a `wait_until_running()` function to Sandbox class that wraps the `_get_task_id()` function for a discoverable method to check if your sandbox has been assigned to a worker.

```python
import modal
import time

s=time.time()
sb.wait_until_running()
e=time.time()
print(f"Sandbox assigned in {e-s:.2f} seconds"
```

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -320,7 +320,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         ] = None,  # Experimental controls over fine-grained scheduling (alpha).
         client: Optional[_Client] = None,
         environment_name: Optional[str] = None,  # *DEPRECATED* Optionally override the default environment
-        wait_until_running: bool = False,  # Enable waiting for the sandbox to be assigned to a worker before returning
+        wait_until_running: bool = False,  # Wait for the sandbox to start running before returning
     ) -> "_Sandbox":
         """
         Create a new Sandbox to run untrusted, arbitrary code.
@@ -421,7 +421,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         ] = None,  # Experimental controls over fine-grained scheduling (alpha).
         client: Optional[_Client] = None,
         verbose: bool = False,
-        wait_until_running: bool = False,  # Whether to wait for the sandbox to be assigned to a worker before returning
+        wait_until_running: bool = False,  # Wait for the sandbox to start running before returning
     ):
         # This method exposes some internal arguments (currently `mounts`) which are not in the public API
         # `mounts` is currently only used by modal shell (cli) to provide a function's mounts to the
@@ -499,7 +499,7 @@ class _Sandbox(_Object, type_prefix="sb"):
         await resolver.load(obj)
 
         if wait_until_running:
-            await obj.wait_until_running()
+            await obj._get_task_id()
 
         return obj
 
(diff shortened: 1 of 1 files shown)


modal/sandbox.py
```diff
@@ -320,6 +320,7 @@ async def create(
         ] = None,  # Experimental controls over fine-grained scheduling (alpha).
         client: Optional[_Client] = None,
         environment_name: Optional[str] = None,  # *DEPRECATED* Optionally override the default environment
+        wait_for_scheduled: bool = False,  # Enables waiting for the sandbox to be assigned to worker before returning
```
mwaskom: Would `wait_until_running` be a better name? I'm having trouble parsing `wait_for_scheduled`.

aadit-juneja: Yeah `wait_for_scheduled` could be confusing. I will change it to `wait_until_running`.

modal/sandbox.py
```diff
@@ -320,7 +320,7 @@ async def create(
         ] = None,  # Experimental controls over fine-grained scheduling (alpha).
         client: Optional[_Client] = None,
         environment_name: Optional[str] = None,  # *DEPRECATED* Optionally override the default environment
-        wait_for_scheduled: bool = False,  # Enables waiting for the sandbox to be assigned to worker before returning
+        wait_until_running: bool = False,  # Enable waiting for the sandbox to be assigned to a worker before returning
```
mwaskom:
```suggestion
        wait_until_running: bool = False,  # Wait for the sandbox to start running before returning
```

The "worker" concept is part of our internal ontology but not something we really expose to users. And it's redundant to say "Enable ..." when you have a boolean parameter

modal/sandbox.py
```diff
@@ -706,8 +706,8 @@ async def _get_task_id(self) -> str:
                 await asyncio.sleep(0.5)
         return self._task_id
 
-    async def wait_for_scheduled(self) -> None:
-        """Allows user to wait for the sandbox to be scheduled onto a worker."""
+    async def wait_until_running(self) -> None:
```
mwaskom: Is there a usecase for a public method here if we're going to have a boolean for it in the constructor? Could the constructor call `self._get_task_id()` directly?

modal/sandbox.py
aadit-juneja: I see, I can get rid of the references to a worker in the comments and the use of enable.

modal/sandbox.py
aadit-juneja: I see, that makes sense. I can get rid of the extra method and just directly call `get_task_id()` in the `_create()` function.

paulgb:
We discussed this on the Agent Runtime team and are in favor of this being a separate method rather than a sandbox parameter for these reasons:

- The goal here is to increase the discoverability of the functionality. The more intuitive place for someone to look is in the tab-completion list of a sandbox instance, rather than in the constructor IMHO.
- The argument list of a constructor typically takes attributes of that object; adding additional blocking behavior based on it feels unintuitive.
- It doesn't support the pattern of:
  - Start a sandbox as part of initialization and await to ensure that the request is processed
  - Do some work unrelated to the sandbox
  - Wait for the sandbox to start

My vote is to remove the constructor arg and support only the method.

PR #3550, opened 2025-09-12
Add `pty` param to `Sandbox.create` and `.exec`

ehdr:
Also deprecate `pty_info` and `_pty_info`.

---

## Compatibility checklist

</details>

## Changelog

- Added an option to enable a PTY for Sandboxes via `Sandbox.create(..., pty=True)` and `Sandbox.exec(..., pty=True)`. Also deprecated the old `pty_info` parameters.

### DIFF
--- a/test/sandbox_test.py
+++ b/test/sandbox_test.py
@@ -639,6 +639,7 @@ def test_sandbox_create_pty(app, servicer):
         assert req.definition.pty_info is not None
         assert req.definition.pty_info.enabled is True
         assert req.definition.pty_info.pty_type == api_pb2.PTYInfo.PTY_TYPE_SHELL
+        assert req.definition.pty_info.no_terminate_on_idle_stdin is True
 
 
 @skip_non_subprocess
@@ -651,3 +652,4 @@ def test_sandbox_exec_pty(app, servicer):
         assert req.pty_info is not None
         assert req.pty_info.enabled is True
         assert req.pty_info.pty_type == api_pb2.PTYInfo.PTY_TYPE_SHELL
+        assert req.pty_info.no_terminate_on_idle_stdin is True


modal/sandbox.py
```diff
@@ -177,6 +183,9 @@ def _new(
         cloud_bucket_mounts = [(k, v) for k, v in validated_volumes if isinstance(v, _CloudBucketMount)]
         validated_volumes = [(k, v) for k, v in validated_volumes if isinstance(v, _Volume)]
 
+        if pty:
+            pty_info = _Sandbox._default_pty_info()
```
thomasjpfan, line 187: That means we are using the parameters from the local shell for pty. I'm okay with that.

modal/sandbox.py
```diff
@@ -370,59 +387,51 @@ async def create(
...
             client=client,
             verbose=verbose,
+            pty_info=pty_info,
         )
 
     @staticmethod
     async def _create(
-        *args: str,  # Set the CMD of the Sandbox, overriding any CMD of the container image.
-        # Associate the sandbox with an app. Required unless creating from a container.
+        *args: str,
```
thomasjpfan, line 395: Are you removing the comments because this is private?

modal/sandbox.py
```diff
@@ -743,8 +754,9 @@ async def exec(
         # Control line-buffered output.
         # -1 means unbuffered, 1 means line-buffered (only available if `text=True`).
         bufsize: Literal[-1, 1] = -1,
-        # Internal option to set terminal size and metadata
-        _pty_info: Optional[api_pb2.PTYInfo] = None,
+        pty: bool = False,  # Enable a PTY for the command
```
thomasjpfan, line 757: Since we do not have `*` to make these keyword only, I do not think we can change the order of these parameters without the possibility of breaking code.

test/sandbox_test.py
```diff
@@ -628,3 +628,26 @@ def test_sandbox_volume(app, servicer, read_only):
...
+
+@skip_non_subprocess
+def test_sandbox_create_pty(app, servicer):
+    with servicer.intercept() as ctx:
+        Sandbox.create("echo", "hi", pty=True, app=app)
+        req = ctx.pop_request("SandboxCreate")
+
+        assert req.definition.pty_info is not None
+        assert req.definition.pty_info.enabled is True
+        assert req.definition.pty_info.pty_type == api_pb2.PTYInfo.PTY_TYPE_SHELL
```
thomasjpfan, line 641: Quick check to make sure `no_terminate_on_idle_stdin` is true?

test/sandbox_test.py
```diff
@@ -628,3 +628,26 @@ def test_sandbox_volume(app, servicer, read_only):
...
+@skip_non_subprocess
+def test_sandbox_exec_pty(app, servicer):
+    with servicer.intercept() as ctx:
+        sb = Sandbox.create("sleep", "infinity", app=app)
+        sb.exec("echo", "hello", pty=True)
+        req = ctx.pop_request("ContainerExec")
+
+        assert req.pty_info is not None
+        assert req.pty_info.enabled is True
+        assert req.pty_info.pty_type == api_pb2.PTYInfo.PTY_TYPE_SHELL
```
thomasjpfan, line 654: Same here for `no_terminate_on_idle_stdin`

modal/sandbox.py
ehdr: Yes exactly, if they are available, and else they are left as absent.

modal/sandbox.py
ehdr: Yes, just to reduce duplication, since the method is anyway private.

modal/sandbox.py
ehdr: Yes but (as @mwaskom [pointed out to me](https://github.com/modal-labs/modal-client/pull/3548#discussion_r2344468809) just the other day), the `*args` variadic args effectively forces the rest to be kwargs!

PR #3559, opened 2025-09-16
Add support for `modal shell sb-id` and `--image im-id`

ehdr:
## Describe your changes

So you can connect to a running Sandbox, or start a shell from a Modal Image ID (in addition to a image registry tag as today).

## Changelog

- Added `modal shell` support for connecting to a running Sandbox (`modal shell sb-id`), and starting a shell with a Modal Image ID (`--image im-id`) (in addition to image registry tag as today).

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -517,30 +517,31 @@ def serve(
 
 
 def shell(
-    container_or_function: Optional[str] = typer.Argument(
+    ref: Optional[str] = typer.Argument(
         default=None,
         help=(
-            "ID of running container, or path to a Python file containing a Modal App."
-            " Can also include a function specifier, like `module.py::func`, if the file defines multiple functions."
+            "ID of running container or Sandbox, or path to a Python file containing an App."
+            " Can also include a Function specifier, like `module.py::func`, if the file defines multiple Functions."
         ),
-        metavar="REF",
     ),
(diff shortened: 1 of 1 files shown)


mwaskom:
Hm, I'm not sure about `--image=image-id`

ehdr:
> Hm, I'm not sure about `--image=image-id`

Why not?

mwaskom:
> should we then also do shell im-id vo-id1 vo-id2 
I would say no on the basis that Volume IDs aren't user-facing. Image IDs aren't exactly either but without a concept of "named images" they're currently the _only_ way to get a CLI-friendly handle for an Image object.

modal/cli/run.py
```diff
@@ -613,19 +618,31 @@ def shell(
...
+        # `modal shell` with a sandbox ID gets the task_id, that's then handled by the `ta-*` flow below.
+        if ref.startswith("sb-") and len(ref[3:]) > 0 and ref[3:].isalnum():
+            from ..sandbox import Sandbox
+
+            try:
+                sandbox = Sandbox.from_id(ref)
+                task_id = sandbox._get_task_id()
+                ref = task_id
+            except Exception as e:
+                from ..exception import NotFoundError
```
mwaskom: Looks like we already import `modal.exception` in global scope for this module so not sure why it's necessary to defer the import and make the exception catching more awkward.

ehdr: Yeah good question, fixed!

PR #3560, opened 2025-09-16
Add AutoscalerConfiguration proto

kareem-modal:
## Describe your changes

- PRD-751 Allow autoscaler overrides from the web app

Introduces a new `AutoscalerConfiguration` message that represents the coalesced state of autoscaler settings. This includes tracking current settings, defaults, static initialization values, overrides, and metadata (UserActionInfo) about who applied overrides and when.

This is not used by the client currently, but added here in anticipation of future use, and to centralize autoscaler related protos.

---

## Compatibility checklist

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -628,23 +628,6 @@ message AuthTokenGetResponse {
   string token = 1;
 }
 
-message AutoscalerSettings {
-  // A collection of user-configurable settings for Function autoscaling
-  // These are used for static configuration and for dynamic autoscaler updates
-
-  // Minimum containers when scale-to-zero is not desired; pka "keep_warm" or "warm_pool_size"
-  optional uint32 min_containers = 1;
-  // Limit on the number of containers that can be running for each Function; pka "concurrency_limit"
-  optional uint32 max_containers = 2;
-  // Additional container to spin up when Function is active
-  optional uint32 buffer_containers = 3;
-  // Currently unused; a placeholder in case we decide to expose scaleup control to users
-  optional uint32 scaleup_window = 4;
-  // Maximum amount of time a container can be idle before being scaled down, in seconds; pka "container_idle_timeout"
-  optional uint32 scaledown_window = 5;
-  reserved 6;
-}
-
 // Message representing the current (coalesced) state of the autoscaler configuration
 // As well as the different sources that were used to create it.
 message AutoscalerConfiguration {
@@ -664,6 +647,23 @@ message AutoscalerConfiguration {
   modal.client.AutoscalerSettings static_settings = 4;
   // The merge of all overrides that were used to create the current configuration.
   modal.client.AutoscalerSettings override_settings = 5;
+}
+
+message AutoscalerSettings {
+  // A collection of user-configurable settings for Function autoscaling
+  // These are used for static configuration and for dynamic autoscaler updates
+
+  // Minimum containers when scale-to-zero is not desired; pka "keep_warm" or "warm_pool_size"
+  optional uint32 min_containers = 1;
+  // Limit on the number of containers that can be running for each Function; pka "concurrency_limit"
+  optional uint32 max_containers = 2;
+  // Additional container to spin up when Function is active
+  optional uint32 buffer_containers = 3;
+  // Currently unused; a placeholder in case we decide to expose scaleup control to users
+  optional uint32 scaleup_window = 4;
+  // Maximum amount of time a container can be idle before being scaled down, in seconds; pka "container_idle_timeout"
+  optional uint32 scaledown_window = 5;
+  reserved 6;
 }
 
 // Used for flash autoscaling
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -632,7 +632,7 @@ message AutoscalerSettings {
   // A collection of user-configurable settings for Function autoscaling
   // These are used for static configuration and for dynamic autoscaler updates
 
-  // Minimum containers when scale-to-zero is not deisired; pka "keep_warm" or "warm_pool_size"
+  // Minimum containers when scale-to-zero is not desired; pka "keep_warm" or "warm_pool_size"
   optional uint32 min_containers = 1;
   // Limit on the number of containers that can be running for each Function; pka "concurrency_limit"
   optional uint32 max_containers = 2;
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -628,11 +628,32 @@ message AuthTokenGetResponse {
...
+// As well as the different sources that were used to create it.
+message AutoscalerConfiguration {
+  // Used for capturing context about an action performed by a user
+  message UserActionInfo {
+    string user_id = 1;
+    double timestamp = 2;
+  }
+
+  // The settings that are currently in effect.
+  modal.client.AutoscalerSettings settings = 1;
```
mwaskom: Minor but we can remove the namespacing now I think (i.e. just reference `AutoscalerSettings` directly

modal_proto/api.proto
```diff
@@ -628,11 +628,32 @@ message AuthTokenGetResponse {
   string token = 1;
 }
 
+// Message representing the current (coalesced) state of the autoscaler configuration
+// As well as the different sources that were used to create it.
+message AutoscalerConfiguration {
+  // Used for capturing context about an action performed by a user
+  message UserActionInfo {
```
mwaskom: Should this be in global scope? I think the original idea was to have a general message to track "who did what when" that we could reuse elsewhere (although this is the only place we currently use it).

modal_proto/api.proto
```diff
@@ -628,11 +628,32 @@ message AuthTokenGetResponse {
...
+  // The settings that are currently in effect.
+  modal.client.AutoscalerSettings settings = 1;
+  // For tracking the source of the overridden value; keys correspond to fields in `settings`.
+  map<string, UserActionInfo> override_events = 2;
+  // The default settings that are used when no static settings are provided and no overrides are in effect.
+  modal.client.AutoscalerSettings default_settings = 3;
+  // The static settings that were used to initialize the configuration.
+  modal.client.AutoscalerSettings static_settings = 4;
+  // The merge of all overrides that were used to create the current configuration.
+  modal.client.AutoscalerSettings override_settings = 5;
```
mwaskom: How is this different from the `settings` field?

modal_proto/api.proto
```diff
@@ -628,11 +628,32 @@ message AuthTokenGetResponse {
...
+  }
+
+  // The settings that are currently in effect.
+  modal.client.AutoscalerSettings settings = 1;
+  // For tracking the source of the overridden value; keys correspond to fields in `settings`.
+  map<string, UserActionInfo> override_events = 2;
+  // The default settings that are used when no static settings are provided and no overrides are in effect.
+  modal.client.AutoscalerSettings default_settings = 3;
+  // The static settings that were used to initialize the configuration.
+  modal.client.AutoscalerSettings static_settings = 4;
```
mwaskom: Curious how you're thinking we'll use these?

modal_proto/api.proto
kareem-modal: yeah i moved it there cause it wasn't used anywhere else, but can make it global for future sharing

modal_proto/api.proto
kareem-modal: this will hold **only** the overrides, so any static values and any defaults will not be present.

`default: { task_idle_timeout: 60 }`
`static: { min_containers: 10 }`
`override: { max_containers: 20 }`
`settings: { min_containers: 10, max_containers: 20, task_idle_timeout: 60 }`

modal_proto/api.proto
kareem-modal: we want to show in the UI the following 

- the current settings
- for each setting, where it comes from (modal default, override, static def'n) 
- if you edit (say clear) we want to be able to show what the effective setting will be (clearing will revert to default)

modal_proto/api.proto
mwaskom: Oh I see. Is there some redundancy though? I believe (check me) that the presence of a field in the `override_events` map indicates that that field is "in an override state"

modal_proto/api.proto
mwaskom: > for each setting, where it comes from (modal default, override, static def'n)

IMO it's not super important to distinguish between "modal default" and "static definition". I could be wrong about that but it feels like it's potentially overloading the user with information. (One relevant point that honestly could go in either direction is that for fields like `CPU request`  we just say "Default" instead of what the actual value is).

> clearing will revert to default

Hm I think that if you want to have a "clear" operation in the UI, it should reset back to any static configuration if present (otherwise to the server default). That said it might be sufficient to just allow editing these values (without a concept of clear/reset). We have an edit / update API in the SDK and nobody has asked for a "reset" operation AFAIK. Though I could see how it's useful!

Food for thought: if someone "clears" the values in the UI, do you attribute the current state to the default / static configuration, or to the person who hit the "clear" button? I kind of think it's more sensible to to consider all changes made outside of code+deployment as "overrides" (or "updates") so that you don't have to draw such distinctions.

modal_proto/api.proto
kareem-modal: no you're right, the presence of the field in the `override_events` map indicates that fields is in an override state. i chose to add a bit of redundancy for consistency (all 3 are just settings, no special logic for overrides) and ergonomics (direct field access vs checking a map). also the settings object provides better type safety than string keys.

override_events is good for telling who did the override, but not determining the overridden value. i also think we'll need to evolve it to provide a meaningful override history, by providing *all* the overrides in effect, in priority order, so we know the effect of expiring/deleting one once we have TTLs

modal_proto/api.proto
kareem-modal: > IMO it's not super important to distinguish between "modal default" and "static definition". I could be wrong about that but it feels like it's potentially overloading the user with information. (One relevant point that honestly could go in either direction is that for fields like CPU request we just say "Default" instead of what the actual value is).

cc @aksh-at who suggested we should show modal default vs user-defined. i see your point, but haven't interacted with enough modal customers to have a confident opinion

> Hm I think that if you want to have a "clear" operation in the UI, it should reset back to any static configuration if present (otherwise to the server default). That said it might be sufficient to just allow editing these values (without a concept of clear/reset). We have an edit / update API in the SDK and nobody has asked for a "reset" operation AFAIK. Though I could see how it's useful!

i figured because we *stack* overrides, and plan to support TTL, it will be necessary to know the stack of overrides. so another way of thinking of "clear" is "after this override expires what happens" - even without an explicit "clear" button. but yeah, we could just clear all overrides i suppose, that would simplify the UI and avoid the need to get all the overrides in effect.

> Food for thought: if someone "clears" the values in the UI, do you attribute the current state to the default / static configuration, or to the person who hit the "clear" button? I kind of think it's more sensible to to consider all changes made outside of code+deployment as "overrides" (or "updates") so that you don't have to draw such distinctions.

we could avoid clear with an additional override / update, but if we have TTLs it seems natural to "expire" an override you don't want anymore versus overriding it with the underlying static / default value.

mwaskom: > i figured because we stack overrides, and plan to support TTL, it will be necessary to know the stack of overrides. so another way of thinking of "clear" is "after this override expires what happens" 

My intuition here is that a "reset" (or "clear") button in the UI should reset the state to what it was at deployment (modal default + static config) but that a TTL expiring should reset the state to the moment the override with the TTL went into effect. I'm trying to decide if that's inconsistent. I'm pretty sure you _don't_ want the "reset" / "clear" interaction to walk backwards through the entire set of overrides though.

kareem-modal: that makes sense. to be clear, we're not introducing reset/clear as a part of this work, only creating new overrides. you can ignore that third bullet, really i'm bringing up things that are outside the current scope. the primary purpose of override_settings at this point is to show an overridden value.

kareem-modal:
> Don't want to be blocking here. Worried this is a little overcomplicated but if we end up not using the extra fields we can remove them later.

agreed, thanks.

PR #3562, opened 2025-09-16
Adds cloud bucket mounts to with_options

thomasjpfan:
## Describe your changes

- Adds cloud bucket mounts to `with_options`.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- `Cls.with_options` supports `CloudBucketMount` in `volumes`.

```python
cloud_bucket = modal.CloudBucketMount("my-bucket", secret=aws_secret)
MyClsCloud = modal.Cls.from_name("my-app", "MyCls").with_options(
        volumes={"/mnt": cloud_bucket}
)
```

### DIFF
--- a/modal/cloud_bucket_mount.py
+++ b/modal/cloud_bucket_mount.py
@@ -1,6 +1,6 @@
 # Copyright Modal Labs 2022
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, Sequence
 from urllib.parse import urlparse
 
 from modal_proto import api_pb2
@@ -118,8 +118,13 @@ class _CloudBucketMount:
     read_only: bool = False
     requester_pays: bool = False
 
+    @property
+    def is_hydrated(self):
+        # Cloud bucket mounts do not have a server side representation.
+        return True
 
-def cloud_bucket_mounts_to_proto(mounts: list[tuple[str, _CloudBucketMount]]) -> list[api_pb2.CloudBucketMount]:
+
+def cloud_bucket_mounts_to_proto(mounts: Sequence[tuple[str, _CloudBucketMount]]) -> list[api_pb2.CloudBucketMount]:
     """Helper function to convert `CloudBucketMount` to a list of protobufs that can be passed to the server."""
     cloud_bucket_mounts: list[api_pb2.CloudBucketMount] = []
 
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -1,9 +1,9 @@
 # Copyright Modal Labs 2022
 import dataclasses
 import inspect
-import os
 import typing
 from collections.abc import Collection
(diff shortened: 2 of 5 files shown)


modal/cloud_bucket_mount.py
```diff
@@ -118,8 +118,13 @@ def f():
     read_only: bool = False
     requester_pays: bool = False
 
+    @property
+    def is_hydrated(self):
+        # Cloud bucket mounts do not have a server side representation.
+        return True
```
mwaskom: 🥴

modal/cloud_bucket_mount.py
```diff
@@ -118,10 +118,8 @@ def f():
     read_only: bool = False
     requester_pays: bool = False
 
-    @property
-    def is_hydrated(self):
-        # Cloud bucket mounts do not have a server side representation.
-        return True
+    # Cloud bucket mounts do not have a server side object_id.
+    is_hydrated: ClassVar[bool] = True
```
mwaskom, line 122: Would it make more sense to change the logic where we're reading this, e.g. to only check `if_hydrated` if the object is a subclass of `modal._Object`, or if it has an `is_hydrated` attribute or something?

modal/cls.py
```diff
@@ -751,7 +751,7 @@ async def _load_from_base(new_cls, resolver, existing_object_id):
             # But the (instance) service function for each Obj will be different
             # since it will rebind to whatever `_options` have been assigned on
             # the particular Cls parent
-            if not self.is_hydrated:
+            if isinstance(self, _Object) and not self.is_hydrated:
```
mwaskom, line 754: 😵‍💫  how are we ending up with `CloudBucketMount` instance as `self` inside a `Cls` method?

thomasjpfan: Ah this is not needed.

mwaskom:
Yes that seems wrong.

thomasjpfan:
I adjusted the logic in the backend: https://github.com/modal-labs/modal/pull/27151. If `cloud_bucket_mounts` is provided and they have secrets, we always extend the secrets.

thomasjpfan:
I think is different but consistent.

`MyCls.with_options(secrets=[new_secret], volumes={"/mnt", modal.CloudBucketMount})  # extends decorator secrets`

The above would replace the `cls`'s secrets with `[new_secret]` and then extend it with the cloud bucket mount secret. The behavior is:

1. Passing `secrets=[]` will **always** replace.
2. Passing a secret through cloud bucket mount will **always** extend.

mwaskom:
Got it. That makes sense. I'd describe the behavior as "replace and then extend".

mwaskom (approved):
LGTM

PR #3565, opened 2025-09-17
Fix input plane attempt retry handling.

danielshaar:
## Describe your changes

We have a bug in which we create excessive input plane attempts due to accidentally submitting AttemptRetry on every AttemptAwait timeout of 55s. We haven't caught it because most input plane calls take less than a minute. This will be patched server-side.

Also fixes a separate bug where internal failures can eat into client specified retries.

SVC-863

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -495,7 +495,7 @@ class _InputPlaneInvocation:
                     # We immediately retry internal failures and the failure doesn't count towards the retry policy.
                     self.attempt_token = await self._retry_input(metadata)
                     continue
-            elif delay_ms := user_retry_manager.get_delay_ms():
+            elif (delay_ms := user_retry_manager.get_delay_ms()) is not None:
                 # We still have user retries left, so sleep and retry.
                 await asyncio.sleep(delay_ms / 1000)
                 self.attempt_token = await self._retry_input(metadata)
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -477,34 +477,34 @@ class _InputPlaneInvocation:
                 metadata=metadata,
             )
 
-            if await_response.HasField("output"):
-                if await_response.output.result.status in TERMINAL_STATUSES:
-                    return await _process_result(
-                        await_response.output.result, await_response.output.data_format, self.client.stub, self.client
-                    )
+            # Keep awaiting until we get an output.
+            if not await_response.HasField("output"):
+                continue
 
-                if await_response.output.result.status == api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE:
-                    internal_failure_count += 1
-                    # Limit the number of times we retry
-                    if internal_failure_count < MAX_INTERNAL_FAILURE_COUNT:
-                        # For system failures on the server, we retry immediately,
-                        # and the failure does not count towards the retry policy.
-                        self.attempt_token = await self._retry_input(metadata)
-                        continue
+            # If we have a final output, return.
+            if await_response.output.result.status in TERMINAL_STATUSES:
+                return await _process_result(
+                    await_response.output.result, await_response.output.data_format, self.client.stub, self.client
+                )
 
-                # We add delays between retries for non-internal failures.
-                delay_ms = user_retry_manager.get_delay_ms()
-                if delay_ms is None:
-                    # No more retries either because we reached the retry limit or user didn't set a retry policy
-                    # and the limit defaulted to 0.
-                    # An unsuccessful status should raise an error when it's converted to an exception.
-                    # Note: Blob download is done on the control plane stub not the input plane stub!
-                    return await _process_result(
-                        await_response.output.result, await_response.output.data_format, self.client.stub, self.client
-                    )
+            # We have a failure (internal or application), so see if there are any retries left, and if so, retry.
+            if await_response.output.result.status == api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE:
+                internal_failure_count += 1
+                # Limit the number of times we retry internal failures.
+                if internal_failure_count < MAX_INTERNAL_FAILURE_COUNT:
+                    # We immediately retry internal failures and the failure doesn't count towards the retry policy.
+                    self.attempt_token = await self._retry_input(metadata)
+                    continue
+            elif delay_ms := user_retry_manager.get_delay_ms():
+                # We still have user retries left, so sleep and retry.
                 await asyncio.sleep(delay_ms / 1000)
+                self.attempt_token = await self._retry_input(metadata)
+                continue
 
-            await self._retry_input(metadata)
+            # No more retries left.
+            return await _process_result(
+                await_response.output.result, await_response.output.data_format, self.client.stub, self.client
+            )
 
     async def _retry_input(self, metadata: list[tuple[str, str]]) -> str:
         retry_request = api_pb2.AttemptRetryRequest(
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -513,7 +513,6 @@ class _InputPlaneInvocation:
             input=self.input_item,
             attempt_token=self.attempt_token,
         )
-        # TODO(ryan): Add exponential backoff?
         retry_response = await retry_transient_errors(
             self.stub.AttemptRetry,
             retry_request,
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -316,12 +316,18 @@ class MockClientServicer(api_grpc.ModalClientBase):
         self.function_get_server_warnings = None
         self.resp_jitter_secs: float = 0.0
         self.port = port
-        # AttemptAwait will return a failure until this is 0. It is decremented by 1 each time AttemptAwait is called.
-        self.attempt_await_failures_remaining = 0
+        # Set a list of custom, fixed responses for the AttemptAwait RPC.
+        # This mock server will return responses in order. The regular behavior will resume once this list is exhausted.
+        self.attempt_await_responses: list[api_pb2.AttemptAwaitResponse] = []
         # Value returned by AuthTokenGet
         self.auth_token = jwt.encode({"exp": int(time.time()) + 3600}, "my-secret-key", algorithm="HS256")
         self.auth_tokens_generated = 0
         self.function_id_to_definition_id: dict[str, str] = {}
+        # Number of times AttemptAwait was called.
+        self.attempt_await_count = 0
+        # Number of times the user's function was called.
+        self.function_call_count = 0
+        self.function_call_result: Any = None
 
         @self.function_body
         def default_function_body(*args, **kwargs):
@@ -2319,28 +2325,29 @@ class MockClientServicer(api_grpc.ModalClientBase):
         )
 
     async def AttemptAwait(self, stream):
+        self.attempt_await_count += 1
         # TODO(dxia): implement attempt token logic
 
-        # To test client retries for internal failures, tests can configure outputs to fail some number of times.
+        if len(self.attempt_await_responses) > 0:
+            await stream.send_message(self.attempt_await_responses.pop(0))
+            return
+
         status = api_pb2.GenericResult.GENERIC_STATUS_SUCCESS
-        if self.attempt_await_failures_remaining > 0:
-            status = api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE
-            self.attempt_await_failures_remaining = self.attempt_await_failures_remaining - 1
 
         if self.function_call_inputs:
             function_call_id = f"fc-{self.fcidx}"
             (idx, input_id, retry_count), (args, kwargs) = self.function_call_inputs.pop(function_call_id)[0]
             self.fcidx -= 1
             try:
-                res = self._function_body(*args, **kwargs)
+                self.function_call_count += 1
+                self.function_call_result = self._function_body(*args, **kwargs)
             except Exception as e:
-                res = e
+                self.function_call_result = e
                 status = api_pb2.GenericResult.GENERIC_STATUS_FAILURE
         else:
             input_id = "in-1"
             idx = 0
             retry_count = 0
-            res = "attempt_await_bogus_response"
 
         await stream.send_message(
             api_pb2.AttemptAwaitResponse(
@@ -2349,7 +2356,7 @@ class MockClientServicer(api_grpc.ModalClientBase):
                     idx=idx,
                     result=api_pb2.GenericResult(
                         status=status,
-                        data=serialize_data_format(res, api_pb2.DATA_FORMAT_PICKLE),
+                        data=serialize_data_format(self.function_call_result, api_pb2.DATA_FORMAT_PICKLE),
                     ),
                     data_format=api_pb2.DATA_FORMAT_PICKLE,
                     retry_count=retry_count,
--- a/test/function_test.py
+++ b/test/function_test.py
@@ -6,20 +6,30 @@ import os
 import pytest
 import time
 import typing
-from contextlib import contextmanager
+from contextlib import contextmanager, nullcontext
 
 from grpclib import Status
 from synchronicity.exceptions import UserCodeException
 
 import modal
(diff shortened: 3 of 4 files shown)


modal/_functions.py
```diff
@@ -479,34 +479,34 @@ async def run_function(self) -> Any:
...
-                if await_response.output.result.status == api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE:
-                    internal_failure_count += 1
-                    # Limit the number of times we retry
-                    if internal_failure_count < MAX_INTERNAL_FAILURE_COUNT:
-                        # For system failures on the server, we retry immediately,
-                        # and the failure does not count towards the retry policy.
-                        self.attempt_token = await self._retry_input(metadata)
-                        continue
+            # Keep awaiting until we get an output.
+            if not await_response.HasField("output"):
```
thomasjpfan, line 481: +100 on reducing the indentation

modal/_functions.py
```diff
@@ -479,34 +479,34 @@ async def run_function(self) -> Any:
...
+                internal_failure_count += 1
+                # Limit the number of times we retry internal failures.
+                if internal_failure_count < MAX_INTERNAL_FAILURE_COUNT:
+                    # We immediately retry internal failures and the failure doesn't count towards the retry policy.
+                    self.attempt_token = await self._retry_input(metadata)
+                    continue
+            elif delay_ms := user_retry_manager.get_delay_ms():
+                # We still have user retries left, so sleep and retry.
                 await asyncio.sleep(delay_ms / 1000)
+                self.attempt_token = await self._retry_input(metadata)
```
thomasjpfan, line 501: Is the core of this fix to make sure we store the attempt_token when there are retries left?

modal/_functions.py
```diff
@@ -479,34 +479,34 @@ async def run_function(self) -> Any:
...
-                    )
+            # We have a failure (internal or application), so see if there are any retries left, and if so, retry.
+            if await_response.output.result.status == api_pb2.GenericResult.GENERIC_STATUS_INTERNAL_FAILURE:
+                internal_failure_count += 1
+                # Limit the number of times we retry internal failures.
+                if internal_failure_count < MAX_INTERNAL_FAILURE_COUNT:
+                    # We immediately retry internal failures and the failure doesn't count towards the retry policy.
+                    self.attempt_token = await self._retry_input(metadata)
+                    continue
+            elif delay_ms := user_retry_manager.get_delay_ms():
```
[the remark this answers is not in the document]
davidxia: This is intentional.

modal/_functions.py
danielshaar: the core of the fix is not retrying every single time we exceed attemptawait deadline and there is no output (the last line of the function was not indented enough). But I'm also restructuring the code to make it impossible to have that issue

modal/_functions.py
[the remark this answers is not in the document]
danielshaar: hmm, good point, let me double check the code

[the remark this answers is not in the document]
danielshaar: fixed

modal/_functions.py
rohansingh: -4 on reducing the indentation :trollface:

modal/_functions.py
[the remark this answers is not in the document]
davidxia: @danielshaar Oh wow, nice catch on treating 0ms as retry immediately instead of don't retry.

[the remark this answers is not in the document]
danielshaar: I somehow totally missed your "this is intentional" comment... My good point response was to cursor lol

thomasjpfan (approved):

test/conftest.py
```diff
@@ -316,12 +316,17 @@ def __init__(self, blob_host, blobs, blocks, files_sha2data, credentials, port):
         self.function_get_server_warnings = None
         self.resp_jitter_secs: float = 0.0
         self.port = port
-        # AttemptAwait will return a failure until this is 0. It is decremented by 1 each time AttemptAwait is called.
-        self.attempt_await_failures_remaining = 0
```
davidxia, line 320: I refactored unit tests to not require this attribute.

test/function_test.py
```diff
@@ -91,7 +322,7 @@ def test_map(client, servicer, slow_put_inputs):
 
 @pytest.mark.parametrize("slow_put_inputs", [False, True])
 @pytest.mark.timeout(120)
-def test_map_inputplane(client, servicer, slow_put_inputs):
+def test_map_input_plane(client, servicer, slow_put_inputs):
```
davidxia, line 295: This change and below are unrelated and simply fix camel-casing of "input plane" to be consistent.

test/input_plane_test.py
```diff
@@ -35,25 +32,3 @@ def test_lookup_foo(client: Client, servicer: MockClientServicer):
...
-    servicer.attempt_await_failures_remaining = 1
-    servicer.function_body(foo.get_raw_f())
-    with app.run(client=client):
-        assert foo.remote() == "foo"
-    # We don't have a great way to verify the call was actually retried. We can at least check that the servicer
-    # decremented the attempts_to_fail counter, which indicates that the call was retried.
-    assert servicer.attempt_await_failures_remaining == 0
-
-
-def test_retry_limit_on_internal_error(client: Client, servicer: MockClientServicer, monkeypatch):
```
davidxia, line 51: [covered here](https://github.com/modal-labs/modal-client/pull/3565/files#diff-be2c0fead75c30eb0a406aefb59b16eccd29934c579d1744ee666988c797b878R141-R142)

test/input_plane_test.py
```diff
@@ -35,25 +32,3 @@ def test_lookup_foo(client: Client, servicer: MockClientServicer):
     assert f.remote() == "foo"
     assert f._get_metadata().input_plane_url is not None
     assert f._get_metadata().input_plane_region == "us-east"
-
-
-def test_retry_on_internal_error(client: Client, servicer: MockClientServicer):
```
davidxia, line 40: [covered here](https://github.com/modal-labs/modal-client/pull/3565/files#diff-be2c0fead75c30eb0a406aefb59b16eccd29934c579d1744ee666988c797b878R124-R125)

davidxia:
I manually tested many of the unit test cases against prod. The client behaved as I expected.

test/function_test.py
```diff
@@ -45,13 +56,233 @@ def dummy():
...
+            # Currently MockClientServicer returns this string instead of the function return value
+            # when the function call takes longer than function_utils.OUTPUTS_TIMEOUT to run.
+            nullcontext("attempt_await_bogus_response"),
+            2,
+            1,
+            id="long running",
+        ),
+    ],
+)
+def test_remote_input_plane(
```
thomasjpfan, line 233: Can we reduce the duration of some of these tests?

```
60.53s call     test/function_test.py::test_remote_input_plane[long running]
7.43s call     test/function_test.py::test_remote_input_plane[honor user retry policy]
7.41s call     test/function_test.py::test_remote_input_plane[internal failure does not use up user retries]
```

davidxia: @thomasjpfan done in 254ce51a

```console
❯ pytest test/function_test.py::test_remote_input_plane --durations 0

...
================================================================================ slowest durations ================================================================================
2.65s call     test/function_test.py::test_remote_input_plane[long running]
1.56s call     test/function_test.py::test_remote_input_plane[internal failure does not use up user retries]
1.52s call     test/function_test.py::test_remote_input_plane[immediate retry]
1.49s call     test/function_test.py::test_remote_input_plane[honor user retry policy]
0.55s call     test/function_test.py::test_remote_input_plane[idle timeout]
0.55s call     test/function_test.py::test_remote_input_plane[internal failure max]
0.55s call     test/function_test.py::test_remote_input_plane[timeout]
0.53s call     test/function_test.py::test_remote_input_plane[terminated]
0.52s call     test/function_test.py::test_remote_input_plane[failure]
0.52s call     test/function_test.py::test_remote_input_plane[success]
0.50s call     test/function_test.py::test_remote_input_plane[internal failure]
0.50s call     test/function_test.py::test_remote_input_plane[init failure]
0.50s call     test/function_test.py::test_remote_input_plane[no output]
0.50s call     test/function_test.py::test_remote_input_plane[unspecified]
0.01s setup    test/function_test.py::test_remote_input_plane[success]
0.01s setup    test/function_test.py::test_remote_input_plane[no output]
0.01s setup    test/function_test.py::test_remote_input_plane[unspecified]
0.01s setup    test/function_test.py::test_remote_input_plane[terminated]
0.01s setup    test/function_test.py::test_remote_input_plane[internal failure]
```

davidxia (approved):
@danielshaar's changes LGTM. Someone should review my unit tests.

test/function_test.py
```diff
@@ -45,13 +55,253 @@ def dummy():
...
+
+@app.function(
+    experimental_options={"input_plane_region": "us-east"},
+    retries=modal.Retries(max_retries=1, initial_delay=0, backoff_coefficient=1.0),
+)
+def input_plane_func_with_immediate_retry():
+    raise ValueError()
+
+
+# We set this to more than the outputs_timeout_overrride of the "long running" test_remote_input_plane() test case.
```
rculbertson: minor nit: `outputs_timeout_overrride` should be `outputs_timeout_override` in a few spots :)

davidxia: thanks, fixed!

test/conftest.py
```diff
@@ -2312,19 +2317,22 @@ async def AttemptStart(self, stream):
...
 
     async def AttemptAwait(self, stream):
+        self.attempt_await_count += 1
         # TODO(dxia): implement attempt token logic
 
+        if len(self.attempt_await_responses) > 0:
+            await stream.send_message(self.attempt_await_responses.pop(0))
+            return
+
         # To test client retries for internal failures, tests can configure outputs to fail some number of times.
```
rculbertson: I think this comment can be removed, since `attempt_await_failures_remaining` was deleted?

test/function_test.py
```diff
@@ -45,13 +55,253 @@ def dummy():
...
+            input_plane_func_long_running,
+            [],
+            1,
+            # Currently MockClientServicer returns this string instead of the function return value
+            # when the function call takes longer than function_utils.OUTPUTS_TIMEOUT to run.
+            nullcontext("attempt_await_bogus_response"),
+            2,
+            1,
+            id="long running",
+        ),
```
rculbertson, line 230: Ah I see. So what's happening is we make one call to attempt_await which takes a long time because the function sleeps. While the function is running, the call from the client times out, and the client calls attempt_await again. This second call returns immediately with `attempt_await_bogus_response`, as the first call continues to run. The main assertion we care about here is that `attempt_await_count` is 2, because that proves that the call timed out, and we called it again. Does that sound right?

I wish we had a better way to simulate this in the dummy server. I can't think of a good way to do this without bigger changes. Maybe we could rename `attempt_await_bogus_response` to something more meaningful for now, and leave a comment explaining how it's a special value that should only get returned in this one test case?

rculbertson: Also, (something for a separate PR) this might be a good use case to have an integration test for, since simulating this in the dummy server is a little tricky.

test/conftest.py
davidxia: yes, removed

test/function_test.py
davidxia: > The main assertion we care about here is that attempt_await_count is 2, because that proves that the call timed out, and we called it again. Does that sound right?

yes, and to verify the 1 here that means the function itself was only called once.

> Maybe we could rename `attempt_await_bogus_response` to something more meaningful for now, and leave a comment...

sounds good, open to suggestions here

davidxia: updated in [b0a7cbb](https://github.com/modal-labs/modal-client/pull/3565/commits/b0a7cbb1c698a41e6e32cb6bfd2ad99dd97a750f), lmk what you think

davidxia:
@cursor review

test/conftest.py
rculbertson: Nice, this approach is much more flexible.

test/function_test.py
rculbertson: Oh cool, nice that the actual result gets returned now. I think it would be nice to get rid of this entire `else` block, but I can try to do that in a separate PR.

```
        else:
            input_id = "in-1"	            input_id = "in-1"
            idx = 0	            idx = 0
            retry_count = 0	            retry_count = 0
```

rculbertson (approved):

test/function_test.py
```diff
@@ -45,13 +55,253 @@ def dummy():
...
     assert len(servicer.cleared_function_calls) == 0
     with app.run(client=client):
         assert foo.remote(2, 4) == 20
         assert len(servicer.cleared_function_calls) == 1
 
 
+@pytest.mark.parametrize(
+    "func, attempt_await_responses, outputs_timeout_override, expectation, attempt_await_count, function_call_count",
+    [
+        # If the function runs sucessfully the first time, the client should call the
```
rculbertson:
```suggestion
        # If the function runs successfully the first time, the client should call the
```

test/function_test.py
rculbertson: [Here's a PR](https://github.com/modal-labs/modal-client/pull/3581) which fixes up the dummy server some more. It adds proper handling of attempt_tokens, and changes how we process inputs/outputs so we can handle these cases a little more cleanly.

test/function_test.py
```diff
@@ -45,13 +55,253 @@ def dummy():
...
+@app.function(
+    experimental_options={"input_plane_region": "us-east"},
+    retries=modal.Retries(max_retries=1, initial_delay=0, backoff_coefficient=1.0),
+)
+def input_plane_func_with_immediate_retry():
+    raise ValueError()
+
+
+# We set this to more than the outputs_timeout_override of the "long running" test_remote_input_plane() test case.
+duration = 2
```
danielshaar: is this duration used anywhere besides the function below? If not, I'd just put the comment in there. If yes, then I'd make this a constant that's more specificly named

test/function_test.py
```diff
@@ -45,13 +55,253 @@ def dummy():
...
+    "func, attempt_await_responses, outputs_timeout_override, expectation, attempt_await_count, function_call_count",
+    [
+        # If the function runs successfully the first time, the client should call the
+        # AttemptAwait RPC once and the user's function should be called once.
+        pytest.param(input_plane_func, [], None, nullcontext("DEADBEEF"), 1, 1, id="success"),
+        # The client should call the AttemptAwait RPC until it receives an AttemptAwaitResponse
+        # with an output attribute and then return a successful AttemptAwaitResponse.
+        pytest.param(
+            input_plane_func,
+            [api_pb2.AttemptAwaitResponse(), api_pb2.AttemptAwaitResponse()],
```
danielshaar: [api_pb2.AttemptAwaitResponse()] * 2

test/function_test.py
```diff
@@ -45,13 +55,253 @@ def dummy():
...
+            None,
+            nullcontext("DEADBEEF"),
+            3,
+            1,
+            id="no output",
+        ),
+        # The client should raise a RemoteError if the AttemptAwaitResponse has a GENERIC_STATUS_TERMINATED status.
+        pytest.param(
+            input_plane_func,
+            [
```
danielshaar: maybe worth making a helper that creates these response lists to keep the parameterizations easier to read (fewer overall lines)

danielshaar:
LGTM aside from nits

test/function_test.py
davidxia: done in [d775ad9](https://github.com/modal-labs/modal-client/commit/d775ad9fcbacb6f33fec32ac38a265e6c1826728)

test/function_test.py
davidxia: done in [d775ad9](https://github.com/modal-labs/modal-client/commit/d775ad9fcbacb6f33fec32ac38a265e6c1826728)

test/function_test.py
davidxia: done in [d775ad9](https://github.com/modal-labs/modal-client/commit/d775ad9fcbacb6f33fec32ac38a265e6c1826728)

PR #3456, opened 2025-08-12
Ensure client honors user's retry policy for `.remote()` when inputs fail

davidxia:
contributes to SVC-602

<details> <summary>Checklists</summary>

---

## Compatibility checklist

Check these boxes or delete any item (or this section) if not relevant for this PR.

- [x] Client+Server: this change is compatible with old servers
- [x] Client forward compatibility: this change ensures client can accept data intended for later versions of itself

Note on protobuf: protobuf message changes in one place may have impact to
multiple entities (client, server, worker, database). See points above.

</details>

PR #3566, opened 2025-09-18
Enable C901 ruff linting rule

davidxia:
> Checks for functions with a high McCabe complexity

https://docs.astral.sh/ruff/rules/complex-structure/

Motivated by bugs like the one introduced in #3456 and fixed
in #3565.

Ignore existing Ruff C901 violations.


mwaskom:
I don't have strong feelings. Might come down on either side of the annoying / helpful boundary :D

PR #3568, opened 2025-09-18
Patch options_pb2 based on protobuf version

thomasjpfan:
## Describe your changes

- Fixes https://linear.app/modal-labs/issue/SDK-38/modal-import-fails-on-some-systems-with-protobuf-5

I tested using:

```python
import modal

image = modal.Image.debian_slim().env({"PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION": "python"})
app = modal.App(image=image)


@app.function()
def check_protobuf():
    import google.protobuf as pb

    print(pb.__version__)
    return pb.__version__
```

Which works with this PR and `MODAL_SYNC_ENTRYPOINT=1`

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Improves compatibility with protobuf >= 5.

### DIFF
--- a/protoc_plugin/options_pb2.py.patch
+++ b/protoc_plugin/options_pb2.py.patch
@@ -1,0 +1,30 @@
+import google.protobuf
+try:
+  pb_older_than_5 = int(google.protobuf.__version__.split(".")[0]) < 5
+except Exception:
+  pb_older_than_5 = False
+
+if pb_older_than_5:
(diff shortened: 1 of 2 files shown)


mwaskom:
Can we document in the source file that we're doing this too? I'm concerned this hack is going to stick around for a while and then someone can add some new options that are silently ignored because of the patch you're applying.

PR #3570, opened 2025-09-18
Add get_tags method to Sandbox

ehdr:
## Describe your changes

This allows retrieve tags previously set on a Sandbox using `.set_tags()`.

Depends on modal-labs/modal-client#3569 and modal-labs/modal#27283.

Partly resolves SDK-457.

---

## Compatibility checklist

</details>

## Changelog

- Added a `.get_tags()` method to Sandbox, enabling fetching tags that were previously set using `.set_tags()`.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -574,12 +574,14 @@ class _Sandbox(_Object, type_prefix="sb"):
 
     async def set_tags(self, tags: dict[str, str], *, client: Optional[_Client] = None) -> None:
         """Set tags (key-value pairs) on the Sandbox. Tags can be used to filter results in `Sandbox.list`."""
+        environment_name = _get_environment_name()
         if client is None:
             client = await _Client.from_env()
 
         tags_list = [api_pb2.SandboxTag(tag_name=name, tag_value=value) for name, value in tags.items()]
 
         req = api_pb2.SandboxTagsSetRequest(
+            environment_name=environment_name,
             sandbox_id=self.object_id,
             tags=tags_list,
         )
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -564,11 +564,7 @@ class _Sandbox(_Object, type_prefix="sb"):
 
     async def get_tags(self):
         """Fetches any tags (key-value pairs) attached to this Sandbox from the server."""
-        environment_name = _get_environment_name()
-        req = api_pb2.SandboxTagsGetRequest(
-            sandbox_id=self.object_id,
-            environment_name=environment_name,
-        )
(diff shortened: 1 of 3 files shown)


modal/sandbox.py
```diff
@@ -562,6 +562,23 @@ async def from_id(sandbox_id: str, client: Optional[_Client] = None) -> "_Sandbo
 
         return obj
 
+    async def get_tags(self, client: Optional[_Client] = None):
```
mwaskom, line 565: Regarding accepting a `client`: I see you're copying what's in `set_tags` below, but shouldn't a hydrated object already have a `client` attached and use that?

modal/sandbox.py
```diff
@@ -562,6 +562,23 @@ async def from_id(sandbox_id: str, client: Optional[_Client] = None) -> "_Sandbo
...
+        """Fetches any tags (key-value pairs) attached to this Sandbox from the server."""
```
mwaskom, line 566:
```suggestion
        """Fetches any tags (key-value pairs) currently attached to this Sandbox from the server."""
```

Tiny suggestion that makes this slightly more accurate and also helps answer any "why does this have to fetch from the server" questions that it might raise.

mwaskom: "My 'This method does IO' shirt is raising a lot of questions already answered by my shirt'"

modal/sandbox.py
```diff
@@ -562,6 +562,23 @@ async def from_id(sandbox_id: str, client: Optional[_Client] = None) -> "_Sandbo
...
+        environment_name = _get_environment_name()
```
mwaskom, line 567: 😖 I know this is also mostly following what we already have but the environment scoping for sandbox tags seems super broken. The "current environment" in the location you call `.get_tags()` should not factor into what tags you retrieve for an existing Sandbox.

modal/sandbox.py
```diff
@@ -562,16 +562,24 @@ async def from_id(sandbox_id: str, client: Optional[_Client] = None) -> "_Sandbo
...
+        except GRPCError as exc:
+            raise InvalidError(exc.message) if exc.status == Status.INVALID_ARGUMENT else exc
+
+        return {tag.tag_name: tag.tag_value for tag in resp.tags}
+
     async def set_tags(self, tags: dict[str, str], *, client: Optional[_Client] = None):
         """Set tags (key-value pairs) on the Sandbox. Tags can be used to filter results in `Sandbox.list`."""
-        environment_name = _get_environment_name()
         if client is None:
             client = await _Client.from_env()
```
mwaskom, line 578: Can we make a small ticket to follow up on deprecating client and using the attached one here?

ehdr: 👍 #3572

PR #3572, opened 2025-09-19
Deprecate `client` in Sandbox `set_tags()`

ehdr:

---

## Compatibility checklist

</details>

## Changelog

- Deprecated the `client` parameter to `Sandbox.set_tags()`, and the `environment_name` parameter to `Sandbox.from_name()`.

### DIFF
--- a/modal/sandbox.py
+++ b/modal/sandbox.py
@@ -575,8 +575,12 @@ class _Sandbox(_Object, type_prefix="sb"):
     async def set_tags(self, tags: dict[str, str], *, client: Optional[_Client] = None) -> None:
         """Set tags (key-value pairs) on the Sandbox. Tags can be used to filter results in `Sandbox.list`."""
         environment_name = _get_environment_name()
-        if client is None:
-            client = await _Client.from_env()
+        if client is not None:
+            deprecation_warning(
+                (2025, 9, 18),
+                "The `client` parameter is deprecated. Set `client` when creating the Sandbox instead "
+                "(in e.g. `Sandbox.create()`/`.from_id()`/`.from_name()`).",
+            )
 
         tags_list = [api_pb2.SandboxTag(tag_name=name, tag_value=value) for name, value in tags.items()]
 
@@ -586,7 +590,7 @@ class _Sandbox(_Object, type_prefix="sb"):
             tags=tags_list,
         )
         try:
-            await retry_transient_errors(client.stub.SandboxTagsSet, req)
+            await retry_transient_errors(self._client.stub.SandboxTagsSet, req)
         except GRPCError as exc:
             raise InvalidError(exc.message) if exc.status == Status.INVALID_ARGUMENT else exc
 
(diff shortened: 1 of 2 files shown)


modal/sandbox.py
```diff
@@ -535,6 +535,13 @@ async def from_name(
         Raises a modal.exception.NotFoundError if no running sandbox is found with the given name.
         A Sandbox's name is the `name` argument passed to `Sandbox.create`.
         """
+        if environment_name is not None:
+            deprecation_warning(
```
mwaskom, line 539: I think we still need this: it scopes the app-by-name part of the lookup

test/sandbox_test.py
```diff
@@ -408,7 +408,7 @@ def test_sandbox_list_app(client, servicer):
 def test_sandbox_list_tags(app, client, servicer):
     sb = Sandbox.create("bash", "-c", "sleep 10000", app=app)
     assert sb.get_tags() == {}
-    sb.set_tags({"foo": "bar", "baz": "qux"}, client=client)
```
mwaskom, line 411: Surprised this test passes without you adding the client in the `Sandbox.create()` call 🤔

modal/sandbox.py
```diff
@@ -584,7 +595,7 @@ async def set_tags(self, tags: dict[str, str], *, client: Optional[_Client] = No
             tags=tags_list,
         )
         try:
-            await retry_transient_errors(client.stub.SandboxTagsSet, req)
+            await retry_transient_errors(self._client.stub.SandboxTagsSet, req)
```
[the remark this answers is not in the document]
mwaskom: Ya dawg that's the idea

modal/sandbox.py
```diff
@@ -575,8 +575,12 @@ async def get_tags(self) -> dict[str, str]:
     async def set_tags(self, tags: dict[str, str], *, client: Optional[_Client] = None) -> None:
         """Set tags (key-value pairs) on the Sandbox. Tags can be used to filter results in `Sandbox.list`."""
         environment_name = _get_environment_name()
```
mwaskom, line 577: Were we going to remove this part here too? Or are you deferring that for later. Fine to break it up but don't want it to get dropped;  I think this is a legit bug with potential harmful consequences. (Might need some backend work to fix first though; ideally we would drop the `environment` from the `sandbox_tag` table altogether.

ehdr: Yes exactly, we have to fix it on the backend first, I confirmed it does not work as intended if we just stop sending it here. :(

test/sandbox_test.py
ehdr: Because `.set_tags()` now falls back to the `client` associated with the `app` passed in at `Sandbox.create()`,  and _it's_ [initiated with the same `client`](https://github.com/modal-labs/modal-client/blob/a05836a1e51a64344160fb75968389d492485ee2/test/sandbox_test.py#L22), I think?

mwaskom: Woah. That's hard to understand. I'm not sure what's best here.

PR #3575, opened 2025-09-19
Implement SandboxRouter protos

saltzm:
## Describe your changes

This PR adds new proto definitions for a service that will run on the worker called the Sandbox Router Server. This service will accept sandbox-related commands and forward them to a Sandbox Command Server that runs in `modal-runtime`. The service currently is responsible for `exec` requests.

Follow-up PRs will create a new Python client object for accessing this service and integrate it into the Sandbox API.

</details>

### DIFF
--- a/modal_proto/sandbox_router.proto
+++ b/modal_proto/sandbox_router.proto
@@ -1,0 +1,126 @@
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+import "modal_proto/api.proto";
+
+package modal.sandbox_router;
+
+enum SandboxExecStderrConfig {
+  // The output will be discarded.
+  SANDBOX_EXEC_STDERR_CONFIG_DEVNULL = 0;
+  // The output will be streamed to the client.
+  SANDBOX_EXEC_STDERR_CONFIG_PIPE = 1;
+  // A special value that can be used to indicate that the stderr stream should
+  // be merged with the stdout stream.
+  SANDBOX_EXEC_STDERR_CONFIG_STDOUT = 2;
+}
+
+enum SandboxExecStdioFileDescriptor {
+  // Read from stdout.
+  SANDBOX_EXEC_STDIO_FILE_DESCRIPTOR_STDOUT = 0;
+  // Read from stderr.
+  SANDBOX_EXEC_STDIO_FILE_DESCRIPTOR_STDERR = 1;
+}
+
+enum SandboxExecStdoutConfig {
+  // The output will be discarded.
+  SANDBOX_EXEC_STDOUT_CONFIG_DEVNULL = 0;
+  // The output will be streamed to the client.
+  SANDBOX_EXEC_STDOUT_CONFIG_PIPE = 1;
+}
+
+message SandboxExecStartRequest {
+  // The task ID of the sandbox to execute the command in.
+  string task_id = 1;
+  // Execution ID. This ID will be used to identify the execution for other
+  // requests and ensure exec commands are idempotent.
+  //
+  // TODO(saltzm): Could instead have a separate idempotency key from the exec_id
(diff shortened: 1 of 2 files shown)


modal_proto/sandbox_router.proto
```diff
@@ -0,0 +1,126 @@
...
+service SandboxRouter {
+  // Execute a command in the sandbox.
+  rpc SandboxExecStart(SandboxExecStartRequest) returns (SandboxExecStartResponse);
+ // Write to the stdin stream of an exec'd command.
+  rpc SandboxExecStdinWrite(SandboxExecStdinWriteRequest) returns (SandboxExecStdinWriteResponse);
+  // Get a stream of output from the stdout or stderr stream of an exec'd command.
+  rpc SandboxExecStdioRead(SandboxExecStdioReadRequest) returns (stream SandboxExecStdioReadResponse);
+  // Wait for an exec'd command to exit and return the exit code.
+  rpc SandboxExecWait(SandboxExecWaitRequest) returns (SandboxExecWaitResponse);
+}
```
rolyatmax: nit: file should end with a newline

modal_proto/sandbox_router.proto
```diff
@@ -0,0 +1,126 @@
...
+  // Command arguments to execute.
+  repeated string command_args= 3;
+  // Configures how the stdout of the command will be handled.
+  SandboxExecStdoutConfig stdout_config = 4;
+  // Configures how the stderr of the command will be handled.
+  SandboxExecStderrConfig stderr_config = 5;
+  // Deadline for the exec'd command to exit. If the command does not exit before
+  // the deadline, the command will be killed. This is NOT the deadline for the
+  // ExecStartRequest RPC to complete.
+  google.protobuf.Timestamp deadline = 6;
```
rolyatmax: Should this maybe be given as a duration? I'm assuming this timestamp would be generated using the client's clock but then would be compared with the worker's clock when timing out?

modal_proto/sandbox_router.proto
```diff
@@ -0,0 +1,126 @@
...
+
+message SandboxExecWaitResponse {
+  oneof exit_status {
+    // The exit code of the command.
+    int32 code = 3;
+    // The signal that terminated the command.
+    int32 signal = 4;
+  }
+  // TODO(saltzm): Give a way for the user to distinguish between normal exit
+  // and termination by Modal (due to sandbox idle timeout, exec exceeded deadline, etc.)
```
rolyatmax: FWIW, if there's an exec in progress, the sandbox should not hit the idle timeout. Agree that giving more detail about how an exec terminated is good, though!

rolyatmax:
A few comments there, but looks pretty good!

modal_proto/sandbox_router.proto
saltzm: Yeah I was worried about this too but the timestamp is from the epoch which is defined in UTC. I prefer deadlines to timeouts since if somehow the message gets stuck in transit for a while or something gets stuck before execstart happens, the clock doesn't start over when the message arrives

modal_proto/sandbox_router.proto
saltzm: Ah true, fixed comment

modal_proto/sandbox_router.proto
rolyatmax: My concern isn't so much about different timezones, it's about using a client clock that has drifted or is set wrong. Also, I think I'd actually want the countdown to start when the exec starts, not when the request is sent. I'm probably setting a timeout based on how long I expect the process to run and I wouldn't want the time it takes to send the request over a network to count against the process's timeout.

saltzm: Chatted offline, going with Taylor's suggestion

PR #3576, opened 2025-09-21
Split AutoscalerOverride out from AutoscalerSettings

kareem-modal:
## Describe your changes

- Adds a new AutoscalerOverrides protobuf to distinguish between settings and overrides
- Introduces clear flags for all AutoscalerSettings fields to support an instruction to revert to the function definition / modal default value 

---

## Compatibility checklist

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -658,7 +658,30 @@ message AutoscalerSettings {
   optional uint32 scaleup_window = 4;
   // Maximum amount of time a container can be idle before being scaled down, in seconds; pka "container_idle_timeout"
   optional uint32 scaledown_window = 5;
-  reserved 6;
+  reserved 6, 7, 8, 9, 10;  // reserved for compatibility with autoscaler overrides
+}
+
+message AutoscalerOverride {
+  // Autoscaler override contains settings
+  // As well as flags for clearing previous overrides and restoring defaults
+
+  // Minimum containers when scale-to-zero is not desired; pka "keep_warm" or "warm_pool_size"
+  optional uint32 min_containers = 1;
+  // Limit on the number of containers that can be running for each Function; pka "concurrency_limit"
+  optional uint32 max_containers = 2;
+  // Additional container to spin up when Function is active
+  optional uint32 buffer_containers = 3;
+  // Currently unused; a placeholder in case we decide to expose scaleup control to users
+  optional uint32 scaleup_window = 4;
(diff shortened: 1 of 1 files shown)


kareem-modal:
> I'd be a little unhappy about the duplication here. Even if you want to explicitly communicate "reset these fields" so that you don't need to disambiguate `0` as a a desired state from `0` as a "reset" command, I'm not sure why you'd need to have separate messages for "override" and "settings" _values_. Have you considered just having something like `AutoscalerReset` with the boolean fields?
> 
> Or — not certain this is a good idea — could you even just send an `AutoscalerSettings` _as_ a representation of a reset command, i.e. setting 1 when you want to reset that field and 0 otherwise?

That's a totally fair point, and I initially did have the reset fields in settings. But as I thought about it more, I think AutoscalerOverride really is a different concept from AutoscalerSettings.

Overrides have semantics that don’t make sense in base settings (e.g. clear flags, and in the future, things like TTLs). By trying to fit all those into AutoscalerSettings, we’re introducing fields that are meaningless in the “real settings” context, for the sake of DRY.

Using AutoscalerSettings as a “reset command” with magic values also feels dangerous, since you can’t easily tell whether a 0 means “the user wants zero containers” or “reset this field.” The ambiguity we have right now (did they just not put a value, or do they want to unset/reset the field) is what I'm hoping to fix.

The tradeoff is a bit of duplication, but it lets us evolve overrides distinctly from the base settings, which are basically just the current values.

Regarding whether we need separate messages for settings values and settings overrides, we don't strictly need it. We could just put `settings AutoscalerSettings` in the AutoscalerOverride, and I agree that would be better. The only reason I didn't is to avoid breaking client by letting them send an `AutoscalerSettings` as an override. But I don't really love having to send two distinct messages to do an override, AutoscalerReset AND AutoscalerSettings. Reseting is just a part of an override operation, imo.

Another alternative here is just to support deleting overrides. That would remove the need for clear flags, but we'd then need to perform multiple operations via API if we're doing to not use flags. And since our API is built around RPCs not REST, introducing DELETE isn't really consistent. 

Open to being wrong though, obviously I'm new and may be missing context. Can hop on a huddle too if its' easier.

mwaskom:
Ok thanks for explaining. It's not immediately obvious why needing to have an explicit representation of "reset" intent (makes sense) requires us to have two separate representations of the desired settings. But I haven't thought about it as deeply.

PR #3578, opened 2025-09-22
Add `image_delete` to `modal.experimental`

ehdr:
## Describe your changes

SDK-610

---

## Compatibility checklist

</details>

## Changelog

- Added `modal.experimental.image_delete()` to allow deleting Images (e.g. Sandbox FS snapshot images).

### DIFF
--- a/modal/experimental/__init__.py
+++ b/modal/experimental/__init__.py
@@ -392,10 +392,17 @@ async def image_delete(
     *,
     client: Optional[_Client] = None,
 ) -> None:
-    """Delete an image by its ID.
+    """Delete an Image by its ID.
 
-    Warning: This is an experimental API that may change.
     Deletion is irreversible and will prevent Apps from using the Image.
+
+    This is an experimental interface for a feature that we will be adding to
+    the main Image class. The stable form of this interface may look different.
+
+    Note: When building an Image, each chained method call will create an
+    intermediate Image layer, each with its own ID. Deleting an Image will not
+    delete any of its intermediate layers, only the image identified by the
+    provided ID.
     """
     if client is None:
         client = await _Client.from_env()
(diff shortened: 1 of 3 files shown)


modal/experimental/__init__.py
```diff
@@ -384,3 +384,21 @@ async def update_autoscaler(
...
+@synchronizer.create_blocking
+async def image_delete(
+    image_id: str,
+    *,
+    client: Optional[_Client] = None,
+) -> None:
+    """Delete an image by its ID.
+
+    Warning: This is an experimental API that may change.
+    Deletion is irreversible and will prevent Apps from using the Image.
```
mwaskom, line 397: Not sure whether we want to add any additional information here re: the conversation we're having on Slack.

modal/experimental/__init__.py
```diff
@@ -384,3 +384,21 @@ async def update_autoscaler(
...
     await retry_transient_errors(client.stub.FunctionUpdateSchedulingParams, request)
+
+
+@synchronizer.create_blocking
+async def image_delete(
+    image_id: str,
+    *,
+    client: Optional[_Client] = None,
+) -> None:
+    """Delete an image by its ID.
```
mwaskom: Nit
```suggestion
    """Delete an Image by its ID.
```

modal/experimental/__init__.py
```diff
@@ -384,3 +384,21 @@ async def update_autoscaler(
...
+
+    Warning: This is an experimental API that may change.
```
mwaskom: If we know at the time when we add the experimental functionality I think it's helpful to be explicit about whether the _interface_ will change vs. whether the _functionality_ will be removed. It helps people decide whether to use it, since you might be fine knowing that you'll need to update your code to a new spelling in the future but that you'd need to completely revise the architecture if you can no longer use that functionality.

In practice, probably once we add a functionality in an experimental API we'll find it difficult to take it away in the future, even if we're very clear in the documentation. But it helps to manage expectations.

ehdr:
@danielnorberg FYI

PR #3589, opened 2025-09-24
Add volume mounts proto to image run_commands

thomasjpfan:
## Describe your changes

Implementing for adding `Image.run_commands`

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- `Image.run_commands` supports `volumes` for mounting a volume for build caching

```python
cache_vol = modal.Volume.from_name("cache-mount")
cmd_using_cache = "..."
image = modal.Image.debian_slim().run_commands(cmd_using_cache, volumes={"/cache": cache_vol})
```

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -39,6 +39,7 @@ from ._utils.docker_utils import (
 )
 from ._utils.function_utils import FunctionInfo
 from ._utils.grpc_utils import RETRYABLE_GRPC_STATUS_CODES, retry_transient_errors
+from ._utils.mount_utils import validate_volumes
 from .client import _Client
 from .cloud_bucket_mount import _CloudBucketMount
 from .config import config, logger, user_config_path
@@ -499,12 +500,16 @@ class _Image(_Object, type_prefix="im"):
         context_mount_function: Optional[Callable[[], Optional[_Mount]]] = None,
         force_build: bool = False,
         build_args: dict[str, str] = {},
+        volumes: Optional[dict[Union[str, PurePosixPath], _Volume]] = None,
         # For internal use only.
         _namespace: "api_pb2.DeploymentNamespace.ValueType" = api_pb2.DEPLOYMENT_NAMESPACE_WORKSPACE,
         _do_assert_no_mount_layers: bool = True,
     ):
         if base_images is None:
             base_images = {}
+
+        if volumes is None:
+            volumes = {}
 
         if secrets is None:
             secrets = []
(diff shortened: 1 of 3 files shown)


thomasjpfan:
> Are there other Image methods where it would make sense to support Volume mounting?

Apart from `run_commands`, some candidates are:

1. `uv_pip_install`, `pip_install` (For mounting the uv/pip local cache)
2. `dockerfile_commands`: A little weird since it only works with the `RUN` command. I can see someone running into the gotcha of using `COPY` to copy data from a volume into the image.

modal/image.py
```diff
@@ -508,12 +513,22 @@ def _from_args(
         if build_function and len(base_images) != 1:
             raise InvalidError("Cannot run a build function with multiple base images!")
 
+        validated_volumes = validate_volumes(volumes)
+        validated_volumes_: list[tuple[str, _Volume]] = []
+
+        for k, v in validated_volumes:
+            if isinstance(v, _CloudBucketMount):
+                raise InvalidError("Image builds only support mounting modal.Volume")
```
mwaskom: Dang, I feel like we just plugged this gap in `.with_options` and now there's a new one :(

modal/image.py
```diff
@@ -508,12 +513,22 @@ def _from_args(
...
+            validated_volumes_.append((k, v))
```
mwaskom: I'm not following why we need to build a separate list here.

modal/image.py
mwaskom: Super minor but regarding the error message language, `.run_function` _does_ support CloudBucketMount. (Also, potentially more important: I'm not sure if we would end up here with a `.run_function` layer but any chance this would break that?)

mwaskom: I guess probably not since there the Volume spec would be packaged into the `build_function` definition?

modal/image.py
thomasjpfan: This is mostly to keep the type checker happy. I refactor the code to not need this anymore.

modal/image.py
thomasjpfan: `run_function` does not end up here. It passes `volumes` to `_Function.from_local` and not passed to `_Image._from_args`.

I refactored the code to have the caller validated volumes before passing it into `_Image._from_args`.

modal/image.py
```diff
@@ -1712,6 +1732,7 @@ def build_dockerfile(version: ImageBuilderVersion) -> DockerfileSpec:
             secrets=secrets,
             gpu_config=parse_gpu_config(gpu),
             force_build=self.force_build or force_build,
+            validated_volumes=validate_only_modal_volumes(volumes, "Image.run_commands"),
```
mwaskom, line 1735: Sorry for belaboring the point here but from a type checking perspective it seems like this is superfluous; won't the values of the dict always have a _Volume type from the checker's perspective? Adding a runtime check is helpful for users, but what am I missing about the need for the type transformation  type guard stuff?

thomasjpfan: > Adding a runtime check is helpful for users, but what am I missing about the need for the type transformation type guard stuff?

Yes, this is for runtime checking.

As for the type guard stuff, calling `validate_volumes` returns a `Sequence[tuple[str, Union[_Volume, _CloudBucketMount]]]`, so we'll need to use a `TypeGuard` or a `typing.cast` to restrict the type. I went with a `TypeGuard` because I want to raise a nicer error during runtime anyways.

An alternative is to reimplement `validate_volumes` that does the type checking and returns a `Sequence[tuple[str, _Volume]]`.

mwaskom: Ah got it, the role of `validate_volumes` here makes sense, thanks.

PR #3590, opened 2025-09-24
Add secret to sandbox definition and SandboxGetDirectAccess protos

saltzm:
## Describe your changes
[Architectural context](https://www.notion.so/modal-com/Client-to-Sandbox-Direct-Communication-v2-2631e7f16949807ba5a9f329200ccb12?source=copy_link#2631e7f169498038a6f3e25bc8efbeb1)

My biggest concern here is one @rolyatmax brought up, which is whether it's a good idea to store the auth secret in the Sandbox definition. Currently we never send the Sandbox definition back to the client, it's only used for Sandbox creation, but if we ever did in the future, we'd need to make sure not to return this secret.

In addition, the way the secret will be communicated to the worker is through its receipt of the Sandbox definition. Is this secure enough? I don't know - would love reviewer input

<details> <summary>Checklists</summary>.

---

## Compatibility checklist

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2704,12 +2704,6 @@ message Sandbox {
 
   // Optional idle timeout in seconds. If set, the sandbox will be terminated after being idle for this duration.
   optional uint32 idle_timeout_secs = 33;
-
-  // Set only by the server. Must not be transmitted to the client. Used to
-  // authenticate the client when directly connecting to a sandbox server running on
-  // the modal-worker to issue exec commands (and other operations as they
-  // become available).
-  optional bytes sandbox_command_router_auth_secret = 34;
 }
 
 message SandboxCreateConnectTokenRequest {
@@ -2735,11 +2729,11 @@ message SandboxCreateResponse {
 // Used to get a JWT and URL for direct access to a sandbox router server
 // running on the modal-worker, so the client can issue exec commands (and other
 // operations as they become available) directly to the worker.
-message SandboxGetCommandRouterAccessInfoRequest {
+message SandboxGetCommandRouterAccessRequest {
   string sandbox_id = 1;
 }
 
-message SandboxGetCommandRouterAccessInfoResponse {
+message SandboxGetCommandRouterAccessResponse {
   string jwt = 1;
   string url = 2;
 }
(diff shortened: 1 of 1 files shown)


saltzm:
[Slack thread for posterity](https://modal-com.slack.com/archives/C097TF3N20J/p1758744973996449)

modal_proto/api.proto
```diff
@@ -2726,6 +2732,18 @@ message SandboxCreateResponse {
   string sandbox_id = 1;
 }
 
+// Used to get a JWT and URL for direct access to a sandbox router server
+// running on the modal-worker, so the client can issue exec commands (and other
+// operations as they become available) directly to the worker.
+message SandboxGetCommandRouterAccessInfoRequest {
```
rolyatmax: Maybe `Credentials` instead of `AccessInfo`?

modal_proto/api.proto
```diff
@@ -2704,6 +2704,12 @@ message Sandbox {
 
   // Optional idle timeout in seconds. If set, the sandbox will be terminated after being idle for this duration.
   optional uint32 idle_timeout_secs = 33;
+
+  // Set only by the server. Must not be transmitted to the client. Used to
+  // authenticate the client when directly connecting to a sandbox server running on
+  // the modal-worker to issue exec commands (and other operations as they
+  // become available).
+  optional bytes sandbox_command_router_auth_secret = 34;
```
pawalt: I'm not really on board with making this symmetric secret put in the database and passed to every worker. IMO, the Sandbox definition should be something we can use in investigations, give to users, etc. It should not have secret information in it.

I think we should use our existing OIDC infrastructure to mint tokens and validate them from the Worker.

modal_proto/api.proto
saltzm: I thought you had that debate on your PR and decided against it 😅 I was just going to say UrlAndToken instead of AccessInfo but went to AccessInfo just in case we ever needed a different field and there was name drift. I can change it to credentials if you feel strongly, just didn't think `url` was really included in the idea of credentials. 

A quick brainstorm with ChatGPT came up with:

SandboxGetCommandRouterConnection
SandboxGetCommandRouterAccess
SandboxCreateCommandRouterConnection
SandboxCreateCommandRouterConnection
...MintCommandRouterSession
...IssueCommandRouterCredentials

I'll go with SandboxIssueCommandRouterCredentials, since then it's like... issuing credentials and telling you the URL

saltzm: Decideed on `SandboxGetCommandRouterAccess`

modal_proto/api.proto
saltzm: Going to go with OIDC approach for now

PR #3594, opened 2025-09-25
Add direct_sandbox_commands_enabled field to sandbox definition

saltzm:
## Describe your changes

We want to decide on sandbox creation whether to use the sandbox command router for exec commands or not based on a config parameter - we need to store this decision in the definition. Otherwise we could swap exec implementations even within the same sandbox which sounds confusing.

---

## Compatibility checklist

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2707,6 +2707,11 @@ message Sandbox {
 
   // Optional idle timeout in seconds. If set, the sandbox will be terminated after being idle for this duration.
   optional uint32 idle_timeout_secs = 33;
+
+  // If set, the sandbox will be created with direct sandbox commands enabled.
+  // Exec commands for the sandbox will be issued directly to the sandbox
+  // command router running on the Modal worker.
+  bool direct_sandbox_commands_enabled = 34;
 }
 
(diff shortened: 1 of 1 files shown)


modal_proto/api.proto
```diff
@@ -2707,6 +2707,11 @@ message Sandbox {
 
   // Optional idle timeout in seconds. If set, the sandbox will be terminated after being idle for this duration.
   optional uint32 idle_timeout_secs = 33;
+
+  // If set, the sandbox will be created with direct sandbox commands enabled.
+  // Exec commands for the sandbox will be issued directly to the sandbox
+  // command router running on the Modal worker.
+  bool direct_sandbox_commands_enabled = 34;
```
paulgb, line 2714: do we always create `Sandbox` on the server, or is it created on the client?

saltzm: We create it on the client, but then the server sometimes[ modifies it before storing it ](https://github.com/modal-labs/modal/blob/c935ce4fc6aca3e7f26345378c4c436daf5b4423/server/modal_server/objects/sandbox.py#L643-L689). It's weird

rolyatmax (approved):

PR #3597, opened 2025-09-25
Add App constructor tags

mwaskom:
Minimal client-side implementation of SDK-622.

This only lets you define tags in the App constructor.

I think it would also make sense to let you supply tags when programmatically running or deploying the App. This will support some workflows that have a shared App codebase that's used by different people or for different workflows such that it would be helpful to define the tags dynamically. The main reason I held off was lack of clarity about needing to have `tags: dict[str, str]` and `tag: str` in the same Function signature 😬.

---

## Compatibility checklist

</details>

## Changelog

- Arbitrary key-value metadata can now be attached to Apps by setting `modal.App(tags={...})`. The tags can be useful for tracking information that may be relevant to your organization, such as the team that owns the App. We'll support the inclusion of tags in some forthcoming APIs related to cost insights.

### DIFF
--- a/modal/_utils/name_utils.py
+++ b/modal/_utils/name_utils.py
@@ -31,10 +31,25 @@ def is_valid_environment_name(name: str) -> bool:
     return len(name) <= 64 and re.match(r"^[a-zA-Z0-9][a-zA-Z0-9-_.]+$", name) is not None
 
 
-def is_valid_tag(tag: str) -> bool:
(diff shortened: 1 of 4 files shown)


ehdr (approved):
👌 

But the changelog entry buries the lede that this is about tags a bit? Maybe `Arbitrary key-value metadata tags can now...` or something?

PR #3603, opened 2025-09-26
[GPU Snapshot] Remove unnecessary CUDA state checks and increase `cuda-checkpoint` timeout

mattnappo:
It is **impossible** for a snapshotted container to have any processes with any CUDA state other than `checkpointed`. This is because we assert that the CUDA state is `checkpointed` for all saved CUDA pids before taking the memory snapshot (we raise otherwise, after retrying 3 times). The current code, on restore, checks that the PID is `checkpointed` before restoring it, _twice_. This is a costly unnecessary check:

Statistics for a _single_ CUDA state check of a process that allocated a 2GiB random tensor.
```
Mean:   2.66884
Var:    1.969
Min:    0.383
Max:    7.119
Median: 2.525
Sample: 25
```

This PR also increases the timeout on running `cuda-checkpoint`. We have found that for large checkpoints, cuda-checkpoint requires a long time to run. Cuda-checkpoint taking a while is not a signal that it will fail.

---

> [!NOTE]
> Removes restore-time CUDA state validations and first refresh, and introduces higher, centralized timeouts for cuda-checkpoint operations.
> 
> - **Runtime: GPU memory snapshot (`modal/_runtime/gpu_memory_snapshot.py`)**
>   - **Timeouts**:
>     - Add `CUDA_CHECKPOINT_TOGGLE_TIMEOUT` and `CUDA_CHECKPOINT_TIMEOUT`; use them for toggle loop and subprocess calls.
>     - Replace per-call constants (`30s`/`10s`) with `CUDA_CHECKPOINT_TIMEOUT`; enforce overall toggle limit via `CUDA_CHECKPOINT_TOGGLE_TIMEOUT`.
>   - **Toggle behavior**:
>     - Change `toggle(...)` signature: remove `timeout_secs`, add `skip_first_refresh`.
>     - `_should_continue_toggle(...)` now supports optional refresh and uses global toggle timeout.
>   - **Restore flow**:
>     - Remove pre-restore state validation loop; call `toggle(RUNNING, skip_first_refresh=True)` per process.
>   - **Misc**:
>     - Note quick 5s state check in `_check_cuda_session` comment.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 52a791bf48c03862b26d13216aa6f11835e11556. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_runtime/gpu_memory_snapshot.py
+++ b/modal/_runtime/gpu_memory_snapshot.py
@@ -61,9 +61,9 @@ class CudaCheckpointProcess:
         max_retries = 3
 
         attempts = 0
-        while self._should_continue_toggle(target_state, start_time):
-            if not (skip_first_refresh and attempts == 0):
-                self.refresh_state()
+        while self._should_continue_toggle(
+            target_state, start_time, refresh=not (skip_first_refresh and attempts == 0)
+        ):
             attempts += 1
             try:
                 self._execute_toggle_command()
@@ -82,8 +82,13 @@ class CudaCheckpointProcess:
 
         logger.debug(f"PID: {self.pid} Target state {target_state.value} reached")
 
-    def _should_continue_toggle(self, target_state: CudaCheckpointState, start_time: float) -> bool:
+    def _should_continue_toggle(
+        self, target_state: CudaCheckpointState, start_time: float, refresh: bool = True
+    ) -> bool:
         """Check if toggle operation should continue based on current state and timeout."""
+        if refresh:
+            self.refresh_state()
+
         if self.state == target_state:
             return False
 


modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -121,7 +131,7 @@ def refresh_state(self) -> None:
                 check=True,
                 capture_output=True,
                 text=True,
-                timeout=10,
+                timeout=CUDA_CHECKPOINT_TIMEOUT,
```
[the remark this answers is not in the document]
mattnappo: Valid for `_check_cuda_session` but not `refresh_state`. Calling `--get-state` on a non-checkpointed pid is typically very fast.

modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -44,7 +46,9 @@ class CudaCheckpointProcess:
     pid: int
     state: CudaCheckpointState
 
-    def toggle(self, target_state: CudaCheckpointState, timeout_secs: float = 5 * 60.0) -> None:
+    def toggle(
+        self, target_state: CudaCheckpointState, timeout_secs: float = 5 * 60.0, skip_first_refresh: bool = False
+    ) -> None:
```
luiscape: Do we need a `timeout_secs` parameter? What about just using the constant `CUDA_CHECKPOINT_TIMEOUT` everywhere?

luiscape (approved):
lgtm

modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -85,7 +93,7 @@ def _should_continue_toggle(
             raise CudaCheckpointException(f"PID: {self.pid} CUDA process state is {self.state}")
 
         elapsed = time.monotonic() - start_time
-        if elapsed >= timeout_secs:
+        if elapsed >= CUDA_CHECKPOINT_TIMEOUT:
```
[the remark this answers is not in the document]
mattnappo: Good catch!

modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -18,6 +18,12 @@
 
 CUDA_CHECKPOINT_PATH: str = config.get("cuda_checkpoint_path")
 
+# Maximum total duration for an entire toggle operation.
+CUDA_CHECKPOINT_TOGGLE_TIMEOUT: float = 5 * 60.0
+
+# Maximum total duration for each individiaul `cuda-checkpoint` invocation.
```
thundergolfer:
```suggestion
# Maximum total duration for each individual `cuda-checkpoint` invocation.
```

modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -54,12 +60,17 @@ def toggle(self, target_state: CudaCheckpointState, timeout_secs: float = 5 * 60
...
+        while self._should_continue_toggle(
+            target_state, start_time, refresh=not (skip_first_refresh and is_first_attempt)
+        ):
+            is_first_attempt = False
             try:
                 self._execute_toggle_command()
                 # Use exponential backoff for retries
                 sleep_time = min(0.1 * (2**retry_count), 1.0)
-                time.sleep(sleep_time)
+                if sleep_time > 0:
```
thundergolfer: `time.sleep(0)` doesn't error, so I find this an unnecessary complication. `sleep_time` can't be negative can it?

modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -73,10 +84,11 @@ def toggle(self, target_state: CudaCheckpointState, timeout_secs: float = 5 * 60
         logger.debug(f"PID: {self.pid} Target state {target_state.value} reached")
 
     def _should_continue_toggle(
-        self, target_state: CudaCheckpointState, start_time: float, timeout_secs: float
+        self, target_state: CudaCheckpointState, start_time: float, refresh: bool = True
     ) -> bool:
         """Check if toggle operation should continue based on current state and timeout."""
-        self.refresh_state()
+        if refresh:
+            self.refresh_state()
```
thundergolfer, line 90: Could this code be simpler if this wasn't embedded inside `_should_continue_toggle`? I don't see why it should be inside, other than initial convenience.

modal/_runtime/gpu_memory_snapshot.py
mattnappo: That's correct.

modal/_runtime/gpu_memory_snapshot.py
mattnappo: Subtly, we need to call refresh after the first toggle to check that it succeeded. So I think it makes sense for the refresh_state call to be within `should_continue_toggle`

modal/_runtime/gpu_memory_snapshot.py
```diff
@@ -256,20 +269,11 @@ def restore(self) -> None:
...
-                    f"PID {proc.pid}: CUDA session not in {CudaCheckpointState.CHECKPOINTED.value} state. "
-                    f"Current state: {proc.state.value}"
-                )
-
         # See checkpoint() for rationale about parallelism.
         start = time.perf_counter()
 
         def restore_process(proc: CudaCheckpointProcess) -> None:
-            proc.toggle(CudaCheckpointState.RUNNING)
+            proc.toggle(CudaCheckpointState.RUNNING, skip_first_refresh=True)
```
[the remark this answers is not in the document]
mattnappo: I don't think cursor understands this state machine.

PR #3617, opened 2025-09-29
Client Update for Multiple open StreamWriter's fix [ART-8]

aadit-juneja:
## This issue fixes an issue in which multiple `stdin` StreamWriters cannot be opened to a Sandbox. Previously, only one StreamWriter could be opened to write messages to `stdin`. Our new fix, at a high level, more explicitly enforces only one open StreamWriter, assigning a `uuid` to each writer and adds an ephemeral `active_writer_id` attribute for sandboxes.

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2606,6 +2606,7 @@ message RuntimeInputMessage {
   bytes message = 1;
   uint64 message_index = 2;
   bool eof = 3;
+  string writer_id = 4;
 }
 
 message RuntimeOutputBatch {
@@ -2876,6 +2877,7 @@ message SandboxStdinWriteRequest {
   bytes input = 2;
   uint32 index = 3;
   bool eof = 4;
+  string writer_id = 5;
 }
 
 message SandboxStdinWriteResponse {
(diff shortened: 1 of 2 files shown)


modal_proto/api.proto
```diff
@@ -2616,6 +2606,7 @@ message RuntimeInputMessage {
   bytes message = 1;
   uint64 message_index = 2;
   bool eof = 3;
+  string writer_id = 4;
```
rolyatmax, line 2609: I think there are two spots where we create `RuntimeInputMessage`s in the client. Do those need to be updated to take this new `writer_id` value, too?

aadit-juneja: I included the field because `RuntimeInputMessage` was used in server code for the code flow of a `SandboxStdinWriteRequest` and it needs to be sent over to the worker for deduplication ([monorepo PR](https://github.com/modal-labs/modal/pull/27868)). The us of `RuntimeInputMessage` PR in the client looks like it's once for non-sandbox tasks and another for `modal shell`.

I think that if the field is not included in when it is sent it gets set to a default string value ("") and it'll be ignored on the code paths for those two.

PR #3621, opened 2025-10-01
Add new SDK for workspace billing report

mwaskom:
Client side, closes SDK-624

---

## Compatibility checklist

## Changelog

---

> [!NOTE]
> Introduce `modal.billing.workspace_billing_report` (sync wrapper over async) to fetch workspace billing data via `WorkspaceBillingReport` stream and return typed rows.
> 
> - **Billing API**
>   - Add `modal.billing.workspace_billing_report` (sync wrapper of `_workspace_billing_report`).
>   - Implements client call to `WorkspaceBillingReport` (unary_stream), returning rows with `object_id`, `description`, `environment_name`, `interval_start` (UTC), `cost` (`Decimal`), and `tags`.
>   - Validates `start`/`end` timezone (UTC or naive treated as UTC), defaults `end` to `now`, supports `resolution` and `tag_names`.
> - **Tests**
>   - New `test/billing_test.py` verifies request emission, UTC handling, and types.
>   - Extend mock servicer in `test/conftest.py` with `WorkspaceBillingReport` to return a sample item.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ce282dcae3ad5e2f70454eb3b0c8b90478da2335. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


modal/billing.py
```diff
@@ -0,0 +1,74 @@
...
+    even if they were added or removed at some point within it.
+
+    """
+    if client is None:
+        client = await _Client.from_env()
+
+    tag_names = tag_names or []
+
+    start_dt = datetime.fromisoformat(start) if isinstance(start, str) else start
+    end_dt = datetime.fromisoformat(end) if isinstance(end, str) else datetime.now() if end is None else end
```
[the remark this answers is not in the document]
mwaskom: 🤔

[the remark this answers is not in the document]
mwaskom: I hate time zones man

modal/billing.py
```diff
@@ -0,0 +1,74 @@
...
+
+__all__ = ["workspace_billing_report"]
+
+
+@synchronizer.create_blocking
+async def workspace_billing_report(
+    *,
+    start: Union[datetime, str],  # Start of the report, inclusive
+    end: Optional[Union[datetime, str]] = None,  # End of the report, exclusive
+    resolution: str = "d",  # Resolution, e.g. "d" for daily or "h" for hourly
```
ehdr: Could maybe be
```suggestion
    resolution: Literal["d", "h"] = "d",  # Resolution, e.g. "d" for daily or "h" for hourly
```
?

modal/billing.py
```diff
@@ -0,0 +1,74 @@
...
+    The result will be a list of dictionaries for each interval (determined by `resolution`)
+    between the `start` and `end` limits. The dictionary represents a single Modal object
+    that billing can be attributed to (e.g., an App) along with metadata (including user-defined
+    tags) for identifying that object.
+
+    The `start` and `end` parameters can be strings or datetime objects. The parameters are
+    required to have a UTC timezone or to be timezone-naive (interpreted as UTC times). The
+    timestamps in the result will be in UTC. Cost will be reported for full intervals, even
+    if the provided `start` or `end` parameters are partial: `start` will be rounded to the
+    beginning of its interval, while partial `end` intervals will be excluded.
```
ehdr: Maybe add a simple example here to illustrate?

modal/billing.py
```diff
@@ -0,0 +1,74 @@
...
+from .client import _Client
+from .exception import InvalidError
+
+__all__ = ["workspace_billing_report"]
+
+
+@synchronizer.create_blocking
+async def workspace_billing_report(
+    *,
+    start: Union[datetime, str],  # Start of the report, inclusive
```
ehdr: Personally not a fan of supporting `str` here but I see the value for convenience.

ehdr (approved):
LGTM, except Cursor's timezone comment and failing test!

modal/billing.py
mwaskom: Currently, the server defines what we accept here. Maybe it would be better to define that in the client and use a literal type so it's more obvious. It would be annoying if we added support for more resolutions in the future and required users to upgrade their client to get access, since all the client needs to do is pass this parameter up to the server. OTOH, it's unlikely that we'll support something like minute resolution in this interface, so perhaps we don't need to design for future expansion.

modal/billing.py
mwaskom: OK, dropped support for `str`.

PR #3613, opened 2025-09-29
Stop changelog extraction at end marker

ehdr:
## Describe your changes

Instead of making the bots adapt to our process, make our process adapt to any description-appending bot?

By stopping changelog extraction at `<!-- end-changelog -->`.

<details> <summary>Checklists</summary>

---

## Compatibility checklist

Check these boxes or delete any item (or this section) if not relevant for this PR.

- [x] Client+Server: this change is compatible with old servers
- [x] Client forward compatibility: this change ensures client can accept data intended for later versions of itself

Note on protobuf: protobuf message changes in one place may have impact to
multiple entities (client, server, worker, database). See points above.

</details>

PR #3635, opened 2025-10-03
Exclude cursor comments from changelog extraction

mwaskom:
Alternate approach to https://github.com/modal-labs/modal-client/pull/3613

A little less robust since we're depending on Cursor identifying itself in the comment it adds. But OTOH it doesn't require us to carefully preserve any tokens while editing the changelog, which I was expecting to be a headache.

The risk of cursor changing something that breaks this is fairly low since the changelog updates get staged these days anyway.

---

> [!NOTE]
> Adjusts changelog parsing in `update_changelog` to ignore Cursor auto-comments and strip HTML comments from the extracted update.
> 
> - **Changelog parsing (`tasks.py:update_changelog`)**:
>   - Update regex to capture text after `## Changelog` up to a Cursor auto-comment or end: `r"## Changelog\s*(.+?)(?:<!--\s*\w*CURSOR\w*\s*-->|$)"`.
>   - Strip HTML comments from the extracted changelog section only (not the entire PR description).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 62e4b7b7abdb4b5513e7f32d0a8225be5dc5c4c3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -459,7 +459,7 @@ def update_changelog(ctx, sha: str = ""):
     # Parse the PR description to get a changelog update, which is all text between
     # the changelog header and any auto comments appended by Cursor
 
-    changelog_pattern = r"## Changelog\s*(.+)(?:<!--\s*\w*CURSOR\w*\s*-->|$)"
+    changelog_pattern = r"## Changelog\s*(.+?)(?:<!--\s*\w*CURSOR\w*\s*-->|$)"
     m = re.search(changelog_pattern, pr_description, flags=re.DOTALL)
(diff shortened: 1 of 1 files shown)


ehdr (approved):
Looks good! I considered this approach too but was worried we'd end up playing whackamole with Cursor, Devin, Claude etc as we add more, but it's admittedly premature generalization so we can solve that if/when it comes to it I guess!

PR #3636, opened 2025-10-05
Add generator output format to generators for correctness

freider:
Since generators output both PICKLE/CBOR and GENERATOR_DONE (proto) outputs, it should be present in supported_output_formats if we want to be strict about it. Among other things, it simplifies output format validation in the worker if we can look at *only* supported_output_formats to determine what is allowed

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -845,7 +845,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
                         method_output_formats = [api_pb2.DATA_FORMAT_CBOR]
                     else:
                         method_output_formats = [api_pb2.DATA_FORMAT_PICKLE, api_pb2.DATA_FORMAT_CBOR]
-                    if is_generator:
+                    if partial_function.params.is_generator:
                         method_output_formats.append(api_pb2.DATA_FORMAT_GENERATOR_DONE)
 
                 method_definition = api_pb2.MethodDefinition(
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -845,6 +845,8 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
                         method_output_formats = [api_pb2.DATA_FORMAT_CBOR]
                     else:
                         method_output_formats = [api_pb2.DATA_FORMAT_PICKLE, api_pb2.DATA_FORMAT_CBOR]
+                    if is_generator:
+                        method_output_formats.append(api_pb2.DATA_FORMAT_GENERATOR_DONE)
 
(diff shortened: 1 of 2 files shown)


modal/_functions.py
```diff
@@ -838,13 +838,15 @@ def from_local(
                 )
                 if is_web_endpoint:
                     method_input_formats = [api_pb2.DATA_FORMAT_ASGI]
-                    method_output_formats = [api_pb2.DATA_FORMAT_ASGI]
+                    method_output_formats = [api_pb2.DATA_FORMAT_ASGI, api_pb2.DATA_FORMAT_GENERATOR_DONE]
```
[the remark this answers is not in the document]
freider: This is intentional - all web endpoints are implicitly also generators (they emit their output in multiple chunks)

modal/_functions.py
```diff
@@ -880,15 +889,11 @@ def _deps(only_explicit_mounts=False) -> list[_Object]:
...
-        elif webhook_config is not None:
-            supported_input_formats = [api_pb2.DATA_FORMAT_ASGI]
-            supported_output_formats = [api_pb2.DATA_FORMAT_ASGI]
         else:
-            supported_input_formats = [api_pb2.DATA_FORMAT_PICKLE, api_pb2.DATA_FORMAT_CBOR]
-            if restrict_output:
-                supported_output_formats = [api_pb2.DATA_FORMAT_CBOR]
-            else:
-                supported_output_formats = [api_pb2.DATA_FORMAT_PICKLE, api_pb2.DATA_FORMAT_CBOR]
+            is_web_endpoint = webhook_config is not None and webhook_config.type != api_pb2.WEBHOOK_TYPE_UNSPECIFIED
```
ehdr, line 893: What's expected if `webhook_config is not None and webhook_config.type == api_pb2.WEBHOOK_TYPE_UNSPECIFIED`? Is that an error state?

ehdr (approved):
Looks good, just one question!

modal/_functions.py
freider: Just me being overly cautious about protobuf filling in "zero valued" objects rather than null for sub-messages. In this case I think webhook_config is usually passed directly by the caller, so it can actually be None, but when used as an unset sub message, it's actually not set to None by the python protobuf lib:
```
f = api_pb2.Function()
assert not f.HasField("webhook_config")  # this checks presence, is not present
assert f.webhook_config is not None  # but is not None when accessed!
assert f.webhook_config.type == api_pb2.WEBHOOK_TYPE_UNSPECIFIED  # and all fields are zeroed

```

PR #3637, opened 2025-10-06
Make container_tests use function definitions instead of duplicating attributes

freider:
Using a new _run_containers_auto that skips all of the manual duplication of function/app definition metadata in tests, and instead relies on collecting that data in a fixture that does a "deploy" of the test.supports.functions App

Rebase that supersedes #3582 

---

## Compatibility checklist

---

## Release checklist

---

</details>


test/container_test.py
```diff
@@ -2731,20 +2649,13 @@ def test_mirrored_input_payload_simple_cls_method(servicer, data_format):
         api_pb2.FunctionGetInputsResponse(inputs=[api_pb2.FunctionGetInputsItem(kill_switch=True)]),
     ]
 
-    ret = _run_container(
+    ret = _run_container_auto(
         servicer,
-        "test.supports.functions",
-        "SimpleCbor.*",
+        "SimpleCls.*",
+        deployed_support_function_definitions,
```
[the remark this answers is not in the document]
freider: This was a bug before - using SimpleCbor would *prevent* mirroring on input to output format, and instead always return cbor. This worked before because the function definition in the test overrode the supported_output_formats

test/container_test.py
```diff
@@ -747,117 +882,67 @@ def test_webhook_setup_failure(servicer):
...
+    #  this would require feeding the servicer with more information about the function
+    #  since it would re-bind parameters to the class service function etc.
+    ret = _run_container_auto(
         servicer,
-        "",
         "SerializedClassWithParams.*",
-        is_class=True,
+        deployed_support_function_definitions,
         inputs=_get_inputs(((), {}), method_name="method"),
-        definition_type=api_pb2.Function.DEFINITION_TYPE_SERIALIZED,
```
freider, line 786: This is the kind of huge duplicate definition stuff that we can omit when using `_run_container_auto` which instead relies on the already recorded definition (by the function decorator + deploy) instead of having devs add hand crafted container args.

This also prevents drift in definition between what the function deploy code (_load) does and what the test executes the container with

test/container_test.py
```diff
@@ -501,22 +630,30 @@ def _get_web_inputs(path="/", method_name=""):
 
 
 @skip_github_non_linux
-def test_success(servicer):
+def test_success(servicer, deployed_support_function_definitions):
```
ehdr, line 633: This one is identical to `test_success_auto` I think?

ehdr (commented):
Tricky to review in depth but overall LGTM!

PR #3639, opened 2025-10-06
Set output_created_at on failed and terminated outputs

rculbertson:
The `output_created_at` field is being set on `FunctionGetOutputsItem` when the output is successful, but not when the output is failed or terminated. We should set in those case too. That field is used to populated the `finished_at` and `event_time` fields in the Input Events that go to clickhouse. If those fields aren't set it will break things in the UI, and anything else powered by input events. 

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/_runtime/container_io_manager.py
+++ b/modal/_runtime/container_io_manager.py
@@ -283,6 +283,7 @@ class IOContext:
             api_pb2.FunctionPutOutputsItem(
                 input_id=input_id,
                 input_started_at=started_at,
+                output_created_at=time.time(),
                 result=api_pb2.GenericResult(status=api_pb2.GenericResult.GENERIC_STATUS_TERMINATED),
                 retry_count=retry_count,
             )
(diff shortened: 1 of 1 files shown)


modal/_runtime/container_io_manager.py
```diff
@@ -283,6 +283,7 @@ async def output_items_cancellation(self, started_at: float):
             api_pb2.FunctionPutOutputsItem(
                 input_id=input_id,
                 input_started_at=started_at,
+                output_created_at=time.time(),
```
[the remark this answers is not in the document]
rculbertson: That is a good point. Fixed.

modal/_runtime/container_io_manager.py
```diff
@@ -278,11 +278,13 @@ async def call_generator_async(self) -> AsyncGenerator[Any, None]:
         logger.debug(f"Finished generator input {self.input_ids}")
 
     async def output_items_cancellation(self, started_at: float):
+        output_created_at = time.time()
         # Create terminated outputs for these inputs to signal that the cancellations have been completed.
         return [
             api_pb2.FunctionPutOutputsItem(
                 input_id=input_id,
                 input_started_at=started_at,
+                output_created_at=output_created_at,
```
thomasjpfan, line 287: Do you see a good/quick place to test this somewhere in https://github.com/modal-labs/modal-client/blob/main/test/container_app_test.py ?

rculbertson: Yes, just pushed a commit adding a test!

PR #3641, opened 2025-10-06
SDK support for App.get_tags() and App.get_tags()

mwaskom:
## Describe your changes

Closes SDK-625

We need this primarily to support tagging Apps created via `modal.App.lookup`, although there are other cases where it might be useful to set or change tags on an App without going through a full deployment.

I considered two additional features but decided against adding them for now (we can add them later if we decide they're useful):
- `retrospective: bool`, which would set the `created_at` time on the tag to match the creation time of the App (rather than the time that `.set_tags()` was called
- `merge: bool`, which would add or overwrite tags in the new set without removing any that are currently attached but absent from the set.

Both ideas suggested enough additional conceptual or implementation complexity that I decided to punt on them for now.

I'll pull the proto changes out into a separate PR (having backtracked on the two "additional features" they're trivial now; originally I thought I might want to discuss the e2e PR before merging the proto changes).

---

## Compatibility checklist

</details>

## Changelog

- Also adds `App.set_tags()` and `App.get_tags()` which should be rolled into the changelog update for tags, generally speaking.


modal/app.py
```diff
@@ -1094,6 +1097,41 @@ def main():
 
         return self
 
+    async def set_tags(self, tags: Mapping[str, str], *, client: Optional[_Client] = None) -> None:
```
mwaskom, line 1094: See discussion on slack about whether we want `client: ...` in here long term, but I think we need it for now.

ehdr (approved):
Setting side the bigger discussion about setting tags before AppPublish, this LGTM!

PR #3643, opened 2025-10-07
Updated docstring for uv_sync

freider:
Updated after some user feedback on this being confusing (which I agree it is)

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Clarifies uv_sync docstring: it does not install the project itself, suggests adding local Python sources separately, explains dependency rebuild avoidance, and notes uv workspaces aren’t supported.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ce8cc6bebc4cae5cd8964ed9277bcfa36947ed6a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/image.py
+++ b/modal/image.py
@@ -1441,10 +1441,12 @@ class _Image(_Object, type_prefix="im"):
         The `pyproject.toml` and `uv.lock` in `uv_project_dir` are automatically added to the build context. The
         `uv_project_dir` is relative to the current working directory of where `modal` is called.
 
-        NOTE: This does *not* install the project or workspace itself into the environment (this is equivalent to the
-        `--no-install-workspace` flag in the `uv sync` command) and you would be expected to add any local python source
-        files using `Image.add_local_python_source` or similar methods after this call. This ensures that updates to
-        your code wouldn't invalidate the uv_sync image cache.
+        NOTE: This does *not* install the project itself into the environment (this is equivalent to the
+        `--no-install-project` flag in the `uv sync` command) and you would be expected to add any local python source
+        files using `Image.add_local_python_source` or similar methods after this call.
+
+        This ensures that updates to your project code wouldn't require reinstalling third-party dependencies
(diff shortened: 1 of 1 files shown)


modal/image.py
```diff
@@ -1441,6 +1441,11 @@ def uv_sync(
         The `pyproject.toml` and `uv.lock` in `uv_project_dir` are automatically added to the build context. The
         `uv_project_dir` is relative to the current working directory of where `modal` is called.
 
+        NOTE: This does *not* install the project or workspace itself into the environment (this is equivalent to the
+        `--no-install-workspace` flag in the `uv sync` command) and you would be expected to add any local python source
+        files using `Image.add_local_python_source` or similar methods after this call. This ensures that updates to
+        your code wouldn't invalidate the uv_sync image cache.
```
mwaskom: Should we actually document that "workspaces" are not supported _at all_ (currently it will error if they are defined).

Also FWIW the final sentence is maybe not a super compelling rationale. Could you rephrase to make it more about, like, "won't require reinstalling all third-party dependencies"?

(That said, this feels more like a limitation to our implementation / consequences of using this in a Modal Image build context, because it's not something you'd need to worry about when using `uv sync` locally.).

freider: Yeah that makes sense. The same thing applies when using `uv` in docker, for which there is a whole article in astral's docs: https://docs.astral.sh/uv/guides/integration/docker/#caching

mwaskom: Yep — but I think it's reasonable to expect that our `uv_sync` would just solve this for users rather than having this purposeful limitation. (Which I think we roughly plan to once we make it possible to add mount layers in the middle of the Image recipe, right?)

thomasjpfan (approved):
This seems reasonable.

PR #3645, opened 2025-10-07
Reorder changelog processing

mwaskom:
Followup to #3635; the fight against slop continues. We're no longer adding cursor comments to the changelog, but we are now creating empty changelog entries because I incorrectly adjusted the order of operations around removing comments and checking if the update is empty.


mwaskom:
@prbot approve

mwaskom:
ugh this didn't work somhow

PR #3646, opened 2025-10-07
modal volume get: Make parallelism configurable and shared across files

dflemstr:
## Describe your changes

Hopefully, it speeds up file transfers of many small files in version 2 volumes by doing more file transfers in parallel.

We already have a reasonable parallel download approach in place when downloading each file from version 2 volumes, but this was handled entirely separately from the download parallelism across files.

### DIFF
--- a/modal/cli/_download.py
+++ b/modal/cli/_download.py
@@ -1,6 +1,7 @@
 # Copyright Modal Labs 2023
 import asyncio
 import functools
+import multiprocessing
 import os
 import shutil
 import sys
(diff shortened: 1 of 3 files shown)


modal/cli/volume.py
```diff
@@ -75,6 +76,9 @@ async def get(
     local_destination: str = Argument("."),
     force: bool = False,
     env: Optional[str] = ENV_OPTION,
+    parallelism: int = Option(
+        default=multiprocessing.cpu_count(), help="Max number of parallel transfers; defaults to number of CPUs"
```
danielnorberg: Number of CPUs is generally a too low concurrency for high-latency remote IO.

I also believe the flag should be named `concurrency`.

danielnorberg:
In the future we'd ideally be smart here and not require users to tune a knob, but allowing users to explicitly control concurrency is probably a good quick fix.

modal/volume.py
```diff
@@ -670,7 +670,12 @@ async def iter_urls() -> AsyncGenerator[str]:
 
     @live_method
     async def read_file_into_fileobj(
-        self, path: str, fileobj: typing.IO[bytes], progress_cb: Optional[Callable[..., Any]] = None
+        self,
+        path: str,
+        fileobj: typing.IO[bytes],
+        concurrency: int = multiprocessing.cpu_count(),
+        download_semaphore: Optional[asyncio.Semaphore] = None,
+        progress_cb: Optional[Callable[..., Any]] = None,
```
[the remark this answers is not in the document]
dflemstr: This is intentional

modal/cli/_download.py
```diff
@@ -23,12 +24,19 @@ async def _volume_download(
     remote_path: str,
     local_destination: Path,
     overwrite: bool,
-    progress_cb: Callable,
+    concurrency: int = max(128, 2 * multiprocessing.cpu_count()),
```
thomasjpfan: Can we make the default `None` and then call `max(128, 2 * multiprocessing.cpu_count())` in the body?

In the current implementation, `max(...)` will be call at import time even if you are not using `_volume_download`.

dflemstr: Sounds good

PR #3647, opened 2025-10-07
Fix changelog extraction more good maybe

mwaskom:
Follow up to #3645


ehdr (approved):
Maybe we're approaching the point where it's good return on time to add tests for this?! 😅

PR #3648, opened 2025-10-08
Remove ability to reference methods as Functions directly on Cls

freider:
This will now return a `PartialFunction` instead.

Not sure if we should be adding some helper `__getattr__` on `PartialFunction` to help users who will undoubtedly still make this mistake, thinking they have gotten a `Function` they can use?

---

## Compatibility checklist

---

## Release checklist

---

</details>

### DIFF
--- a/modal/cls.py
+++ b/modal/cls.py
@@ -863,16 +863,6 @@ More information on class parameterization can be found here: https://modal.com/
         )
 
     def __getattr__(self, k):
-        # TODO: remove this method - access to attributes on classes (not instances) should be discouraged
-        if not self._is_local() or k in self._method_partials:
-            # if not local (== k *could* be a method) or it is local and we know k is a method
-            deprecation_warning(
-                (2025, 1, 13),
-                "Calling a method on an uninstantiated class will soon be deprecated; "
-                "update your code to instantiate the class first, i.e.:\n"
-                f"{self._name}().{k} instead of {self._name}.{k}",
-            )
-            return getattr(self(), k)
         # non-method attribute access on local class - arguably shouldn't be used either:
         return getattr(self._user_cls, k)
 


mwaskom:
> Not sure if we should be adding some helper __getattr__ on PartialFunction to help users who will undoubtedly still make this mistake, thinking they have gotten a Function they can use?

Do you mean so that something like

```
MyCls.run.remote()
```

Says something useful instead of just "PartialFunction has no attribute 'remote'"?

modal/cls.py
```diff
@@ -863,18 +863,28 @@ def __call__(self, *args, **kwargs) -> _Obj:
...
-            deprecation_warning(
-                (2025, 1, 13),
-                "Calling a method on an uninstantiated class will soon be deprecated; "
-                "update your code to instantiate the class first, i.e.:\n"
-                f"{self._name}().{k} instead of {self._name}.{k}",
+        if self._user_cls is not None:
+            # local class, we can check if there are static attributes and let the user access them
+            # except if they are PartialFunction (i.e. methods)
+            v = getattr(self._user_cls, k)
+            if not isinstance(v, modal.partial_function.PartialFunction):
```
mwaskom, line 870: 😵‍💫 double checking that we wouldn't expect these to be the internal `_PartialFunction` type here.

modal/cls.py
```diff
@@ -863,18 +863,28 @@ def __call__(self, *args, **kwargs) -> _Obj:
...
+                "You can't access methods on a Cls directly - Did you forget to instantiate the class first?\n"
+                "e.g. instead of MyClass.method.remote(), do MyClass().method.remote()"
             )
-            return getattr(self(), k)
-        # non-method attribute access on local class - arguably shouldn't be used either:
-        return getattr(self._user_cls, k)
+
+        return _Function._from_loader(
+            method_loader,
+            rep=f"Maybe({self._name}.{k})",
```
mwaskom: :/

test/cls_test.py
```diff
@@ -1135,21 +1135,32 @@ class C:
         def method(self):
             pass
 
-    assert len(recwarn) == 0
-    with pytest.raises(AttributeError):
-        C.blah  # type: ignore   # noqa
-    assert len(recwarn) == 0
+    # The following should warn since it's accessed on the class directly
+    with pytest.raises(AttributeError, match="Did you forget to instantiate the class first?"):
+        C.method.remote()  # noqa  # triggers a deprecation warning
```
mwaskom: Why is there a deprecation *warning* still?

modal/cls.py
```diff
@@ -863,18 +863,28 @@ def __call__(self, *args, **kwargs) -> _Obj:
...
+            # except if they are PartialFunction (i.e. methods)
+            v = getattr(self._user_cls, k)
+            if not isinstance(v, modal.partial_function.PartialFunction):
+                return v
+
+        # We create a synthetic dummy Function that is guaranteed to raise an AttributeError when
+        # a user tries to use any of its "live methods" - this lets us raise exceptions for users
+        # only if they try to access methods on a Cls as if they were methods on the instance.
+        async def method_loader(fun: _Function, resolver: Resolver, existing_object_id: Optional[str]):
+            raise AttributeError(
```
mwaskom, line 877: Not sure what the right exception type is. Something to clean up eventually?

mwaskom: I'm not sure `AttributeError` is right. For normal Python classes this is a TypeError: the method attribute exists, but it's not been bound and so the arglist is incomplete.

I'm still not sure exactly what error type to captures the modal-specific issue. It turns out that `modal.Cls` is really not exactly like a class....

modal/cls.py
freider: Yeah this is a bit tbh 😵 But `@method` is itself a synchronicity-wrapped method, so it will output *wrapped* types (PartialFunction in this case) that will be part of the class namespace of the underlying class, which is what we're getting here. It would break the tests if this wasn't so, so even if this was to change we would detect it in CI :)

modal/cls.py
```diff
@@ -863,18 +863,28 @@ def __call__(self, *args, **kwargs) -> _Obj:
...
-                (2025, 1, 13),
-                "Calling a method on an uninstantiated class will soon be deprecated; "
-                "update your code to instantiate the class first, i.e.:\n"
-                f"{self._name}().{k} instead of {self._name}.{k}",
+        if self._user_cls is not None:
+            # local class, we can check if there are static attributes and let the user access them
+            # except if they are PartialFunction (i.e. methods)
+            v = getattr(self._user_cls, k)
+            if not isinstance(v, modal.partial_function.PartialFunction):
+                return v
```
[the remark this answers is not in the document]
freider: Intentional, we want a regular AttributeError in those cases

modal/cls.py
freider: Ah this wasn't intended to be the final name :) What about `UnboundMethod` ?

test/cls_test.py
freider: There isn't, but there is a comment about there being one :P Removing

PR #3649, opened 2025-10-08
Suppress a number of App properties from the docs

mwaskom:
There are a number of properties on the `App` object that are exposed publicly despite not really being intended to provide user-facing utility. Many of these are used to hold state while building the App and don't reflect the server-side data model. This has become a more acute source of confusion after the introduction of `modal.App.lookup`.

I think that we're basically on the path to deprecating all of these properties. For some of them (e.g., `modal.App.registered_functions`) I think it makes sense to develop a more complete public API that works as expected in the lookup case, i.e., you should be able to look up an App and then retrieve all of its Functions. Other properties should probably just be removed, i.e. `App.image` reveals a convenience attribute that is not really part of the App data model.

As a first step I'm just going to suppress these from the docs and add notes discouraging their use.


modal/app.py
```diff
@@ -227,7 +227,12 @@ def name(self) -> Optional[str]:
 
     @property
     def is_interactive(self) -> bool:
-        """Whether the current app for the app is running in interactive mode."""
+        """mdmd:hidden
+        Whether the current app for the app is running in interactive mode.
+
+        Note: this method will like the deprecated in the future.
```
ehdr:
```suggestion
        Note: this method will likely be deprecated in the future.
```

modal/app.py
```diff
@@ -515,37 +533,48 @@ def _init_container(self, client: _Client, running_app: RunningApp):
...
 
     @property
     def registered_entrypoints(self) -> dict[str, _LocalEntrypoint]:
-        """All local CLI entrypoints registered on the app.
+        """mdmd:hidden
+        All local CLI entrypoints registered on the app.
 
         Note: this property is populated only during the build phase, and it is not
         expected to work when a deplyoed App has been retrieved via `modal.App.lookup`.
+        This method is likely to be deprecated in the future in favor of a different
```
freider: a different... ?

freider (approved):
lgtm

modal/app.py
mwaskom: whoops didn't hit [tab] enough times

PR #3661, opened 2025-10-11
Split out all local/definition state for an App into a its own struct

freider:
Driveby improvement/bike sheding while working on a more consistent "client attachment" logic.

This moves all of the variables that are only used in the scope that has the App definition locally available, not necessarily with a "running" app (i.e. mostly things that are available after loading the module that an App is defined in)

This is by no means a public API and things can move around between this state and other state vars in the future as we add more functionality to App.from_name/lookup for example, but I think this will make navigating the code and making changes much easier since you don't have state that is possibly set differently in different scenarios, and it makes it clear in which contexts you'll have access to each of these variables.

I think we want something similar to this for Function (and Cls/Obj...) too

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Introduce `_LocalAppState` to encapsulate local-only App state and refactor App, Cls, and Runner to read/write through it, including publish/create flows and web endpoint tracking.
> 
> - **App (`modal/app.py`)**:
>   - **New struct**: Add dataclass `_LocalAppState` holding local-only state (`functions`, `classes`, `image_default`, `secrets_default`, `volumes_default`, `include_source_default`, `web_endpoints`, `local_entrypoints`, `tags`).
>   - **Refactor `_App`**:
>     - Store local state in `_local_state_attr` with `_local_state` accessor; `lookup()` sets it to `None` for non-local apps.
>     - Read/write defaults via local state (image, secrets, volumes, include_source); update registration and getters to use `local_state`.
>     - Update `_add_function`/`_add_class`, hydration, and registered* accessors to use `local_state`.
>     - `_get_default_image`, `function()`/`cls()` decorators use local defaults from `local_state`.
>     - Track class method web endpoints in `local_state.web_endpoints`; `include()` merges objects and `tags` via local state.
> - **Cls (`modal/cls.py`)**:
>   - Adjust `from_local` to only register lifecycle methods; move web endpoint registration out (handled in `App.cls`).
> - **Runner (`modal/runner.py`)**:
>   - Update typing to reference `modal.app._App`.
>   - Change `_create_all_objects`/`_publish_app` to accept/use `_LocalAppState` (functions/classes/tags) and update all call sites.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a05b6f26dda7b89ef1b06b7f7ccbce3873aa836e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


modal/app.py
```diff
@@ -114,6 +115,21 @@ def __call__(
     def __call__(self, func): ...
 
 
+@dataclass
+class _LocalAppState:
+    """All state for apps that's part of the local/definition state"""
+
+    functions: dict[str, _Function]
+    classes: dict[str, _Cls]
```
mwaskom, line 123: `clss`

modal/app.py
```diff
@@ -169,6 +177,11 @@ def foo():
 
     _include_source_default: Optional[bool] = None
```
thomasjpfan: `_include_source_default` should be removed here too?

modal/app.py
```diff
@@ -169,6 +177,11 @@ def foo():
...
 
+    def _get_local_state(self) -> _LocalAppState:
```
mwaskom: Minor but kind of annoying that this is a method and not a property, since it's cheap

modal/app.py
```diff
@@ -169,6 +177,11 @@ def foo():
...
+        if self._local_state is None:
+            raise InvalidError("Local state is not initialized")
```
mwaskom: Is this expected to ever be user facing?

modal/app.py
```diff
@@ -114,6 +115,21 @@ def __call__(
...
+
+    functions: dict[str, _Function]
+    classes: dict[str, _Cls]
+    image: Optional[_Image]
+    secrets: Sequence[_Secret]
+    volumes: dict[Union[str, PurePosixPath], _Volume]
+    web_endpoints: list[str]  # Used by the CLI
+    local_entrypoints: dict[str, _LocalEntrypoint]
+    tags: dict[str, str]
+    include_source_default: bool
```
mwaskom, line 129: Naming nit but it's weird to call this one out as a `default` since that's the case for e.g. `image` too.

modal/app.py
freider: No, the error is purely for us to detect regressions in tests in a straight forward manner. Maybe making it a property and raising an AttributeError instead would make sense

modal/app.py
freider: I agree it's inconsistent (I just kept the old names). I think all of them should be called `_default` though, since that more clearly communicates what they are used for - and that changing them post-function creation will have no effect etc.

modal/app.py
freider: ah yes!

modal/app.py
freider: Done

modal/app.py
freider: Fixed

modal/app.py
freider: Renamed the others to `_default` too now. It was pretty nice to have things scoped under this type - made it trivial to rename across the code base :)

PR #3666, opened 2025-10-13
Implement TaskCommandRouterClient

saltzm:
## Describe your changes

This implements the bulk of the RPC logic for the new exec implementation. Eventual usage for init will be something like:

```py
def exec(...):
       # ...
        kwargs = {
            "task_id": task_id,
            "pty_info": pty_info,
            "stdout": stdout,
            "stderr": stderr,
            "timeout": timeout,
            "workdir": workdir,
            "secrets": secrets,
            "text": text,
            "bufsize": bufsize,
        }
        # NB: This must come after the task ID is set, since the sandbox must be
        # scheduled before we can create a router client.
        if (command_router_client := await self._get_command_router_client(task_id)) is not None:
            kwargs["command_router_client"] = command_router_client
            return await self._exec_through_command_router(*args, **kwargs)
        else:
            return await self._exec_through_server(*args, **kwargs)
 ```

Retry logic was mostly ported from [`retry_transient_errors`](https://github.com/modal-labs/modal-client/blob/55355bfb5c1f0f95d1436f06dfdd9213b43d09f3/modal/_utils/grpc_utils.py#L177-L272) which expects a `UnaryUnaryWrapper`.

Draft PR showing how this is used in the new StreamReader implementation if it helps for context (still polishing that PR):
https://github.com/modal-labs/modal-client/pull/3667

---

## Compatibility checklist

</details> 

---

> [!NOTE]
> Introduce a direct TaskCommandRouter client with auth-refresh and robust retry logic for exec start/stdio/stdin/poll/wait, plus config and exception additions.
> 
> - **Client**:
>   - **`modal/_utils/task_command_router_client.py`**: New `TaskCommandRouterClient` for direct worker router access.
>     - Supports `exec_start`, `exec_stdio_read` (streaming with offset resume), `exec_stdin_write`, `exec_poll`, and `exec_wait`.
>     - Built-in exponential backoff for transient errors; special handling for `StreamTerminatedError` and grpclib quirks.
>     - Automatic JWT attachment, proactive refresh (background loop), and on-demand refresh on `UNAUTHENTICATED`.
>     - Per-call deadlines and `ExecTimeoutError` on deadline exceed.
>     - Secure channel setup with optional insecure TLS via config.
> - **Config**:
>   - Add `task_command_router_insecure` setting to allow insecure TLS for local/testing.
> - **Exceptions**:
>   - Add `ExecTimeoutError` for exec deadline/timeouts.
> - **Tests**:
>   - **`test/task_command_router_client_test.py`**: Extensive unit tests covering streaming, retries, auth refresh, deadlines, and error paths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fa293f3409e4f2932cabdee150615be5fad384e0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_utils/task_command_router_client.py
+++ b/modal/_utils/task_command_router_client.py
@@ -43,7 +43,7 @@ def _parse_jwt_expiration(jwt_token: str) -> Optional[float]:
             return float(exp)
     except Exception:
         # Avoid raising on malformed tokens; fall back to server-driven refresh logic.
-        logger.warning(f"Failed to parse JWT expiration: {jwt_token}")
+        logger.warning("Failed to parse JWT expiration")
         return None
     return None
 
@@ -388,7 +388,8 @@ class TaskCommandRouterClient:
         except GRPCError as exc:
             if exc.status == Status.UNAUTHENTICATED:
                 await self._refresh_jwt()
-                return await func()
+                # Retry with the original arguments preserved
+                return await func(*args, **kwargs)
             raise
 
     async def _jwt_refresh_loop(self) -> None:
--- a/test/task_command_router_client_test.py
+++ b/test/task_command_router_client_test.py
@@ -656,3 +656,56 @@ async def test_exec_stdio_read_deadline_exceeded_on_open_raises_exec_timeout_err
     # Should not significantly exceed the deadline (allow small overhead)
     assert elapsed_secs < 0.2
     await client.close()
+
+
+@pytest.mark.asyncio
+async def test_exec_wait_succeeds_after_auth_retry(monkeypatch):
+    client = TaskCommandRouterClient(
+        server_client=None,
+        task_id="sb-1",
+        server_url="https://router.test",
+        jwt="t",
+        channel=create_dummy_channel(),
+    )
+
+    calls: List[tuple[sr_pb2.TaskExecWaitRequest, Optional[float]]] = []
+
+    class _WaitMethod:
+        def __init__(self):
+            self._first = True
+
+        async def __call__(self, request: sr_pb2.TaskExecWaitRequest, *, timeout: Optional[float] = None):
+            calls.append((request, timeout))
+            if self._first:
+                self._first = False
+                raise GRPCError(Status.UNAUTHENTICATED, "auth required")
+            # Return a successful response on retry
+            return sr_pb2.TaskExecWaitResponse(code=0)
+
+    class _Stub2:
+        def __init__(self):
+            self.TaskExecWait = _WaitMethod()
+
+    refreshes = 0
+
+    async def _refresh_jwt():
+        nonlocal refreshes
+        refreshes += 1
+
+    client._refresh_jwt = _refresh_jwt  # type: ignore[assignment]
+    client._stub = _Stub2()  # type: ignore[assignment]
+
+    # No explicit deadline -> internal call uses timeout=60 and should preserve it across auth retry
+    resp = await client.exec_wait("task-1", "exec-1")
+    assert isinstance(resp, sr_pb2.TaskExecWaitResponse)
+    assert resp.code == 0
+
+    # Should have attempted twice: once failing with UNAUTHENTICATED, once succeeding.
+    assert len(calls) == 2
+    req1, to1 = calls[0]
+    req2, to2 = calls[1]
+    assert req1.task_id == "task-1" and req1.exec_id == "exec-1"
+    assert req2.task_id == "task-1" and req2.exec_id == "exec-1"
+    assert to1 == 60 and to2 == 60
+    assert refreshes == 1
+    await client.close()

### CODE REVIEW

File: modal/_utils/task_command_router_client.py [hunk 1]
```diff
@@ -0,0 +1,517 @@
...
+                logger.debug(f"Command router access is not enabled for task {task_id}")
+                return None
+            raise
+
+        logger.debug(f"Using command router access for task {task_id}")
+
+        # Build and connect a channel to the task command router now that we have access info.
+        o = urllib.parse.urlparse(resp.url)
+        if o.scheme != "https":
+            raise ValueError(f"Task router URL must be https, got: {resp.url}")
```
saltzm, line 138: Wondering if I should wrap this and any other errors that might escape this in some more user-friendly kind of error that they can catch

modal/_utils/async_utils.py
```diff
@@ -878,3 +878,36 @@ async def async_chain(*generators: AsyncGenerator[T, None]) -> AsyncGenerator[T,
...
+@asynccontextmanager
+async def timeout_at(deadline: Optional[float]):
+    """Cancel the current task when the absolute deadline is reached.
+
+    Python 3.9-compatible alternative to asyncio.timeout_at(). Converts CancelledError to
+    asyncio.TimeoutError when the deadline fires, so callers can catch TimeoutError.
+    """
+    if deadline is None:
+        yield
+        return
```
paulgb: might be worth adding a comment on the behavior when `deadline = None`. I understand why it's helpful for ergonomics when used as a context manager, but it might not be obvious to a reader why it's useful to be able to run this as a no-op.

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,517 @@
...
+        parts = jwt_token.split(".")
+        if len(parts) != 3:
+            return None
+        payload_b = _b64url_decode(parts[1])
+        payload = json.loads(payload_b)
+        exp = payload.get("exp")
+        if isinstance(exp, (int, float)):
+            return float(exp)
+    except Exception:
+        # Avoid raising on malformed tokens; fall back to server-driven refresh logic.
```
paulgb, line 45: I expect this to be rare; it would be nice to log a little warning so that if we ever suspect that this branch is being hit we can verify it.

modal/_utils/async_utils.py
```diff
@@ -878,3 +878,36 @@ async def async_chain(*generators: AsyncGenerator[T, None]) -> AsyncGenerator[T,
                 logger.exception(f"Error closing async generator: {e}")
         if first_exception is not None:
             raise first_exception
+
+
+@asynccontextmanager
+async def timeout_at(deadline: Optional[float]):
```
saltzm: In the process of removing this because I found a bug that meant it didn't do what I wanted it to

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,515 @@
...
+        async with self._jwt_refresh_lock:
+            if self._closed:
+                return
+            resp = await fetch_command_router_access(self._server_client, self._task_id)
+            # Ensure the server URL remains stable for the lifetime of this client.
+            assert resp.url == self._server_url, "Task router URL changed during session"
+            self._jwt = resp.jwt
+            self._jwt_exp = _parse_jwt_expiration(resp.jwt)
+            # Wake up the background loop to recompute its next sleep.
+            self._jwt_refresh_event.set()
```
paulgb, line 383: the lock will prevent us from refreshing multiple times _concurrently_, but there is still a chance that multiple callers will stack up on the lock and cause refreshes when they are not necessary.

Probably the simplest solution is to check `self._jwt_exp` immediately after the lock is acquired to see if it still needs refreshing.

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,515 @@
...
+from modal_proto import api_pb2, task_command_router_pb2 as sr_pb2
+from modal_proto.task_command_router_grpc import TaskCommandRouterStub
+
+from .grpc_utils import RETRYABLE_GRPC_STATUS_CODES, connect_channel, retry_transient_errors
+
+
+def _b64url_decode(data: str) -> bytes:
+    """Decode a base64url string with missing padding tolerated."""
+    padding = "=" * (-len(data) % 4)
+    return base64.urlsafe_b64decode(data + padding)
```
paulgb, line 27: (no action required) so annoying that python makes you do this when unpadded urlsafe b64 is generally RFC compliant.

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,515 @@
...
+            # these attributes.
+            task = getattr(self, "_jwt_refresh_task", None)
+            if task is not None:
+                try:
+                    task.cancel()
+                except Exception:
+                    pass
+            channel = getattr(self, "_channel", None)
+            if channel is not None:
+                channel.close()
```
thomasjpfan: Can channel be `None`? If not:

```suggestion
            self._channel.close()
```

Otherwise, was `close` suppose to set `self._channel = None`?

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,515 @@
...
+        except Exception:
+            pass
+
+    async def close(self) -> None:
+        """Close the client and stop the background JWT refresh task."""
+        self._closed = True
```
thomasjpfan, line 224: Should this early exit if `self._closed` is `True`?

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,515 @@
...
+    except Exception:
+        # Avoid raising on malformed tokens; fall back to server-driven refresh logic.
+        return None
+    return None
+
+
+async def call_with_retries_on_transient_errors(
+    func,
+    *,
+    base_delay: float = 0.01,
```
paulgb: this applies generally to several places in the PR: anywhere that a number is used as a unit of time, I would encourage adding the unit in the name, e.g. `base_delay_secs`

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,515 @@
...
+                await self._refresh_jwt()
+            except asyncio.CancelledError:
+                logger.debug(f"Cancelled JWT refresh loop for exec with task ID {self._task_id}")
+                break
+            except Exception as e:
+                logger.warning(f"Background JWT refresh failed for exec with task ID {self._task_id}: {e}")
+                try:
+                    await asyncio.sleep(1.0)
+                except Exception:
+                    # Ignore sleep issues; loop will re-check closed flag.
```
thomasjpfan, line 435: When can sleep raise an exception here?

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,515 @@
...
+        if o.scheme != "https":
+            raise ValueError(f"Task router URL must be https, got: {resp.url}")
+
+        host, _, port_str = o.netloc.partition(":")
+        port = int(port_str) if port_str else 443
+        ssl_context = ssl.create_default_context()
+
+        # Allow insecure TLS when explicitly enabled via config.
+        if config["task_command_router_insecure"]:
+            logger.warning("Using insecure TLS for task command router due to MODAL_TASK_COMMAND_ROUTER_INSECURE")
```
thomasjpfan, line 146: Is there any risk for a user doing this?

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,515 @@
...
+            self._jwt_refresh_event.set()
+
+    async def _call_with_auth_retry(self, func):
+        try:
+            return await func()
+        except GRPCError as exc:
+            if exc.status == Status.UNAUTHENTICATED:
+                await self._refresh_jwt()
+                return await func()
+            raise
```
thomasjpfan: Nit: Can we change this to?

```suggestion
    async def _call_with_auth_retry(self, func, *args, **kwargs):
        try:
            return await func(*args, **kwargs)
        except GRPCError as exc:
            if exc.status == Status.UNAUTHENTICATED:
                await self._refresh_jwt()
                return await func()
            raise
```

So that we do not need double lambdas for the call sites:

```
await call_with_retries_on_transient_errors(
    lambda: self._call_with_auth_retry(self._stub.TaskExecPoll, request)
)
```

test/task_command_router_client_test.py
```diff
@@ -0,0 +1,667 @@
...
+
+    fake_stub = _Stub(_open)
+    client._stub = fake_stub  # type: ignore[assignment]
+
+    out: List[bytes] = []
+    async for item in client.exec_stdio_read("task-1", "exec-1", api_pb2.FILE_DESCRIPTOR_STDOUT):
+        out.append(item.data)
+
+    assert out == pieces
+    await client.close()
```
thomasjpfan, line 96: If the exception gets raised, then the client would not close. Although, I do not think it impacts the test.

Also, should there be test to assert behavior of a closed client?

modal/_utils/async_utils.py
saltzm: I ended up removing this because there was a bug in exec_stdio_read that basically caused this to not work properly which sort of rendered it... useless, since the whole point of it was to act as a backstop for bugs or missing logic in exec_stdio_read. So now the deadline is handled entirely in exec_stdio_read

modal/_utils/task_command_router_client.py
saltzm: Good call, fixed

modal/_utils/task_command_router_client.py
saltzm: Yeah this was LLM generated and I was scratching my head and looked it up, seems silly

modal/_utils/task_command_router_client.py
saltzm: Good call, I also prefer this, just forgot. Think I caught all the locations and fixed them

modal/_utils/task_command_router_client.py
saltzm: I doubt it, it just seemed worth logging even for local/dev use. Can remove it if you prefer

modal/_utils/task_command_router_client.py
saltzm: Good question. I think channel would be `None` here if the attribute didn't exist, if `__init__` threw before setting `_channel`. Reordered things in `__init__` so the `getattr` stuff isn't needed anymore and this is much simpler now

modal/_utils/task_command_router_client.py
saltzm: Good call. Did that here and in `__del__`

modal/_utils/task_command_router_client.py
saltzm: I think if the task is canceled by `close()` or `__del__` it'll raise an `asyncio.CancelledError`

modal/_utils/task_command_router_client.py
saltzm: Nice, done

test/task_command_router_client_test.py
saltzm: Hmm good question, I think since close() is only going to be used in tests (since as we discussed IRL there's no good place to close it in sandbox.py) we don't need to test that behavior, but if it were ever used outside of tests I think it's worth doing

modal/_utils/task_command_router_client.py
```diff
@@ -0,0 +1,524 @@
...
+            self._jwt_refresh_event.set()
+
+    async def _call_with_auth_retry(self, func, *args, **kwargs):
+        try:
+            return await func(*args, **kwargs)
+        except GRPCError as exc:
+            if exc.status == Status.UNAUTHENTICATED:
+                await self._refresh_jwt()
+                return await func()
+            raise
```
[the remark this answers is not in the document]
saltzm: Added test for this and fixed

modal/_utils/task_command_router_client.py
thomasjpfan: I'm okay with keeping it.

PR #3667, opened 2025-10-14
Implement new StreamReader that uses task command router

saltzm:
## Describe your changes

Implements new StreamReader functionality for the new exec path.

---

## Compatibility checklist

---

</details>

---

> [!NOTE]
> Introduce StreamReader implementations for task command router (bytes/text, by-line), add incremental UTF-8 decoding and line-splitting utilities, adjust container attach to use underlying impl, and compile new proto in CI.
> 
> - **I/O Streams**:
>   - **New StreamReader architecture**: Add `_StreamReaderThroughServer`, `_TextStreamReaderThroughCommandRouter`, `_BytesStreamReaderThroughCommandRouter`, and `_DevnullStreamReader`, with wrapper `_StreamReader` delegating based on `TaskCommandRouterClient`.
>   - **Helpers**: Add `_decode_bytes_stream_to_str` (incremental UTF-8 decoding) and `_stream_by_line` (line buffering).
>   - **APIs/Errors**: Support `ExecTimeoutError`; expose `file_descriptor`, `read`, async iteration, and `aclose` via delegation.
> - **Container Process**:
>   - `attach()`: route reads through underlying stream implementation (`_impl._get_logs`) for correct boot/liveness handling.
> - **Tests**:
>   - Add unit tests for `_stream_by_line` and `_decode_bytes_stream_to_str`; update imports and coverage for timeouts and retries.
> - **CI**:
>   - Protoc step now also compiles `modal_proto/task_command_router.proto`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b3aed72a1f052401a1e3dc1e8fc592838bd0afdc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -176,7 +176,7 @@ jobs:
           elif [ "${{ matrix.python-version }}" == "3.13" ]; then
             pip install grpcio-tools==1.66.2 grpclib==0.4.7;
           fi
-          python -m grpc_tools.protoc --python_out=. --grpclib_python_out=. --grpc_python_out=. -I . modal_proto/api.proto
+          python -m grpc_tools.protoc --python_out=. --grpclib_python_out=. --grpc_python_out=. -I . modal_proto/api.proto modal_proto/task_command_router.proto
           python -m grpc_tools.protoc --plugin=protoc-gen-modal-grpclib-python=protoc_plugin/plugin.py --modal-grpclib-python_out=. -I . modal_proto/api.proto
           deactivate
 


modal/io_streams.py
```diff
@@ -166,19 +145,7 @@ def file_descriptor(self) -> int:
...
-        ```python fixture:running_app
-        from modal import Sandbox
-
-        sandbox = Sandbox.create("echo", "hello", app=running_app)
-        sandbox.wait()
-
-        print(sandbox.stdout.read())
-        ```
-        """
+        """Fetch the entire contents of the stream until EOF."""
```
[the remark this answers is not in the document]
saltzm: Not a bug, __aiter__ is only really needed in the wrapper class, and it's used in the other command router implementation classes because the `read()` function makes use of it

saltzm:
These unit test failures don't happen locally once I have all my changes in, only when I have this one branch selected without the rest of the changes. Need to dig in and figure out what's missing here

saltzm:
Okay, should be fixed now

modal/io_streams.py
```diff
@@ -133,10 +117,6 @@ def __init__(
         if object_type == "sandbox" and not text:
             raise ValueError("Sandbox streams must have text mode enabled.")
 
-        # line-buffering is only supported when text=True
-        if by_line and not text:
-            raise ValueError("line-buffering is only supported when text=True")
-
```
paulgb, line 139: is this not still relevant?

modal/io_streams.py
```diff
@@ -356,6 +319,289 @@ async def aclose(self):
...
+    """Yield complete lines only (ending with \n), buffering partial lines until complete."""
+    line_buffer = b""
+    async for message in stream:
+        assert isinstance(message, bytes)
+        line_buffer += message
+        while b"\n" in line_buffer:
+            line, line_buffer = line_buffer.split(b"\n", 1)
+            yield line + b"\n"
+
+    if line_buffer:
```
paulgb, line 349: nit: I prefer making explicit comparisons instead of relying on truthiness of non-boolean values, i.e. `if len(line_buffer) > 0` or `if line_buffer != b""`

To me this makes the types and intent of the code clearer to a reader, and (though not relevant in this case) avoids a certain set of bugs.

modal/io_streams.py
```diff
@@ -83,27 +86,8 @@ async def _container_process_logs_iterator(
...
-        "-c",
-        "for i in $(seq 1 10); do echo foo; sleep 0.1; done",
-        app=running_app,
-    )
-    for message in sandbox.stdout:
-        print(f"Message: {message}")
-    ```
-    """
+class _StreamReaderThroughServer(Generic[T]):
+    """A StreamReader implementation that reads from the server."""
```
paulgb, line 90: I'm curious why this is generic over T. Isn't T basically `str | bytes` in any case?

modal/io_streams.py
saltzm: Moved it to the top-level `_StreamReader` [here](https://github.com/modal-labs/modal-client/pull/3667/files#diff-b4571b0cd2747ddf2af84eb133d7d5ae204abc2d9d368b69766463686558adf4R529-R530) so it applies in both paths

modal/io_streams.py
saltzm: No idea honestly, maybe as an easy way to have `read(self) -> T` have a specific dynamically chosen return type?

modal/io_streams.py
saltzm: Yeah, I agree with you, I just assumed doing this was more "Pythonic" or something, though I don't know if that's a great reason to keep doing it or not

saltzm: I do think it's generally well-defined to do this for an array/container type so less bug-prone than doing it for something that's e.g. an optional array or something

saltzm: I asked Michael and he said this was the idiomatic/safe way to do this, so I just left it as is

PR #3669, opened 2025-10-15
Make all "constructors" take a `client` optional arg

freider:
## Goals of this patch:
* Allow all `.from_name()`, `.from_id()` etc. user-facing "constructors" to take an optional `client` object
* Get rid of all `set_env_client` fixtures in tests - all test code should be functioning by passing in `Client` objects explicitly through various means

Closes SDK-632

## Why this is hard
* `Cls` and `Function` can either be hydrated as part of an `app.deploy(client=...)/run(client=...)` *or* lazily
* In the deploy hydration case ("non-lazy"), the client isn't actually available when the `Function` object is created, since the App isn't "live" at decoration time
* There are various ways in which we create lazy "clones" of object that should "inherit" their original's `client`, e.g. `Cls.with_options()`. This clone can also be created *before* `client` has been set on its parent, and once the client is set on the parent it should transfer to the clones as well 😵 
* Objects may initialize "load" of independent other objects when they are themselves loaded, in which case they should transfer their own `Client` (and environment etc.) to the independent "dependency" (e.g. a `Sandbox.create()` with a set of `Secret`s as deps.
* If a "dependency" lacks a specific client, it should use its parent's `Client`. If it gets a client from its parent it should not create a `from_env` client unnecessarily

## Bonus goal
* The same rules for `client` above applies to `environment_name` as well where it's applicable, and to some degree `app_id` - these are also load-time metadata that can either be explicitly injected or inferred from parents or the runtime.

## Change summary
- Add a new `LoadContext` type and concept, that encapsulates optional values of `client`, `environment_name` and `app_id` that were previously part of the `Resolver`
- Remove the `client`, `environment_name` and `app_id` fields from `Resolver`
- Add optional overrides of all LoadContext vars into `modal._Objects` via a `object._load_context` field on unhydrated construction (typically from provided `client` and `environment_name` values in `.from_*` constructors.
- `App` also gets an internal `LoadContext` (initially empty on `App()` construction)
- The signature of the `_load()` loader method on `_Object` now receives both a `Resolver` and a `LoadContext` object and load-code should use `load_context.client` etc. for rpcs and to set the `_Object._client` when hydrating.
The new "loader function" signature is:
```py
def _load(instance: _Object, resolver: Resolver, load_context: LoadContext, existing_object_id: Optional[str]) -> None
```
The `LoadContext` passed is guaranteed to contain a non-optional `Client` and `environment_name`, and an `app_id` if one exists.
- The `Resolver` instance that is now passed to loaders only serves to load other instances explicitly via `.load(other_obj)`. `Resolver.load()` call will now take both an object and a "parent LoadContext". The Resolver then "merges" into each loaded Object's passed parent load context with whatever the object has optionally specified as overrides in `self._load_context`. This makes sure that we prioritize the context overrides of each object, but allow the "parent" context to propagate to any child that doesn't have overrides. Most of the dependency loading is still handled automatically via "deps" passed to the `Object._from_loader()` contstructor.

## Slightly hacky: `LoadContext` sharing for `Cls`/`Function` "clones"
While normally we give each object its own LoadContext overrides instance, in the case of `Cls` (and a future `Function.with_options()`) we need to make sure that all related entities share the same LoadContext *reference*.
* This is because the "load dag" of Cls->Obj->Method is "inverted" from a load context perspective - the optional `client` would be injected at `Cls.from_name(client=...)`, but the "leaf" _Object (a `Function` representing a method on the class) has no such override mechanims in itself. The method `Function` is however top level/"root" object when resolving dependencies during lazy loading, so in order for it to get the 
* By making a `Cls` reuse its load context from the `Cls` itself when creating its methods, we make sure that whatever client specified as an override on Cls is also used when loading the method `Function`. 
* In addition we also make any "app creation" Functions and Cls use a reference to an app associated LoadContext, which means this load context gets referenced by all Functions, Cls, Obj and Methods from that app. We then make sure `app.run` updates this load context in place, effectively propagating any load_context overrides from app launch to the lazy-loaded "Method" objects that might get referenced by users. This fixes the messy case of:

```
app = App()
@app.cls()
class F:
     @method
     def foo(self): ...

f = F()

with app.run(client=SomeClient):
    
```

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Introduce LoadContext to carry client/environment/app_id, refactor loading to use it, and add optional client to all from_name/from_id constructors; update App/Resolver/Object/Cls/Function flows and tests accordingly.
> 
> - **Core architecture**:
>   - **New `LoadContext`**: Encapsulates `client`, `environment_name`, and `app_id`; supports merging, defaults, and in-place upgrades.
>   - **Resolver refactor**: Removes embedded `client/env/app_id`; `Resolver.load/preload` now accept a parent `LoadContext` and merge with each object’s `load_context_overrides`.
>   - **Object loading API**: All loaders now use signature `(_obj, resolver, load_context, existing_object_id)` and must use `load_context.client/...` for RPCs; `_Object.hydrate` updated accordingly.
> - **App integration**:
>   - **App-level root context**: `App` maintains a shared `LoadContext` referenced by created `Function`/`Cls`; `run/deploy` populate it in place so descendants lazily inherit updated client/env/app_id.
> - **Functions/Classes**:
>   - Propagate `load_context_overrides` through `Function.from_local`, parameter-bound methods, `_Function._from_name`, and class method loaders.
>   - `Cls.from_name`/method bindings/`with_options`/`with_concurrency`/`with_batching` share the parent `LoadContext` reference to ensure lazy children use the same client/env.
> - **Constructors API**:
>   - All user-facing constructors (`*.from_name`, `*.from_id`, etc.) now accept optional `client` and pass it via `LoadContext` (e.g., `Function`, `Cls`, `Dict`, `Queue`, `Volume`, `NetworkFileSystem`, `Mount`, `Proxy`, `Image`, `SandboxSnapshot`, `FunctionCall`).
> - **CLI updates**:
>   - Commands (e.g., `dict create`, `queue create`) construct a `LoadContext` and call `Resolver.load` with it.
> - **Tests**:
>   - Remove `set_env_client` reliance; pass `client` explicitly; add tests for `LoadContext` propagation and lazy behaviors.
> - **Misc**:
>   - `_get_environment_name` simplified; numerous RPC calls switched from `resolver.client` to `load_context.client`; preload/load paths updated across modules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d6a6feac661b41db3f9eaa715fb9d150afc509e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/app.py
+++ b/modal/app.py
@@ -314,6 +314,7 @@ class _App:
         app._local_state_attr = None  # this is not a locally defined App, so no local state
         app._app_id = response.app_id
         app._client = client
+        app._root_load_context = LoadContext(client=client, environment_name=environment_name, app_id=response.app_id)
         app._running_app = RunningApp(response.app_id, interactive=False)
         return app
 


modal/_functions.py
```diff
@@ -895,12 +896,12 @@ def _deps(only_explicit_mounts=False) -> list[_Object]:
                 is_web_endpoint, is_generator, restrict_output
             )
 
-        async def _preload(self: _Function, resolver: Resolver, existing_object_id: Optional[str]):
-            assert resolver.client and resolver.client.stub
-
-            assert resolver.app_id
+        async def _preload(
+            self: _Function, resolver: Resolver, load_context: LoadContext, existing_object_id: Optional[str]
```
thomasjpfan, line 887: Does `_preload` need the resolve anymore?

modal/app.py
```diff
@@ -237,6 +239,7 @@ def __init__(
         # Client is special - needed to be set just before the app is "hydrated" or running at the latest
         # Guaranteed to be set for running apps, but also needed to actually *hydrate* the app and make it running
         self._client = None
+        self._load_context = LoadContext()  # Initialize empty LoadContext
```
thomasjpfan: For consistency?

```suggestion
        self._load_context = LoadContext.empty()  # Initialize empty LoadContext
```

test/cls_test.py
```diff
@@ -727,7 +727,18 @@ class FooUnhydrated:
     def bar(self, x): ...
 
 
-def test_unhydrated():
+def test_unhydrated(set_env_client):
+    # TODO: get rid of set_env_client here.
```
freider, line 731: One of the few places I had to keep the cursed env_client injections. It's a bit of an odd case and I have some ideas for how to fix it if really want to, but shouldn't be a huge deal since it's only for a particular error case

modal/cli/dict.py
```diff
@@ -29,8 +30,10 @@ async def create(name: str, *, env: Optional[str] = ENV_OPTION):
     """
     d = _Dict.from_name(name, environment_name=env, create_if_missing=True)
     client = await _Client.from_env()
-    resolver = Resolver(client=client)
-    await resolver.load(d)
+    resolver = Resolver()
+
+    parent_metadata = LoadContext(client=client, environment_name=env)
```
mwaskom: Why is this "parent metadata"? What's the parent here?

modal/cli/queues.py
```diff
@@ -39,8 +40,9 @@ async def create(name: str, *, env: Optional[str] = ENV_OPTION):
     """
     q = _Queue.from_name(name, environment_name=env, create_if_missing=True)
     client = await _Client.from_env()
-    resolver = Resolver(client=client)
-    await resolver.load(q)
+    resolver = Resolver()
+    parent_metadata = LoadContext(client=client, environment_name=env)
+    await resolver.load(q, parent_metadata)
```
mwaskom: BTW feels like we could be using `modal.Queue.objects.create()` here?

modal/_functions.py
```diff
@@ -669,7 +670,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
     @staticmethod
     def from_local(
         info: FunctionInfo,
-        app,
+        app: Optional["modal.app._App"],  # App here should only be None in case of Image.run_function
```
mwaskom, line 660: 😵‍💫

modal/_load_context.py
```diff
@@ -0,0 +1,97 @@
...
+    _environment_name: Optional[str] = None
+    _app_id: Optional[str] = None
+
+    def __init__(
+        self,
+        *,
+        client: Optional[_Client] = None,
+        environment_name: Optional[str] = None,
+        app_id: Optional[str] = None,
+        _apply_defaults: bool = True,
```
mwaskom: Why do we have a "private" parameter to this (already private?) object

modal/cls.py
```diff
@@ -154,11 +155,15 @@ def _deps():
 
     rep = f"Method({cls._name}.{method_name})"
 
+    # Bound methods should *reference* their parent Cls's LoadContext
+    # so that it can be modified in place on the parent and be reflected in the method
+    print("METHOD METADATA", method_name, cls._load_context)
```
mwaskom: Hello

modal/_load_context.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class LoadContext:
+    """Encapsulates optional metadata values used during object loading.
+
+    This metadata is set during object construction and propagated through
+    parent-child relationships (e.g., App -> Function, Cls -> Obj -> bound methods).
+    """
+
+    _client: Optional[_Client] = None
```
thomasjpfan, line 15: At a glance, it looks like `LoadContext` is used as a "builder" to combine contexts. Can we have `LoadContext.apply_defaults` return a `ResolvedContext`, where the parameters are not None:

```python
@dataclass
class ResolvedContext:
   client: _Client
   environment_name: str
   app_id: Optiona[str]
```

I would even rename `LoadContext` into `ContextBuilder`. With `ResolvedContext`, we can be sure that the client is defined with the type system.

modal/_object.py
```diff
@@ -281,19 +297,20 @@ async def hydrate(self, client: Optional[_Client] = None) -> Self:
                 # on restore to handle staleness.
                 logger.debug(f"rehydrating {self} after snapshot")
                 self._is_hydrated = False  # un-hydrate and re-resolve
-                c = client if client is not None else await _Client.from_env()
-                resolver = Resolver(c)
-                await resolver.load(typing.cast(_Object, self))
+                # Set the client on LoadContext before loading
+                parent_metadata = LoadContext(client=client)
```
thomasjpfan: Nit:

```suggestion
                parent_context = LoadContext(client=client)
```

modal/_object.py
```diff
@@ -281,19 +297,20 @@ async def hydrate(self, client: Optional[_Client] = None) -> Self:
...
-                logger.debug(f"rehydrated {self} with client {id(c)}")
+                logger.debug(f"rehydrated {self} with client {id(self.client)}")
         elif not self._hydrate_lazily:
-            # TODO(michael) can remove _hydrate lazily? I think all objects support it now?
             self._validate_is_hydrated()
         else:
-            c = client if client is not None else await _Client.from_env()
-            resolver = Resolver(c)
+            # Set the client on LoadContext before loading
+            parent_metadata = LoadContext(client=client)
```
thomasjpfan:
```suggestion
            parent_context = LoadContext(client=client)
```

modal/_functions.py
freider: Probably not in practice, but technically something that preloads should be able to preload other objects

modal/app.py
freider: 👍

modal/_load_context.py
freider: Yeah that could be nice!

One argument against would be if we think ahead to when the Resolver doesn't exist anymore and all contexts are just propagated from their "parents" explicitly (instead of using `deps`). In that scenario, I don't think we want to `apply_default()` anymore, and instead just apply each property as it's needed from within the loader (if it's not already set on the load_context_overrides.

Making each property resolve individually would fix the case where we have loaders that may raise exceptions immediately without actually *needing* a functional context (like calling methods directly on un-instantiated classes). Maybe we should actually already make that change, e.g. having LoadContext have `async def get_client() -> Client` etc. so that loaders use those (even though we currently have to make the resolver apply the defaults first due to `deps`...)

modal/_object.py
freider: Nice catch! I initially called the object LoadMetadata, but must have missed this rename

modal/_functions.py
freider: Might not be super clear, but this was a driveby typing fix, my branch doesn't change this behavior (it was already optional)

modal/_load_context.py
freider: Removing this attribute, it's for a single (stupid) use case anyway, so I'm just patching that particular thing instead (method references on uninstantiated class-related)

modal/cli/queues.py
freider: Yeah there are probably more of these cases in the cli, deferring that for another patch though

PR #3688, opened 2025-10-22
Incrementally decode utf-8 from sandbox container process streams

saltzm:
## Describe your changes

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Refactors stream reading to incrementally decode UTF‑8 and improve line-buffered/text vs. bytes behavior, with tests covering split multibyte chunks and iteration semantics.
> 
> - **Streams/IO**:
>   - Refactor `StreamReader` to support incremental UTF-8 decoding via `_decode_bytes_stream_to_str`, preserving multibyte characters across chunk boundaries.
>   - Simplify async iteration (`__anext__`) and unify text/bytes handling; add `_stream_by_line` and adjust `by_line` buffering to emit complete lines and flush at EOF.
>   - Update server/command-router implementations to produce consistent byte streams, move decoding to a single layer, and tweak container process buffering/iteration.
> - **Sandbox**:
>   - Initialize `stdout`/`stderr` with updated `StreamReader` in `by_line` text mode.
> - **Tests**:
>   - Add parametrized tests ensuring split multibyte sequences (e.g., café) decode correctly in both text and bytes modes, and validate by-line vs. raw chunk iteration behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 03f96c235766dfb4b04403237ec2967b51b72298. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/sandbox_test.py
+++ b/test/sandbox_test.py
@@ -1,5 +1,4 @@
 # Copyright Modal Labs 2022
-import asyncio
 import hashlib
 import pytest
 import time
@@ -716,7 +715,9 @@ def test_sandbox_stdout_server_read_incremental_decode(servicer, client, by_line
         async def streamer(servicer, stream):
             req: api_pb2.ContainerExecGetOutputRequest = await stream.recv_message()
             if req.file_descriptor != api_pb2.FileDescriptor.FILE_DESCRIPTOR_STDOUT or len(queued_responses) == 0:
-                await stream.send_message(api_pb2.RuntimeOutputBatch(exit_code=0),)
+                await stream.send_message(
+                    api_pb2.RuntimeOutputBatch(exit_code=0),
+                )
                 return
 
             await stream.send_message(queued_responses.popleft())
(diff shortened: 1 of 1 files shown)


modal/io_streams.py
```diff
@@ -89,7 +89,7 @@ async def _container_process_logs_iterator(
 class _StreamReaderThroughServer(Generic[T]):
     """A StreamReader implementation that reads from the server."""
 
-    _stream: Optional[AsyncGenerator[Optional[bytes], None]]
+    _stream: Optional[Union[AsyncGenerator[bytes, None], AsyncGenerator[str, None]]]
```
freider: This could be `AsyncGenerator[T, None]` right?

modal/io_streams.py
```diff
@@ -275,43 +273,34 @@ async def _get_logs(self, skip_empty_messages: bool = True) -> AsyncGenerator[Op
                         continue
                 raise
 
-    async def _get_logs_by_line(self) -> AsyncGenerator[Optional[bytes], None]:
+    async def _get_logs_by_line(self) -> AsyncGenerator[bytes, None]:
```
freider, line 280: Probably unrelated to this PR, but since _get_logs_by_line is only supported for text=True, it feels like this should be at a "higher" level in the abstraction

modal/io_streams.py
saltzm: I get confused about this - I don't think we even properly set T anywhere when we construct this thing so I don't know how it works

saltzm: For exec/ContainerProcess anyway

modal/io_streams.py
```diff
@@ -76,11 +75,13 @@ async def _container_process_logs_iterator(
             break
         except StopAsyncIteration:
             break
+
+        for item in batch.items:
```
freider, line 79: Just in case someone sends actual data together with the exit code like I did in this test, we don't want to drop the data before closing the stream

modal/io_streams.py
freider: T is purely a static typing thing so you don't really set it at runtime, but it helps the static checkers in the places where we do specify the types

PR #3689, opened 2025-10-22
flash sdk 

molocule:


modal/_container_entrypoint.py
```diff
@@ -491,13 +488,36 @@ def breakpoint_wrapper():
...
+            # TODO: Check for flash web_server and then add add customer flash enter method
+            flash_configs = [
+                partial_method.params.flash_config
+                for partial_method in _find_partial_methods_for_user_cls(
+                    type(service.user_cls_instance), _PartialFunctionFlags.FLASH_WEB_INTERFACE
+                ).values()
+                if partial_method.params.flash_config
+            ]
+            from modal.experimental import flash_forward
+
```
molocule: (todo claudia) refactor into fn

modal/_container_entrypoint.py
```diff
@@ -491,13 +488,36 @@ def breakpoint_wrapper():
...
             post_snapshot_methods = _find_callables_for_obj(
                 service.user_cls_instance, _PartialFunctionFlags.ENTER_POST_SNAPSHOT
             )
             call_lifecycle_functions(event_loop, container_io_manager, list(post_snapshot_methods.values()))
 
+            # TODO: Check more than one
+            processes = [p for p in vars(service.user_cls_instance).values() if isinstance(p, flash_process)]
+            assert len(processes) <= 1
+            process = processes[0] if processes else None
```
molocule: (todo claudia) support more than one process

modal/_container_entrypoint.py
```diff
@@ -491,13 +488,36 @@ def breakpoint_wrapper():
 
         sys.breakpointhook = breakpoint_wrapper
 
+        from modal.experimental.flash import FlashManager, flash_process
+
+        # Identify all "enter" methods that need to run before we snapshot.
+        flash_managers: dict[int, FlashManager] = {}
         # Identify the "enter" methods to run after resuming from a snapshot.
         if service.user_cls_instance is not None and not is_auto_snapshot:
+            # TODO: Check for flash web_server and then add add customer flash enter method
```
molocule: remove

modal/app.py
```diff
@@ -1052,6 +1052,22 @@ def wrapper(wrapped_cls: Union[CLS_T, _PartialFunction]) -> CLS_T:
...
+                if partial_method.params.flash_config
+            ]
+            # TODO: Get the super set of regions
+            assert len(flash_configs) == 1
+            flash_config = flash_configs[0]
+
+            # TODO: Check that there are not additional flash_options in experimental_options
+            # TODO: Actual use `experimental_options` from the callable
+            experimental_options = {}
+            experimental_options["flash"] = flash_config.region
```
molocule: validate all the flash params
1. only 1 flash server method for now
2. same region
3. region overridden in decorator

modal/experimental/flash.py
```diff
@@ -621,3 +622,30 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
     req = api_pb2.FlashContainerListRequest(function_id=fn.object_id)
     resp = await retry_transient_errors(client.stub.FlashContainerList, req)
     return resp.containers
+
+
+def _flash_web_server(port: int, *, region: Optional[Union[str, Literal[True]]] = None):
```
gongy: @molocule @thomasjpfan @mwaskom are we open to flash http server? thinking abt maybe separate path for web sockets / L4 in future

thomasjpfan: I'm okay with renaming this to `flash_http_server` to have better separation with L4.

What would be the name for L4, something like `flash_l4_server` or `flash_l4_proxy`? This assumes that a user understands what L4 means and when to use it. Also, what arguments would this new function take?

gongy (commented):
btw does this `flash_web_server` also handle the "exit grace period" where you wait for requests to drain?

modal/experimental/flash.py
molocule: update: we're keeping flash_web_server for now -- this is because it's easier to adapt into different server types and we're not sure what we want the params for flash_l4_server to be/whether it will be ergonomic

modal/_container_entrypoint.py
molocule: done

modal/app.py
molocule: only 1 flash server method for now ✅ 
same region ✅ 
region overridden in decorator ✅

modal/_container_entrypoint.py
```diff
@@ -467,7 +468,6 @@ def main(container_args: api_pb2.ContainerArguments, client: Client):
                 function_def._experimental_group_size,
             )
 
-        # Identify all "enter" methods that need to run before we snapshot.
```
thomasjpfan, line 470: Can you put this back? (I accidentally deleted this comment here)

modal/_container_entrypoint.py
```diff
@@ -491,13 +491,25 @@ def breakpoint_wrapper():
 
         sys.breakpointhook = breakpoint_wrapper
 
+        from modal.experimental.flash import FlashManager, flash_process
+
+        # Identify all "enter" methods that need to run before we snapshot.
```
thomasjpfan: Can you remove this comment as well?

modal/_container_entrypoint.py
```diff
@@ -491,13 +491,25 @@ def breakpoint_wrapper():
...
+
             post_snapshot_methods = _find_callables_for_obj(
                 service.user_cls_instance, _PartialFunctionFlags.ENTER_POST_SNAPSHOT
             )
             call_lifecycle_functions(event_loop, container_io_manager, list(post_snapshot_methods.values()))
 
+            processes = [p for p in vars(service.user_cls_instance).values() if isinstance(p, flash_process)]
+
+            for flash_config in flash_configs:
+                flash_managers[flash_config.port] = flash_forward(flash_config.port, process=processes)
```
thomasjpfan: I put the flash logic into `_container_entrypoint` to get things working and it's simple.

It feels a little weird to keep flash logic here, but I do not really want to have more abstractions.

modal/_partial_function.py
```diff
@@ -47,6 +48,9 @@ class _PartialFunctionFlags(enum.IntFlag):
     CONCURRENT = 128
     CLUSTERED = 256  # Experimental: Clustered functions
 
+    # # Lifecycle method flags
```
thomasjpfan:
```suggestion
    # Lifecycle method flags
```

molocule: fixed

modal/_container_entrypoint.py
molocule: fixed

modal/_container_entrypoint.py
molocule: fixed

modal/experimental/flash.py
```diff
@@ -24,12 +25,16 @@
...
+    pass
+
+
 class _FlashManager:
     def __init__(
         self,
         client: _Client,
         port: int,
-        process: Optional[subprocess.Popen] = None,
+        process: Union[list[subprocess.Popen], list[flash_process]] = [],
```
thomasjpfan, line 38: Given that a single flash manager has a single port, in what use cases would it have multiple processes?

modal/_container_entrypoint.py
molocule: fixed in #3714

modal/experimental/flash.py
molocule: There could be instances where there's separate worker processes in addition to some main server process? Also i think this makes it easier to extend to multiple flash servers

modal/experimental/flash.py
```diff
@@ -621,3 +644,63 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
     req = api_pb2.FlashContainerListRequest(function_id=fn.object_id)
     resp = await retry_transient_errors(client.stub.FlashContainerList, req)
     return resp.containers
+
+
+def _flash_web_server(
+    port: int,
+    *,
+    region: Union[str, Literal[True]] = True,
+    target_concurrent_requests: Optional[int] = None,
```
thomasjpfan: If `@modal.concurrent` works, I'll prefer to not have a `target_concurrent_requests`, to be consistent with all other modal functions.

PR #3695, opened 2025-10-28
[SVC-803 1/N] Add nonpreemptible field to SchedulerPlacement proto

eggag32:
Adds a `nonpreemptible` boolean to `SchedulerPlacement` in modal_proto/api.proto.

This SchedulerPlacement is part of the function definition and is how the client indicates the Function's containers should be scheduled onto nonpreemptible instances.

---

## Compatibility checklist

---

> [!NOTE]
> Adds a `nonpreemptible` boolean to `SchedulerPlacement` in `modal_proto/api.proto`.
> 
> - **Proto definitions**:
>   - `modal_proto/api.proto`:
>     - `SchedulerPlacement`: add boolean field `nonpreemptible` (`= 6`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4f0331229178b0d3e789864f88099d55af1d2b6f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2983,12 +2983,11 @@ message Schedule {
 }
 
 message SchedulerPlacement {
-  bool nonpreemptible = 6; 
   repeated string regions = 4;
-  // TODO(irfansharif): Make these two repeated.
-  optional string _zone = 2; // admin-only
-  optional string _lifecycle = 3; // admin-only, "on-demand" or "spot", else ignored
-  repeated string _instance_types = 5; // admin-only
+  optional string _zone = 2; // TODO: Deprecate
+  optional string _lifecycle = 3; // TODO: Deprecate
+  repeated string _instance_types = 5; // TODO: Deprecate
+  bool nonpreemptible = 6;
 
   reserved 1;
 }


modal_proto/api.proto
```diff
@@ -2971,6 +2971,7 @@ message SchedulerPlacement {
   // - Allow specifying list of zones, cloud, fallback and alternative
   //   GPU types.
 
+  bool nonpreemptible = 6;
```
thomasjpfan: Nit: Can this be moved to be under `_instance_types`?


I normally prefer positive naming for bools: "preemptible". Is this named "nonpreemptible" to more easily support older clients and not use optional? Specifically, old client messages will have `nonpreemptible=False`, because the default is `False`.

eggag32: Sure.

Yes, I think that is part of it (in general, we seem to use `False` as default). I think on-demand was another suggestion. This is by no means final, I just went with it as a placeholder.

mwaskom: Note that there's a bunch of prior discussion of the naming on the notion doc about this. It's definitely vexing.

mwaskom (approved):
✅

PR #3700, opened 2025-10-29
Speed up tests in blob_test.py

thomasjpfan:
## Describe your changes

This PR speeds up `blob_test.py` by disabling the retires. With `pytest test/blob_test.py -k failure`, with this PR the runtime is 0.04s and on `main` the runtime is ~17s.

---

## Compatibility checklist

---

## Release checklist

---

</details>

---

> [!NOTE]
> Adds a `RETRY_N_ATTEMPTS_OVERRIDE` hook to the `retry` decorator and monkeypatches it in blob failure tests to run without retries, significantly speeding them up.
> 
> - **Async utils**:
>   - Add `RETRY_N_ATTEMPTS_OVERRIDE` to control `retry` decorator attempts.
>   - Update retry loop and debug messaging to use the override when set.
> - **Tests**:
>   - In `test/blob_test.py`, monkeypatch `modal._utils.async_utils.RETRY_N_ATTEMPTS_OVERRIDE = 1` for failure cases to avoid retries.
>   - Minor import added to enable monkeypatch target.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 318fb7abbe389182a7f9038709b132090b307f77. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_utils/async_utils.py
+++ b/modal/_utils/async_utils.py
@@ -79,12 +79,13 @@ def retry(direct_fn=None, *, n_attempts=3, base_delay=0, delay_factor=2, timeout
     def decorator(fn):
         @functools.wraps(fn)
         async def f_wrapped(*args, **kwargs):
-            nonlocal n_attempts
             if RETRY_N_ATTEMPTS_OVERRIDE is not None:
-                n_attempts = RETRY_N_ATTEMPTS_OVERRIDE
+                local_n_attempts = RETRY_N_ATTEMPTS_OVERRIDE
+            else:
+                local_n_attempts = n_attempts
 
             delay = base_delay
-            for i in range(n_attempts):
+            for i in range(local_n_attempts):
                 t0 = time.time()
                 try:
                     return await asyncio.wait_for(fn(*args, **kwargs), timeout=timeout)
@@ -92,12 +93,12 @@ def retry(direct_fn=None, *, n_attempts=3, base_delay=0, delay_factor=2, timeout
                     logger.debug(f"Function {fn} was cancelled")
                     raise
                 except Exception as e:
-                    if i >= n_attempts - 1:
+                    if i >= local_n_attempts - 1:
                         raise
                     logger.debug(
                         f"Failed invoking function {fn}: {e}"
                         f" (took {time.time() - t0}s, sleeping {delay}s"
-                        f" and trying {n_attempts - i - 1} more times)"
+                        f" and trying {local_n_attempts - i - 1} more times)"
                     )
                 await asyncio.sleep(delay)
                 delay *= delay_factor


test/blob_test.py
```diff
@@ -29,13 +32,25 @@ async def test_blob_put_get(servicer, blob_server, client):
...
 
 @pytest.mark.asyncio
-async def test_blob_put_failure(servicer, blob_server, client):
+async def test_blob_put_failure(servicer, blob_server, client, monkeypatch):
+    # Remove retries in `_upload_to_s3_url` to fail faster
+    monkeypatch.setattr(
+        modal._utils.blob_utils,
+        "_upload_to_s3_url",
+        _upload_to_s3_url.__wrapped__,  # type: ignore
+    )
```
mwaskom: Nice!! This seems like a sort of roundabout / not super robust way of accomplishing your goal here (e.g. what if someone comes along and adds another decorator?) Is there a more direct way to disable retries from tests?

thomasjpfan: I tried another approach by adding a global config option that can change the number of attempts. It should only be used for testing.

mwaskom: Hm that's tricky since the config is "public" and users will find + use it. "MODAL_RETRY_ATTEMPTS" is also probably a little misleading since we have many distinct retry concepts?

thomasjpfan: Would you be okay with using a non-public environment variable that controls all the retry mechanisms? i.e. the `retry` decorator and `retry_transient_errors`

thomasjpfan: I updated to have a Python variable defined in `async_utils` that can get monkeypatched

modal/_utils/async_utils.py
```diff
@@ -51,6 +51,10 @@ def synchronize_api(obj, target_module=None):
     return synchronizer.create_blocking(obj, blocking_name, target_module=target_module)
 
 
+# Used for testing to configure the `n_attempts` that `retry` will use.
+RETRY_N_ATTEMPTS: Optional[int] = None
```
mwaskom: Naming nit: `RETRY_N_ATTEMPS_OVERRIDE` would maybe make it more clear that this is "exceptional"

modal/_utils/async_utils.py
```diff
@@ -75,21 +79,26 @@ async def may_fail_delay():
     def decorator(fn):
         @functools.wraps(fn)
         async def f_wrapped(*args, **kwargs):
+            if RETRY_N_ATTEMPTS is not None:
+                n_attempts_ = RETRY_N_ATTEMPTS
```
mwaskom: Do you need the trailing underscore here? Typing-wise we should know that `RETRY_N_ATTEMPTS` is an int and it's safe to overwrite no? Just worried about annoying bugs since it's easy to miss the difference.

test/blob_test.py
mwaskom: Hm yeah, that's probably the best approach. Still seems a little risky to have a parameter that just gets ignored due to global state. But I made a naming suggestion that might help a little.

modal/_utils/async_utils.py
thomasjpfan: I can not reuse `n_attempts` again here, because it'll be unbounded:

Screenshot 2025-10-29 at 7 39 21 PM

When `n_attempts` is defined in the local scope, Python will not use the `n_attempts` from the nonlocal scope. Specifically, if `RETRY_N_ATTEMPTS_OVERRIDE is None`, then Python raises an `UnboundLocalError`.

I updated the PR to use `local_n_attempts` so the difference is more drastic.

mwaskom: Oh I see. Isn't that what the `nonlocal` keyword is for? 🤓 . Not a big deal either way tho.

modal/_utils/async_utils.py
```diff
@@ -75,6 +79,10 @@ async def may_fail_delay():
     def decorator(fn):
         @functools.wraps(fn)
         async def f_wrapped(*args, **kwargs):
+            nonlocal n_attempts
+            if RETRY_N_ATTEMPTS_OVERRIDE is not None:
+                n_attempts = RETRY_N_ATTEMPTS_OVERRIDE
```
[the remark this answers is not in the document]
mwaskom: OK my suggestion was bad sorry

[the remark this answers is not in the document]
mwaskom: I guess this is sort of a gnarly pattern regardless.

PR #3702, opened 2025-10-29
Speed up test_mid_build_modifications tests

thomasjpfan:
## Describe your changes

This speeds up `test_mid_build_modifications`, by patching the resolver with a build_time that is always older than any file. When running:

```bash
pytest test/runner_test.py::test_mid_build_modifications
```

I get ~26s on main and 4s with this PR.

I'm a little skeptical of this PR, because it changes the nature of the test by pretending that the app started at a later time. (But saving ~22s is pretty good.)

---

## Compatibility checklist

---

## Release checklist

---

</details>

---

> [!NOTE]
> Patches `Resolver` to force old build start and simplifies test data, while removing the simulated upload delay from the mock `MountPutFile`.
> 
> - **Tests**:
>   - Patch `modal.runner.Resolver` in `test_mid_build_modifications` to return an old `build_start`, triggering validation deterministically.
>   - Reduce test data in `large_files` to a single file and remove delayed file-touch logic; simplify run context.
> - **Test server/mock**:
>   - Remove simulated slow upload (previous 2s sleep for data starting with `b"large"`) from `MockClientServicer.MountPutFile` in `test/conftest.py`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cfe733cba8ded5580edb8d4aa1861bf5c5fc9865. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/runner_test.py
+++ b/test/runner_test.py
@@ -126,8 +126,7 @@ async def test_mid_build_modifications(servicer, client, tmp_path, monkeypatch, 
     monkeypatch.setattr("modal.runner.Resolver", PatchedResolver)
 
     (large_dir := tmp_path / "large_files").mkdir()
-    for i in range(8 + 1):  # Equivalent to file upload concurrency
-        (large_dir / f"{i:02d}.txt").write_bytes(f"large {i:02d}".encode())
+    (large_dir / "1.txt").write_bytes("large 1".encode())
 
     image = modal.Image.debian_slim().add_local_dir(large_dir, "/root/large_files")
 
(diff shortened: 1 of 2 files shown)


mwaskom:
I don't think the resolver monkey patching is too bad though so we might be able to make it even faster if we take out the sleeping altogether (shouldn't need it if you're just hardcoding the "start time").

test/runner_test.py
```diff
@@ -117,20 +117,23 @@ def test_deploy_without_rich(servicer, client, no_rich):
...
+    class PatchedResolver(Resolver):
+        @property
+        def build_start(self) -> float:
+            return 0.0
+
+    monkeypatch.setattr("modal.runner.Resolver", PatchedResolver)
+
     (large_dir := tmp_path / "large_files").mkdir()
-    for i in range(512 + 1):  # Equivalent to file upload concurrency
+    for i in range(8 + 1):  # Equivalent to file upload concurrency
```
mwaskom: I'm not even sure we need multiple files at all with your approach right?

PR #3704, opened 2025-10-29
Usage Image.build to accelerate image_test.py

mwaskom:
We can cut off a ton of time in `image_test.py` by eagerly building the test Images instead of making them a dependency of a dummy Function. This is based on @thomasjpfan observation that building the entrypoint mount (which in this case is the full client repo) is the major factor in test time. This cuts the entire `image_test.py` down to ≈20s on my devbox (down from ≈200s on main)!

I think I want to do another cleanup pass here but putting up for review first as the diff itself will get larger with the cleanup.

---

> [!NOTE]
> Refactors image tests to build images directly via a helper instead of hydrating through dummy functions, updating assertions and adding a focused Image.build test for faster, simpler execution.
> 
> - **Tests (`test/image_test.py`)**
>   - Introduce `build_image(image, client)` helper to eagerly build images within `App().run(...)`.
>   - Refactor tests to call `build_image` and read layers/commands from `servicer`, removing dummy function/app wiring across the file.
>   - Update expectations accordingly (e.g., servicer function counts, scheduler placement and cloud fields, context files for different builder versions).
>   - Add minimal helper function `f()` and use `App(include_source=False)` where needed; set `parallel_app = App(include_source=False)`.
>   - Simplify/inline negative tests to raise errors without running an app.
>   - Replace prior hydration helpers with direct builds; adjust stability hash helpers to use `build_image`.
>   - Add `test_build` to verify `Image.build` behavior (two ImageGetOrCreate requests).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 31b8394e1de5b8f03a63069a50a1f3a8fe59c7bd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -47,6 +47,11 @@ def test_supported_python_series():
         assert SUPPORTED_PYTHON_SERIES[builder_version] <= list(PYTHON_STANDALONE_VERSIONS)
 
 
+def build_image(image: Image, client: Client) -> None:
+    with App().run(client=client) as app:
+        image.build(app)
+
+
 def get_image_layers(image_id: str, servicer) -> list[api_pb2.Image]:
     """Follow pointers to the previous image recursively in the servicer's list of images,
     and return a list of image layers from top to bottom."""
@@ -133,7 +138,6 @@ def test_dockerhub_python_version(builder_version):
 
 
 def test_image_base(builder_version, servicer, client, test_dir):
-    app = App()
     constructors = [
         (Image.debian_slim, ()),
         (Image.from_registry, ("ubuntu",)),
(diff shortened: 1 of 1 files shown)


test/image_test.py
```diff
@@ -271,11 +264,9 @@ def test_image_uv_python_requirements_error():
 
 
 def test_run_commands_secrets_type_validation(builder_version, servicer, client):
-    app = App()
-    image = Image.debian_slim().run_commands(
-        "echo hi", secrets=[Secret.from_dict({"xyz": "123"}), Secret.from_name("foo")]
-    )
-    app.function(image=image)(dummy)
+    # Happy case
+    Image.debian_slim().run_commands("echo hi", secrets=[Secret.from_dict({"xyz": "123"}), Secret.from_name("foo")])
```
thomasjpfan, line 261: Should there be a `image.build` here to actually build the image?

test/image_test.py
```diff
@@ -743,14 +734,13 @@ def test_image_run_function_with_region_selection(servicer, client):
...
     app = App()
-    app.image = Image.debian_slim().run_function(run_f, cloud="oci")
-    app.function()(dummy)
 
     with app.run(client=client):
-        pass
+        image.build(app)
 
-    assert len(servicer.app_functions) == 2
+    assert len(servicer.app_functions)
```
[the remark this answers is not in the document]
mwaskom: be chill bot it's fine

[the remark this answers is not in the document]
thomasjpfan: Cursor has a point here:

```suggestion
    assert len(servicer.app_functions) == 1
```

test/image_test.py
```diff
@@ -1989,14 +1968,16 @@ def test_add_local_lazy_vs_copy(client, servicer, set_env_client, supports_on_pa
     assert all(fn.startswith("/root/pkg_a/") for fn in copied_files)
 
 
+def f():
```
thomasjpfan, line 1759: Very extra nit:

```suggestion
def noop():
```

(So it's clear from the test, the function does nothing)

test/image_test.py
mwaskom: Not really needed since we do this validation eagerly (see the error case below). Maybe good to throw one on though.

thomasjpfan (approved):
Reviewed the latest commit [`31b8394` (#3704)](https://github.com/modal-labs/modal-client/pull/3704/commits/31b8394e1de5b8f03a63069a50a1f3a8fe59c7bd). Diff is bigger, but I like having less indenting. 😆

LGTM as long as the test pass.

PR #3712, opened 2025-10-31
Allow specific error codes to be retried indefinitely

mwaskom:
The idea here is that there might be contexts where you want to retry specific error codes indefinitely, e.g. for rate limiting cases that should be solved by the client backing off, without indefinitely retrying problems like connection issues.

This PR makes a relatively small enhancement to our `retry_transient_errors` helper to accept status codes that should be treated in this way and do separate bookkeeping for total retries (which still matter for backoff) and "limited" retries (which matter for eventually failing the attempt).

---

## Compatibility checklist

</details>

---

> [!NOTE]
> Adds `indefinite_status_codes` and separates total vs limited retries so only non-indefinite failures count toward `max_retries`, updating metadata, logging, and warning logic.
> 
> - **grpc_utils.retry_transient_errors**:
>   - Add `indefinite_status_codes` and include them in retryable `status_codes`.
>   - Track retries with `n_total_retries` and `n_limited_retries`; only limited retries apply to `max_retries`.
>   - Use total retries for `x-retry-attempt` and `x-retry-delay`; keep idempotency key and timestamp.
>   - Update logging and warning checks to reference total retries.
>   - Adjust increment logic so limited retries are not increased for indefinite-status errors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8747467da62f5e50e66bbd5debead5d67f1ee63e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -182,6 +182,7 @@ async def retry_transient_errors(
     delay_factor: float = 2,
     max_retries: Optional[int] = 3,
     additional_status_codes: list = [],
+    indefinite_status_codes: list = [],
     attempt_timeout: Optional[float] = None,  # timeout for each attempt
     total_timeout: Optional[float] = None,  # timeout for the entire function call
     attempt_timeout_floor=2.0,  # always have at least this much timeout (only for total_timeout)
@@ -192,9 +193,9 @@ async def retry_transient_errors(
     If max_retries is None, retry forever."""
 
     delay = base_delay
-    n_retries = 0
+    n_total_retries = n_limited_retries = 0
 
-    status_codes = [*RETRYABLE_GRPC_STATUS_CODES, *additional_status_codes]
+    status_codes = {*RETRYABLE_GRPC_STATUS_CODES, *additional_status_codes, *indefinite_status_codes}
 
     idempotency_key = str(uuid.uuid4())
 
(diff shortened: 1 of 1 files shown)


modal/_utils/grpc_utils.py
```diff
@@ -256,19 +257,24 @@ async def retry_transient_errors(
...
-
-            n_retries += 1
+            logger.debug(
+                f"Retryable failure {repr(exc)} {n_total_retries=} {delay=} for {fn.name} ({idempotency_key[:8]})"
+            )
 
             if (
                 retry_warning_message
-                and n_retries % retry_warning_message.warning_interval == 0
+                and n_total_retries % retry_warning_message.warning_interval == 0
```
mwaskom, line 266: Note that neither `n_total_retries` nor `n_limited_retries` is exactly right here since the warning logic should warn about a _separate_ set of specific error types!

modal/_utils/grpc_utils.py
```diff
@@ -256,19 +257,24 @@ async def retry_transient_errors(
...
                 and isinstance(exc, GRPCError)
                 and exc.status in retry_warning_message.errors_to_warn_for
             ):
                 logger.warning(retry_warning_message.message)
 
             await asyncio.sleep(delay)
+
+            n_total_retries += 1
```
mwaskom, line 274: There's also a slight behavioral change here to move the retry increment after the warning, that way we'll always warn the first time we hit the warnable error type (but still only warn intermittently to avoid spam). I think this makes sense — if we're retrying because you're getting rate limited it's going to add latency and you may want to know so that you can take some action.

mwaskom:
Going in a different direction here.

PR #3717, opened 2025-11-03
Implement handling for STDOUT StreamType in sandbox exec

saltzm:
## Describe your changes

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Adds STDOUT support for sandbox exec through the task command router by streaming to client and printing directly to local stdout (unreadable), with updated stream readers and tests.
> 
> - **Sandbox exec (router path)**
>   - Support `stdout=StreamType.STDOUT` by routing as `PIPE` and printing locally in the client.
> - **IO Streams**
>   - Add `_StdoutPrintingStreamReaderThroughCommandRouter` that prints to local stdout (text or bytes) and disallows `read`/iteration.
>   - Update `StreamReader` to use the new printing reader when `StreamType.STDOUT` is set; refine router readers to concrete `bytes`/`str` types and add necessary casts.
> - **Tests**
>   - Parametrize exec backend (`server`/`router`) and add assertions that `STDOUT` prints immediately (text, line-buffered, and bytes modes) and that reading from `stdout` raises `InvalidError`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16a3d44a88cca7bf4125a3279f5d67291386a03c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -11,7 +11,6 @@ from typing import (
     Generic,
     Literal,
     Optional,
-    Protocol,
     TypeVar,
     Union,
     cast,
@@ -418,12 +417,6 @@ class _BytesStreamReaderThroughCommandRouter:
         if self._stream:
             await self._stream.aclose()
 
-    @staticmethod
-    def print_chunk(chunk: bytes) -> None:
-        """Print a chunk of stream output to the console. Used for StreamType.STDOUT streams."""
-        sys.stdout.buffer.write(chunk)
-        sys.stdout.buffer.flush()
-
 
 class _TextStreamReaderThroughCommandRouter:
     """
@@ -469,22 +462,6 @@ class _TextStreamReaderThroughCommandRouter:
         if self._stream:
             await self._stream.aclose()
 
-    @staticmethod
-    def print_chunk(chunk: str) -> None:
-        """Print a chunk of stream output to the console. Used for StreamType.STDOUT streams."""
-        print(chunk, end="")
-
-
-class _StreamReaderProtocol(Protocol[T]):
-    @property
-    def file_descriptor(self) -> int: ...
-
-    def __aiter__(self) -> AsyncIterator[T]: ...
-
-    async def aclose(self) -> None: ...
-
-    def print_chunk(self, chunk: T) -> None: ...
-
 
 class _StdoutPrintingStreamReaderThroughCommandRouter(Generic[T]):
     """
@@ -494,11 +471,11 @@ class _StdoutPrintingStreamReaderThroughCommandRouter(Generic[T]):
     the local stdout immediately and is not readable via StreamReader methods.
     """
 
-    _reader: _StreamReaderProtocol[T]
+    _reader: Union[_TextStreamReaderThroughCommandRouter, _BytesStreamReaderThroughCommandRouter]
 
     def __init__(
         self,
-        reader: _StreamReaderProtocol[T],
+        reader: Union[_TextStreamReaderThroughCommandRouter, _BytesStreamReaderThroughCommandRouter],
     ) -> None:
         self._reader = reader
         self._task: Optional[asyncio.Task[None]] = None
@@ -513,8 +490,25 @@ class _StdoutPrintingStreamReaderThroughCommandRouter(Generic[T]):
     def _start_printing_task(self) -> None:
         async def _run():
             try:
-                async for chunk in self._reader:
-                    self._reader.print_chunk(chunk)
+
+                def print_text_part(part: Union[str, bytes]) -> None:
+                    assert isinstance(part, str)
+                    print(cast(str, part), end="")
+
+                def print_bytes_part(part: Union[str, bytes]) -> None:
+                    assert isinstance(part, bytes)
+                    sys.stdout.buffer.write(cast(bytes, part))
+                    sys.stdout.buffer.flush()
+
+                if isinstance(self._reader, _BytesStreamReaderThroughCommandRouter):
+                    print_part = print_bytes_part
+                elif isinstance(self._reader, _TextStreamReaderThroughCommandRouter):
+                    print_part = print_text_part
+                else:
+                    raise ValueError("Unsupported reader type")
+
+                async for part in self._reader:
+                    print_part(part)
             except Exception as e:
                 logger.exception(f"Error printing stream: {e}")
             finally:
@@ -624,7 +618,7 @@ class _StreamReader(Generic[T]):
                     reader = _BytesStreamReaderThroughCommandRouter(params)
 
                 if stream_type == StreamType.STDOUT:
-                    self._impl = _StdoutPrintingStreamReaderThroughCommandRouter(cast(_StreamReaderProtocol[T], reader))
+                    self._impl = _StdoutPrintingStreamReaderThroughCommandRouter(reader)
                 else:
                     self._impl = reader
 


modal/io_streams.py
```diff
@@ -461,6 +462,63 @@ async def aclose(self):
...
+    def __init__(
+        self,
+        params: _StreamReaderThroughCommandRouterParams,
+        by_line: bool,
+    ) -> None:
+        self._reader: Optional[_TextStreamReaderThroughCommandRouter[str]] = None
+        self._task: Optional[asyncio.Task[None]] = None
+        self._file_descriptor = params.file_descriptor
+        # Kick off a background task that reads from the underlying text stream and prints to stdout.
+        self._start_printing_task(params, by_line)
```
freider: It would be great if we can avoid kicking off this task in the constructor and instead have some kind of delegation method we use from whatever constructs this in an async context to make the "asyncness" of this function explicit, e.g. have an a `async def _init_stream()` method on all StreamReader/StreamWriter objects that's called by their constructor to kick off any potential background tasks.

This helps with both testing and making sure refactors don't accidentally push this construction calling into some non-async scope where create_task wouldn't work

modal/io_streams.py
```diff
@@ -461,6 +462,63 @@ async def aclose(self):
...
+    """
+
+    def __init__(
+        self,
+        params: _StreamReaderThroughCommandRouterParams,
+        by_line: bool,
+    ) -> None:
+        self._reader: Optional[_TextStreamReaderThroughCommandRouter[str]] = None
+        self._task: Optional[asyncio.Task[None]] = None
+        self._file_descriptor = params.file_descriptor
```
freider: should we maybe just add `self._params = params` and then reference that from the rest of the implementation code?

modal/io_streams.py
```diff
@@ -461,6 +462,63 @@ async def aclose(self):
...
+        # Kick off a background task that reads from the underlying text stream and prints to stdout.
+        self._start_printing_task(params, by_line)
+
+    @property
+    def file_descriptor(self) -> int:
```
freider, line 487: This would be a `"api_pb2.FileDescriptor.ValueType"` right?

modal/io_streams.py
```diff
@@ -461,6 +462,63 @@ async def aclose(self):
...
+        return self._file_descriptor
+
+    def _start_printing_task(self, params: _StreamReaderThroughCommandRouterParams, by_line: bool) -> None:
+        async def _run():
+            self._reader = _TextStreamReaderThroughCommandRouter[str](params, by_line)
```
freider: Is there a case for printing non-str (binary) data to stdout sometimes?

modal/io_streams.py
saltzm: I had that at first but changed it because I generally like to only keep data that's actually needed

modal/io_streams.py
saltzm: True, the `_StreamReader` class has always had `int` as the return type here though

modal/io_streams.py
saltzm: Yeah good question, I wasn't sure, what do you think? I sort of assumed people wouldn't want to do that

modal/io_streams.py
saltzm: Thanks, yeah I remember you commenting on this when you were working on the other issue. I'll give this a try

saltzm: Hmm, it looks like this would unfortunately be pretty invasive. We instantiate StreamReaders pretty much only from sync functions, e.g. `Sandbox._hydrate_metadata` ([here](https://github.com/modal-labs/modal-client/blob/ad1a74ab7413694be7dbd20b9f97cee4a45755f0/modal/sandbox.py#L519-L524)) and in `ContainerProcess*.__init__`, e.g. [here](https://github.com/modal-labs/modal-client/blob/ad1a74ab7413694be7dbd20b9f97cee4a45755f0/modal/container_process.py#L48-L69). So we'd have to do a lot of rearranging to make this workable. 

Let me know if you think it's still worth doing. But I think we really need to redesign a lot of the Sandbox API and do proper resource management via context managers and the like at some point anyway...

freider: I see what you mean... Yeah the similar thing I ran across in Italy was ContainerProcess also doing something similar - starting asyncio Tasks from the constructor. And `_hydrate_metadata()` should also not be doing anything with event loops, that's risky 😬 

I think we should clean this up sooner rather than later, without necessarily adding the contextmanager support (that should probably be optional anyway to not break backwards compatibility or similarities with Popen), but I'm ok with it being a followup on this since it's not directly related.

modal/io_streams.py
freider: I think the full params are "needed" if you count them being used to pass into `_TextStreamReaderThroughCommandRouter` in the `_start_printing_task` - esp. if we would defer that task creation to some followup step that's not part of the contstrcutor (then it wouldn't take this as an argument but have to find it as some instance var)

modal/io_streams.py
freider: I see... so many mini cleanups. Maybe it makes sense to add it as a quick driveby improvement to all the _StreamReaders?

modal/io_streams.py
freider: I could see some potential use cases for it - e.g if you want to run commands that output terminal control sequences which aren't necessarily text-encodable. I guess we could have different underlying _reader implementations here depending on text=True | False ?

Btw, why is `_TextStreamReaderThroughCommandRouter` and `_BytesStreamReader...` - wouldn't they always use `str` as the generic type var for the first and `bytes` for the second? 🤔

modal/io_streams.py
```diff
@@ -461,6 +462,63 @@ async def aclose(self):
...
+                    print(part, end="")
+            finally:
+                reader, self._reader = self._reader, None
+                if reader:
+                    await reader.aclose()
+
+        self._task = asyncio.create_task(_run())
+
+    async def read(self) -> T:
+        raise InvalidError("Logs can only be retrieved using the PIPE stream type.")
```
freider: Nit, maybe use "Output can only be read..." rather than the "Logs can only be retrieved" - command output isn't necessarily seen as "logs"? Maybe that's just me though

saltzm: No I agree with you, this was copied from an existing error message but I prefer output so I'll change it

modal/io_streams.py
```diff
@@ -461,6 +462,63 @@ async def aclose(self):
...
+    def __aiter__(self) -> AsyncIterator[T]:
+        raise InvalidError("Logs can only be retrieved using the PIPE stream type.")
+
+    async def __anext__(self) -> T:
+        raise InvalidError("Logs can only be retrieved using the PIPE stream type.")
+
+    async def aclose(self):
+        if self._task is not None:
+            self._task.cancel()
+            with contextlib.suppress(Exception):
```
freider: If this is to suppress the expected asyncio.CancelledError specifically, can we use that type instead?

modal/io_streams.py
saltzm: I think maybe a word is missing there, are you asking why they're generic? If that's the question, then there's no good reason, I should make them concrete

modal/io_streams.py
saltzm: This is me not examining LLM generated code closely enough, I'll see if I can figure out the intent and scope it more specifically

modal/io_streams.py
```diff
@@ -523,35 +582,26 @@ def __init__(
...
-                    )
                 params = _StreamReaderThroughCommandRouterParams(
                     file_descriptor, task_id, object_id, command_router_client, deadline
                 )
-                if text:
-                    self._impl = _TextStreamReaderThroughCommandRouter(params, by_line)
+                if stream_type == StreamType.STDOUT:
+                    if not text:
+                        raise ValueError("StreamType.STDOUT is only supported when text=True")
+                    self._impl = _StdoutPrintingStreamReaderThroughCommandRouter(params, by_line)
```
freider: Now that I think about it (considering the bytes vs str reading etc.) - could we implement the _StdoutPrintingStreamReaderThroughCommandRouter as accepting another streamreader as the underlying buffer - passing the `_TextStreamReaderThroughCommandRouter` or `_BytesStreamReaderThroughCommandRouter` created below into the constructor of _StdoutPrintingStreamReaderThroughCommandRouter ? Then we don't have to duplicate the logic of determining the underlying stream reader by looking at params etc.?

modal/io_streams.py
freider: Ah yeah, missed a word 🤦 Generic yes, they shouldn't need to be generic

modal/io_streams.py
saltzm: yeah that's better

modal/io_streams.py
saltzm: Fixed

modal/io_streams.py
saltzm: oh i just realized... maybe the issue here is we don't want to leak this protobuf value type to users who might do type-checking?

modal/io_streams.py
saltzm: Did your other suggested change to pass in the reader type and this went away

modal/io_streams.py
freider: I guess 🤷 But also a bit weird that the file descriptor anonymous int is exposed in the public api to begin with :D

modal/io_streams.py
```diff
@@ -437,30 +439,105 @@ def __init__(
...
+
+    @property
+    def file_descriptor(self) -> int:
+        return self._reader.file_descriptor
+
+    def _start_printing_task(self) -> None:
+        async def _run():
+            try:
+
+                def print_text_part(part: Union[str, bytes]) -> None:
```
freider: What if we just had the underlying stream types expose an internal method
```async def _pipe_to_stream(output_stream: TextIO):
    async for part in self._reader:
        print_part(part)
...
```
Then we can skip all of the isinstance-checking and just rely on polymorphism

freider: (to be clear, the printing task would still be "owned" by this wrapper _StdoutPrintingStreamReaderThroughCommandRouter, but it would defer the actual iteration and stdout-writing loop to self._reader itself

modal/io_streams.py
```diff
@@ -523,35 +601,26 @@ def __init__(
...
                 params = _StreamReaderThroughCommandRouterParams(
                     file_descriptor, task_id, object_id, command_router_client, deadline
                 )
                 if text:
-                    self._impl = _TextStreamReaderThroughCommandRouter(params, by_line)
+                    reader = _TextStreamReaderThroughCommandRouter(params, by_line)
+                else:
+                    reader = _BytesStreamReaderThroughCommandRouter(params)
+
+                if stream_type == StreamType.STDOUT:
```
freider, line 620: Nice!

modal/io_streams.py
```diff
@@ -437,30 +439,105 @@ def __init__(
...
+        raise InvalidError("Output can only be retrieved using the PIPE stream type.")
+
+    def __aiter__(self) -> AsyncIterator[T]:
+        raise InvalidError("Output can only be retrieved using the PIPE stream type.")
+
+    async def __anext__(self) -> T:
+        raise InvalidError("Output can only be retrieved using the PIPE stream type.")
+
+    async def aclose(self):
+        if self._task is not None:
```
freider, line 531: nit, perhaps we should add a test that the internal task finishes (eventually) when a sandbox/containerprocess has terminated, just so we don't accidentally introduce some kind of asyncio task leak (esp. since people will most likely never call .close() on their output streams until we have proper context manager support)

modal/io_streams.py
saltzm: True, idk who even uses this

modal/io_streams.py
saltzm: Trying this now and the problem is `part` is of type `str | bytes` so the type-checker still complains...

saltzm: LLM figured out a way to do it using Protocols, and I think it's better

modal/io_streams.py
saltzm: I feel like the task is simple enough that the test would essentially be checking that `self._reader` returns a closed stream once the sandbox is finished, which is the case but you made me realize I don't have a good integration for it. I'll add that test server-side - not sure if it's worth doing client-side since I'd have to mock all of the interesting parts of that

```py
    def _start_printing_task(self) -> None:
        async def _run():
            try:
                async for chunk in self._reader:
                    self._reader.print_chunk(chunk)
            except Exception as e:
                logger.exception(f"Error printing stream: {e}")
            finally:
                closed, self._closed = self._closed, True
                if not closed:
                    await self._reader.aclose()
```

modal/io_streams.py
saltzm: That introduced a type-checking error that's over my head:
```
modal/io_streams.pyi:207: error: Signature of "__subclasshook__" incompatible with supertype "object"  [override]
modal/io_streams.pyi:207: note:      Superclass:
modal/io_streams.pyi:207: note:          @classmethod
modal/io_streams.pyi:207: note:          def __subclasshook__(cls, type, /) -> bool
modal/io_streams.pyi:207: note:      Subclass:
modal/io_streams.pyi:207: note:          def __subclasshook__(other) -> Any
Found 1 error in 1 file (checked 327 source files)
```

LLM says:
```
The synchronicity tool (which generates the .pyi stub files) seems to have a bug where it generates an incorrect signature for __subclasshook__ on Protocol classes unless it is explicitly defined. It was generating def __subclasshook__(other): ... instead of the required @classmethod signature, causing the conflict with object.
```

I'm just going to go ahead and push the original way I had it for now - it's not super clean but doesn't seem that bad to me either - it's clearly correct at least

PR #3719, opened 2025-11-04
Add `modal shell im-*` to shell into an image id

thomasjpfan:
## Describe your changes

Every so often, when debugging an image, I find myself creating a function just so I can shell into it. With this PR, you can now run `modal shell im-*`.

**Aside**: One downside is that Image IDs do not contain your mounts (by design). From a DevX prospective, it'll be nice to also have a `modal shell my_mod.py::my_image`, where it builds the image and shells into it with all your mounts. It'll make it faster to iterate on an image definition.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- `modal shell` now supports Image IDs for debugging image builds or exploring the files on your image.

### DIFF
--- a/modal/cli/run.py
+++ b/modal/cli/run.py
@@ -613,6 +613,12 @@ def shell(
     ```
     modal shell sb-abc123xyz
     ```
+
+    Launch a shell to explore an Image by ID:
+
+    ```
+    modal shell im-abc123xyz
+    ```
     """
     env = ensure_env(env)
 


mwaskom:
There might be a conflict here with some broader refactoring that @ehdr is doing, let's check in with him to avoid any issues before merging.

> One downside is that Image IDs do not contain your mounts (by design).

I think this design is subject to change by the way.

modal/cli/run.py
```diff
@@ -624,7 +630,29 @@ def shell(
...
+            interactive_shell,
+            image=modal_image,
+            cpu=cpu,
+            memory=memory,
+            gpu=gpu,
+            cloud=cloud,
+            volumes=volumes,
+            secrets=secrets,
+            region=region.split(",") if region else [],
+            pty=pty,
```
mwaskom, line 653: The thing that bothers me about this (and where we got stuck in @ehdr's previous effort, motivating his attempt at a refactor) is that it makes it harder to explain the modal shell parameterization. Currently we have a (relatively) understandable heuristic that you pass REF _or_ you pass a specification in the flags. This adds complexity because you need to say "If not passing REF, or if REF is an image ID".

OTOH it's not actively in conflict like it is with a task / sandbox ID (Function reference is more ambiguous since you could define "override" semantics for the additional flags, we just don't currently do that) and I can see how it might be useful to drop into a specific Image with a GPU etc...

mwaskom: I think I'd be 👍 if we can plan to implement "override" semantics in the function reference case, so that the additional flags work where they "could" and it's not too hard to explain why you can't, e.g., start a shell into an existing container and add a GPU.

modal/cli/run.py
mwaskom, line 531: We need some doc improvements too right?

PR #3728, opened 2025-11-06
Add Status to get details in GRPCError

thomasjpfan:
## Describe your changes

This PR adds a `Status` which will be used with `grpclib` to decode `GRPCError.details`.

This is fully compatible with the default codec shipped with grpclib.

The decoder skips anything that it does not recognize. This way we can pass a list of messages to `GRPCError(details=[..])` and the receiver would only decode the ones it recognizes. An unknown proto message can be accidentally passed in if the proto is not defined in `modal_proto`.

The follow up to this PR to to introduce a new `RetryErrorDetails` (name TBD) and pass that through `GRPCError.details`.

---

## Compatibility checklist

---

## Release checklist

---

</details>

---

> [!NOTE]
> Adds a custom gRPC status details codec and `RPCStatus` proto to decode `GRPCError.details`, and wires the codec into client/server channels with tests.
> 
> - **gRPC utils**:
>   - Introduce `CustomProtoStatusDetailsCodec` with encode/decode to `api_pb2.RPCStatus`; cache `SymbolDatabase` via `_sym_db()`.
>   - Export `custom_detail_codec` and pass `status_details_codec=custom_detail_codec` when creating `grpclib.client.Channel` (unix/http/https).
> - **Proto**:
>   - Add import of `google/protobuf/any.proto` and new message `RPCStatus { code, message, repeated Any details }` in `modal_proto/api.proto`.
> - **Tests/Fixtures**:
>   - Use `custom_detail_codec` in `grpclib.server.Server`.
>   - Add tests validating codec round-trip, handling unknown messages, compatibility with grpclib default codec, and integration via channel (`GRPCError.details`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8090ac901740138b14e0652a810c86c68877d384. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/grpc_utils_test.py
+++ b/test/grpc_utils_test.py
@@ -179,7 +179,11 @@ def test_CustomProtoStatusDetailsCodec_unknown():
 def test_CustomProtoStatusDetailsCodec_google_common_proto_compat():
     """Check that rpc's encoded with the default GRPC codec works with the
     CustomProtoStatusDetailsCodec decoder."""
+
+    # ProtoStatusDetailsCodec requires `googleapis-common-protos` to be installed,
+    # which installs `google.rpc`.
     pytest.importorskip("google.rpc")
+
     from grpclib.encoding.proto import ProtoStatusDetailsCodec
 
     blob_msg = api_pb2.BlobCreateResponse(blob_id="abc")


requirements.dev.txt
```diff
@@ -38,3 +38,4 @@ pyright==1.1.351
 python-json-logger==2.0.7  # unpinned transitive dependency of jupytext breaking on later versions
 console-ctrl==0.1.0
 synchronicity[compile]==0.10.3
+googleapis-common-protos~=1.72.0
```
mwaskom: Looks like these protos were build with 4.25 which is newer than the minimal version we test on?

mwaskom: may just need to conditionally skip some tests?

modal/_utils/grpc_utils.py
```diff
@@ -107,6 +111,52 @@ def close(self):
...
 
+@cache
+def _sym_db() -> SymbolDatabase:
+    from google.protobuf.symbol_database import Default
+
+    return Default()
+
+
+# From https://github.com/vmagamedov/grpclib/blob/b841b4e861c4ac82c7e1f3bfa0aea03dd0b9ba18/grpclib/encoding/proto.py#L66
+# which uses `api_pb2.Status`.
```
mwaskom: Since what we're doing here is a little fussy might be helpful to document more thoroughly for the future. My understanding is that the codec is designed to decode GRPCError details that are serialized on the server using the `google.rpc.Status` wrapper into our own `api_pb2.RPCStatus` (not `api_pb2.Status`), but using messages that _we_ define on both sides? I.e. not the `google-common-protos` messages. I'm not sure that's 100% accurate but laying out the objectives and assumptions would be good.

modal/_utils/grpc_utils.py
```diff
@@ -125,15 +175,17 @@ def create_channel(
     )
 
     if o.scheme == "unix":
-        channel = grpclib.client.Channel(path=o.path, config=config)  # probably pointless to use a pool ever
+        channel = grpclib.client.Channel(
+            path=o.path, config=config, status_details_codec=custom_detail_codec
+        )  # probably pointless to use a pool ever
```
mwaskom: I wonder if we still need this comment

modal/_utils/grpc_utils.py
thomasjpfan: > messages that we define on both sides?

The `details` field can be a list of any message, but for client to decode it, the messages should be defined in `modal_proto`.

In the current implementation, if the client does not recognize the message, then it'll ignore it. This way, if the server accidentally sends over an message that the client does not know how to decode, we do not crash the container.

test/grpc_utils_test.py
```diff
@@ -145,6 +148,73 @@ async def wrapped_blob_create(req, **kwargs):
...
+
+
+def test_CustomProtoStatusDetailsCodec_unknown():
+    encoded_details = [
+        Any(type_url="abc", value=b"bad"),
+    ]
+    encoded_msg = api_pb2.RPCStatus(details=encoded_details).SerializeToString()
+    codec = CustomProtoStatusDetailsCodec()
+
+    decoded_msg = codec.decode(Status.OK, None, encoded_msg)
```
mwaskom: In normal grpclib usage is there any way to access details for an OK status? My understanding was they only are used to populate the `GRPCError` metadata (even though it's just more trailing metadata on the wire).

test/grpc_utils_test.py
```diff
@@ -145,6 +148,73 @@ async def wrapped_blob_create(req, **kwargs):
...
+    details = [api_pb2.BlobCreateResponse(blob_id="abc")]
+
+    async def raise_error(servicer, stream):
+        raise GRPCError(Status.INTERNAL, "Function create failed", details=details)
+
+    request = api_pb2.FunctionCreateRequest(app_id="xyz")
+
+    @synchronize_api
+    async def wrapped_function_create(req, **kwargs):
+        return await client.stub.FunctionCreate(req, **kwargs)
```
mwaskom: Do we need to be using synchronicity here? (In our expected usage the details won't be propagating out through the public API right?)

requirements.dev.txt
```diff
@@ -38,3 +38,7 @@ pyright==1.1.351
 python-json-logger==2.0.7  # unpinned transitive dependency of jupytext breaking on later versions
 console-ctrl==0.1.0
 synchronicity[compile]==0.10.3
+# Although we use a older `googleapis-common-proto` version of testing protobuf 3.19, as long as the server uses
+# protobuf >= 4.0, a client with protobuf 3.19 will still be able to decode the message.
+googleapis-common-protos~=1.56.4;python_version<='3.9'  # TODO: remove when we drop client support for Protobuf 3.19
```
mwaskom: Hm this is a little tricky as we'll almost certainly drop Python 3.9 support before we drop support for Protobuf 3.19; this is going to potentially cause issues when we do the former. But maybe we can just deal with it then. I didn't completely follow the direction of the compatibility issues; if we installed the old `googleapis-common-protos` on 3.10 and had 3.10 builds that exercised both older and newer protobuf runtimes, would that cause any issues?

test/grpc_utils_test.py
thomasjpfan: Yea, if the status is OK, then there would not have been an error. I updated the PR to use `Status.INTERNAL` instead.

test/grpc_utils_test.py
thomasjpfan: Yea, I updated the PR to not use `synchronicity`.

requirements.dev.txt
thomasjpfan: As long as the encoder, (i.e. the server), uses a new `googleapis-common-protos`, then our custom codec with protobuf 3.19 can still decode it. (I tested this locally to make sure it works)

For CI testing, the issue is that an new `googleapis-common-protos` + protobuf 3.19 can not **encode** the details. We need to go back to `googleapis-common-protos==1.56.4` which is the latest version that supported protobuf 3.19. I updated the PR to not pin `googleapis-common-protos` and let `protobuf` determine which version to use in the CI here:

https://github.com/modal-labs/modal-client/blob/8229af930861add6e32441b9bc33a24ac551cc8b/.github/workflows/ci-cd.yml#L117

An alternative is to use the custom encoder in the server by passing it here:

https://github.com/modal-labs/modal-client/blob/8229af930861add6e32441b9bc33a24ac551cc8b/test/conftest.py#L2855

But I wanted to test using the default grpclib codec's encoder with the custom codec's decoder.

mwaskom: Are you at all worried about us having `google-common-protos` always available during CI despite not having it part of the library's dependencies? Feels like there is some risk of us accidentally introducing a dependency on it and not noticing ... but not sure.

PR #3735, opened 2025-11-07
[SVC-933 8/N] Protobuf changes to support reading/writing max concurrent tasks and GPUs for environments

devennavani:
## Describe your changes

Proto changes to support reading and writing the max concurrent tasks and GPUs for each environment.

Using `google.protobuf.Int32Value` so we can distinguish between the field being un-set (unlimited tasks/gpus for environment) and `0` (enforce max 0 tasks/gpus for environment).

## Compatibility checklist

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -1289,8 +1289,8 @@ message EnvironmentListItem {
   bool default = 4;
   bool is_managed = 5;
   string environment_id = 6;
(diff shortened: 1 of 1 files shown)


devennavani:
@prbot approve @howard36

mwaskom:
j/w. what's the difference between `google.protobuf.Int32Value` and `optional int32`?

PR #3737, opened 2025-11-09
[SVC-933 18/N] Add current concurrent tasks and GPUs to `EnvironmentListItem`

devennavani:
This is so that the frontend can display the current concurrent containers and GPUs for each environment.

---

## Compatibility checklist


modal_proto/api.proto
```diff
@@ -1291,6 +1291,8 @@ message EnvironmentListItem {
   string environment_id = 6;
   optional int32 max_concurrent_tasks = 7;
   optional int32 max_concurrent_gpus = 8;
+  int32 current_concurrent_tasks = 9;
```
mwaskom, line 1294: This is a measurement, not a parameter?

PR #3744, opened 2025-11-12
Flash sdk v1

molocule:
## flash sdk v1 

based off of a lot of @thomasjpfan work here: https://github.com/modal-labs/modal-client/pull/3689
more flash sdk details/discussion here: https://www.notion.so/modal-com/Flash-SDK-2941e7f1694980c1a8b1f1b199421a7f#29b1e7f16949800080c2df9f0d9e2fc4

summary of what we've implemented:
- `@modal.experimental.http_server(port: int, region: Literal['us-east', 'us-west', 'ap-south', 'all'], exit_grace_period: Optional[int])` decorator to run flash proxy  -- NOT YET USED IN CONTAINER ENTRYPOINT
- fetch and store http config on function definition

---

> [!NOTE]
> Introduce `modal.experimental.http_server` for Flash services, add HTTP web-interface support, and wire `http_config` through decorators, class registration, and function creation with startup/heartbeat improvements.
> 
> - **Experimental Flash HTTP server**:
>   - Add `modal.experimental.http_server(port, proxy_regions, startup_timeout, exit_grace_period)` decorator.
>   - New `_PartialFunctionFlags.HTTP_WEB_INTERFACE` and validation to prevent mixing with callable decorators or method-level usage.
> - **Core plumbing**:
>   - Thread `http_config` (`api_pb2.HTTPConfig`) through `_PartialFunctionParams`, `App.cls()` to `_Function.from_local()`, and into `api_pb2.Function`/`FunctionData`.
>   - Update `is_flash_object(...)` to treat presence of `http_config` as Flash.
> - **Concurrency validation**:
>   - `verify_concurrent_params` now considers Flash when `http_config` is set (require `target_inputs`, disallow `max_inputs`).
> - **Flash manager/runtime**:
>   - Add `startup_timeout` to `_FlashManager` and `flash_forward`; heartbeat defers failures during startup and deregisters after timeout; minor robustness tweaks.
> - **Tests**:
>   - New/updated tests for heartbeat behavior, startup timeout, and removal of deprecated flash URL test; adjust mocks to use `startup_timeout`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fded4b0217da18fc7a8b7e6679c7e283a86d4674. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/experimental/flash.py
+++ b/modal/experimental/flash.py
@@ -52,15 +52,14 @@ class _FlashManager:
 
         start_time = time.monotonic()
 
-        def check_process_is_running():
+        def check_process_is_running() -> Optional[Exception]:
             if process is not None and process.poll() is not None:
-                return False, Exception(f"Process {process.pid} exited with code {process.returncode}")
-            return True, None
+                return Exception(f"Process {process.pid} exited with code {process.returncode}")
+            return None
 
         while time.monotonic() - start_time < timeout:
             try:
-                running, error = check_process_is_running()
-                if not running:
+                if error := check_process_is_running():
                     return False, error
                 with socket.create_connection(("localhost", self.port), timeout=0.5):
                     return True, None


modal/_container_entrypoint.py
```diff
@@ -539,9 +546,11 @@ def breakpoint_wrapper():
                     # Identify "exit" methods and run them.
                     # want to make sure this is called even if the lifespan manager fails
                     if service.user_cls_instance is not None and not is_auto_snapshot:
+                        flash_entry.stop()
                         exit_methods = _find_callables_for_obj(service.user_cls_instance, _PartialFunctionFlags.EXIT)
                         call_lifecycle_functions(event_loop, container_io_manager, list(exit_methods.values()))
-
+                        # this runs only in container event loop
+                        event_loop.run(flash_entry.close())
```
molocule: verify this

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+        if unwrapped._options.http_config is not None:
+            return unwrapped._options.http_config
+        return unwrapped.http_config
+    else:
+        return None
+
+
+class _FlashContainerEntry:
+    def __init__(self):
+        self.flash_manager: Optional[FlashManager] = None  # type: ignore
```
freider: what's the `type: ignore` for here?

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+        self.exit_grace_period = 0
+
+    def validate_flash_configs(self, flash_configs: list[_HTTPConfig]):
+        assert len(flash_configs) == 1, "Only one @http_server decorator is supported"
+        flash_config = flash_configs[0]
+
+    def enter(self, service):
+        http_config = get_http_config(service)
+        print(f"http_config: {http_config}")
```
freider: printy printy

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+                exit_grace_period=http_config.exit_grace_period,
+            )
+
+    def stop(self):
+        if self.flash_manager:
+            self.flash_manager.stop()
+
+    async def close(self):
+        if self.flash_manager:
+            await asyncio.sleep(self.exit_grace_period)  # TODO(claudia): write a test for this!
```
freider: What are we waiting for here? If possible, it would be nice if the grace period is treated as a max but we try to wait for whatever condition it is we actually want to achieve

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+        pf.validate_obj_compatibility("`http_server`")
+        return pf
+
+    return wrapper
+
+
+http_server = synchronize_api(_http_server, target_module=__name__)
+
+
+def get_http_config(cls_or_user_cls: Union[type[Any], Any]) -> Optional[_HTTPConfig]:
```
freider: For the Cls case we could use the `Cls` type in the type annotations. Also - if I understand the code correctly wouldn't this always return None for non-Cls/_Cls instances?

I'd prefer if this just took a `_Cls` or just took a `Cls` and we made whatever caller uses this make sure to send the right data type by wrapping/unwrapping etc. if necessary, rather than try to cater to all input data types

modal/_runtime/user_code_imports.py
```diff
@@ -334,8 +334,22 @@ def import_class_service(
...
 
+    # UPDATE TO USE with_http_config from ImportedClass in USER_CODE_IMPORTs.py
+    # Instead of extracting from experimental_options, extract from protobuf:
+    from modal._partial_function import _HTTPConfig
+
+    http_config: Optional[_HTTPConfig] = None
+    if function_def.http_config:
+        http_config = _HTTPConfig(
+            port=function_def.http_config.port,
+            proxy_region=function_def.http_config.proxy_region,  # type: ignore
```
freider: Why `type: ignore`?

modal/_partial_function.py
```diff
@@ -64,6 +66,16 @@ def interface_flags() -> int:
         return _PartialFunctionFlags.CALLABLE_INTERFACE | _PartialFunctionFlags.WEB_INTERFACE
 
 
+@dataclass
+class _HTTPConfig:
+    port: int
+    proxy_region: Literal[
+        "us-east", "us-west", "ap-south", "True"  # True means all regions
```
freider: curious - Why do we use the word "True" rather than "all" or something? 🤔

modal/_container_entrypoint.py
```diff
@@ -491,13 +491,19 @@ def breakpoint_wrapper():
 
         sys.breakpointhook = breakpoint_wrapper
 
+        from modal.experimental.flash import _FlashContainerEntry
```
freider: It would be nice if this import and usage was conditional on the imported user workload.

I think we should be able to refactor user_code_imports to always return a "context manager" or similar that encapsulates all of the enter/exit logic for the user's workload and we can then call all lifecycle stuff from container_entrypoint in a single call, instead of having a bunch of conditionals here

modal/cls.py
```diff
@@ -879,6 +881,38 @@ def _deps():
         cls._options.merge_options(batching_options)
         return cls
 
+    def with_http_server(self: "_Cls", *, port: int, proxy_region: Literal["us-east", "us-west", "ap-south", "True"], startup_timeout: Optional[int] = None, exit_grace_period: Optional[int] = None) -> "_Cls":
```
freider: Is this something we need at this point? What would be the use case - something like adding separate http servers for different proxy regions?

From a client perspective I feel like it would be more ergnomical to just specify the list or regions when you deploy (which would also kind of remove the need for the "True" option), but maybe there are infastructural parts preventing this?

modal/_partial_function.py
mwaskom: Oh yeah — this was something we'd discussed a bit on the design doc but it got overshadowed. I agree that `proxy_region="True"` is very weird.

modal/_runtime/user_code_imports.py
```diff
@@ -334,8 +334,22 @@ def import_class_service(
...
+            startup_timeout=function_def.http_config.startup_timeout or None,
+            exit_grace_period=function_def.http_config.exit_grace_period or None,
+        )
+        _cls._options.http_config = http_config  # type: ignore
```
mwaskom: Why are we writing to the `_cls._options` here? That's normally used client side for `modal.Cls.with_options`; I don't think we have any precedent for using it to hold state inside the container runtime.

modal/app.py
```diff
@@ -972,6 +972,7 @@ def cls(
         container_idle_timeout: Optional[int] = None,  # Replaced with `scaledown_window`
         allow_concurrent_inputs: Optional[int] = None,  # Replaced with the `@modal.concurrent` decorator
         _experimental_buffer_containers: Optional[int] = None,  # Now stable API with `buffer_containers`
+        http_config: Optional[api_pb2.HTTPConfig] = None,  # HTTP config to use for the class
```
mwaskom: Why is this ending up in the `@app.cls()` decorator — doesn't the config come in through the additional `http_server()` decorator?

modal/app.py
```diff
@@ -999,12 +1000,26 @@ def cls(
...
+            http_config_ = None
             # Check if the decorated object is a class
             if isinstance(wrapped_cls, _PartialFunction):
                 wrapped_cls.registered = True
                 user_cls = wrapped_cls.user_cls
+                if wrapped_cls.flags & _PartialFunctionFlags.HTTP_WEB_INTERFACE:
+                    http_config = wrapped_cls.params.http_config
+                    if http_config:
+                        flash_region = http_config.proxy_region
+                        experimental_options_["flash"] = flash_region
```
mwaskom: Are we still going to be communicating to the server via `experimental_options`? Trying to understand how load bearing this is. My general preference is that `experiemental_options` is "clean" and used only for "pure server-side config".

modal/app.py
```diff
@@ -999,12 +1000,26 @@ def cls(
...
+                        http_config_ = api_pb2.HTTPConfig(
+                            port=http_config.port,
+                            proxy_region=http_config.proxy_region,
+                            startup_timeout=http_config.startup_timeout or 0,
+                            exit_grace_period=http_config.exit_grace_period or 0,
+                        )  # Store for later use
```
mwaskom: Why are we copying fields from the `HTTPConfig` constructed in the decorator to a new one here (if that's what is happening?)

mwaskom: Oh wait, I see — the `http_config` contributed by the `http_config` decorator is a dataclass not a protobuf message (as it is in the case of `api_pb2.WebConfig`). Hm, would it make sense to just use a message to hold the state? Alternatively (if there's a good reason to have a dataclass), it could be nice to move this block to a `.to_proto()` method.

modal/app.py
```diff
@@ -1058,10 +1074,27 @@ def wrapper(wrapped_cls: Union[CLS_T, _PartialFunction]) -> CLS_T:
...
+                method.registered = True  # Avoid warning about not registering the method (hacky!)
+                raise InvalidError(
+                    "The `@modal.http_server` decorator cannot be used on methods; decorate the class instead."
+                )
+
             info = FunctionInfo(None, serialized=serialized, user_cls=user_cls)
 
             i6pn_enabled = i6pn or cluster_size is not None
 
+            # Convert _HTTPConfig dataclass to protobuf message
```
mwaskom: I'm confused about why we're doing this again

modal/app.py
```diff
@@ -1090,12 +1123,13 @@ def wrapper(wrapped_cls: Union[CLS_T, _PartialFunction]) -> CLS_T:
...
                 restrict_modal_access=restrict_modal_access,
                 max_inputs=max_inputs,
+                http_config=http_config_proto,
                 scheduler_placement=scheduler_placement,
                 i6pn_enabled=i6pn_enabled,
                 cluster_size=cluster_size,
                 rdma=rdma,
                 include_source=include_source if include_source is not None else local_state.include_source_default,
-                experimental_options={k: str(v) for k, v in (experimental_options or {}).items()},
+                experimental_options={k: str(v) for k, v in (experimental_options_).items()},
```
mwaskom: I'm a little unnerved by this change since the function is so long and it feels easy to miss that we have this "resolved" version of `experimental_options` if anyone else comes along and needs to inspect or mutate it.

modal/cls.py
mwaskom: +1, and I'm not sure how you'd use this. What would cause an actual new "instance" of the Function with the different configuration to end up created in our server and addressable by incoming requests? In the `.with_options()` analog, we don't create the Function until you invoke it via the `.remote()` etc. interface.

modal/cls.py
```diff
@@ -879,6 +881,38 @@ def _deps():
...
+        """Create an instance of the Cls with HTTP server enabled or overridden with new values.
+
+        **Usage:**
+
+        ```python notest
+        Model = modal.Cls.from_name("my_app", "Model")
+        ModelUsingGPU = Model.with_options(gpu="A100").with_http_server(port=8000, proxy_region="True")
+        ModelUsingGPU().generate.remote(42)  # will run on an A100 GPU with HTTP server enabled
```
mwaskom: How will this work if `http_server` classes can't otherwise have methods?

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
     return resp.containers
+
+
+def _http_server(
+    port: int,
+    *,
+    proxy_region: Literal[
+        "us-east", "us-west", "ap-south", "True"
+    ],  # Not defaulting since changing user behavior after allowing a default is tricky
+    startup_timeout: Optional[int] = None,
```
mwaskom: What's going to be the interaction between the `startup_timeout` on this decorator and the `startup_timeout` on the `app.cls()` decorator?

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
     resp = await client.stub.FlashContainerList(req)
     return resp.containers
+
+
+def _http_server(
+    port: int,
+    *,
+    proxy_region: Literal[
+        "us-east", "us-west", "ap-south", "True"
+    ],  # Not defaulting since changing user behavior after allowing a default is tricky
```
mwaskom: While it's nice to have literal types, do you ever anticipate running the Flash proxy in other regions? (Sure we'll need an EU one for compliance at some point right?). Are we going to want users to have to update their clients to gain access to a new region (which otherwise doesn't require any client-side logic AFAIU?)

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+    """Decorator for Flash-enabled HTTP servers on Modal classes.
+
+    Args:
+        port: The local port to forward to the Flash service.
+        proxy_region: The region to proxy the Flash service to.
+        startup_timeout: The maximum time to wait for the Flash service to start.
+        exit_grace_period: The time to wait for the Flash service to exit gracefully.
+
+    This is a highly experimental decorator that can break or be removed at any time without warning.
+    Do not use this decorator unless explicitly instructed to do so by Modal support.
```
mwaskom: Is this accurate? I thought this push was to get things into a "mostly stable" state.

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+    startup_timeout: Optional[int] = None,
+    exit_grace_period: Optional[int] = None,
+):
+    """Decorator for Flash-enabled HTTP servers on Modal classes.
+
+    Args:
+        port: The local port to forward to the Flash service.
+        proxy_region: The region to proxy the Flash service to.
+        startup_timeout: The maximum time to wait for the Flash service to start.
+        exit_grace_period: The time to wait for the Flash service to exit gracefully.
```
mwaskom: Is the "Flash service" the thing that's starting and stopping if you're using this on a "normal" web server function?

modal/experimental/flash.py
```diff
@@ -616,3 +643,108 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
     req = api_pb2.FlashContainerListRequest(function_id=fn.object_id)
     resp = await client.stub.FlashContainerList(req)
     return resp.containers
+
+
+def _http_server(
+    port: int,
```
mwaskom: I just noticed that out `web_server` violates the pattern where we catch "uncalled" decorators, i.e.
```
@modal.http_server
class Foo:
    ...
```

by using a sentinel positional argument. In cases where a positional argument would be ambiguous (which `port` ... arguably is?) I can see what that is a little annoying, but perhaps we should try to match the pattern.

modal/_partial_function.py
```diff
@@ -46,6 +47,7 @@ class _PartialFunctionFlags(enum.IntFlag):
     BATCHED = 64
     CONCURRENT = 128
     CLUSTERED = 256  # Experimental: Clustered functions
+    HTTP_WEB_INTERFACE = 512  # Experimental: HTTP server
```
mwaskom, line 49: Hm this is maybe a little annoying as a temporary state since it implies that the other `WEB_INTERFACE` does _not_ use `HTTP` which would be inaccurate. I'd say the new thing you're adding is closer to both _having_ a WEB_INTERFACE *and* _being_ an HTTP_SERVER or something...

modal/_partial_function.py
```diff
@@ -158,10 +171,19 @@ def validate_flag_composition(self) -> None:
...
+        has_http_web_interface = self.flags & _PartialFunctionFlags.HTTP_WEB_INTERFACE
         has_callable_interface = self.flags & _PartialFunctionFlags.CALLABLE_INTERFACE
-        if has_web_interface and has_callable_interface:
+        if (has_web_interface or has_http_web_interface) and has_callable_interface:
             self.registered = True  # Hacky, avoid false-positive warning
             raise InvalidError("Callable decorators cannot be combined with web interface decorators.")
+        if has_web_interface and has_http_web_interface:
+            self.registered = True  # Hacky, avoid false-positive warning
+            raise InvalidError(
+                "Web interface decorators cannot be combined with HTTP web interface decorators. Please only use one."
```
mwaskom: I think this error message is too confusing for a user: it's leaking some (hopefully transitional) internal details about what we're calling things. I might say we don't need to run this check since currently you can _only_ use `http_server` on a class and you _can't_ use the other "web interface" decorators (`web_server` / `asgi_app`/ etc.) at the class level. We'll catch that bad pattern through other means.

modal/_partial_function.py
```diff
@@ -158,10 +171,19 @@ def validate_flag_composition(self) -> None:
...
+            )
+        if has_http_web_interface and (self.flags & _PartialFunctionFlags.BATCHED):
+            self.registered = True  # Hacky, avoid false-positive warning
+            raise InvalidError("HTTP web interface decorators cannot be combined with batched decorators.")
```
mwaskom: Similar to the above, since this is otherwise illegal (`batched` goes on a function or method and `http_server` goes on a class), we might get away with not running the check.

mwaskom (commented):
Bunch of comments — sorry wasn't able to get through everything quite yet.

modal/experimental/flash.py
molocule: it's because FlashManager is a variable where 
```
FlashManager = synchronize_api(_FlashManager)
```

and types can't be variables. but i think it makes sense to just set the type to be `_FlashManager`?

molocule: if i set the type to be `_FlashManager`, then do i just synchronize api with this class (_FlashContainerEntry) instead?

modal/experimental/flash.py

modal/experimental/flash.py
molocule: we need this for now to wait for any currently running requests on the user's container to finish. we don't have introspection into when this is complete until we add l7 control onto containers.

modal/_runtime/user_code_imports.py
molocule: it's because proxy_region is a Literal and `function_def.http_config.proxy_region` is a string/scalar from protobuf. is this common usage pattern?

modal/_partial_function.py
molocule: ah this was a remanent of past work. I preferred using a list but couldn't because `experimental_options` is only of type `dict[str, str]`. now that I have HTTPConfig, I will make proxy_region a string

modal/cls.py
molocule: agree! fixed

modal/app.py
molocule: ah good point, removing this dependency

modal/app.py
molocule: yes -- sorry this is remnant of some desperation when my testing was failing, I don't actually use this

molocule: removed

modal/app.py
molocule: what does using a message to hold the state mean? as in, convert it to a proto message after receiving it from the decorator? does that allow for easy type checking? I think I didn't notice other decoraators doing this.

modal/app.py
molocule: fixed

modal/app.py
molocule: i don't think i need this anymore, i will revert this

modal/cls.py
molocule: re: with_options usage, I did not realize it was a completely different analog, I thought I could use it as a helper method in ImportedClass, but I think this understanding was wrong. Going to remove

modal/_runtime/user_code_imports.py
molocule: ohhh this totally makes sense now. I didn't know how to pass it into _container_entrypoint via service without using mro, so I tacked it onto `_cls._options`. But I realize now I can just store and retrieve from function_def

modal/app.py
mwaskom: Yeah I think the closest analog would be the `_PartialFunctionParams.webhook_config` which is a `api_pb2.WebhookConfig` and avoids having multiple representations with conversion steps. Using a proto message also gives you type safety (although optionals etc. are handled differently of course). I think compared to your current approach we'd lose the ability to type a string _literal_, although as mentioned elsewhere I'm not sure that we necessarily want to be constraining `region` values in the client.

Perhaps there are other reasons that the proto message isn't a good fit for the client-side representation? But since we need to get there anyway, keeping the internal config local to the decorator seems useful.

modal/experimental/flash.py
molocule: good point, i'll move the checking into server side and make this type str

modal/experimental/flash.py
molocule: yeah, i'll change naming of this to HTTP server?

modal/experimental/flash.py
molocule: this timeout configures how long to poll for before timing out
the startup_timeout on app.cls specifies how long the enter function can run for (so i guess app.cls startup timeout has to be greater)

modal/experimental/flash.py
molocule: ahh yes you're right, but upon further reflection, after i added state to function_def, I realized I don't need to read from cls._options anymore. I will just remove this.

modal/_container_entrypoint.py
molocule: same here, after i added state to function_def, I realized I don't need to read from cls._options anymore. I will just remove this.

is there any value in still keeping this? i don't think so

molocule: update: i will work on this

modal/experimental/flash.py
molocule: removed -- is there any template we want to use for saying that this decorator might change in the future? or not saying anything is also fine

modal/_partial_function.py
molocule: ah got it!

modal/_partial_function.py
molocule: fixed

modal/experimental/flash.py
mwaskom: The thing that's felt important to communicate about "experimental" features is primarily whether the _functionality_ is experimental or whether we're committed to the functionality but don't have a finalized interface yet. My sense is that this case fits the latter — users should feel ok starting to design their systems around the existence of a low-latency proxy with the characteristics we're introducing here, but expect some changes to the SDK over time.

In the Flash case there's maybe also a dimension about reliability / scalability of the underlying tech a bit — but I wouldn't be quite sure what to say there.

modal/experimental/flash.py
mwaskom: Should we aim to deconflict the terminology?

modal/_container_entrypoint.py
molocule: this is done!

modal/experimental/flash.py
molocule: confused here -- aren't they the same?

modal/_partial_function.py
molocule: made this a list

modal/app.py
molocule: fixed -- we just use the proto now

modal/experimental/flash.py
molocule: is it ok to do so via the comments 😓

modal/_partial_function.py
molocule: i am confused by this one too ;-;

modal/experimental/flash.py
mwaskom: If possible, I suspect using a different term is best since:

- It's hard to communicate with users when names are shared in the main decorator and in supplemental decorators (we've learned a hard lesson here with `max_inputs`)
- If we eventually consolidate into an `@app.server()` decorator, we'd need to have both in there (if I'm understanding correctly)

molocule: i think the naming is ok because it is the same as with @web_server

modal/experimental/flash.py
mwaskom: Elsewhere we try to catch this with the `_warn_parentheses_missing` sentinel ([example](https://github.com/modal-labs/modal-client/blob/main/modal/_partial_function.py#L386-L390)).

Then again, since you check the type of the first argument, the error when you forget to cal the decorator isn't too bad

```
InvalidError: First argument of `@http_server` must be a local port, such as `@http_server(8000)`.
```

modal/experimental/flash.py
mwaskom: Generally we want to write the library code against the "implementation" (the private, non-wrapped type). The types will get "unwrapped" as they pass through the interface layer — this is fairly magical though!

modal/experimental/flash.py
thomasjpfan: We try to always enforce users to write decorators with a closing bracket. Implementation-wise we use a `_warn_parentheses_missing=None` to enforce it:

https://github.com/modal-labs/modal-client/blob/30476b2f87ab003697a52b57920b6b3b33d23367/modal/_partial_function.py#L578-L583

I suspect we can do the same thing here with `port=None` and raising when it's `None`.

modal/experimental/flash.py
```diff
@@ -618,3 +645,77 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+            )
+
+    def stop(self):
+        if self.flash_manager:
+            self.flash_manager.stop()
+
+    async def close(self):
+        if self.flash_manager:
+            await asyncio.sleep(self.exit_grace_period)
+            self.flash_manager.close()
```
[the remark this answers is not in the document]
thomasjpfan: I think cursor has a point here.

modal/experimental/flash.py
```diff
@@ -122,12 +134,17 @@ async def _run_heartbeat(self, host: str, port: int):
...
-                    logger.error(
-                        f"[Modal Flash] Deregistering container {self.task_id} on {self.tunnel.url} "
-                        f"due to error: {port_check_error}, num_failures: {self.num_failures}"
-                    )
-                    self.num_failures += 1
-                    await self.client.stub.FlashContainerDeregister(api_pb2.FlashContainerDeregisterRequest())
+                    if first_registration and (
+                        self.startup_timeout is None or time.monotonic() - start_time < self.startup_timeout
+                    ):
+                        continue
```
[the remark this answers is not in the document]
molocule: this is fine -- we want it to keep continuing until the startup function fails

modal/_partial_function.py
mwaskom: Yeah you can just leave this one. The status quo is pretty confusing, but it's all internal.

modal/experimental/flash.py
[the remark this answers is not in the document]
molocule: but isn't _FlashManager synchronized?

modal/experimental/flash.py
molocule: revisit this in next pr

modal/experimental/flash.py
molocule: fixed!! thanks

modal/experimental/flash.py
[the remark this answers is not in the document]
molocule: will address in another pr

modal/experimental/flash.py
```diff
@@ -618,3 +639,57 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+        port: The local port to forward to the HTTP server.
+        proxy_regions: The regions to proxy the HTTP server to.
+        startup_timeout: The maximum time to wait for the HTTP server to start.
+        exit_grace_period: The time to wait for the HTTP server to exit gracefully.
+
+    """
+    if port is None:
+        raise InvalidError(
+            "Positional arguments are not allowed. Did you forget parentheses? Suggestion: `@modal.http_server()`."
+        )
```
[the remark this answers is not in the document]
molocule: i think this is just the convention (e.g. with all other examples)

modal/experimental/flash.py
```diff
@@ -49,10 +52,16 @@ async def is_port_connection_healthy(
...
+                return False, Exception(f"Process {process.pid} exited with code {process.returncode}")
+            return True, None
+
         while time.monotonic() - start_time < timeout:
             try:
-                if process is not None and process.poll() is not None:
-                    return False, Exception(f"Process {process.pid} exited with code {process.returncode}")
+                running, error = check_process_is_running()
+                if not running:
+                    return False, error
```
thomasjpfan: Can `check_process_is_running` just return the exception, then we can do this:

```suggestion
                if (error := check_process_is_running()):
                    return False, error
```

thomasjpfan (approved):
Minor nit, otherwise LGTM

modal/experimental/flash.py
molocule: ah nice!

modal/experimental/flash.py
```diff
@@ -618,3 +638,57 @@ async def flash_get_containers(app_name: str, cls_name: str) -> list[dict[str, A
...
+        port: The local port to forward to the HTTP server.
+        proxy_regions: The regions to proxy the HTTP server to.
+        startup_timeout: The maximum time to wait for the HTTP server to start.
+        exit_grace_period: The time to wait for the HTTP server to exit gracefully.
+
+    """
+    if port is None:
+        raise InvalidError(
+            "Positional arguments are not allowed. Did you forget parentheses? Suggestion: `@modal.http_server()`."
+        )
```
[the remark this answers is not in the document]
mwaskom: I think cursor is right here — what you're trying to catch is the function getting invoked as a decorator syntactically and being passed a function object as its first positional.

PR #3748, opened 2025-11-14
Add `modal shell --add-local dir/` for mounting local files

ehdr:
## Describe your changes

Allows adding both files and directories, and can be passed multiple times.

Like with `--volume` you can't choose the mount point, so we default to `/{basename}` just like the `Image.add_local_dir()` default.

## Changelog

- Added `modal shell --add-local dir/` for mounting local files and directories.

---

> [!NOTE]
> Adds `--add-local` to `modal shell` to mount local files or directories at `/mnt/{basename}`, supporting multiple entries.
> 
> - **CLI: `modal shell`**
>   - **New option**: `--add-local PATH` to mount local files or directories inside the shell at `/mnt/{basename}`; can be specified multiple times.
>   - **Mount handling**: Resolves local paths and builds mounts via `_Mount._from_local_dir`/`_Mount._from_local_file`, then passes `mounts` to `interactive_shell` when no function ref is provided.
>   - Retains existing volume/secret options and image selection flow when not using a function ref.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ced1a70c0436b618ec1407a87aa73a545ae3c89c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/cloud_bucket_mount.py
+++ b/modal/cloud_bucket_mount.py
@@ -133,7 +133,7 @@ def cloud_bucket_mounts_to_proto(mounts: Sequence[tuple[str, _CloudBucketMount]]
             elif parse_result.hostname.endswith("storage.googleapis.com"):
                 bucket_type = api_pb2.CloudBucketMount.BucketType.GCP
             else:
-                logger.warning(
+                logger.info(
                     "CloudBucketMount received unrecognized bucket endpoint URL. "
                     "Assuming AWS S3 configuration as fallback."
                 )
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -367,6 +367,7 @@ message AppGetLogsRequest {
   float timeout = 2;
   string last_entry_id = 4;
   string function_id = 5;
+  string parametrized_function_id = 11;
   string input_id = 6;
   string task_id = 7;
   string function_call_id = 9;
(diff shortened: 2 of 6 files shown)


modal/cli/run.py
```diff
@@ -692,9 +707,24 @@ def shell(
         modal_image = Image.from_registry(image, add_python=add_python) if image else None
         volumes = {} if volume is None else {f"/mnt/{vol}": Volume.from_name(vol) for vol in volume}
         secrets = [] if secret is None else [Secret.from_name(s) for s in secret]
+
+        mounts = []
+        if add_local:
+            for local_path_str in add_local:
+                local_path = Path(local_path_str).expanduser().resolve()
+                # mount at /{basename}, matching the default of Image.add_local_dir()
```
mwaskom: Which default of `.add_local_dir` do you mean? Currently we actually require that you set an (absolute) `remote_path`.

Note that `modal shell --volume` will mount the Volume root at `/mnt/{name}`; not sure if we want to match that.

ehdr: Seems _I've_ started hallucinating now too! Will go for `/mnt` then to keep it consistent.

PR #3749, opened 2025-11-14
Remove async constructor FunctionCall.from_id

freider:
What should preferably have been trivial (remove the async keyword) becomes messy because:
* We don't want the coming "sync in async warnings" to trigger (and we don't want to use the synchronizer thread since it introduces some blocking to wait for that)
* We want to support backwards compatibility with a user that might have done `FunctionCall.from_id.aio()` in their code base, for a while
* We want to emit a deprecation when that happens but gracefully still return a function call as is expected

Luckily there was a precedent in the codebase for "manually" crafting .aio-enabled methods instead of delegating it to synchronicity, which is what we're using here.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Make `FunctionCall.from_id` a synchronous classmethod with a deprecated `.aio` alias, add a general `deprecate_aio_usage` helper, and adjust tests/utilities accordingly.
> 
> - **API/Functions**:
>   - `FunctionCall.from_id` converted to synchronous `@classmethod` returning a wrapped handle; provides `.aio` alias via deprecation with guidance. Uses `synchronizer` translate in/out.
> - **Utils**:
>   - Add `deprecate_aio_usage` in `modal/_utils/async_utils.py` to create `.aio` proxies with deprecation warnings.
> - **Testing**:
>   - Add test ensuring `FunctionCall.from_id` does not invoke synchronicity runners and that `.aio` warns but works.
> - **gRPC Testing**:
>   - Patch mock servicer fallback per-method (`fallback(name)`) when abstract methods are unimplemented.
> - **Client/Types**:
>   - Tighten typing/initialization for `_Client` (`client_type` typing, `_stub` default, event loop attr type).
> - **Tooling**:
>   - Minor path match tweak in `tasks.py` copyright check.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b238a6e5a1862be5cd08ec75610f10cce35020d5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -274,7 +274,7 @@ def check_copyright(ctx, fix=False):
                 # third-party code (i.e., in a local venv) has a different copyright
                 and "/site-packages/" not in root
                 and "/build/" not in root
-                and "/.venv/" not in root
+                and "/.venv" not in root
                 and not re.search(r"/venv[0-9]*/", root)
             )
         ]


modal/_utils/async_utils.py
```diff
@@ -1107,3 +1109,25 @@ async def async_chain(*generators: AsyncGenerator[T, None]) -> AsyncGenerator[T,
...
+            sync_implementation = sync_implementation.__func__
+            is_classmethod = True
+        else:
+            is_classmethod = False
+
+        async def _async_proxy(*args, **kwargs):
+            deprecation_warning(
+                deprecation_date,
+                f"""The async constructor {readable_sync_call}.aio(...) will be deprecated in a future version of Modal.
+                Please use {readable_sync_call}(...) instead (it doesn't perform any IO, and is safe in async contexts)
```
mwaskom: Is this going to end up with weird indentations in the warning message?

modal/_functions.py
```diff
@@ -2037,14 +2042,15 @@ async def from_id(function_call_id: str, client: Optional[_Client] = None) -> "_
...
-        fc = _FunctionCall.from_id(fc_id)
+        fc = FunctionCall.from_id(fc_id)
         result = fc.get()
         ```
 
         Note that it's only necessary to re-instantiate the `FunctionCall` with this method
         if you no longer have access to the original object returned from `Function.spawn`.
 
         """
+        _client = typing.cast(_Client, synchronizer._translate_in(client))
```
thomasjpfan, line 2064: Should `synchronizer._translate_in` be a public function?

I did not know this existed. Previously, I have done this manually by inspecting `Object.__dict__` and finding the key with `_sync_original_*`.

freider: Possibly - it really shouldn't be used and when it is it's an indication of a shortcoming in synchronicity where we manually have to do part of the wrapping/unwrapping. But since those use cases do pop up here and there I guess it would make sense to make it public. Probably in synchronicity 2 :)

PR #3750, opened 2025-11-14
Synchronicity sync in async-usage warnings

freider:
Will emit warnings when running sync Modal APIs from within running event loops, which would block the event loop and is usually a mistake (since you can just use the sync variant)

* Does not warn when running inside an interactive ipython shell or notebook
### `.from_x` constructors shouldn't be async
This is blocked on getting rid of the async "from_x" constructors we have. Which is going on in followup branches:
- [x] `FunctionCall.from_id` #3749 
- [x] `Image.from_id` #3751 
- [x] `SandboxSnapshot.from_id` #3796

These sandbox constructors should possibly also be made lazy IMO (for consistency), but that would be a breaking change since they currently validate the argument on construction and raise errors, so lets postpone that until client 1.4:
* `Sandbox.from_name` 
* `Sandbox.from_id`

### custom context managers
async functions returning custom context managers end up with "double asyncness" where you have to await in the async with statement
- [x] Volume.batch_upload
- [ ] Sandbox.open (when used as context manager)

### Others
- [x] Fix suggestion/code rewrite for modal dict's setattr/getattr 🤯 

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog
* The SDK will now emit warnings if an IO-performing Modal SDK function is called from an asyncio event loop without using the async API, and suggest what the correct async API is. E.g. if `Sandbox.create()`is called from an async function, a warning will be emitted suggesting the use of `await Sandbox.create.aio()` instead, to prevent locking up the event loop.
* The async variant of `Volume.batch_upload` now uses the syntax `async with volume.batch_upload():` rather than `async with await volume.batch_upload.aio():`. The old syntax is no longer supported.
* The async `Image.from_id.aio()` has been deprecated, in favor of using the non-blocking `Image.from_id()` directly
* The async `FunctionCall.from_id.aio()` has been deprecated, in favor of using non-blocking `FunctionCall.from_id()` directly

---

> [!NOTE]
> Adds intelligent warnings for using sync Modal APIs in async contexts, deprecates async `.from_id.aio()` in favor of non-blocking constructors, updates Volume `batch_upload` to a true async context manager, and bumps synchronicity/deps.
> 
> - **Runtime/Async**:
>   - Add `BlockingInAsyncContextWarning` with context-aware suggestions (`rewrite_sync_to_async`) and suppress in interactive IPython.
>   - Wire warning into `synchronizer` callback; refine generator-not-consumed wrapper.
>   - New `deprecate_aio_usage` helper for transitioning APIs.
> - **API Changes**:
>   - `FunctionCall.from_id`, `Image.from_id`, `SandboxSnapshot.from_id`: now non-async classmethods (lazy), with `.aio()` use deprecated.
>   - `modal.volume.batch_upload`: now a real async context manager (`async with vol.batch_upload():`) via `live_method_contextmanager`.
>   - Replace `is_notebook` with `is_interactive_ipython`.
> - **Client/Infra**:
>   - Minor client typing/initialization cleanups; improve grpc testing fallbacks per-method.
> - **Tests**:
>   - Extensive tests for sync-in-async warnings, IPython detection, updated async patterns, and volume behavior.
> - **CI/Deps**:
>   - Bump `synchronicity` to `0.10.4` (workflows, pyproject, dev reqs); update `mypy`, `nbclient`, `jupytext`.
>   - Add pytest filter for `BlockingInAsyncContextWarning`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 76f7b19d50aafeb397fac9702f432c2609c19981. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -37,4 +37,4 @@ jupytext==1.18.1
 pyright==1.1.351
 python-json-logger==2.0.7  # unpinned transitive dependency of jupytext breaking on later versions
 console-ctrl==0.1.0
-synchronicity[compile]==0.11.0
+synchronicity[compile]==0.10.4


modal/_utils/async_utils.py
```diff
@@ -38,7 +42,212 @@
...
+
+    if original_func and call_frame and call_frame.line:
+        func_name = getattr(original_func, "__name__", str(original_func))
+        code_line = call_frame.line.strip()
+
+        # Use the unified rewrite function for all patterns
+        _, suggestion = rewrite_sync_to_async(code_line, func_name)
+
+    message_parts.append(
+        "\n\nThis may cause performance issues or bugs. Consider using Modal's async interfaces for async contexts."
```
mwaskom, line 279: Annoying nit but putting the "This may cause ..." _after_ the suggestion makes it a little ambiguous whether you are warning about the code as it exists or about potential risks of making the change you're suggesting. I mean, pragmatically, it would be weird for us to be suggesting a change that causes bugs. But just in terms of a naive user reading it quickly, seems like a risk.

Also I wonder if we should link out to our guide page on the async API here? And I wonder if we'd benefit from a more distinct terminology for what this thing is (ie. "Modal's async interfaces" could mean several things) to facilitate searching for more information.

modal/_ipython.py
```diff
@@ -2,10 +2,20 @@
...
+    Returns True for IPython shells (including Jupyter notebooks), False otherwise.
+    """
+    try:
+        # Check if IPython is available and get the current instance
+        ipython = sys.modules.get("IPython")
+        if ipython is None:
+            return False
+
+        # Try to get the active IPython instance
+        shell = ipython.get_ipython()
```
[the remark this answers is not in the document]
freider: The code works, there is an integration test for it too

modal/_utils/async_utils.py
```diff
@@ -38,7 +42,221 @@
     # quick workaround for deadlocks on shutdown - need to investigate further
     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
 
-synchronizer = synchronicity.Synchronizer()
+
+class BlockingInAsyncContextWarning(Warning):
```
mwaskom, line 48: Should the base class here be `UserWarning`?

mwaskom: Also since the warning class name competes with the message for space, per

Image

Does it make sense to try and make it less verbose?

modal/_utils/async_utils.py
```diff
@@ -38,7 +44,285 @@
...
+
+    # Handle __aenter__ pattern: with obj as x -> async with obj as x
+    if func_name == "__aenter__" and code_line.startswith("with "):
+        suggestion = code_line.replace("with ", "async with ", 1)
+        return (True, suggestion)
+
+    # Handle __setitem__ pattern: dct['key'] = value -> suggest alternative
+    if func_name == "__setitem__":
+        # Try to extract the object and key from the bracket syntax
+        setitem_match = re.match(r"(\w+)\[([^\]]+)\]\s*=\s*(.+)", code_line.strip())
```
thomasjpfan, line 91: Nit: Given this function is called many times, we can `re.compile` this regex as a global and use it here?

(Same comment for all the `re.match` use cases)

modal/client.py
```diff
@@ -70,15 +70,16 @@ class _Client:
     _client_from_env: ClassVar[Optional["_Client"]] = None
     _client_from_env_lock: ClassVar[Optional[asyncio.Lock]] = None
     _cancellation_context: TaskContext
-    _cancellation_context_event_loop: Optional[asyncio.AbstractEventLoop] = None
-    _stub: Optional[modal_api_grpc.ModalClientModal]
+    _cancellation_context_event_loop: Optional[Optional[asyncio.AbstractEventLoop]] = None
```
thomasjpfan, line 73: Is this a typo or is this type actually a nested `Optional[Optional[...]`?

freider: Typo, thanks! :D

modal/_utils/async_utils.py
mwaskom: Curious if we have any sense of performance impact in general here.

modal/_utils/async_utils.py
```diff
@@ -38,7 +44,285 @@
...
+        func_name = getattr(original_func, "__name__", str(original_func))
+        if func_name in ("__aexit__", "__anext__"):
+            # These dunders would typically already have caused a warning on the __aenter__ or __aiter__ respectively
+            return
+
+    # Extract the call frame from the stack
+    call_frame = _extract_user_call_frame()
+
+    # Build detailed warning message with location and function first
+    message_parts = ["Blocking Modal interface used from within an async code block"]
```
mwaskom, line 267: Nit: Missing a period here.

modal/_utils/async_utils.py
mwaskom: Not sure about this but should we call these Modal's "aio interface"?

modal/_utils/async_utils.py
```diff
@@ -38,7 +44,285 @@
...
+
+    message_parts.append(
+        "\n\nThis may cause performance issues or bugs. Consider using Modal's async interfaces for async contexts."
+    )
+    # Add suggestion in "change X to Y" format
+    if suggestion and code_line:
+        # this is a bit ugly, but the warnings formatter will show the offending source line
+        # on the last line regardless what we do, so we add this to not make it look out of place
+        message_parts.append(
+            f"\n\nSuggestion:\n  {suggestion}\n\n(This warning was triggered by the line shown below.)"
```
mwaskom, line 286: Could maybe make this less ugly if it were something like

```
Suggestion:
    await blah.aio(...)

Offending line:
    blah(...)
```

Not completely certain the best title though.

modal/_utils/async_utils.py
mwaskom: "Consider using Modal's async interfaces for async contexts" is sort of clumsily redundant.

modal/_utils/async_utils.py
freider: My feeling is that this shouldn't be a hot path since it's specifically for bad user patterns, and the consequence of using the sync interface itself is probably much worse than the performance impact of the warning

PR #3758, opened 2025-11-19
Add RPCStatusPolicy message for server side retrying.

thomasjpfan:
## Describe your changes

- Adds `RPCRetry` for server side retrying

If the server sends a retry instruction, then we'll use that as precedence for retrying. What the server stops sending the instruction, then we let the client resume control flow. Note that any retries will counts toward the number of retires that client attempts.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

- Improve reliability with server side instruction used to retry RPC calls.

---

> [!NOTE]
> Implements server-driven retry logic using new `RPCRetryPolicy`, adds throttled retry metadata/logging, introduces `max_throttle_wait` config, and updates tests accordingly.
> 
> - **Retry logic (client)**:
>   - Handle server-driven retries in `_retry_transient_errors` via `RPCRetryPolicy.retry_after_secs` with backoff sleep, warning throttling, and new metadata (`x-throttle-retry-attempt`, `x-throttle-retry-delay`).
>   - Refactor retry exception handling into `process_exception_before_retry`; add `get_server_retry_policy`.
>   - Use `config.get("max_throttle_wait")` to bound total wait when server instructs retries; add `SERVER_RETRY_WARNING_TIME_INTERVAL`.
> - **Config**:
>   - New setting `max_throttle_wait` (parsed as positive int or `None`).
> - **Protobuf**:
>   - Add `RPCRetryPolicy` message with `retry_after_secs` for use in `GRPCError.details`.
> - **Tests**:
>   - Update servicer to raise `GRPCError` instances and add coverage for server-driven retry paths, throttled retry metadata/logging, total-timeout via `MODAL_MAX_THROTTLE_WAIT`, and config parsing of `max_throttle_wait`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6d8fc0c1c8b46eda894d0e8d36c8da36c7f9c35b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -344,6 +344,9 @@ async def _retry_transient_errors(
         ]
         if n_retries > 0:
             attempt_metadata.append(("x-retry-delay", str(time.time() - t0)))
+        if n_throttled_retries > 0:
+            attempt_metadata.append(("x-throttle-retry-delay", str(time.time() - t0)))
+
         timeouts = []
         if retry.attempt_timeout is not None:
             timeouts.append(retry.attempt_timeout)
--- a/test/grpc_utils_test.py
+++ b/test/grpc_utils_test.py
@@ -284,7 +284,7 @@ async def test_retry_transient_errors_grpc_retry(servicer, client, caplog, monke
         await client.stub.BlobCreate(req)
 
     assert servicer.blob_create_metadata.get("x-idempotency-key")
-    assert servicer.blob_create_metadata.get("x-retry-attempt") == "10"
+    assert servicer.blob_create_metadata.get("x-throttle-retry-attempt") == "10"
 
     # With an interval of 0.3 sec and retrying for a 1.0 sec, warning message is shown at time 0.0, 0.3, 0.6, 0.9
     assert caplog.text.count("foobar-message. Will retry in 0.10 seconds") == 4
@@ -307,4 +307,4 @@ async def test_retry_transient_errors_grpc_retry_total_timeout(servicer, client,
     with pytest.raises(GRPCError):
         await client.stub.BlobCreate(req)
 
-    assert servicer.blob_create_metadata.get("x-retry-attempt") == "0"
+    assert servicer.blob_create_metadata.get("x-throttle-retry-attempt") == "0"


modal_proto/api.proto
```diff
@@ -2634,6 +2634,11 @@ message QueuePutRequest {
   int32 partition_ttl_seconds = 6;
 }
 
+message RPCRetry {
+  float retry_after_secs = 1;
+  string warning_message = 2;
```
mwaskom: In practice this will be highly redundant with the `message` in the GRPCError. I think we discussed and there's arguments for and against only using `message` (mainly that if we want the warning to indicate something about retries / latency, just using the error message will be confusing for users on older client that don't have this behavior).

I wonder if it makes sense to construe this as like a "suffix", e.g. the `message` has the main information about what's getting limited and then the suffix indicates something about retries and latency.

Or maybe it's sufficient to just hardcode the suffix in the client? When would there be variation in the additional information that we want to add in the context of a retry?

modal_proto/api.proto
```diff
@@ -2634,6 +2634,11 @@ message QueuePutRequest {
...
+}
```
mwaskom, line 2641: I think the other thing to figure out with warning messages is how to avoid really spamming the client by showing the same message repeatedly. Do you think the server needs any role in that or can we keep the "show the warning every [x seconds / n retries / etc.]" logic entirely in the client?

modal_proto/api.proto
```diff
@@ -2634,6 +2634,11 @@ message QueuePutRequest {
   int32 partition_ttl_seconds = 6;
 }
 
+message RPCRetry {
```
mwaskom: Can we get a comment on what this is used for?

(Also "Retry" feels like the wrong noun ... it's more like RetryInstructions or something?)

mwaskom (approved):

modal_proto/api.proto
thomasjpfan: My original intention is for a non-empty `warning_message` to mean "show the message" which means the logic is on the server.

I'm okay with having the logic on the server.

modal_proto/api.proto
thomasjpfan: > When would there be variation in the additional information that we want to add in the context of a retry?

From recently memory, maybe `FunctionGetInputs` where we show no warning or say something nice about "This may take longer than normal."

modal/_utils/grpc_utils.py
```diff
@@ -307,6 +322,29 @@ async def _retry_transient_errors(
...
+                        f"Warning: Received {exc.status}: {exc.message}. Will retry in {retry_after_sec:0.2f} seconds."
+                    )
+
+                n_retries += 1
+                logger.debug(
+                    f"Retryable failure {repr(exc)} {n_retries=} {retry_after_sec=} "
+                    f"for {fn.name} ({idempotency_key[:8]})"
+                )
+                await asyncio.sleep(retry_after_sec)
+                continue
```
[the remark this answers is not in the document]
thomasjpfan: Ignoring `total_deadline` was on purpose. "If the server tells the client to retry, then you must retry again".

modal/_utils/grpc_utils.py
```diff
@@ -307,22 +323,36 @@ async def _retry_transient_errors(
...
                 return await fn_callable(req, metadata=attempt_metadata, timeout=timeout)
         except (StreamTerminatedError, GRPCError, OSError, asyncio.TimeoutError, AttributeError) as exc:
-            if isinstance(exc, GRPCError) and exc.status not in status_codes:
+            server_delay = None
+            if isinstance(exc, GRPCError) and (server_retry_policy := get_server_retry_policy(exc)):
+                server_delay = server_retry_policy.retry_after_secs
+
+                # Override total_deadline with `max_throttle_wait` when a server signals to retry.
+                if total_deadline is None and (max_throttle_wait := config.get("max_throttle_wait")):
+                    total_deadline = t0 + max_throttle_wait
```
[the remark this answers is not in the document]
thomasjpfan: >  If a user configured total_timeout in their Retry object, that deadline persists even when the server instructs retries, 

This is on purpose. For `Queue`, a user can explicitly set a `total_timeout` which should be respected.

test/grpc_utils_test.py
```diff
@@ -242,3 +246,65 @@ async def wrapped_flash_container_deregister(req, **kwargs):
...
+
+    with pytest.raises(GRPCError):
+        await client.stub.BlobCreate(req)
+
+    assert servicer.blob_create_metadata.get("x-idempotency-key")
+    assert servicer.blob_create_metadata.get("x-retry-attempt") == "10"
+
+    # With an interval of 0.3 sec and retrying for a 1.0 sec, warning message is shown at time 0.0, 0.3, 0.6, 0.9
+    assert caplog.text.count("foobar-message. Will retry in 0.10 seconds") == 4
+
```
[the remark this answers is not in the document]
thomasjpfan: Server side is instructed the retry, so the retry still happens.

modal/_utils/grpc_utils.py
```diff
@@ -307,6 +355,38 @@ async def _retry_transient_errors(
...
+                    now - last_server_retry_warning_time >= SERVER_RETRY_WARNING_TIME_INTERVAL
+                ):
+                    last_server_retry_warning_time = now
+                    logger.warning(
+                        f"Warning: Received failure {exc.status}: {exc.message}. "
+                        f"Will retry in {server_delay:0.2f} seconds."
+                    )
+
+                await asyncio.sleep(server_delay)
+                continue
```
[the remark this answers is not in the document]
thomasjpfan: This is tracked separately in `x-throttle-retry-attempt`.

modal/_utils/grpc_utils.py
```diff
@@ -307,6 +360,39 @@ async def _retry_transient_errors(
...
+            if isinstance(exc, GRPCError) and (server_retry_policy := get_server_retry_policy(exc)):
+                server_delay = server_retry_policy.retry_after_secs
+
+                # When `total_deadline` is not defined and server gives instruction for retries, then use
+                # `max_throttle_wait`.
+                if total_deadline is None and (max_throttle_wait := config.get("max_throttle_wait")):
+                    total_deadline = t0 + max_throttle_wait
+
+                final_attempt = (
+                    total_deadline is not None
```
mwaskom: Should we remove the two cases I could find where we're apparently unnecessarily passing `total_timeout`? I.e. it has a functional role in the Queue operations because we're setting `max_retries=None` there, but it's passed in `FunctionGetCurrentStats` and `ContainerExecPutInputRequest` in a way that doesn't currently have any effect but might if we add throttling to those endpoints.

modal/_utils/grpc_utils.py
```diff
@@ -307,6 +360,39 @@ async def _retry_transient_errors(
...
+                    and time.time() + server_delay + retry.attempt_timeout_floor >= total_deadline
+                )
+                with suppress_tb_frames(1):
```
mwaskom: Just checking that it's correct to use this suppression context manager both outside the `process_exception_before_retry` helper and inside it?

PR #3759, opened 2025-11-20
Fix post-snapshot rehydration of non loadables

freider:
After a memory snapshot, the client object is stale and shouldn't be used. The rehydration logic previously assumed that the underlying _Object is lazily loadable, but that's not always the case. In case it's not, we will now not attempt to "load" the object, and we instead just replace the `._client` property with a fresh Client.from_env.

This previously surfaced as an issue if any `._new_hydrated()` objects were captured by a memory snapshot and then attempted to be used. Loader-based objects were actually also potentially broken in case a user supplied a custom client on instantiation (since that client would now be stale), but that's also fixed by this patch by having `LoadContext.apply_default()` disregard any snapshotted/stale clients and create a new Client.from_env as if the client was unspecified.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

* Fixed an issue where some modal Objects (like FunctionCall as returned by Function.spawn()) that were captured by memory snapshots were not usable in resumed containers, resulting in `has no loader function` errors when these objects were used.

---

> [!NOTE]
> Fixes object rehydration after memory snapshots (reload lazy handles; replace client for non-lazy) and makes LoadContext ignore snapshotted clients; also enforces auth metadata for input-plane calls and updates typings/tests.
> 
> - **Object hydration/rehydration**:
>   - For already-hydrated objects with snapshotted clients: reload lazy objects via `Resolver` with `LoadContext.empty()`; for non-lazy, swap `_client` with a fresh `Client.from_env()`.
>   - Default `_load` to `None` for `_Object`.
> - **LoadContext**:
>   - `apply_defaults()` now disregards snapshotted clients and fetches a fresh env client.
> - **Client**:
>   - Clean up/clarify Optional annotations and defaults; `anonymous()` returns `AsyncGenerator`.
>   - Add `get_input_plane_metadata`/auth token usage (and assert manager is present).
> - **Functions (input plane)**:
>   - Require auth token in request metadata when using the input plane.
> - **Tests**:
>   - Add rehydration tests for both loader and non-loader objects; verify `FunctionCall.from_id(..., client=...)` works.
>   - Resolver tests cast synchronized client; minor cleanup.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 74aa931f40ad7e02f7628dced8896f78c26e6a60. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -1677,9 +1677,7 @@ Use the `Function.get_web_url()` method instead.
             input_queue,
             self.client,
         )
-        fc: _FunctionCall[ReturnType] = _FunctionCall._new_hydrated(
-            function_call_id, self.client, None, rep="modal.Function._spawn_map(...)"
-        )
+        fc: _FunctionCall[ReturnType] = _FunctionCall._new_hydrated(function_call_id, self.client, None)
         fc._num_inputs = num_inputs  # set the cached value of num_inputs
         return fc
 
@@ -1866,7 +1864,7 @@ Use the `Function.get_web_url()` method instead.
             )
 
         fc: _FunctionCall[ReturnType] = _FunctionCall._new_hydrated(
-            invocation.function_call_id, invocation.client, None, rep="modal.Function._experimental_spawn(...)"
+            invocation.function_call_id, invocation.client, None
         )
         fc._is_generator = self._is_generator if self._is_generator else False
         return fc
@@ -1897,7 +1895,7 @@ Use the `Function.get_web_url()` method instead.
             invocation = await self._call_function_nowait(args, kwargs, api_pb2.FUNCTION_CALL_INVOCATION_TYPE_ASYNC)
 
         fc: _FunctionCall[ReturnType] = _FunctionCall._new_hydrated(
-            invocation.function_call_id, invocation.client, None, rep="modal.Function.spawn(...)"
+            invocation.function_call_id, invocation.client, None
         )
         return fc
 


modal/_functions.py
```diff
@@ -1676,7 +1677,9 @@ async def _spawn_map(self, input_queue: _SynchronizedQueue) -> "_FunctionCall[Re
             input_queue,
             self.client,
         )
-        fc: _FunctionCall[ReturnType] = _FunctionCall._new_hydrated(function_call_id, self.client, None)
+        fc: _FunctionCall[ReturnType] = _FunctionCall._new_hydrated(
```
freider: quick quality of life improvement that would have made debugging this a bit easier (no repr = defaults to saying it is .from_id() which isn't technically true)

modal/_load_context.py
```diff
@@ -66,8 +66,9 @@ async def apply_defaults(self) -> "LoadContext":
 
         Returns a new instance (no in place mutation)"""
 
+        is_valid_client = self._client is not None and not self._client._snapshotted
```
freider, line 69: This is the fix

modal/_object.py
```diff
@@ -308,11 +308,16 @@ async def hydrate(self, client: Optional[_Client] = None) -> Self:
                 # memory snapshots capture references which must be rehydrated
                 # on restore to handle staleness.
                 logger.debug(f"rehydrating {self} after snapshot")
-                self._is_hydrated = False  # un-hydrate and re-resolve
-                # Set the client on LoadContext before loading
-                root_load_context = LoadContext(client=client)
-                resolver = Resolver()
-                await resolver.load(typing.cast(_Object, self), root_load_context)
+                if self._hydrate_lazily:
```
freider, line 311: This is another part of the fix - only ._load if there is an actual loader that should be called

modal/_object.py
```diff
@@ -308,11 +308,16 @@ async def hydrate(self, client: Optional[_Client] = None) -> Self:
...
+                    logger.debug(f"reloading lazy {self} from server")
+                    self._is_hydrated = False  # un-hydrate and re-resolve
+                    # client will be defaulted by resolver that ignores stale clients
+                    root_load_context = LoadContext.empty()
+                    resolver = Resolver()
+                    await resolver.load(typing.cast(_Object, self), root_load_context)
+                else:
+                    logger.debug(f"reloading non-lazy {self} by replacing client")
```
freider, line 320: If there is no loader - just replace the client and hope for the best (this should work as long as there aren't any stale data on the object itself)

modal/client.py
```diff
@@ -5,7 +5,7 @@
 import sys
 import urllib.parse
 import warnings
-from collections.abc import AsyncGenerator, AsyncIterator, Collection, Mapping
+from collections.abc import AsyncGenerator, Collection, Mapping
```
freider, line 8: All of the changes to this file are just lint fixes

test/function_test.py
```diff
@@ -637,6 +637,8 @@ async def test_function_future_async(client, servicer):
         servicer.function_is_running = False
         assert await future.get.aio(0.01) == "hello"
         assert future.object_id not in servicer.cleared_function_calls  # keep results around a bit longer for futures
+        fc2 = modal.FunctionCall.from_id(future.object_id, client=client)
```
freider, line 640: This didn't actually catch the old error but was my first attempt at reproing and I think we aren't testing this specific code path elsewhere?

test/object_test.py
```diff
@@ -41,3 +45,70 @@ def test_types():
     assert not _Dict._is_id_type("qu-123")
     assert _Queue._is_id_type("qu-123")
     assert not _Queue._is_id_type("di-123")
+
+
+@pytest.mark.asyncio
+async def test_object_rehydration_new_hydrated(blob_server, container_env):
```
freider, line 51: These two new added tests were previously raising errors (prior to this fix) so they can hopefully work as regression tests, even if they rely a bit on the "internals" of the _Object model/live_method

test/resolver_test.py
```diff
@@ -2,16 +2,19 @@
...
 from modal._load_context import LoadContext
 from modal._object import _Object
 from modal._resolver import Resolver
+from modal._utils.async_utils import synchronizer
 
 
 @pytest.mark.flaky(max_runs=2)
 @pytest.mark.asyncio
 async def test_multi_resolve_sequential_loads_once(client):
+    _client = cast(modal.client._Client, synchronizer._translate_in(client))
```
mwaskom, line 17: :sob:

freider: this is only for a test, in most tests we want the user-exposed wrapper but here we are testing internal functionality so we want the unwrapped instance. It worked with the wrong data type before these changes but it would have been fragile

PR #3761, opened 2025-11-20
Document new throttle config

mwaskom:
The docstring of `modal.config` [ends up on the website](https://modal.com/docs/reference/modal.config) as the source of truth for environment variable documentation.

---

> [!NOTE]
> Adds `max_throttle_wait` config and uses it to cap server-throttled gRPC retries; tweaks retry warning text and clarifies AttributeError handling comments.
> 
> - **Retry logic (`modal/_utils/grpc_utils.py`)**:
>   - Respect `config.get("max_throttle_wait")` to set a total deadline when server returns retry-after (throttling).
>   - Refine warning message wording when receiving server retry status.
>   - Add clarifying comments around temporary `AttributeError` catch for grpclib<=0.4.7.
> - **Config (`modal/config.py`)**:
>   - Document and expose new `max_throttle_wait` (`.toml`/`MODAL_MAX_THROTTLE_WAIT`).
>   - Add `_SETTINGS["max_throttle_wait"]` with int-or-None parsing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d7060a28dc6e4d85cd776bd56672038b61815a2f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_utils/grpc_utils.py
+++ b/modal/_utils/grpc_utils.py
@@ -292,6 +292,8 @@ def process_exception_before_retry(
             # StreamTerminatedError are not properly raised in grpclib<=0.4.7
             # fixed in https://github.com/vmagamedov/grpclib/issues/185
             # TODO: update to newer version (>=0.4.8) once stable
+            # Also be sure to remove the AttributeError from the set of exceptions
+            # we handle in the retry logic once we drop this check!
             raise exc
 
     logger.debug(f"Retryable failure {repr(exc)} {n_retries=} {delay=} for {fn_name} ({idempotency_key[:8]})")
--- a/modal/config.py
+++ b/modal/config.py
@@ -52,7 +52,7 @@ Other possible configuration options are:
   Number of seconds to wait for logs to drain when closing the session,
   before giving up.
 * `max_throttle_wait` (in the .toml file) / `MODAL_MAX_THROTTLE_WAIT` (as an env var).
-  Defaults to None.
+  Defaults to None (no limit).
   Maximum number of seconds to wait when requests are being throttled (i.e., due
   to rate limiting or other cases that can normally be resolved through backoff).
 * `force_build` (in the .toml file) / `MODAL_FORCE_BUILD` (as an env var).


modal/config.py
```diff
@@ -51,6 +51,10 @@
   Defaults to 10.
   Number of seconds to wait for logs to drain when closing the session,
   before giving up.
+* `max_throttle_wait` (in the .toml file) / `MODAL_MAX_THROTTLE_WAIT` (as an env var).
+  Defaults to None.
```
thomasjpfan: Default `None` is feels weird as an environment variable, where `MODAL_MAX_THROTTLE_WAIT=0` means a Python `None`.

Should we just make the default 0 which gets interpreted as "no waiting"?

mwaskom: Wouldn't a default of 0 mean _no retries_?

mwaskom: In practice we want the default to be "infinite". Agree that describing that as None is a little weird in the env var case. Maybe just "null"?

modal/_utils/grpc_utils.py
```diff
@@ -384,7 +389,7 @@ async def _retry_transient_errors(
                 ):
                     last_server_retry_warning_time = now
                     logger.warning(
-                        f"Warning: Received failure {exc.status}: {exc.message}. "
+                        f"Warning: Received {exc.status} status: {exc.message}. "
```
mwaskom, line 392: Was worried that people would see "failure" and be overly alarmed.

modal/config.py
thomasjpfan: Yea, I'm okay with "null". I've seen some libraries use `-1` to mean "null".

> Wouldn't a default of 0 mean no retries?

Currently, `MODAL_MAX_THROTTLE_WAIT=0` is the same as `None`, which means "infinite". Do you think it's better for it to mean "no retries"?

mwaskom: Oh! Yes that was the intention. We want users to be able to disable throttling / revert to the current state in case their app logic was relying on it.

PR #3770, opened 2025-11-24
Cleanup of sandbox stream handling

freider:
* Use polymorphism delegation to print stdout streams in byte/str implemenations to avoid excessive type checking in proxy stream type
* Move generator/stream initialization to `__aiter__` rather than `__anext__` (maintaining "raw" anext compatibility for now, although would have been preferred to require aiter)
* Store generators rather than streams
* Add a bunch of `with aclosing` to generator usage to ensure proper/timely cancellation propagation
* Type improvement: use AsyncGenerator when using generators (implies aclose() existance etc)

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Switch stream APIs to AsyncGenerator with proper closure via aclosing, initialize in __aiter__, and delegate stdout printing to readers; update router client and tests accordingly.
> 
> - **Streams**:
>   - Initialize and cache read generators in `StreamReader.__aiter__`; deprecate direct `__anext__`; `aclose()` now closes the cached generator.
>   - Replace `AsyncIterator` with `AsyncGenerator` across readers; return composed generators directly from `__aiter__`.
>   - Delegate stdout printing via `_print_all` in text/bytes readers, removing ad-hoc type checks in the stdout proxy reader.
>   - Ensure timely cancellation/closure using `aclosing` in `_stdio_stream_from_command_router`, text decoding, and line-buffering paths; `_stream_by_line` now closes the underlying stream.
> - **Task Command Router**:
>   - `exec_stdio_read` and `_stream_stdio` now return `AsyncGenerator`; wrap streaming with `aclosing` for proper cleanup; import `aclosing`.
> - **Tests**:
>   - Adjust `_stream_by_line` misuse test to use `aclosing` context manager.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 97b636a17975a4948b30d567ee4aeee18a381873. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/io_streams.py
+++ b/modal/io_streams.py
@@ -334,7 +334,7 @@ async def _decode_bytes_stream_to_str(stream: AsyncGenerator[bytes, None]) -> As
 async def _stream_by_line(stream: AsyncGenerator[bytes, None]) -> AsyncGenerator[bytes, None]:
     """Yield complete lines only (ending with \n), buffering partial lines until complete.
 
-    When this generator is returns, the underlying generator is closed.
+    When this generator returns, the underlying generator is closed.
     """
     line_buffer = b""
     try:


modal/io_streams.py
```diff
@@ -631,21 +603,25 @@ async def read(self) -> T:
         """Fetch the entire contents of the stream until EOF."""
         return cast(T, await self._impl.read())
 
-    # TODO(saltzm): I'd prefer to have the implementation classes only implement __aiter__
```
freider, line 634: @saltzm not sure what synchronicity issues you ran into, but this seems to work now. I still maintained `__anext__` on this top level proxy StreamReader class since we use it with next() directly in tests, but I think we eventually want to deprecate direct next() usage on the StreamReader in favor of the usual iter() -> next() pattern

test/io_streams_test.py
```diff
@@ -383,8 +383,9 @@ async def _bad_stream():
         yield "not-bytes"  # type: ignore[misc]
 
     with pytest.raises(AssertionError):
-        async for _ in _stream_by_line(_bad_stream()):
-            pass
+        async with aclosing(_stream_by_line(_bad_stream())) as stream:
```
freider, line 386: This test was leaking an unclosed generator or something, but the aclosing fixed it

modal/io_streams.py
```diff
@@ -336,17 +332,23 @@ async def _decode_bytes_stream_to_str(stream: AsyncGenerator[bytes, None]) -> As
 
 
 async def _stream_by_line(stream: AsyncGenerator[bytes, None]) -> AsyncGenerator[bytes, None]:
-    """Yield complete lines only (ending with \n), buffering partial lines until complete."""
+    """Yield complete lines only (ending with \n), buffering partial lines until complete.
+
+    When this generator is returns, the underlying generator is closed.
```
ehdr:
```suggestion
    When this generator returns, the underlying generator is closed.
```

modal/io_streams.py
```diff
@@ -336,17 +332,23 @@ async def _decode_bytes_stream_to_str(stream: AsyncGenerator[bytes, None]) -> As
...
+            while b"\n" in line_buffer:
+                line, line_buffer = line_buffer.split(b"\n", 1)
+                yield line + b"\n"
 
-    if line_buffer:
-        yield line_buffer
+        if line_buffer:
+            yield line_buffer
+    finally:
+        await stream.aclose()
```
thomasjpfan, line 351: Is this required, since the caller (`_TextStreamReaderThroughCommandRouter.__aiter__`) closes the stream with `aclosing`?:

https://github.com/modal-labs/modal-client/blob/eb9adfd82c4a37ef9d0b55b832bd5e8d6d99b2c1/modal/io_streams.py#L447

freider: Not strictly required no, but I feel like it's probably good to have it in case this helper is used in some other context, esp. since it's "destructive" on the underlying stream by keeping consumed buffer bytes in this generator (so reuse of the underlying generator after reading a line would not necessarily get the subsequent bytes)

modal/io_streams.py
```diff
@@ -445,22 +443,21 @@ async def read(self) -> str:
...
-        # This raises StopAsyncIteration if the stream is at EOF.
-        return await self._stream.__anext__()
+                stream = _decode_bytes_stream_to_str(bytes_stream)
 
-    async def aclose(self):
-        if self._stream:
-            await self._stream.aclose()
+            async with aclosing(stream):
+                async for part in stream:
+                    yield part
```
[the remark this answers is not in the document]
freider: Sorry cursor, you are wrong here - __aiter__ is an async generator function defined using async + yield, so it returns an async generator just like _BytesStreamReaderThroughCommandRouter.__aiter__ (although the latter does it via sync indirection)

modal/_utils/task_command_router_client.py
```diff
@@ -268,8 +269,9 @@ async def exec_stdio_read(
         else:
             raise ValueError(f"Invalid file descriptor: {file_descriptor}")
 
-        async for item in self._stream_stdio(task_id, exec_id, sr_fd, deadline):
-            yield item
+        async with aclosing(self._stream_stdio(task_id, exec_id, sr_fd, deadline)) as stream:
```
saltzm, line 272: Ah, I just assumed that when you used an async for over a stream it'd automatically close it when exited. Didn't know you needed to do this explicitly. Thanks!

modal/io_streams.py
```diff
@@ -404,18 +408,13 @@ async def read(self) -> bytes:
...
-        if self._stream is None:
-            self._stream = _stdio_stream_from_command_router(self._params)
-        # This raises StopAsyncIteration if the stream is at EOF.
-        return await self._stream.__anext__()
-
-    async def aclose(self):
-        if self._stream:
-            await self._stream.aclose()
+    async def _print_all(self, output_stream: TextIO) -> None:
+        async for part in self:
```
saltzm, line 415: Curious why this one doesn't need `aclosing` when we use it in the other `_print_all` call?

modal/io_streams.py
```diff
@@ -490,31 +486,9 @@ def file_descriptor(self) -> int:
...
-                if isinstance(self._reader, _BytesStreamReaderThroughCommandRouter):
-                    print_part = print_bytes_part
-                elif isinstance(self._reader, _TextStreamReaderThroughCommandRouter):
-                    print_part = print_text_part
-                else:
-                    raise ValueError("Unsupported reader type")
-
-                async for part in self._reader:
-                    print_part(part)
+                await self._reader._print_all(sys.stdout)
```
saltzm, line 489: 🙏 Thanks for doing this. I'd tried making `print_part` abstract and ran into issues, didn't think to do this

modal/io_streams.py
```diff
@@ -445,22 +443,21 @@ async def read(self) -> str:
...
-        return self
-
-    async def __anext__(self) -> str:
-        if self._stream is None:
-            bytes_stream = _stdio_stream_from_command_router(self._params)
+    async def __aiter__(self) -> AsyncGenerator[str, None]:
+        async with aclosing(_stdio_stream_from_command_router(self._params)) as bytes_stream:
             if self._by_line:
-                self._stream = _decode_bytes_stream_to_str(_stream_by_line(bytes_stream))
+                stream = _decode_bytes_stream_to_str(_stream_by_line(bytes_stream))
```
saltzm, line 449: This changes the behavior so now if you create a new StreamReader, it'll always start from the beginning of the exec output stream. I kind of prefer this behavior since it's much more intuitive, and that it works currently, but it's different from the old behavior and it constrains us to have to always keep exec output from the beginning and never delete it.

saltzm: Oh nevermind, I see, you changed this caching to happen in the top-level StreamReader. Nice!

modal/io_streams.py
saltzm: I don't remember the exact issues, but I'm glad it works! I thought what was happening was some issue wrapping `__aiter__` into `__iter__` but maybe I was just having other typing issues

PR #3773, opened 2025-11-24
Allow custom Function name without serializing

mwaskom:
Users may want to decorate a Function implementation multiple times to supply different configurations:

```python
def impl(...):
    ...

f_h100 = app.function(gpu="h100", ...)(impl)
f_a100 = app.function(gpu="a100", ...)(impl)

```

Because Functions need to have a unique name inside the App namespace, if you do this you need to pass `name=` in the decorator. And currently, when passing `name=` you are required to set `serialized=True`.

The latter restriction is because we don't differentiate between "the name that identifies the Function in the App namespace" and "the name that is used to locate the Function's implementation in the user's code during container startup. But that's not a deep limitation, just a matter of bookkeeping. This PR adds the necessary bookkeeping so that `name=` can be set without `serialized=True`.

---

## Compatibility checklist

</details>

## Changelog

- It is now possible to set a custom `name=` for a Function without setting `serialized=True`. This can be useful when decorating a function multiple times, e.g. applying multiple Modal configurations to the same implementation.

---

> [!NOTE]
> Separate user-visible function name from implementation qualname, enabling name= without serialized=True and using new implementation_name for imports.
> 
> - **Core API**:
>   - Add `implementation_name` to `api_pb2.Function`/`FunctionData` and populate it from `FunctionInfo`.
>   - `FunctionInfo`: compute `implementation_name` from qualname/class context; set `function_name = name_override or implementation_name` (removes prior name+serialized restriction).
> - **Runtime Import**:
>   - `user_code_imports.import_single_function_service` now resolves callables using `function_def.implementation_name` instead of `function_name`.
> - **Function creation/loading**:
>   - Include `implementation_name` in FunctionCreate/Precreate payloads and in GPU-ranked `FunctionData`.
> - **Tests & fixtures**:
>   - Update test function defs to include `implementation_name`.
>   - Add tests verifying custom `name=` without `serialized=True` and container execution of `custom_name`.
>   - Add support function `impl_for_custom_name` decorated as `@app.function(name="custom_name")`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f5e9bdf7dcfc4ab3531c6abbe21df1072a816d1d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_runtime/user_code_imports.py
+++ b/modal/_runtime/user_code_imports.py
@@ -465,7 +465,7 @@ def import_single_function_service(
     else:
         # Load the module dynamically
         module = importlib.import_module(function_def.module_name)
-        qual_name: str = function_def.import_name
+        qual_name: str = function_def.implementation_name
 
         if not is_global_object(qual_name):
             raise LocalFunctionError("Attempted to load a function defined in a function scope")
--- a/modal/_utils/function_utils.py
+++ b/modal/_utils/function_utils.py
@@ -129,7 +129,7 @@ class FunctionInfo:
 
     raw_f: Optional[Callable[..., Any]]  # if None - this is a "class service function"
     function_name: str
-    import_name: str
+    implementation_name: str
     user_cls: Optional[type[Any]]
     module_name: Optional[str]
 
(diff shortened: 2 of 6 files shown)


modal/_utils/function_utils.py
```diff
@@ -129,6 +129,7 @@ class FunctionInfo:
 
     raw_f: Optional[Callable[..., Any]]  # if None - this is a "class service function"
     function_name: str
+    import_name: str
```
thomasjpfan: During the sync, we spoke about renaming this `implementation_name`

modal/_runtime/user_code_imports.py
```diff
@@ -244,7 +244,7 @@ def import_single_function_service(
     else:
         # Load the module dynamically
         module = importlib.import_module(function_def.module_name)
-        qual_name: str = function_def.function_name
+        qual_name: str = function_def.implementation_name
```
thomasjpfan, line 468: As long as all the workers are using the new proto with `implementation_name` then this will be okay. Specifically, if we ship this client change and the workers have not been updated, then one would not be able to deploy using the new client.

To mitigate, should we merge `modal_proto/api.proto` in first and let workers pick it up and ship with this new argument first?

thomasjpfan (approved):

thomasjpfan (commented):
There is a merge conflict

PR #3771, opened 2025-11-24
refactor context manager 

molocule:
## Describe your changes
- refactor container entrypoint to use a context manager introduced in user code import
---

## Compatibility checklist

Check these boxes or delete any item (or this section) if not relevant for this PR.

- [ x] Client+Server: this change is compatible with old servers
- [ x] Client forward compatibility: this change ensures client can accept data intended for later versions of itself

Note on protobuf: protobuf message changes in one place may have impact to
multiple entities (client, server, worker, database). See points above.

---

> [!NOTE]
> Refactors container entrypoint to delegate execution via Service.execution_context, extracting event loop and lifecycle logic into runtime modules with a helper for serialized defs.
> 
> - **Container runtime**:
>   - Simplifies `modal/_container_entrypoint.py` to use `Service.execution_context(...)` for setup/teardown and calling finalized functions.
>   - Extracts fetching of serialized user class/function into `get_serialized_user_class_and_function`.
>   - Minor typing fixes (`cast`, optional metadata) and cleanup.
> - **Event loop**:
>   - Introduces `modal/_runtime/user_code_event_loop.py` with `UserCodeEventLoop` (moved from entrypoint).
> - **User code imports/lifecycle**:
>   - Expands `modal/_runtime/user_code_imports.py`:
>     - Adds lifecycle helpers: `call_lifecycle_functions`, `maybe_snapshot`, `create_breakpoint_wrapper`, and `_run_service_lifecycle`.
>     - Extends `Service` with `execution_context(...)` and implements it for `ImportedFunction` and `ImportedClass` (handling pre/post-enter, snapshot, ASGI lifespan, exit, and volume commit).
>     - Passes `function_def` into service instances; updates imports/webhook construction; minor API/typing adjustments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ff821e63d1716c1bda33d5653cd3b6c76a825236. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #3778, opened 2025-11-25
refactor _run_service_lifecycle to use context managers

molocule:
## Describe your changes
In response to this comment
https://github.com/modal-labs/modal-client/pull/3771/files#r2560551372

refactor _run_service_lifecycle to define the flow in terms of context managers

---

## Compatibility checklist

- [x ] Client+Server: this change is compatible with old servers
- [x ] Client forward compatibility: this change ensures client can accept data intended for later versions of itself

---

> [!NOTE]
> Refactors service lifecycle into context managers with explicit ASGI lifespan, disables SIGINT/SIGUSR1 during shutdown and volume commits, and adds tests/fixtures to validate full end-to-end lifecycle ordering.
> 
> - **Runtime (user_code_imports.py)**:
>   - **Lifecycle refactor**: Replace `_run_service_lifecycle` with context-managed flow: `lifecycle_presnapshot`, `lifecycle_postsnapshot`, and `lifecycle_asgi`; `Service.execution_context` orchestrates snapshot, breakpoint wrapper, function finalization, and ASGI lifespan.
>   - **Signal handling**: Add `disable_signals()`/`try_enable_signals()`; disable before ASGI shutdown and re-enable after `volume_commit`.
>   - **Volume commits**: Extract `volume_commit(...)`; commit after exits/lifespans, outside lifecycle managers.
>   - **Service API**: `Service` now holds `function_def`; `ImportedFunction`/`ImportedClass` implement new lifecycle hooks; move enter/exit invocations accordingly.
> - **Tests**:
>   - Add `test_full_lifecycle_order_signals_disabled_before_asgi_exit` asserting event order, signal state during ASGI shutdown, and volume commit.
>   - `_run_container_process` accepts `volume_mounts`.
>   - Add `FullLifecycleCls` and lifecycle event tracking in `test.supports.functions` to support the new test.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d86908d045104494af36de35917c10a2bd6540f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_runtime/user_code_imports.py
+++ b/modal/_runtime/user_code_imports.py
@@ -161,15 +161,15 @@ class Service(metaclass=ABCMeta):
         7. Handles cleanup (lifespan shutdown, 'exit' methods)
         """
         int_handler, usr1_handler = None, None
-        # 1. Pre-snapshot Enter
-        with self.lifecycle_presnapshot(event_loop, container_io_manager):
-            # 2. Snapshot -- If this container is being used to create a checkpoint, checkpoint the container after
-            # global imports and initialization. Checkpointed containers run from this point onwards.
-            maybe_snapshot(container_io_manager, self.function_def)
-            # 3. Breakpoint wrapper
-            create_breakpoint_wrapper(container_io_manager)
-            # 4. Post-snapshot Enter
-            try:
+        try:
+            # 1. Pre-snapshot Enter
+            with self.lifecycle_presnapshot(event_loop, container_io_manager):
+                # 2. Snapshot -- If this container is being used to create a checkpoint, checkpoint the container after
+                # global imports and initialization. Checkpointed containers run from this point onwards.
+                maybe_snapshot(container_io_manager, self.function_def)
+                # 3. Breakpoint wrapper
+                create_breakpoint_wrapper(container_io_manager)
+                # 4. Post-snapshot Enter
                 with self.lifecycle_postsnapshot(event_loop, container_io_manager):
                     # Get Functions
                     with container_io_manager.handle_user_exception():
@@ -181,13 +181,13 @@ class Service(metaclass=ABCMeta):
                             yield finalized_functions
                         finally:
                             int_handler, usr1_handler = disable_signals()
+        finally:
+            # 9. Volume commit - runs OUTSIDE all lifecycle managers so exit handlers
+            # have a chance to write to disk before we commit volumes
+            try:
+                volume_commit(container_io_manager, self.function_def)
             finally:
-                # 9. Volume commit - runs OUTSIDE all lifecycle managers so exit handlers
-                # have a chance to write to disk before we commit volumes
-                try:
-                    volume_commit(container_io_manager, self.function_def)
-                finally:
-                    try_enable_signals(int_handler, usr1_handler)
+                try_enable_signals(int_handler, usr1_handler)
 
 
 def construct_webhook_callable(
@@ -308,6 +308,7 @@ class ImportedFunction(Service):
         event_loop: UserCodeEventLoop,
         container_io_manager: "modal._runtime.container_io_manager.ContainerIOManager",
     ):
+        # This is a no-op for imported functions since @enter methods are not supported
         yield
 
     @contextmanager
@@ -316,6 +317,7 @@ class ImportedFunction(Service):
         event_loop: UserCodeEventLoop,
         container_io_manager: "modal._runtime.container_io_manager.ContainerIOManager",
     ):
+        # This is a no-op for imported functions since @enter methods are not supported
         yield
 
 


modal/_runtime/user_code_imports.py
```diff
@@ -340,52 +332,74 @@ def get_finalized_functions(
...
             call_lifecycle_functions(event_loop, container_io_manager, list(post_snapshot_methods.values()))
+        yield
 
-        # 5. Get Functions & Start Lifespan
-        with container_io_manager.handle_user_exception():
-            finalized_functions = self.get_finalized_functions(self.function_def, container_io_manager)
-
-        def exit_callback():
+    @contextmanager
+    def lifecycle_exit(
```
freider: This could just be the post-yield part of the `lifecycle_postsnapshot` I think?

modal/_runtime/user_code_imports.py
```diff
@@ -340,52 +332,74 @@ def get_finalized_functions(
...
+                # 7. Register Exit
+                with self.lifecycle_exit(event_loop, container_io_manager):
+                    # 5. Get Functions & Start Lifespan
+                    with container_io_manager.handle_user_exception():
+                        finalized_functions = self.get_finalized_functions(self.function_def, container_io_manager)
+                    with lifecycle_asgi(event_loop, container_io_manager, finalized_functions):
+                        # 6. Yield Finalized Functions
+                        yield finalized_functions
+                        int_handler, usr1_handler = disable_signals()
+                try:
```
freider: The volume_commit and enable_signals should probably run outside of all lifecycle managers so everything has a chance to wind down before we run any volume commits (in case exit handlers write to disk)

freider: This would btw be good to have a regression container test for, if we don't, e.g. a dummy class that has all kinds of lifecycles and asserts that each step happens in the expected order

1. modal.enter(snap=True)
2. modal.enter(snap=False)
3. asgi enter
4. function call
5. signals disabled
6. asgi exit
7. modal.exit
8. volume commit

modal/_runtime/user_code_imports.py
```diff
@@ -340,52 +332,74 @@ def get_finalized_functions(
...
-        with _run_service_lifecycle(
-            event_loop,
-            container_io_manager,
-            self.function_def,
-            finalized_functions,
-            exit_callback,
-        ):
-            yield finalized_functions
+    @contextmanager
+    def execution_context(
```
freider: One idea to add some more consistency guarantees for how lifecycles work and reduce some code duplication between classes and functions:
We could make the `execution_context` implementation part of the base Service class so this "full lifecycle" handler is always run, but make the lifecycle_pre/postsnapshot, into no-op context managers in case of function services - that should make them functionally identical right?

modal/_runtime/user_code_imports.py
molocule: ah yeah makes sense
i think just have lifecycle_presnapshot and lifecycle_postsnapshot makes sense as well after thinking it through (as opposed to also having lifecycle_exit)

modal/_runtime/user_code_imports.py
molocule: good point, fixed

modal/_runtime/user_code_imports.py
molocule: fixed!

modal/_runtime/user_code_imports.py
```diff
@@ -150,6 +160,37 @@ def execution_context(
...
+                    # 5. Start ASGI lifespan
+                    try:
+                        with lifecycle_asgi(event_loop, container_io_manager, finalized_functions):
+                            # 6. Yield Finalized Functions
+                            yield finalized_functions
+                    finally:
+                        # 8. Disable signals before exit/ASGI shutdown/lifecycle wind-down
+                        # From this point onward, ignore all SIGINT signals that come from
+                        # graceful shutdowns originating on the worker, as well as stray SIGUSR1 signals
+                        int_handler, usr1_handler = disable_signals()
```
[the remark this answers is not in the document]
molocule: fixed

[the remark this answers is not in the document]
molocule: also added test ;-;

modal/_runtime/user_code_imports.py
```diff
@@ -150,6 +160,34 @@ def execution_context(
...
+                        finalized_functions = self.get_finalized_functions(self.function_def, container_io_manager)
+                    # 5. Start ASGI lifespan
+                    with lifecycle_asgi(event_loop, container_io_manager, finalized_functions):
+                        # 6. Yield Finalized Functions
+                        try:
+                            yield finalized_functions
+                        finally:
+                            int_handler, usr1_handler = disable_signals()
+            finally:
+                # 9. Volume commit - runs OUTSIDE all lifecycle managers so exit handlers
```
freider: I think we should move this try/finally outside of the lifecycle_presnapshot (even if that has no exit handling) just for consistency (also in case we ever add something to the presnapshot exit handling)

modal/_runtime/user_code_imports.py
```diff
@@ -265,27 +303,20 @@ def get_finalized_functions(
...
-    ) -> Generator[dict[str, "FinalizedFunction"], None, None]:
-        # If this container is being used to create a checkpoint, checkpoint the container after
-        # global imports and initialization. Checkpointed containers run from this point onwards.
-        maybe_snapshot(container_io_manager, self.function_def)
-        create_breakpoint_wrapper(container_io_manager)
-
-        with container_io_manager.handle_user_exception():
-            finalized_functions = self.get_finalized_functions(self.function_def, container_io_manager)
+    ):
+        yield
```
freider, line 312: Could add some comment about this intentionally being a no-op for non-classes

modal/_runtime/user_code_imports.py
molocule: fixed!

modal/_runtime/user_code_imports.py
molocule: fixed!

PR #3784, opened 2025-11-28
Flash SDK part II (functional changes)

molocule:
## Describe your changes

functional changes to use http config for flash

---

## Compatibility checklist

---

> [!NOTE]
> Integrates Flash class http_server into container lifecycle (enter → stop → exit → close) with exit grace, adds validation to forbid callable methods on http_server classes, and updates tests and helpers accordingly.
> 
> - **Runtime/lifecycle (classes)**:
>   - Add `_FlashContainerEntry` and invoke it during postsnapshot lifecycle: `enter()` after post-snapshot enters, `stop()` before `@exit` handlers, then `close()` after exits.
>   - Wire Flash lifecycle using `function_def.http_config` in `ImportedClass.lifecycle_postsnapshot`.
> - **Validation**:
>   - Disallow mixing callable methods with class-level web interface by checking `http_config` in `@app.cls` and raising on `CALLABLE_INTERFACE` methods.
>   - Simplify flag composition: conflict check now only considers `WEB_INTERFACE` vs `CALLABLE_INTERFACE` in `_PartialFunction.validate_flag_composition`.
> - **Flash manager/API**:
>   - Add `exit_grace_period` to `FlashManager` and `flash_forward`; wait before closing tunnel; make `stop()` cancel heartbeat safely; expose via `synchronize_api(..., target_module=__name__)`.
> - **Tests/fixtures**:
>   - New Flash class tests for lifecycle ordering and RPC register/deregister sequence; validations for invalid decorator usage; options/overrides checks.
>   - Test harness updates to pass `http_config`, track Flash RPC calls, and add Flash sample classes in `test.supports.functions`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b2fcfc15225eb8233a22daade0556f8001e05fa1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal_proto/api.proto
+++ b/modal_proto/api.proto
@@ -2793,6 +2793,11 @@ message Sandbox {
   // Exec commands for the sandbox will be issued directly to the sandbox
   // command router running on the Modal worker.
   bool direct_sandbox_commands_enabled = 34;
+
+  // Internal: restricts sandbox to run on this specific instance type.
+  // Set by server during SandboxRestore to ensure the restored sandbox runs
+  // on the same instance type as the original snapshot.
+  string _restore_instance_type = 35;
 }
 
 message SandboxCreateConnectTokenRequest {
--- a/modal_version/__init__.py
+++ b/modal_version/__init__.py
@@ -1,4 +1,4 @@
 # Copyright Modal Labs 2025
 """Supplies the current version of the modal client library."""
 
-__version__ = "1.2.5.dev13"
+__version__ = "1.2.5.dev14"


modal/_runtime/user_code_imports.py
```diff
@@ -392,18 +393,23 @@ def lifecycle_postsnapshot(
...
         container_io_manager: "modal._runtime.container_io_manager.ContainerIOManager",
     ):
+        flash_entry = _FlashContainerEntry()
         # Identify the "enter" methods to run after resuming from a snapshot.
         if not self.function_def.is_auto_snapshot:
             post_snapshot_methods = _find_callables_for_obj(
                 self.user_cls_instance, _PartialFunctionFlags.ENTER_POST_SNAPSHOT
             )
             call_lifecycle_functions(event_loop, container_io_manager, list(post_snapshot_methods.values()))
+            if self.function_def.HasField("http_config"):
```
freider: It feels like an inconsistency that enter is conditionally called, but stop and close are always called from this scope.

modal/_runtime/user_code_imports.py
```diff
@@ -392,18 +393,23 @@ def lifecycle_postsnapshot(
...
+                flash_entry.enter(self.function_def.http_config)
         try:
             yield
         finally:
             if not self.function_def.is_auto_snapshot:
+                flash_entry.stop()
```
freider, line 408: Is it intentional/important that stop runs before @modal.exit methods and close runs after?

If not, we could possibly structure the entire `FlashContainerEntry` as a context manager instead (like the lifecycle variants)?

modal/_runtime/user_code_imports.py
freider: What if we just passed the http config to the FlashContainerEntry constructor instead, so the conditionality can be solved entirely internal to that context manager based on the value of the http config?

modal/_runtime/user_code_imports.py
molocule: yes, we want to first deregister the container, then run all exit handlers, and finally call close to close the tunnels. but i wasn't sure how to structure this as a context manager

modal/_runtime/user_code_imports.py
molocule: fixed!

modal/experimental/flash.py
```diff
@@ -692,3 +692,26 @@ def wrapper(obj: Union[Callable[..., Any], _PartialFunction]) -> _PartialFunctio
 
 
 http_server = synchronize_api(_http_server, target_module=__name__)
+
+
+class _FlashContainerEntry:
+    def __init__(self):
+        self.flash_manager: Optional[FlashManager] = None  # type: ignore
+        self.exit_grace_period = 0
```
thomasjpfan: Can `_FlashContainerEntry` take in the `http_config` as it's config?

```python
@dataclass
class _FlashContainerEntry:
    http_config: api_pb2.HTTPConfig

    def enter(self):
        if not self.http_config.port:  # assuming we always require a port here
             return
        flash_manager = flash_forward(
            self.http_config.port,
            startup_timeout=self.http_config.startup_timeout,
        )
```

modal/experimental/flash.py
```diff
@@ -692,3 +692,26 @@ def wrapper(obj: Union[Callable[..., Any], _PartialFunction]) -> _PartialFunctio
...
+                startup_timeout=http_config.startup_timeout,
+            )
+
+    def stop(self):
+        if self.flash_manager:
+            self.flash_manager.stop()
+
+    async def close(self):
+        if self.flash_manager:
+            await asyncio.sleep(self.exit_grace_period)
```
thomasjpfan: Should `exit_grace_period` be a parameter for the `FlashManager` object?

modal/experimental/flash.py
molocule: done!

modal/experimental/flash.py
molocule: done!

modal/experimental/flash.py
```diff
@@ -171,19 +173,22 @@ async def close(self):
...
             await self.stop()
 
+        await asyncio.sleep(self.exit_grace_period)
+
         logger.warning(f"[Modal Flash] Closing tunnel on {self.tunnel.url}.")
         await self.tunnel_manager.__aexit__(*sys.exc_info())
 
 
-FlashManager = synchronize_api(_FlashManager)
+FlashManager = synchronize_api(_FlashManager, target_module=__name__)
```
mwaskom, line 185: Does FlashManager need a synchronized interface? Where do we use it from a blocking context?

molocule: FlashManager.close() is asynchronous but i think we want to call it from user event loop via FlashContainerEntry in container entrypoint -- verifying this is the right way to call it

modal/_runtime/user_code_imports.py
freider: Is the idea that the exit handlers may do something with all the open connections (send responses etc.?) before closing the connections?

That makes sense, so lets keep the "three step" thing as is. Can we document the above info on why it is this way in the FlashEntry class so that it's clear in case someone refactors this code at some point in the future?

modal/experimental/__init__.py
```diff
@@ -20,6 +20,7 @@
 from ..image import DockerfileSpec, ImageBuilderVersion, _Image, _ImageRegistryConfig
 from ..secret import _Secret
 from .flash import (  # noqa: F401
+    _FlashContainerEntry,
```
mwaskom: Does this need to be in the main `modal.experimental` namespace?

modal/experimental/__init__.py
```diff
@@ -20,6 +20,7 @@
...
     flash_forward,
     flash_get_containers,
     flash_prometheus_autoscaler,
```
mwaskom, line 25: Also curious about our plan with these. Users don't use them directly anymore right?

modal/experimental/__init__.py
molocule: ah this does not

modal/experimental/__init__.py
molocule: yes, but we're in the process of migrating users off of this. will keep this for now until we fully transition everyone off

test/flash_cls_test.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+        class InvalidFlashClassDecoratorOnMethod:
+            @modal.experimental.http_server(8001, proxy_regions=["us-east", "us-west", "ap-south"])
+            @modal.enter(snap=True)
+            def enter(self):
+                self.entered = True
+                self.process = subprocess.Popen(["python3", "-m", "http.server", "8001"])
+
+
+# def test_invalid_flash_class_method():
```
thomasjpfan: Should this test be uncommented?

modal/_runtime/user_code_imports.py
molocule: done

test/flash_cls_test.py
```diff
@@ -0,0 +1,200 @@
...
+
+#             @modal.enter(snap=True)
+#             def enter(self):
+#                 self.entered = True
+#                 self.process = subprocess.Popen(["python3", "-m", "http.server", "8001"])
+
+#             @modal.method()
+#             def modal_method(self, y: int) -> int:
+#                 return y**2
+
```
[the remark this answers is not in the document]
mwaskom: that's cheating, someone else already called this out!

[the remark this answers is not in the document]
molocule: fixed : ))

test/flash_cls_test.py
molocule: yes, just fixed

modal/_runtime/user_code_imports.py
```diff
@@ -392,18 +393,22 @@ def lifecycle_postsnapshot(
...
             call_lifecycle_functions(event_loop, container_io_manager, list(post_snapshot_methods.values()))
+            flash_entry.enter()
         try:
             yield
         finally:
             if not self.function_def.is_auto_snapshot:
+                flash_entry.stop()
                 exit_methods = _find_callables_for_obj(self.user_cls_instance, _PartialFunctionFlags.EXIT)
                 call_lifecycle_functions(event_loop, container_io_manager, list(exit_methods.values()))
+                flash_entry.close()
```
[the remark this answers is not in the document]
molocule: this is ok -- we need deregistering to succeed

PR #3790, opened 2025-12-02
[flash] add checks for clustered flash functions autoscaling

emilyhanyf:
this makes sure that flash server does not need to worry about rounding up/down to the nearest multiple of `cluster_size` for clustered flash function autoscaling. more info [here](https://www.notion.so/modal-com/Flash-Autoscaler-Overview-2a11e7f16949806588c2e8bd26c87ee5?source=copy_link#2bd1e7f1694980aeb35bed93e662a855)

---

> [!NOTE]
> Enforces `min_containers`, `max_containers`, and `buffer_containers` to be multiples of `cluster_size` for clustered functions, with corresponding tests.
> 
> - **Functions/runtime**:
>   - Add validation in `modal/_functions.py` to ensure `autoscaler_settings` (`min_containers`, `max_containers`, `buffer_containers`) are multiples of `cluster_size` when `cluster_size > 1`; raises `InvalidError` with detailed message.
> - **Tests**:
>   - Add `test_autoscaler_settings_are_multiples_of_cluster_size` in `test/function_test.py` covering invalid and valid configurations.
>   - Import `modal.experimental` for `@modal.experimental.clustered` decorator usage.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a9fa3ccd07949d67c1e1947128cd7cdf1fa561ab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_functions.py
+++ b/modal/_functions.py
@@ -794,7 +794,7 @@ class _Function(typing.Generic[P, ReturnType, OriginalReturnType], _Object, type
                 if value and value % cluster_size != 0:
                     raise InvalidError(
                         f"`{field}` ({value}) must be a multiple of `cluster_size` ({cluster_size}) "
-                        f"for clustered functions"
+                        f"for clustered Functions"
                     )
 
         if _experimental_custom_scaling_factor is not None and (


modal/_functions.py
```diff
@@ -778,6 +779,24 @@ def from_local(
         if scaledown_window is not None and scaledown_window <= 0:
             raise InvalidError("`scaledown_window` must be > 0")
 
+        # For clustered flash functions, container settings must be multiples of cluster_size
+        if cluster_size is not None and cluster_size > 1 and is_flash_object(experimental_options, http_config):
```
molocule: is this true outside of flash functions as well?

mwaskom: +1, what's flash specific here?

emilyhanyf: good point! we should enforce this for non-flash functions as well

modal/_functions.py
```diff
@@ -778,6 +779,24 @@ def from_local(
...
+            if max_containers is not None and max_containers % cluster_size != 0:
+                raise InvalidError(
+                    f"`max_containers` ({max_containers}) must be a multiple of `cluster_size` ({cluster_size}) "
+                    f"for clustered flash functions"
+                )
+            if buffer_containers is not None and buffer_containers % cluster_size != 0:
+                raise InvalidError(
+                    f"`buffer_containers` ({buffer_containers}) must be a multiple of `cluster_size` ({cluster_size}) "
+                    f"for clustered flash functions"
+                )
```
mwaskom: Could probably cut down on some code duplication with a for loop if you do this after building the `autoscaler_settings` message btw.

modal/_functions.py
```diff
@@ -779,31 +778,23 @@ def from_local(
...
         )
 
+        # For clustered functions, container settings must be multiples of cluster_size
+        if cluster_size is not None and cluster_size > 1:
+            for field in ["min_containers", "max_containers", "buffer_containers"]:
+                value = getattr(autoscaler_settings, field)
+                if value and value % cluster_size != 0:
+                    raise InvalidError(
+                        f"`{field}` ({value}) must be a multiple of `cluster_size` ({cluster_size}) "
+                        f"for clustered functions"
```
mwaskom: Super nit: 

```suggestion
                        f"for clustered Functions"
```

mwaskom (approved):
Thanks for this!

modal/_functions.py
```diff
@@ -785,6 +785,16 @@ def from_local(
             scaledown_window=scaledown_window,
         )
 
+        # For clustered functions, container settings must be multiples of cluster_size
+        if cluster_size is not None and cluster_size > 1:
```
thomasjpfan, line 791: I know I suggested this in the doc, but I think this is a breaking change. For me, I'm okay with it breaking.

Do we have users that set `*_containers` to a value that is not a multiple of `cluster_size`?

modal/_functions.py
```diff
@@ -785,6 +785,16 @@ def from_local(
...
+            for field in ["min_containers", "max_containers", "buffer_containers"]:
+                value = getattr(autoscaler_settings, field)
+                if value and value % cluster_size != 0:
+                    raise InvalidError(
```
thomasjpfan, line 795: Also can we add a quick unit test to check this failure?

molocule:
corresponding server side check! https://github.com/modal-labs/modal/pull/30744

modal/_functions.py
emilyhanyf: discussed via dm and think this breaking change is okay. there's a list of user that use clustered functions but might be too long to check each individually.

PR #3792, opened 2025-12-03
Add `_restore_instance_type` to Sandbox definition

devennavani:
I want to fully deprecate the `_instance_types` field on the `SchedulerPlacement` proto so that `SchedulerPlacement` is only used for user-specified scheduling constraints.

The final only place it is being used is in `SandboxRestore` to restrict a restored Sandbox to run on the same instance type as where the snapshot was taken. Let's instead have an internal field on the Sandbox definition.

Server PR #30338 modifies `SandboxRestore` to set this field.


saltzm:
LGTM

PR #3793, opened 2025-12-03
Deprecate file I/O methods

AmitPr:
## Describe your changes

---

## Compatibility checklist

---

</details>

## Changelog

* Removed `replace_bytes` and `delete_bytes` from Sandbox FileIO

---

> [!NOTE]
> Remove `FileIO.delete_bytes`/`replace_bytes` by turning them into deprecation errors and delete related tests.
> 
> - **File I/O API**:
>   - Remove byte-range mutation helpers: `delete_bytes` and `replace_bytes` now raise `deprecation_error((2025, 12, 3), ...)`.
>   - Drop unused `InvalidError` import.
> - **Tests**:
>   - Remove tests and imports related to `delete_bytes` and `replace_bytes`; keep remaining FileIO tests intact.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f777b1911c295e55fc73a07d83e233c8f9c677f3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


modal/file_io.py
```diff
@@ -46,11 +47,13 @@
 
 
 async def _delete_bytes(file: "_FileIO", start: Optional[int] = None, end: Optional[int] = None) -> None:
-    """Delete a range of bytes from the file.
+    """mdmd:hidden
+    Delete a range of bytes from the file.
```
mwaskom: We should also indicate that this method is no longer functional in the docstring

modal/file_io.py
```diff
@@ -46,11 +47,13 @@
...
 
     `start` and `end` are byte offsets. `start` is inclusive, `end` is exclusive.
     If either is None, the start or end of the file is used, respectively.
     """
+    deprecation_error((2025, 12, 3), "delete_bytes is deprecated and will be removed in a future release.")
```
mwaskom: I think the phrasing is a bit wrong here, this is the sort of thing we would say when the method is still functional. I'd say at this point the functionality has been removed.

Normally we'd also try to point the user towards the replacement or other recommendation in this context. I am not sure what's best though, so maybe not necessary here.

PS you can delete the dead code below the error too.

mwaskom (changes requested):

modal/file_io.py
AmitPr: Just to be clear, the server-side changes haven't gone out so the methods still work -- updating the docstring is to reflect _just_ the client-side status, correct?

mwaskom: That's correct. I'm not sure we have any extant examples in the current codebase but typically would want to make the description something like `Deprecated: delete_bytes has been removed` so it's really obvious from IDE popups etc.

PR #3796, opened 2025-12-05
Aio warnings: Make SandboxSnapshot.from_id sync

freider:
Also makes the hydration logic actually do something, if a user potentially calls .hydrate()

It's still a bit weird that for the standard use case of passing SandboxSnapshot.from_id into a Sandbox.from_snapshot call, the hydration isn't actually used - only the object_id, so for most purposes `SandboxSnapshot` could just as well have been a string or a dataclass or something...

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Add async-context blocking warnings, deprecate .aio constructors, refactor `Volume.batch_upload` to an async context manager, and modernize `from_id` APIs; also upgrade synchronicity and tests.
> 
> - **Async utils & warnings**:
>   - Add `BlockingInAsyncContextWarning` and intelligent code rewrite suggestions; configure `synchronizer` with blocking-in-async callback.
>   - Detect interactive IPython via new `is_interactive_ipython()`; suppress warnings in notebooks.
> - **API deprecations & constructors**:
>   - Introduce `deprecate_aio_usage` decorator; deprecate `.aio` variants of constructors.
>   - Make `FunctionCall.from_id`, `Image.from_id`, and `SandboxSnapshot.from_id` synchronous classmethods returning wrapped types; adjust client translation.
> - **Volume**:
>   - Add `live_method_contextmanager`; make `Volume.batch_upload` an async context manager (no manual `__aenter__/__aexit__`); minor typing tweaks (`read_file`).
> - **Client & GRPC testing**:
>   - Refine `_Client` typing/fields; ensure task context loop handling; improve mock servicer fallbacks with method names.
> - **App/notebook behavior**:
>   - Use `is_interactive_ipython()` to hide duplicate function warnings in notebooks; add notebook tests.
> - **Tests**:
>   - Widespread updates to async usage (`async with`/`await`), new tests for sync-in-async rewrite logic and IPython detection; adjust expectations.
> - **Tooling/CI**:
>   - Bump `synchronicity` to `0.10.4`; update `mypy`, `nbclient`, `jupytext`; workflows install new versions.
>   - pytest config treats `BlockingInAsyncContextWarning` as error by default.
> - **Misc**:
>   - Small fixes in `tasks.py` (venv path match, deprecation message handling).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 66a3e6a797c0b9e51b9417e6c83a3780620b9de5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/snapshot.py
+++ b/modal/snapshot.py
@@ -1,12 +1,15 @@
 # Copyright Modal Labs 2024
-from typing import Optional
+from typing import Optional, cast
 
+import typing_extensions
+
+import modal.client
 from modal_proto import api_pb2
 
 from ._load_context import LoadContext
 from ._object import _Object
 from ._resolver import Resolver
-from ._utils.async_utils import synchronize_api
+from ._utils.async_utils import deprecate_aio_usage, synchronize_api, synchronizer
 from .client import _Client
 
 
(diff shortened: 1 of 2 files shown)


modal/snapshot.py
```diff
@@ -19,27 +22,35 @@ class _SandboxSnapshot(_Object, type_prefix="sn"):
...
-        obj._hydrate(sandbox_snapshot_id, client, None)
-
-        return obj
+        obj = _SandboxSnapshot._from_loader(
+            _load, rep, load_context_overrides=LoadContext(client=_client), hydrate_lazily=True
+        )
+        # Setting the object id directly is a bit hacky, but
+        # it avoids hydrating the object fully if it's going
+        # to be used only for its object id anyway
+        obj._object_id = sandbox_snapshot_id
```
freider, line 52: This was mostly so we don't add an extra and unnecessary `SandboxSnapshotGet` rpc by forcing actual hydration of this object in Sandbox.from_snapshot...

freider: I think this might turn out to be a common pattern (@danielnorberg had to add something similar to Image) so we should consider if we want to make "having an object id but not any metadata or client object" one of the possible "hydration states" of any modal object. Needs some thought

mwaskom: I'm open to that, it makes sense as a requirement. I think it's less important to be incredibly strict about the cardinality of hydration states and more important that it's very clear when looking at the code (ideally both from the perspective of a human developer and the perspective of the type checker) what state the object is required to be in.

PR #3800, opened 2025-12-08
Enable `modal shell im-id`, and warn on incompatible arguments

ehdr:
## Describe your changes

- Enable starting a shell using a Modal Image ID: `modal shell im-abc123`
- Issues warnings if you pass combinations of arguments that are not allowed (e.g. `--cpu` together with a Sandbox `sb-...` ref, etc. etc.)
- Refactor quite a bit to improve testability.

---

## Compatibility checklist

</details>

## Changelog

- Enabled starting a shell using a Modal Image ID with `modal shell im-abc123`. `modal shell` now also warns you if you pass invalid combinations of arguments (like `--cpu` together with the ID of an already running Sandbox, etc.).

---

> [!NOTE]
> Adds support for `modal shell im-<id>`, enforces incompatible-argument checks, and refactors shell routing with comprehensive tests.
> 
> - **CLI (`modal shell`)**
>   - Enable launching a shell from a Modal Image ID (`im-...`).
>   - Enforce incompatible-argument validation with clear errors when attaching to running containers/Sandboxes, using function refs, or Image IDs.
>   - Refactor routing logic into helpers:
>     - `_is_valid_modal_id`, `_is_running_container_ref`, `_start_shell_in_running_container`.
>     - `_function_spec_from_ref`, `_start_shell_from_function_spec`, `_start_shell_from_image`.
>     - Parameter utilities: `_params_from_signature`, `_passed_forbidden_args`.
>   - Normalize/explicit option flags (e.g., `--image`, `--add-python`, `--cpu`, `--pty`, etc.); make `env` optional.
> - **Tests**
>   - Add routing tests covering container/Sandbox refs, function refs, image refs/params, and default path.
>   - Add validation tests for forbidden argument combinations and parameter declaration checks.
>   - Add unit tests for helper utilities and error messaging.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2af0ad967f5f4dec9e0cc9e8625b4e1c747218ae. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/cli/shell.py
+++ b/modal/cli/shell.py
@@ -29,13 +29,19 @@ from .run import _get_runnable_list
 from .utils import ENV_OPTION, is_tty
 
 
-def _params_from_signature(func: Callable[..., Any]) -> dict[str, Any]:
+def _params_from_signature(
+    func: Callable[..., Any],
+) -> dict[str, typer.models.ParameterInfo]:
     sig = inspect.signature(func)
-    return {param_name: param.default for param_name, param in sig.parameters.items()}
+    params = {param_name: param.default for param_name, param in sig.parameters.items()}
+    assert all(isinstance(param, typer.models.ParameterInfo) for param in params.values()), (
+        f"All params to {func.__name__} must be of type typer.models.ParameterInfo."
+    )
+    return params
 
 
 def _passed_forbidden_args(
-    param_objs: dict[str, Any],
+    param_objs: dict[str, typer.models.ParameterInfo],
     passed_args: dict[str, Any],
     allowed: Callable[[str], bool],
 ) -> list[str]:
@@ -44,6 +50,8 @@ def _passed_forbidden_args(
     for param_name, param_obj in param_objs.items():
         if allowed(param_name):
             continue
+
+        assert param_obj.param_decls is not None, "All params must be typer.models.ParameterInfo, and have param_decls."
 
         if passed_args.get(param_name) != param_obj.default:
             passed_forbidden.append("/".join(param_obj.param_decls))


modal/cli/shell.py
```diff
@@ -127,111 +307,71 @@ def shell(
...
+    if ref is not None and not _is_valid_modal_id(ref, "im-"):
+        # If ref it not a Modal Image ID, then it's a function reference, and we'll start a new container from its spec.
+        function_spec = _function_spec_from_ref(ref, use_module_mode)
+
+        if passed_forbidden := _passed_forbidden_args(
+            param_objs, locals(), allowed_params={"cmd", "env", "pty", "ref", "use_module_mode"}
+        ):
+            raise ClickException(
+                f"Cannot specify container configuration arguments ({', '.join(passed_forbidden)}) "
+                f"when starting a new container from a function reference ('{ref}')."
```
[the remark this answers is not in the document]
ehdr: Alright!

modal/cli/shell.py
```diff
@@ -27,6 +29,176 @@
 from .utils import ENV_OPTION, is_tty
 
 
+def _params_from_signature(func: Callable[..., Any]) -> dict[str, Any]:
+    sig = inspect.signature(func)
+    return {param_name: param.default for param_name, param in sig.parameters.items()}
+
+
+def _passed_forbidden_args(
+    param_objs: dict[str, Any],
```
thomasjpfan: From reading the code, it's hard to tell that `param.default` is an `click.Parameter` object in `_params_from_signature`, which are passed in here. (At first, I thought it would be some inspect.Parameter` object, which also has a `default`)

Can we somehow get the type of `param_objs` to be `dict[str, click.Parameter]`?

modal/cli/shell.py
```diff
@@ -27,6 +29,176 @@
...
+    if forbidden_params and (overlap := forbidden_params & allowed_params):
+        raise ValueError(f"Parameters cannot be both forbidden and allowed: {overlap}")
+
+    passed_forbidden: list[str] = []
+    for param_name, param_obj in param_objs.items():
+        if param_name in allowed_params:
+            continue
+
+        if forbidden_params and param_name not in forbidden_params:
+            continue
```
thomasjpfan: I know it's more verbose for the `modal shell im-` case, but can we only have `allowed_params` and not have `forbidden_params`? And then the `im-*` would pass in a long list of allow parameters.

I find it a clearer to not need to handle any interaction between `forbidden_params` and `allowed_params` and be a little more verbose.

ehdr: Yeah, I see the problem... it just feels like [the intention becomes much less clear](https://github.com/modal-labs/modal-client/blob/ehdr/shell-async/modal/cli/shell.py#L366), i.e. arguably easier to understand what's going on (you can't combine image params with image IDs) when you see `_passed_forbidden_args(..., forbidden_params={"add_python", "image"})` vs. `_passed_forbidden_args(..., allowed_params={"cpu", "memory", "gpu", "volume", "secret", "cloud", "region", "env", "add-local"})`...

thomasjpfan: I see what you mean. Can this be a lambda so that the image case looks like this?

```python
_passed_forbidden_args(..., allow_name=lambda name: name not in {"add_python", "image"})
```

ehdr: Yes, great idea, much better, thanks!

modal/cli/shell.py
```diff
@@ -27,6 +29,171 @@
...
+    region: Optional[str],
+    pty: bool,
+) -> None:
+    volumes = {f"/mnt/{vol}": Volume.from_name(vol) for vol in volume}
+    secrets = [Secret.from_name(s) for s in secret]
+
+    mounts = []
+    for local_path_str in add_local:
+        local_path = Path(local_path_str).expanduser().resolve()
+        remote_path = PurePosixPath(f"/mnt/{local_path.name}")
```
thomasjpfan, line 171: **Not actionable for this PR**: I know this is in the original implementation, but there is an edge case where a local path name and a volume name can be the same. What happens in that situation?

ehdr: Yeah good catch, this could also happen if you add files with the same name but from different directories I think? I think the last one will win, but we should warn in this case. Will make a follow-up PR for that!

PR #3806, opened 2025-12-10
Raise modal.Error types for all gRPC status codes

mwaskom:
This adds middleware that automatically maps gRPC status responses into our own error types. We previously did this in an ad hoc manner in some places, but had many gaps and leaked `grpclib.GRPCError` out to end users. This is problematic because it ties us to using `grpclib` as our client implementation, since user code may depend on handling the `GRPCError` to respond correctly to Modal error states. This PR aims for a gradual migration where users can be warned about the upcoming change and adapt their code to be future-compatible without having it break.

I'm not super opinionated about the exact mapping from gRPC status to exception type that I chose; feel free to suggest alternative groupings.

---

## Compatibility checklist

</details>

## Changelog

- The Modal SDK will no longer propagate `grpclib.GRPCError` types out to the user; our own `modal.Error` subtypes will be used instead. To avoid disrupting user code that has relied on `GRPCError` exceptions for control flow, we are temporarily making some exception types subclass `GRPCError` so that they will also be caught by `except grpclib.GRPCError` statements. Accessing the `.status` attribute of the exception will issue a deprecation warning, but warnings cannot be issued if the exception object is only caught and there is no other interaction with it. We advise proactively migrating any exception handling to use Modal types. as we will remove the dependency on `grpclib` types entirely in the future.

---

> [!NOTE]
> Introduces middleware to map all gRPC status codes to Modal Error types and refactors SDK/CLI to raise these consistently, updating call sites and tests.
> 
> - **Core / Middleware**:
>   - Add `modal._grpc_client` wrappers with `grpc_error_converter` and `_STATUS_TO_EXCEPTION` mapping all `grpclib.GRPCError` statuses to `modal.exception` types; apply to unary and streaming calls (including retries).
> - **Exceptions**:
>   - Implement Modal exception classes that temporarily subclass `grpclib.GRPCError` with deprecation warnings on GRPC attributes.
> - **Refactors across SDK**:
>   - Replace ad hoc `GRPCError` handling with Modal errors in `functions`, `sandbox`, `volume`, `queue`, `dict`, `io_streams`, `tunnel`, `cls`, runners, and CLI commands.
>   - Simplify CLI error surfaces (remove status-specific panels/handlers) and rely on unified exceptions.
>   - Adjust retry paths to wrap errors and preserve details; keep server-throttle handling in `grpc_utils`.
> - **Behavioral tweaks**:
>   - Map `RESOURCE_EXHAUSTED` to `queue.Full` where appropriate and convert 413 payload errors to `RequestSizeError` with clearer messages.
> - **Tests**:
>   - Update existing tests to expect Modal exceptions (e.g., `InternalError`, `ConflictError`).
>   - Add new exception mapping tests and adjust gRPC utils tests to call `.direct` where needed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9128259adcd2be0a12aaf42956271f3da8bde2f4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_utils/task_command_router_client.py
+++ b/modal/_utils/task_command_router_client.py
@@ -18,6 +18,7 @@ from modal.exception import ConflictError, ExecTimeoutError
 from modal_proto import api_pb2, task_command_router_pb2 as sr_pb2
 from modal_proto.task_command_router_grpc import TaskCommandRouterStub
 
+from .._grpc_client import grpc_error_converter
 from .async_utils import aclosing
 from .grpc_utils import RETRYABLE_GRPC_STATUS_CODES, connect_channel
 
@@ -247,9 +248,10 @@ class TaskCommandRouterClient:
 
     async def exec_start(self, request: sr_pb2.TaskExecStartRequest) -> sr_pb2.TaskExecStartResponse:
         """Start an exec'd command, properly retrying on transient errors."""
-        return await call_with_retries_on_transient_errors(
-            lambda: self._call_with_auth_retry(self._stub.TaskExecStart, request)
-        )
+        with grpc_error_converter():
+            return await call_with_retries_on_transient_errors(
+                lambda: self._call_with_auth_retry(self._stub.TaskExecStart, request)
+            )
 
     async def exec_stdio_read(
         self,
@@ -284,9 +286,10 @@ class TaskCommandRouterClient:
         else:
             raise ValueError(f"Invalid file descriptor: {file_descriptor}")
 
-        async with aclosing(self._stream_stdio(task_id, exec_id, sr_fd, deadline)) as stream:
-            async for item in stream:
-                yield item
+        with grpc_error_converter():
+            async with aclosing(self._stream_stdio(task_id, exec_id, sr_fd, deadline)) as stream:
+                async for item in stream:
+                    yield item
 
     async def exec_stdin_write(
         self, task_id: str, exec_id: str, offset: int, data: bytes, eof: bool
@@ -304,9 +307,10 @@ class TaskCommandRouterClient:
               from the RPC itself.
         """
         request = sr_pb2.TaskExecStdinWriteRequest(task_id=task_id, exec_id=exec_id, offset=offset, data=data, eof=eof)
-        return await call_with_retries_on_transient_errors(
-            lambda: self._call_with_auth_retry(self._stub.TaskExecStdinWrite, request)
-        )
+        with grpc_error_converter():
+            return await call_with_retries_on_transient_errors(
+                lambda: self._call_with_auth_retry(self._stub.TaskExecStdinWrite, request)
+            )
 
     async def exec_poll(
         self, task_id: str, exec_id: str, deadline: Optional[float] = None
@@ -330,15 +334,17 @@ class TaskCommandRouterClient:
         timeout = deadline - time.monotonic() if deadline is not None else None
         if timeout is not None and timeout <= 0:
             raise ExecTimeoutError(f"Deadline exceeded while polling for exec {exec_id}")
-        try:
-            return await asyncio.wait_for(
-                call_with_retries_on_transient_errors(
-                    lambda: self._call_with_auth_retry(self._stub.TaskExecPoll, request)
-                ),
-                timeout=timeout,
-            )
-        except asyncio.TimeoutError:
-            raise ExecTimeoutError(f"Deadline exceeded while polling for exec {exec_id}")
+
+        with grpc_error_converter():
+            try:
+                return await asyncio.wait_for(
+                    call_with_retries_on_transient_errors(
+                        lambda: self._call_with_auth_retry(self._stub.TaskExecPoll, request)
+                    ),
+                    timeout=timeout,
+                )
+            except asyncio.TimeoutError:
+                raise ExecTimeoutError(f"Deadline exceeded while polling for exec {exec_id}")
 
     async def exec_wait(
         self,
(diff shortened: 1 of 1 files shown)


modal/_tunnel.py
```diff
@@ -186,13 +185,10 @@ def some_function():
...
-        if exc.status == Status.ALREADY_EXISTS:
-            raise InvalidError(f"Port {port} is already forwarded")
-        elif exc.status == Status.UNAVAILABLE:
-            raise RemoteError("Relay server is unavailable") from exc
-        else:
-            raise
+    except AlreadyExistsError as exc:
+        raise InvalidError(f"Port {port} is already forwarded")
+    except ServiceError as exc:
+        raise RemoteError("Relay server is unavailable") from exc
```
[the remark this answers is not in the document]
mwaskom: I'm not sure I agree that we need to be propagating connection error states at the level of granularity in the gRPC status model 🤔 .

modal/_functions.py
```diff
@@ -1126,12 +1126,8 @@ async def _load(
...
                 try:
                     response: api_pb2.FunctionCreateResponse = await load_context.client.stub.FunctionCreate(request)
-                except GRPCError as exc:
-                    if exc.status == Status.INVALID_ARGUMENT:
-                        raise InvalidError(exc.message)
-                    if exc.status == Status.FAILED_PRECONDITION:
-                        raise InvalidError(exc.message)
-                    if exc.message and "Received :status = '413'" in exc.message:
+                except Exception as exc:
+                    if "Received :status = '413'" in str(exc):
```
thomasjpfan, line 1130: It'll be nice not to read the message from this exception, but I think this error message comes from `nginx`, so we do not have as much flexibility to adjust the message on the server.

modal/_grpc_client.py
```diff
@@ -29,20 +52,14 @@ def __exit__(self, exc_type, exc, traceback) -> Literal[False]:
...
-                if not use_full_traceback:
-                    # just include the frame in grpclib that actually raises the GRPCError
-                    tb = exc.__traceback__
-                    while tb.tb_next:
-                        tb = tb.tb_next
-                    exc.with_traceback(tb)
-                    raise exc from None  # from None to skip the grpc-internal cause
-                raise exc
+                modal_exc = _STATUS_TO_EXCEPTION[exc.status](exc.message)
+                modal_exc.message = exc.message
```
thomasjpfan: For consistency with the few lines below:

```suggestion
                modal_exc._grpc_message = exc.message
```

(Or the other way around and set `modal_exc.status = exc.status`.)

modal/_tunnel.py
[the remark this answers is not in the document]
thomasjpfan: I would give more granularity if the error promotes different behavior. I see that the 5 status that you mapped to `ServiceError` are the ones we automatically retry.

Do you intend for users to catch the `RemoteError` and retry themselves?

modal/cli/environment.py
```diff
@@ -61,13 +59,7 @@ def list_(json: Optional[bool] = False):
...
 def create(name: Annotated[str, typer.Argument(help="Name of the new environment. Must be unique. Case sensitive")]):
     check_environment_name(name)
-
-    try:
-        environments.create_environment(name)
-    except GRPCError as exc:
-        if exc.status == Status.INVALID_ARGUMENT:
-            raise InvalidError(exc.message)
-        raise
+    environments.create_environment(name)
```
thomasjpfan, line 62: Much cleaner! ✨

modal/cli/network_file_system.py
```diff
@@ -200,14 +194,8 @@ async def rm(
...
     console = make_console()
-    try:
-        await volume.remove_file(remote_path, recursive=recursive)
-        console.print(OutputManager.step_completed(f"{remote_path} was deleted successfully!"))
-
-    except GRPCError as exc:
-        if exc.status in (Status.NOT_FOUND, Status.INVALID_ARGUMENT):
-            raise UsageError(exc.message)
-        raise
+    await volume.remove_file(remote_path, recursive=recursive)
```
thomasjpfan, line 197: The `UsageError` shows up a little nicer:

Screenshot 2025-12-12 at 4 11 52 PM

with this PR:

Screenshot 2025-12-12 at 4 09 17 PM

modal/exception.py
```diff
@@ -26,10 +28,112 @@ class Error(Exception):
...
+
+    def _warn_on_grpc_error_attribute_access(self) -> None:
+        from ._utils.deprecation import deprecation_warning  # Avoid circular import
+
+        exc_type = type(self).__name__
+        deprecation_warning(
+            (2025, 12, 9),
+            "Modal will stop propagating the `grpclib.GRPCError` type in the future. "
+            f"Update your code so that it catches `modal.exception.{exc_type}` directly "
+            "to avoid changes to error handling behavior in the future.",
```
thomasjpfan, line 101: Do we need to document the mapping from GRPCStatus to our own exceptions? (So users can migrate without having to see the deprecation warning)

If a user is depending on a GRPCError status that we grouped together, then they would not get the exact same behavior as before. I feel like it's okay, but it's a slight risk.

modal/_functions.py
mwaskom: Yep that's right, re. nginx. This is a best-effort attempt to raise a more user-friendly error but I don't think it's critical if we end up falling through and propagating a different string with `413` in the message.

modal/exception.py
mwaskom: I think it's self-documenting in the deprecation warning right? It should tell them what type is being used and indicate that they should migrate their except clause to catch that type.

modal/_functions.py
mwaskom: Also I wasn't actually able to trigger this error case any more with some testing, but might have been doing something dumb. Maybe there was a change somewhere else.

modal/_utils/task_command_router_client.py
```diff
@@ -124,11 +124,9 @@ async def try_init(
...
         try:
             resp = await fetch_command_router_access(server_client, task_id)
-        except GRPCError as exc:
-            if exc.status == Status.FAILED_PRECONDITION:
-                logger.debug(f"Command router access is not enabled for task {task_id}")
-                return None
-            raise
+        except ConflictError:
+            logger.debug(f"Command router access is not enabled for task {task_id}")
+            return None
```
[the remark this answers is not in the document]
mwaskom: I think this is fine as we don't currently return Status.ABORTED here.

modal/io_streams.py
```diff
@@ -697,11 +697,8 @@ async def drain(self) -> None:
                         input=api_pb2.RuntimeInputMessage(message=data, message_index=index, eof=self._is_closed),
                     ),
                 )
-        except GRPCError as exc:
-            if exc.status == Status.FAILED_PRECONDITION:
-                raise ValueError(exc.message)
-            else:
-                raise exc
+        except ConflictError as exc:
+            raise ValueError(str(exc))
```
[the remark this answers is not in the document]
mwaskom: I think this is fine in the sense that we don't return Status.ABORTED here. It's unfortunate that this was previously being coerced to ValueError which doesn't really make sense, but I'll err on the side of consistent types.

modal/cli/network_file_system.py
mwaskom: Ah I think this is because VolumeListFiles2 is UnaryStreaming which I haven't handled yet.

modal/exception.py
thomasjpfan: > I think it's self-documenting in the deprecation warning right? 

I'm thinking of writing down the mapping somewhere, so one does not need to run the code and see the deprecation warning. Also FDE's can link users somewhere when they are going through the migration.

The most obvious place to put this info would be at the top of https://modal.com/docs/reference/modal.exception#modalexception .

modal/exception.py
```diff
@@ -26,10 +28,112 @@ class Error(Exception):
...
+
+    # These will be set on the instance in our error handling middleware
+    _grpc_message: str
+    _grpc_status: grpclib.Status
+    _grpc_details: Any
+
+    def __init__(self, message: Optional[str] = None):
+        # Override GRPCError's init and repr to behave more like a regular Exception
+        # (We don't customize these anywhere in our custom error types currently).
+        self._message = message or ""
```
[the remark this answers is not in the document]
mwaskom: This seems wrong to me but it definitely is a little finicky.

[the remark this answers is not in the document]
mwaskom: Added a test that asserts this.

modal/_tunnel.py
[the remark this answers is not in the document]
mwaskom: I'm not exactly sure what users might be doing with this `RemoteError` — as you mention we already retry on these statuses so a persistent issue is unlikely to be solved by more retries, but that won't have changed. I _think_ the phrasing is suitably general as to still be accurate for the other status, though.

modal/exception.py
```diff
@@ -26,10 +64,112 @@ class Error(Exception):
...
+        self._grpc_status = value
+
+    @property
+    def details(self) -> Any:
+        self._warn_on_grpc_error_attribute_access()
+        return self._grpc_details
+
+    @details.setter
+    def details(self, value: Any) -> None:
+        self._grpc_details = value
```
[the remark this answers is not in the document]
mwaskom: This is fine, we don't expect user code catching Modal exception types directly to attempt access of the GRPCError-specific fields.

modal/cli/network_file_system.py
mwaskom: fixed now 
```
$ modal volume ls test-vol bad
Usage: modal volume ls [OPTIONS] VOLUME_NAME [PATH]
Try 'modal volume ls -h' for help.
╭─ Error ───────────────────────────────────────────────────────────────────────────────────╮
│ No such file or directory                                    
╰───────────────────────────────────────────────────────────────────────────────────────────╯
```

PR #3815, opened 2025-12-12
Adds .cleanup methods to Sandbox and ContainerProcess

freider:
We previously relied on `__del__` to be triggered to clean up certain resources, but `__del__` was never reliably called because said resources would also prevent gc most of the time. Also, once `__del__` is eventually triggered (gc collects the circular dependency?), since it ran `.cancel()` on internal tasks, those would trigger non-thread safe event loop modification across threads which could lead to catastrophic races as a result.

This adds an explicit `.cleanup()` method to Sandbox and ContainerProcess. Intentionally named `cleanup` and not `close()` so as not to be confused with something that could terminate the process (this is meant to be run *after* terminating the process, or at least when you don't care about consuming anything from the process anymore)

I don't love this, but it can serve as a stop gap for preventing resource leaks until we enforce context manager usage with .exec and output streams (probably in v1.5, since it would be a breaking change)

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog
* Added a Sandbox.cleanup() and ContainerProcess.cleanup() (for Sandbox.exec processes) that can be used to clean up client-side resources associated with sandboxes and container processes. Use this to prevent resource leakage in long running processes that start and interact with sandboxes.

---

> [!NOTE]
> Adds explicit `cleanup()` methods to `Sandbox` and `ContainerProcess` to close streams/router resources, and removes `__del__` from `TaskCommandRouterClient` in favor of explicit async `close()`.
> 
> - **API**
>   - Add `Sandbox.cleanup()` to close `stdout`/`stderr` streams.
>   - Add `ContainerProcess.cleanup()` delegating to implementation-specific cleanup:
>     - Through server: closes `stdout`/`stderr`.
>     - Through command router: closes `stdout`/`stderr` and calls `TaskCommandRouterClient.close()`.
> - **Internals**
>   - Remove `TaskCommandRouterClient.__del__` and rely on explicit `close()`; `close()` now cancels/awaits the JWT refresh task and closes the channel.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c5e63bc3f1a483a5163375357a29d6963a31622b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_utils/task_command_router_client.py
+++ b/modal/_utils/task_command_router_client.py
@@ -112,9 +112,6 @@ class TaskCommandRouterClient:
     A new instance should be created per task.
     """
 
-    def __del__(self):
-        print("gc commandrouterclient", id(self), type(self))
-
     @classmethod
     async def try_init(
         cls,
@@ -404,7 +401,6 @@ class TaskCommandRouterClient:
         Uses an event to wake early when a manual refresh happens or token changes.
         """
         while not self._closed:
-            print("jank")
             try:
                 exp = self._jwt_exp
                 now = time.time()


modal/_utils/task_command_router_client.py
```diff
@@ -413,6 +404,7 @@ async def _jwt_refresh_loop(self) -> None:
         Uses an event to wake early when a manual refresh happens or token changes.
         """
         while not self._closed:
+            print("jank")
```
mwaskom: :shipit:

modal/container_process.py
```diff
@@ -468,5 +477,15 @@ async def attach(self):
         """mdmd:hidden"""
         await self._impl.attach()
 
+    async def cleanup(self):
```
freider, line 480: This is the new public facing API on the ContainerProcess returned by .exec

PR #3817, opened 2025-12-12
Remove eager JWT background refresh loop

saltzm:
Context [here](https://modal-com.slack.com/archives/C056CGAANRM/p1765541906104129)

TLDR; Calling cancel on an asyncio task from `__del__` is thread-unsafe.

Removing this background loop is also okay since it was just an optimization - we already reactively refresh the token if a request receives `UNAUTHENTICATED`, since there's never a guarantee that the background loop will refresh things on time.

Tests:
* [for streaming](https://github.com/modal-labs/modal-client/blob/79809b3c83fcc6025e493fca4d9c16cdff7e767e/test/task_command_router_client_test.py#L100)
* [for not streaming](https://github.com/modal-labs/modal-client/blob/79809b3c83fcc6025e493fca4d9c16cdff7e767e/test/task_command_router_client_test.py#L662-L711)

Was going to remove `__del__` entirely with the reasoning that `grpclib` already closes the channel in its own `__del__` [implementation](https://github.com/vmagamedov/grpclib/blob/474f0556db5a7fc90e50ac826eadf86fce500c78/grpclib/client.py#L822-L830), but that eventually [also calls cancel](https://github.com/vmagamedov/grpclib/blob/474f0556db5a7fc90e50ac826eadf86fce500c78/grpclib/protocol.py#L240) on an asyncio.TimerHandle 😅 So I wrapped channel.close in `callsoon_threadsafe` instead

---

</details>

## Changelog

* Fixes a thread-safety issue in resource cleanup for sandboxes

---

> [!NOTE]
> Removes the eager JWT refresh task and switches to thread-safe channel cleanup while passing the owning loop/lock into the client; tests updated and a new __del__ cleanup test added.
> 
> - **Core – TaskCommandRouterClient**:
>   - Remove background JWT refresh loop/task and related event/cancellation; rely on reactive refresh on `UNAUTHENTICATED`.
>   - Make cleanup thread-safe: `__del__` now schedules `channel.close()` on the owning loop via `loop.call_soon_threadsafe`; `close()` simplified.
>   - Pass in `loop` and `jwt_refresh_lock` from `try_init()` to ensure they’re created on the correct event loop.
>   - Simplify `_refresh_jwt` (no event signaling) and keep single-flight via `asyncio.Lock`.
>   - Exec stdio method signature uses `api_pb2.FileDescriptor` for `file_descriptor`.
> - **Tests**:
>   - Update constructors to pass `loop` and `jwt_refresh_lock`.
>   - Add test ensuring `__del__` closes the channel via the event loop.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cbce1645a71b35c791a3b6726891d360236462ee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_utils/task_command_router_client.py
+++ b/modal/_utils/task_command_router_client.py
@@ -158,8 +158,10 @@ class TaskCommandRouterClient:
         )
 
         await connect_channel(channel)
+        loop = asyncio.get_running_loop()
+        jwt_refresh_lock = asyncio.Lock()
 
-        return cls(server_client, task_id, resp.url, resp.jwt, channel)
+        return cls(server_client, task_id, resp.url, resp.jwt, channel, loop, jwt_refresh_lock)
 
     def __init__(
         self,
@@ -168,6 +170,8 @@ class TaskCommandRouterClient:
         server_url: str,
         jwt: str,
         channel: grpclib.client.Channel,
+        loop: asyncio.AbstractEventLoop,
+        jwt_refresh_lock: asyncio.Lock,
         *,
         stream_stdio_retry_delay_secs: float = 0.01,
         stream_stdio_retry_delay_factor: float = 2,
@@ -176,7 +180,7 @@ class TaskCommandRouterClient:
         """Callers should not use this directly. Use TaskCommandRouterClient.try_init() instead."""
         # Record the loop this instance is bound to so __del__ can safely schedule cleanup
         # even if finalization happens from a different thread (e.g. via synchronicity).
-        self._loop = asyncio.get_running_loop()
+        self._loop = loop
 
         # Attach bearer token on all requests to the worker-side router service.
         self._server_client = server_client
@@ -191,7 +195,8 @@ class TaskCommandRouterClient:
 
         # JWT refresh coordination
         self._jwt_exp: Optional[float] = _parse_jwt_expiration(jwt)
-        self._jwt_refresh_lock = asyncio.Lock()
+        # This is passed in as an argument to ensure it's created from within the correct event loop.
+        self._jwt_refresh_lock = jwt_refresh_lock
 
         self._closed = False
 
@@ -215,22 +220,24 @@ class TaskCommandRouterClient:
         Use getattr in the event that attributes are not yet initialized or the
         object is in a half-torn-down state.
         """
-        try:
-            if getattr(self, "_closed", False):
-                return
-            self._closed = True
+        if getattr(self, "_closed", False):
+            return
+        self._closed = True
 
-            channel = getattr(self, "_channel", None)
-            if channel is None:
-                return
+        channel = getattr(self, "_channel", None)
+        if channel is None:
+            return
 
-            loop = getattr(self, "_loop", None)
-            if loop is not None and not loop.is_closed():
+        loop = getattr(self, "_loop", None)
+
+        if loop is not None and not loop.is_closed():
+            try:
                 loop.call_soon_threadsafe(channel.close)
-        except Exception:
-            # During interpreter shutdown, the loop may already be torn down, which could raise an exception.
-            # This is safe to ignore, and we don't want to raise an exception from a destructor.
-            pass
+            except Exception:
+                # call_soon_threadsafe could throw if the loop is torn down
+                # after calling is_closed. This is safe to ignore, and we don't
+                # want to raise an exception from a destructor.
+                pass
 
     async def close(self) -> None:
         """Close the client."""
--- a/test/task_command_router_client_test.py
+++ b/test/task_command_router_client_test.py
@@ -56,6 +56,8 @@ async def test_exec_stdio_read_streams_stdout_batches(monkeypatch):
         server_url="https://router.test",
         jwt="t",
         channel=create_dummy_channel(),
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
     )
 
     class _Stream:
@@ -107,6 +109,8 @@ async def test_exec_stdio_read_auth_retry_resumes_from_correct_offset(monkeypatc
         server_url="https://router.test",
         jwt="t",
         channel=create_dummy_channel(),
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
     )
 
     num_attempts_made = 0
@@ -182,6 +186,8 @@ async def test_exec_stdio_read_transient_error_retry_resumes_from_correct_offset
         server_url="https://router.test",
         jwt="t",
         channel=create_dummy_channel(),
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
     )
 
     num_attempts_made = 0
@@ -240,6 +246,8 @@ async def test_exec_stdio_read_auth_fails_twice_raises_auth_error(monkeypatch):
         server_url="https://router.test",
         jwt="t",
         channel=create_dummy_channel(),
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
     )
 
     num_attempts_made = 0
@@ -299,6 +307,8 @@ async def test_exec_stdio_read_unavailable_forever_raises_grpcerror(monkeypatch)
         server_url="https://router.test",
         jwt="t",
         channel=create_dummy_channel(),
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
         stream_stdio_retry_delay_secs=0.001,
         stream_stdio_retry_delay_factor=1.0,
         stream_stdio_max_retries=5,
@@ -346,6 +356,8 @@ async def test_exec_stdio_read_open_raises_once_then_succeeds(monkeypatch):
         server_url="https://router.test",
         jwt="t",
         channel=create_dummy_channel(),
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
     )
 
     raised = False
@@ -397,6 +409,8 @@ async def test_exec_stdio_read_deadline_exceeded_on_send_raises_exec_timeout_err
         server_url="https://router.test",
         jwt="t",
         channel=create_dummy_channel(),
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
     )
 
     class _Stream:
@@ -440,6 +454,8 @@ async def test_exec_stdio_read_deadline_exceeded_on_first_item_raises_exec_timeo
         server_url="https://router.test",
         jwt="t",
         channel=create_dummy_channel(),
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
     )
 
     class _Stream:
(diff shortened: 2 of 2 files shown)


modal/_utils/task_command_router_client.py
```diff
@@ -205,29 +206,38 @@ async def send_request(event: grpclib.events.SendRequest) -> None:
...
-            return
+        """Best-effort cleanup if the caller forgot to close().
 
-        self._jwt_refresh_task.cancel()
+        This object is typically used through synchronicity wrappers, which means this finalizer
+        may run on a different thread than the event loop that owns the channel. Closing the
+        channel is therefore scheduled onto the owning loop using call_soon_threadsafe.
 
+        Use getattr in the event that attributes are not yet initialized or the
+        object is in a half-torn-down state.
```
saltzm, line 221: Did this on guidance from the LLM - the former part makes sense to me, but don't know enough to understand whether the "half-torn-down" state is really possible. 

Even if we currently initialize things in an order so these attributes are always available, it does seem safer to use `getattr` in the event that the order changes somehow. 

But if none of this is necessary please let me know and I'll just remove the `getattr` uses

mwaskom: This sounds right to me, I believe we've run into similar teardown races before

modal/_utils/task_command_router_client.py
```diff
@@ -205,29 +206,38 @@ async def send_request(event: grpclib.events.SendRequest) -> None:
...
+            if loop is not None and not loop.is_closed():
+                loop.call_soon_threadsafe(channel.close)
         except Exception:
+            # During interpreter shutdown, the loop may already be torn down, which could raise an exception.
+            # This is safe to ignore, and we don't want to raise an exception from a destructor.
             pass
 
     async def close(self) -> None:
-        """Close the client and stop the background JWT refresh task."""
+        """Close the client."""
```
thomasjpfan, line 243: Does the `self._channel` in `close` also need to be called with a `loop.call_soon_threadsafe`?

saltzm: I didn't think so since this is an async function and I was assuming that meant it'd be running on the synchronicity event loop. Is that right? Also only called from tests rn

modal/_utils/task_command_router_client.py
freider: Yeah the half torn down state where some attributes become unset can happen when there are circular dependencies between objects (in particular at interpreter shutdown)

modal/_utils/task_command_router_client.py
```diff
@@ -188,11 +192,8 @@ def __init__(
         # JWT refresh coordination
         self._jwt_exp: Optional[float] = _parse_jwt_expiration(jwt)
         self._jwt_refresh_lock = asyncio.Lock()
```
freider: While we are at it, can we move the creation of this asyncio.Lock into an `async def _get_or_create_jwt_lock()` or something, instead of the constructor to get rid of anything that can be implicitly tied to the wrong event loop here

modal/_utils/task_command_router_client.py
```diff
@@ -174,6 +174,10 @@ def __init__(
         stream_stdio_max_retries: int = 10,
     ) -> None:
         """Callers should not use this directly. Use TaskCommandRouterClient.try_init() instead."""
+        # Record the loop this instance is bound to so __del__ can safely schedule cleanup
+        # even if finalization happens from a different thread (e.g. via synchronicity).
+        self._loop = asyncio.get_running_loop()
```
freider: I think it would be better to pass the event loop in an argument of the constructor and extract the loop at the point where we make the `await connect_channel(channel)` above - otherwise there is no guarantee that the loop here is the same as the loop that the Channel is tied to

modal/_utils/task_command_router_client.py
freider: I think we should assume that the async methods of a class will only be called from the same event loop, so this should be fine. It's much riskier with non-async functions interacting with asyncio entities, since it's a much easier mistake to call those from the wrong thread. For example, synchronicity-wrapped non-async functions are just proxy functions that run on whatever thread calls them.

Also the `close` method here is never actually called in the client since we have no notion of end scope for these objects (as @saltzm says, only called in tests)

modal/_utils/task_command_router_client.py
```diff
@@ -407,47 +412,6 @@ async def _call_with_auth_retry(self, func, *args, **kwargs):
                 return await func(*args, **kwargs)
             raise
 
-    async def _jwt_refresh_loop(self) -> None:
```
freider, line 410: 💚

modal/_utils/task_command_router_client.py
```diff
@@ -205,29 +206,38 @@ async def send_request(event: grpclib.events.SendRequest) -> None:
...
+
+            channel = getattr(self, "_channel", None)
+            if channel is None:
+                return
+
+            loop = getattr(self, "_loop", None)
+            if loop is not None and not loop.is_closed():
+                loop.call_soon_threadsafe(channel.close)
         except Exception:
+            # During interpreter shutdown, the loop may already be torn down, which could raise an exception.
```
freider: Nit: Wouldn't the loop.is_closed() check catch that? Wonder if we could narrow the scope of this exception catch to prevent silently hiding bugs in *our* code. At least we could have it just surround the call_soon_threadsafe

freider (commented):
Added some inline comments. Apart from those I also think we should add a small test that exercises the `__del__` explicitly to see that the channel gets closed.
And if we don't already have it, something that tests that jwt renewal works without the loop

modal/_utils/task_command_router_client.py
saltzm: Decided to solve this same as the `get_running_loop` call by instead passing the lock into the constructor from `try_init`

modal/_utils/task_command_router_client.py
saltzm: True, narrowed the scope

saltzm: Actually is there any chance there could be a race where you check if it's closed, it's not, then it gets closed in the background somewhere, and then you try to call `call_soon_threadsafe` and it's closed?

test/task_command_router_client_test.py
```diff
@@ -709,3 +735,35 @@ async def _refresh_jwt():
...
+        jwt="t",
+        channel=channel,
+        loop=asyncio.get_running_loop(),
+        jwt_refresh_lock=asyncio.Lock(),
+    )
+
+    # __del__ schedules channel.close() on the loop - let the loop run so the callback is executed.
+    #
+    # Note: TaskCommandRouterClient registers a request hook with grpclib, which can create a reference
+    # cycle (client -> channel -> callback -> client). Rely on calling __del__ explicitly in the test
```
saltzm, line 764: This is an LLM comment - does this make sense to y'all?

saltzm: I did confirm the test does not work if you just call `del client`

freider (approved):
lgtm, thanks for fixing! we can probably come up with more "integration style" tests for the channel, similar to what we do with the other grpc client, but that's a much bigger task so I wouldn't hold up this PR on that

test/task_command_router_client_test.py
thomasjpfan: **Follow up**: This reference cycle is unfortunate. The simple way to break the cycle is to use a `weakref` in the `TaskCommandRouterClient.__init__`:

```python
weak_self = weakref.ref(self)

async def send_request(event: grpclib.events.SendRequest) -> None:
    inner_self = weak_self()
    if inner_self is None:
         raise ValueError("This should never happen?")
    event.metadata["authorization"] = f"Bearer {inner_self._jwt}"
```

modal/_utils/task_command_router_client.py
```diff
@@ -205,29 +211,40 @@ async def send_request(event: grpclib.events.SendRequest) -> None:
...
+        """Best-effort cleanup if the caller forgot to close().
+
+        This object is typically used through synchronicity wrappers, which means this finalizer
+        may run on a different thread than the event loop that owns the channel. Closing the
+        channel is therefore scheduled onto the owning loop using call_soon_threadsafe.
+
+        Use getattr in the event that attributes are not yet initialized or the
+        object is in a half-torn-down state.
+        """
+        if getattr(self, "_closed", False):
```
erikbern, line 223: This feels a bit too defensive? If we initialize these in the constructor then they should always be set right

saltzm: See here for discussion: https://github.com/modal-labs/modal-client/pull/3817#discussion_r2614794951

saltzm: `__del__` is [called](https://stackoverflow.com/questions/6409644/is-del-called-on-an-object-that-doesnt-complete-init) even when `__init__` throws, so you can end up with a partially initialized object

saltzm:
Ran locally:
```py
import sys
import time

import modal

sandbox = modal.Sandbox.create(
    "sleep",
    "100000",
    app=modal.App.lookup("base44-bug", create_if_missing=True),
    image=modal.Image.debian_slim(),
    timeout=10000
)

proc = sandbox.exec("echo", "hello")
print(proc.stdout.read())
proc.wait()

print("sleeping...")

# Sleep past JWT refresh period
total_seconds = 700
bar_width = 40
start = time.monotonic()
while True:
    elapsed = int(time.monotonic() - start)
    elapsed = min(elapsed, total_seconds)
    frac = elapsed / total_seconds if total_seconds else 1.0
    filled = int(bar_width * frac)
    bar = "#" * filled + "-" * (bar_width - filled)
    sys.stdout.write(f"\r[{bar}] {elapsed}/{total_seconds}s")
    sys.stdout.flush()
    if elapsed >= total_seconds:
        break
    time.sleep(1)
print()  # newline after progress bar
print("done sleeping...")

proc = sandbox.exec("echo", "hello")
print(proc.stdout.read())
proc.wait()

sandbox.terminate()
```

And confirmed refresh works (with some additional local debug logging):
```
(modal) [ec2-user@ip-10-1-2-39 client]$ MODAL_PROFILE=prod MODAL_LOGLEVEL=debug python main.py
[modal-client] 2025-12-12T23:23:35+0000 Connecting to api.modal.com using scheme https
[modal-client] 2025-12-12T23:23:35+0000 Sending request to /modal.client.ModalClient/AppGetOrCreate
[modal-client] 2025-12-12T23:23:35+0000 Sending request to /modal.client.ModalClient/EnvironmentGetOrCreate
[modal-client] 2025-12-12T23:23:35+0000 Created environment with id en-6nbXLj0pkWRVS0fpPsT44V
[modal-client] 2025-12-12T23:23:35+0000 Sending request to /modal.client.ModalClient/ImageGetOrCreate
[modal-client] 2025-12-12T23:23:36+0000 Sending request to /modal.client.ModalClient/SandboxCreate
[modal-client] 2025-12-12T23:23:36+0000 Sending request to /modal.client.ModalClient/SandboxGetTaskId
[modal-client] 2025-12-12T23:23:36+0000 Sending request to /modal.client.ModalClient/TaskGetCommandRouterAccess
[modal-client] 2025-12-12T23:23:36+0000 Using command router access for task ta-01KCAE10A3AF6R552E49GQ3K73
RUNNING LOCAL CLIENT
hello

sleeping...
[########################################] 700/700s
done sleeping...
[modal-client] 2025-12-12T23:35:17+0000 Refreshing JWT for exec with task ID ta-01KCAE10A3AF6R552E49GQ3K73
[modal-client] 2025-12-12T23:35:17+0000 Sending request to /modal.client.ModalClient/TaskGetCommandRouterAccess
hello

[modal-client] 2025-12-12T23:35:17+0000 Sending request to /modal.client.ModalClient/SandboxTerminate
[modal-client] 2025-12-12T23:35:17+0000 Client (139666584892496): closing
```

PR #3819, opened 2025-12-12
Avoid reading all Mount files into memory during upload

mwaskom:
Mounting directories with large files can cause uncontrolled memory accumulation because all files are read into memory in the first pass to hash their contents. With this fix, files are only held in memory if they are above a size threshold, otherwise we defer reading them into memory until they are being uploaded.

</details>

## Changelog

- Improved performance when adding large local directories to an Image.

---

> [!NOTE]
> Streamlines hashing and file uploads by caching only very small files, reading content on-demand, and yielding mount files lazily to bound memory usage.
> 
> - **Mount/Upload pipeline**
>   - Lazily yield `FileUploadSpec` from `_get_files` instead of pre-creating tasks; adds debug logging for checksum computation and per-file mounting.
>   - During non-blob uploads, read content on-demand (`file_spec.read_content`) if not pre-cached.
> - **Blob/File utils (`modal/_utils/blob_utils.py`)**
>   - `FileUploadSpec`: add `content: Optional[bytes] = None` and `read_content()` helper.
>   - `_get_file_upload_spec`:
>     - Use blob for files `>= LARGE_FILE_LIMIT` (unchanged).
>     - For small files (`< 256 KiB`), read once and cache `content` for hashing/upload.
>     - For medium files (256 KiB–4 MiB), compute hashes from the file object without caching.
> - **Volume uploads (`modal/volume.py`)**
>   - Mirror on-demand content read for non-blob uploads.
> - **Tests**
>   - Update `test_get_files` to align with new behavior (no assertion on large-file `content` being `None`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c81dde3f95749dc1363cac3721ab363f5af8a6ca. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -549,8 +549,10 @@ class _Mount(_Object, type_prefix="mo"):
                 logger.debug(
                     f"Uploading file {file_spec.source_description} to {remote_filename} ({file_spec.size} bytes)"
                 )
-                # Read content lazily only when needed for upload
-                content = file_spec.content or await asyncio.to_thread(file_spec.read_content)
+                if file_spec.content is None:
+                    content = await asyncio.to_thread(file_spec.read_content)
+                else:
+                    content = file_spec.content
                 request2 = api_pb2.MountPutFileRequest(data=content, sha256_hex=file_spec.sha256_hex)
 
             start_time = time.monotonic()
--- a/modal/volume.py
+++ b/modal/volume.py
@@ -1056,8 +1056,10 @@ class _VolumeUploadContextManager(_AbstractVolumeUploadContextManager):
                 logger.debug(
                     f"Uploading file {file_spec.source_description} to {remote_filename} ({file_spec.size} bytes)"
                 )
-                # Read content lazily only when needed for upload
-                content = file_spec.content or await asyncio.to_thread(file_spec.read_content)
+                if file_spec.content is None:
+                    content = await asyncio.to_thread(file_spec.read_content)
+                else:
+                    content = file_spec.content
                 request2 = api_pb2.MountPutFileRequest(data=content, sha256_hex=file_spec.sha256_hex)
                 self._progress_cb(task_id=progress_task_id, complete=True)
 


modal/mount.py
```diff
@@ -464,21 +464,36 @@ def _description(entries: list[_MountEntry]) -> str:
     @staticmethod
     async def _get_files(entries: list[_MountEntry]) -> AsyncGenerator[FileUploadSpec, None]:
         loop = asyncio.get_event_loop()
+        # Limit concurrent file spec creation to prevent memory accumulation
+        max_concurrent_specs = 128
```
thomasjpfan: The `async_map` has a [concurrent limit](https://github.com/modal-labs/modal-client/blob/e874cd13d79a6bbf63c86fdd5f7df680c6901240/modal/mount.py#L563-L565) and we have another one here.

In https://github.com/modal-labs/modal-client/pull/3694 we adjusted that number on the client. With this new limit, we'll have two values to think about when thinking about concurrent uploads. (BTW I'm okay with this)

modal/mount.py
```diff
@@ -547,7 +562,9 @@ async def _put_file(file_spec: FileUploadSpec) -> api_pb2.MountFile:
                 logger.debug(
                     f"Uploading file {file_spec.source_description} to {remote_filename} ({file_spec.size} bytes)"
                 )
-                request2 = api_pb2.MountPutFileRequest(data=file_spec.content, sha256_hex=file_spec.sha256_hex)
+                # Read content lazily only when needed for upload
+                content = await asyncio.get_event_loop().run_in_executor(None, file_spec.get_content)
```
thomasjpfan: Also, there is overhead for running everything on another thread. Can we do something like this?

```python
if file_spec.cached_content is not None:
    content = file_spec.cached_content
else:
    content = await asyncio.to_thread(file_spec.read_content)  
```

(Also using `asyncio.to_thread` since it does the same thing as `run_in_executor`)

modal/volume.py
```diff
@@ -1056,7 +1056,9 @@ async def _upload_file(self, file_spec: FileUploadSpec) -> api_pb2.MountFile:
                 logger.debug(
                     f"Uploading file {file_spec.source_description} to {remote_filename} ({file_spec.size} bytes)"
                 )
-                request2 = api_pb2.MountPutFileRequest(data=file_spec.content, sha256_hex=file_spec.sha256_hex)
+                # Read content lazily only when needed for upload
+                content = await asyncio.get_event_loop().run_in_executor(None, file_spec.get_content)
```
thomasjpfan: Same here regarding checking for `cached_content`.

modal/mount.py
```diff
@@ -464,21 +464,36 @@ def _description(entries: list[_MountEntry]) -> str:
     @staticmethod
     async def _get_files(entries: list[_MountEntry]) -> AsyncGenerator[FileUploadSpec, None]:
         loop = asyncio.get_event_loop()
+        # Limit concurrent file spec creation to prevent memory accumulation
+        max_concurrent_specs = 128
+        semaphore = asyncio.Semaphore(max_concurrent_specs)
+
         with concurrent.futures.ThreadPoolExecutor() as exe:
             all_files = await loop.run_in_executor(exe, _select_files, entries)
+            logger.debug(f"Computing checksums for {len(all_files)} files using {exe._max_workers} worker threads")
```
thomasjpfan, line 469: Using the private attribute `exe._max_workers` feels brittle, but this is an existing problem in `main`.

modal/mount.py
mwaskom: Do you mean there's overhead to doing the "check if we already have content" part on the other thread?

thomasjpfan: Yup. If the content is already cached then, there is no need to run a function another another thread.

mwaskom: Yeah makes sense

thomasjpfan (approved):

PR #3822, opened 2025-12-12
Add lints to catch stray `print`s and `breakpoint()`s

ehdr:
To auto-catch when [debug stuff lingers](https://github.com/modal-labs/modal-client/pull/3815#discussion_r2614253791).

---

## Compatibility checklist

</details>

---

> [!NOTE]
> Enable Ruff T10/T20 to flag breakpoint/print; add targeted per-file ignores and # noqa annotations on intentional prints/breakpoints.
> 
> - **Lint configuration (pyproject.toml)**:
>   - Enable Ruff rules `T10` (flake8-debugger) and `T20` (flake8-print).
>   - Add per-file ignores for `T201` (print) across tests and specific module paths; extend existing ignores.
> - **Codebase annotations**:
>   - Add `# noqa: T201` to intentional `print` usages in `modal/_output.py`, `modal/_traceback.py`, `modal/cli/*`, `modal/container_process.py`, `modal/io_streams.py`, `modal/mount.py`, `modal_version/__main__.py`.
>   - Add `# noqa: T100` for intentional `breakpoint()`/`pdb` imports in `modal/_runtime/container_io_manager.py` and `modal/_runtime/user_code_imports.py`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5fe6232c186c391b44b59808ca67b123545637ea. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/mount.py
+++ b/modal/mount.py
@@ -769,13 +769,13 @@ async def _create_single_client_dependency_mount(
     if check_if_exists:
         try:
             await Mount.from_name(mount_name, namespace=api_pb2.DEPLOYMENT_NAMESPACE_GLOBAL).hydrate.aio(client)
-            print(f"➖ Found existing mount {mount_name} in global namespace.")
+            print(f"➖ Found existing mount {mount_name} in global namespace.")  # noqa: T201
             return
         except modal.exception.NotFoundError:
             pass
 
     with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as tmpd:
-        print(f"📦 Building {mount_name}.")
+        print(f"📦 Building {mount_name}.")  # noqa: T201
         requirements = os.path.join(os.path.dirname(__file__), f"builder/{builder_version}.txt")
         cmd = " ".join(
             [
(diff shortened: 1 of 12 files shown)


modal/mount.py
```diff
@@ -769,13 +769,13 @@ async def _create_single_client_dependency_mount(
     if check_if_exists:
         try:
             await Mount.from_name(mount_name, namespace=api_pb2.DEPLOYMENT_NAMESPACE_GLOBAL).hydrate.aio(client)
-            print(f"➖ Found existing mount {mount_name} in global namespace.")
+            print(f"➖ Found existing mount {mount_name} in global namespace.")  # noqa: T201
```
mwaskom, line 772: Probably makes sense to move this stuff out of this directory altogether. I guess another way to avoid sprinkling `noqa: T201` everywhere would be to use always `rich` where we are intending to produce user-facing (although, in this case, modal devs are the "user") output.

ehdr: Thanks, sounds good. Where should it go instead? Ok if I do it in a follow-up PR?

mwaskom: Yeah was kind of an idle thought. Might be an annoying migration as IIRC we depend on this from the monorepo too.

PR #3834, opened 2025-12-17
Ad hoc byte/string optimizations

mwaskom:
Caught a couple of cases where we had iterative string concatenations that could be improved.

---

> [!NOTE]
> Switches string/bytes accumulation to StringIO/BytesIO in multipart upload XML, filesystem watch/wait paths, and stream readers to reduce concatenation overhead.
> 
> - **Performance**:
>   - **Blob uploads (`modal/_utils/blob_utils.py`)**: Build multipart completion XML via list-join instead of iterative string concatenation; minor comment fix.
>   - **Filesystem I/O (`modal/file_io.py`)**:
>     - `_wait`: accumulate output using `BytesIO`.
>     - `watch`: buffer event chunks with `BytesIO` and add `end_of_event` helper for boundary detection.
>   - **Streams (`modal/io_streams.py`)**: Use `StringIO`/`BytesIO` in `StreamReader.read()` for text/binary accumulation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8f6061b95918c631b99da0bd8d96adc38c6b36ab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/file_io.py
+++ b/modal/file_io.py
@@ -459,7 +459,7 @@ class _FileIO(Generic[T]):
             ),
         )
 
-        def end_of_event(item: bytes, buffer: io.BytesIO, boundary_token: bytes = b"\n\n") -> bool:
+        def end_of_event(item: bytes, buffer: io.BytesIO, boundary_token: bytes) -> bool:
             if not item.endswith(b"\n"):
                 return False
             boundary_token_size = len(boundary_token)
@@ -468,7 +468,6 @@ class _FileIO(Generic[T]):
             buffer.seek(-boundary_token_size, io.SEEK_END)
             if buffer.read(boundary_token_size) == boundary_token:
                 return True
-            buffer.seek(0, io.SEEK_END)
             return False
 
         async with TaskContext() as tc:
@@ -484,7 +483,8 @@ class _FileIO(Generic[T]):
                         raise item
                     item_buffer.write(item)
                     assert isinstance(item, bytes)
-                    if end_of_event(item, item_buffer):
+                    # Single events may span multiple messages so we need to check for a special event boundary token
+                    if end_of_event(item, item_buffer, boundary_token=b"\n\n"):
                         try:
                             event_json = json.loads(item_buffer.getvalue().strip().decode())
                             event = FileWatchEvent(


modal/_utils/blob_utils.py
```diff
@@ -152,12 +152,13 @@ async def perform_multipart_upload(
...
 
     # The body of the complete_multipart_upload command needs some data in xml format:
-    completion_body = "<CompleteMultipartUpload>\n"
+    completion_parts = ["<CompleteMultipartUpload>"]
     for part_number, etag in enumerate(part_etags, 1):
-        completion_body += f"""<Part>\n<PartNumber>{part_number}</PartNumber>\n<ETag>"{etag}"</ETag>\n</Part>\n"""
-    completion_body += "</CompleteMultipartUpload>"
+        completion_parts.append(f"""<Part>\n<PartNumber>{part_number}</PartNumber>\n<ETag>"{etag}"</ETag>\n</Part>""")
+    completion_parts.append("</CompleteMultipartUpload>")
+    completion_body = "".join(completion_parts)
```
thomasjpfan: Was there a benefit of going with `list` + `"".join(..)` here instead of `StringIO`?

modal/file_io.py
```diff
@@ -458,23 +458,35 @@ async def watch(
...
+            item_buffer = io.BytesIO()
             while True:
                 if len(self._watch_output_buffer) > 0:
                     item = self._watch_output_buffer.pop(0)
                     if item is None:
                         break
                     if isinstance(item, Exception):
                         raise item
-                    buffer += item
-                    # a single event may be split across multiple messages
```
thomasjpfan, line 473: Can we keep this comment?

For my understanding, who is setting the two newlines to end the stream?

modal/file_io.py
```diff
@@ -458,23 +458,35 @@ async def watch(
                 task_id=self._task_id,
             ),
         )
+
+        def end_of_event(item: bytes, buffer: io.BytesIO, boundary_token: bytes = b"\n\n") -> bool:
```
thomasjpfan: It's a bit unfortunate there is no cleaner way to detect when the steam ends.

modal/file_io.py
```diff
@@ -458,23 +458,35 @@ async def watch(
...
+            if not item.endswith(b"\n"):
+                return False
+            boundary_token_size = len(boundary_token)
+            if buffer.tell() < boundary_token_size:
+                return False
+            buffer.seek(-boundary_token_size, io.SEEK_END)
+            if buffer.read(boundary_token_size) == boundary_token:
+                return True
+            buffer.seek(0, io.SEEK_END)
```
thomasjpfan: After the `buffer.read`, I think the current stream position should already be at the end?

(Given we `seek`ed to -2)

modal/file_io.py
mwaskom: I have no context for this really. I assume this is going to get rewritten in a new FS API implementation.

modal/file_io.py
mwaskom: gp

modal/_utils/blob_utils.py
mwaskom: Not that I've benchmarked, although my sense has always been that `.join()` is the canonical suggestion for eliminating iterative concatenation. I do kind of like the `io` buffer approach though.

modal/_utils/blob_utils.py
```diff
@@ -152,12 +152,13 @@ async def perform_multipart_upload(
...
 
     # The body of the complete_multipart_upload command needs some data in xml format:
-    completion_body = "<CompleteMultipartUpload>\n"
+    completion_parts = ["<CompleteMultipartUpload>"]
     for part_number, etag in enumerate(part_etags, 1):
-        completion_body += f"""<Part>\n<PartNumber>{part_number}</PartNumber>\n<ETag>"{etag}"</ETag>\n</Part>\n"""
-    completion_body += "</CompleteMultipartUpload>"
+        completion_parts.append(f"""<Part>\n<PartNumber>{part_number}</PartNumber>\n<ETag>"{etag}"</ETag>\n</Part>""")
+    completion_parts.append("</CompleteMultipartUpload>")
+    completion_body = "\n".join(completion_parts)
```
thomasjpfan, line 159: **Follow up**: I hope the newlines between the tags do not actually matter.

mwaskom: Hopefully not :)

PR #3837, opened 2025-12-17
Add a `current_retry_count()` function

eggag32:
## Describe your changes

The retry count is already part of `IOContext`, this just exposes it to users.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Expose `current_retry_count()` to user code and propagate retry counts through execution context and container I/O.
> 
> - **API/Runtime**:
>   - Add `current_retry_count()` in `modal/_runtime/execution_context.py` and export via `modal/__init__.py` (`__all__`).
>   - Extend `_set_current_context_ids(...)` to accept `retry_counts` and store in `_current_retry_count` `ContextVar`.
> - **Container**:
>   - Update `call_function` in `modal/_container_entrypoint.py` to pass `io_context.retry_counts` to `_set_current_context_ids`.
> - **Tests**:
>   - Adjust ASGI tests to pass `[0]` `retry_counts` to `_set_current_context_ids` in `test/asgi_wrapper_test.py`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c5e0d9eb5ab14e65ed2385ac0d825babc5c7503c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/_runtime/execution_context.py
+++ b/modal/_runtime/execution_context.py
@@ -102,7 +102,7 @@ def current_attempt_token() -> Optional[str]:
 def _set_current_context_ids(
     input_ids: list[str], function_call_ids: list[str], attempt_tokens: list[str], retry_counts: list[int]
 ) -> Callable[[], None]:
-    assert len(input_ids) == len(function_call_ids) == len(attempt_tokens) and input_ids
+    assert len(input_ids) == len(function_call_ids) == len(attempt_tokens) == len(retry_counts) and input_ids
 
     input_id = input_ids[0]
     function_call_id = function_call_ids[0]
--- a/modal/__init__.py
+++ b/modal/__init__.py
@@ -9,7 +9,13 @@ if sys.version_info[:2] >= (3, 14):
 from modal_version import __version__
 
 try:
-    from ._runtime.execution_context import current_function_call_id, current_input_id, interact, is_local
+    from ._runtime.execution_context import (
+        current_function_call_id,
+        current_input_id,
+        current_retry_count,
+        interact,
+        is_local,
+    )
     from ._tunnel import Tunnel, forward
     from .app import App
     from .client import Client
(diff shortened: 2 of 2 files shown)


modal/_runtime/execution_context.py
```diff
@@ -72,6 +72,25 @@ def process_stuff():
         return None
 
 
+def current_retry_count() -> Optional[int]:
```
thomasjpfan, line 75: What do you think of having a `RunningContext` dataclass that exposes all of the metadata for the function call?

Specifically something like this:

```python
@dataclass
class RunningContext:
    task_id: str
    retry_count: int
    region: str
    ...
```

The above would be returned by a `FunctionCall.current_context()` that can only be called in a container.

mwaskom: Would also prefer that we head in this direction over adding more random functions in the `modal` namespace, although I am not sure about the exact spelling. We'd probably want a spelling that also works for other objects like `modal.Container` and perhaps even `modal.Environment`, etc.

eggag32:
> How is this going to be defined w/r/t various "retry" concepts? Does it include both auto-retries after container failures and any input retries as configured by the client-side `retries=` parameter? If the former, will it be confusing that you can have `current_retry_count() > retries`?

Good question. From the in-person discussion, it sounds like having internal and user code failures counted separately could make sense, but needs more thought on how it is exposed in the SDK.

Thought a bit more about the "how do you know if it is the last retry" logic -- even if we have separate counts, I am not sure how you can tell if the code is running for the last time. For instance, if the number of user retries is at its maximum, if there are still some internal retries left, the code could run again (does that matter?). Maybe I am missing something -- would be nice to know what people would like to use the retry count for.

PR #3840, opened 2025-12-17
Instantiate flash manager exclusively on rank 0 for clustered functions

timmy-feng:
## Describe your changes

Previously, flash requests potentially route to any rank of a clustered function because each task exposes a flash handle. Similarly, health checks relied on acknowledgments from all ranks. This change makes it so that only the first rank exposes a handle, aligning Flash SDK with the idea that only rank 0 receives and responds to flash HTTP requests.

Note: This deviates from the Modal clustered paradigm where inputs are replicated to each rank. Going down this route may pose compatibility issues in the future if we plan to back `.remote()` with Flash Proxy.

---

## Compatibility checklist

---

## Release checklist

---

</details>

## Changelog

---

> [!NOTE]
> Only rank 0 of a clustered Flash function now starts the flash manager; tests confirm non-zero ranks do not expose a Flash handle.
> 
> - **Flash container entry**:
>   - In `modal/experimental/flash.py`, instantiate `flash_forward` only when `get_cluster_info().rank == 0`; fallback to rank `0` if cluster info unavailable.
> - **Tests**:
>   - Add `TestFlashClusteredEntry` in `test/flash_test.py` validating only rank 0 creates the Flash handle and others do not; update imports accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c97b5086a470b0a95709706465fe434cba176320. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/modal/experimental/flash.py
+++ b/modal/experimental/flash.py
@@ -10,6 +10,7 @@ from collections import defaultdict
 from typing import Any, Callable, Optional, Union
 from urllib.parse import urlparse
 
+from modal._clustered_functions import get_cluster_info
 from modal._partial_function import _PartialFunctionFlags
 from modal.cls import _Cls
 from modal.dict import _Dict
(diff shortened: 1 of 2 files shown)


modal/experimental/flash.py
```diff
@@ -720,12 +721,17 @@ def __init__(self, http_config: api_pb2.HTTPConfig):
...
-                startup_timeout=self.http_config.startup_timeout,
-                exit_grace_period=self.http_config.exit_grace_period,
-                h2_enabled=self.http_config.h2_enabled,
-            )
+            try:
+                rank = get_cluster_info().rank
+            except InvalidError:
+                rank = 0
+            if rank == 0:
+                self.flash_manager = flash_forward(
```
thomasjpfan, line 729: Should this be a part of `HTTPConfig` and be configurable by the user? Specifically, is there a use case for `flash_forward`ing to the other ranks?

